Issue #115, opened 2025-02-27
fix: address an issue where sonification is not completely cleared after autoplay stops


jooyoungseo:
# Bug Report

## Description

I hear some white noise sonification when autoplay stops.

## Steps to Reproduce

1. Adjust your autoplay speed to be fast enough
2. Trigger autoplay either from left to right or right to left in `examples/lineplot.html`
3. While autoplay is in progress, interrupt by pressing ctrl key
4. Listen carefully whether you hear some  background noise

## Actual Behavior

Sonification is not completely cleared such that users hear subtle remaining background noise.

## Expected Behavior

Background noise needs to be killed.

## Screenshots

N/A

## Environment

- Operating System: Windows 11
- Browser: Chrome
- Version: universal

## Additional Information

N/A

PR #150, opened 2025-03-14
feat: support distinct audio legend for group levels

### LINKED ISSUE #149: feat: implement audio legend
# Feature Request

## Motivation

Multi-line plot and stacked bar plot have more than one categorical group levels. However, we currently use the same sound color for all the groups within those plots. This makes it difficult to distinguish between the groups when people use sonification.

## Proposed Solution

I propose to implement an audio legend that will provide a unique sound color for each group within the multi-line plot and stacked bar plot. This will help people to distinguish between the groups when they use sonification.

Key consideration is that we need to make this audio legend component scalable in a way that it can be matched with the number of groups within the plot and each of the audio colors needs to be different from each other.

## Additional Context

N/A

## Checklist

- [x] I have searched for similar feature requests and confirmed that this is a new request.
- [x] I have provided a clear and concise description of the feature.
- [x] I have explained the motivation behind this feature request.
- [x] I have outlined a proposed solution or ideas for implementing this feature.
- [x] I have provided any additional context or screenshots if applicable.

jooyoungseo:
Introduce a new command to play audio legends for data groups in multi-line and stacked bar plots. Refactor the command interface to accommodate this feature and enhance audio differentiation by including group indices in relevant models.

Fixes #149 and #115

### DIFF
--- a/src/audio/audio-transitions.ts
+++ b/src/audio/audio-transitions.ts
@@ -4,50 +4,127 @@
  */
 
 /**
- * Applies a fade-in to an audio parameter
+ * Applies a fade-in to an audio parameter using an exponential curve for smoother transitions
+ *
  * @param param - The audio parameter to fade in (e.g., gainNode.gain)
- * @param startValue - The starting value (usually 0)
+ * @param startValue - The starting value (usually near-zero)
  * @param endValue - The target value to fade to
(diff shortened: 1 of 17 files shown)


jooyoungseo:
Closing in favor of #279

PR #336, opened 2025-06-13
feat: add fallback svg highlight for smooth line trace

### LINKED ISSUE #335: fix: Smooth plot tracing
# Bug Report

## Description

Fix, the smooth curve highlight issue in line.ts. The number of points in smooth curve (<path> tag) is lesser than the number of points present in lowess curve. So, the one-to-one mapping fails in line.ts. Identify a fix to this issue.

line.ts :  211-219
```
      if (coordinates.length !== this.lineValues[r].length) {
        if (coordinates.length < this.lineValues[r].length) {
          while (coordinates.length < this.lineValues[r].length) {
            coordinates.push({ x: Number.NaN, y: Number.NaN });
          }
        } else if (coordinates.length > this.lineValues[r].length) {
          coordinates.length = this.lineValues[r].length;
        }
      }

```

nk1408:
# Pull Request

## Description

Enhancement: Robust Highlighting for Smooth Curves

Improved the highlight feature for smooth curves by ensuring a true one-to-one mapping between data points and highlight circles. The code now directly uses svg_x and svg_y values (passed from py-maidr) for each data point, guaranteeing that highlights correspond exactly to the points rendered in SVG space. This approach eliminates mismatches between real data space and SVG space, providing accurate keyboard navigation and highlighting for all smooth plots.

## Related Issues

Closes #335 

## Changes Made

src/model/smooth.ts

Refactored the highlight logic to use svg_x and svg_y from each data point for creating highlight circles,
Now uses the new SmoothPoint type for type safety and clarity.

src/model/line.ts

Changed the points property from private to protected so subclasses (like SmoothTrace) can access it directly and safely.
Changed mapToSvgElements from private to protected to extend in subclass

src/type/grammar.ts

Added a new SmoothPoint interface to define the expected structure for smooth plot data points, including both real (x, y) and SVG (svg_x, svg_y) coordinates.

## Checklist


jooyoungseo:
@nk46-cloud Using `svg_x` and `svg_y` properties violates the design consistency across other plots.

1. Please add this `svg_x`/`svg_y` approach as a fallback logic.

2. If this approach is applicable to other plots, let's refactor other plots' hilight (file an issue).

nk1408:
> @nk46-cloud Using `svg_x` and `svg_y` properties violates the design consistency across other plots.
> 
> 1. Please add this `svg_x`/`svg_y` approach as a fallback logic.
> 2.

@jooyoungseo , Do you think it would be better to override mapToSvgElements in smooth.ts or possibly cover that fallback in line.ts ? The later might cover cases where logic possibly can fail in lines but we might have to make a lot changes downstream to acomodate this svg change. The other option is we can apply this fallback logic only to smooth lines, with a very minimal blast radius. Thanks!

jooyoungseo:
@nk46-cloud For now, let's limit the fallback scope to smooth line trace. We need to make a follow-up refactor for lineplot anyhow because currently, highlight doesn't work well in [multi-lineplot](https://py.maidr.ai/examples.html#multiline-plot).

krishnanand5:
LGTM!

nk1408:
> > @nk46-cloud Using `svg_x` and `svg_y` properties violates the design consistency across other plots.
> > 
> > 1. Please add this `svg_x`/`svg_y` approach as a fallback logic.
> 
> @jooyoungseo , Do you think it would be better to override mapToSvgElements in smooth.ts or possibly cover that fallback in line.ts ? The later might cover cases where logic possibly can fail in lines but we might have to make a lot changes downstream to acomodate this svg change. The other option is we can apply this fallback logic only to smooth lines, with a very minimal blast radius. Thanks!

I've added fallback logic in latest commit. Please review if this would be good. Thanks

jooyoungseo (approved):

krishnanand5:
Merging!

PR #354, opened 2025-06-17
feat: add command palette

### LINKED ISSUE #324: feat: create a UI component to pickup keyboard shortcuts directly
Similar to Google Docs and VSCode, we need a UI element that allows the user to type in or select from a list of commands and execute that directly.

krishnanand5:
# Pull Request

## Description

This PR aims to introduce a command palette that users can utilize to run a command on the active plot.

## Related Issues

closes #324 

## Changes Made

The following changes have been made in this PR:
1. Created a new keybinding `Ctrl + Shift + p` to open the newly created `CommandPalette` component.
2. The new command palette component consists of a search box and list of commands available in the current scope.
3. When the `CommandPalette` is invoked, it only shows the list of commands available in the current scope.

## Checklist

- [X] I have read the [Contributor Guidelines](../CONTRIBUTING.md).
- [X] I have performed a self-review of my own code and ensured it follows the project's coding standards.
- [X] I have tested the changes locally following `ManualTestingProcess.md`, and all tests related to this pull request pass.
- [ ] I have commented my code, particularly in hard-to-understand areas.


krishnanand5:
Suggestions from meeting on 06/26/2025:
1. Replace `Describe` in label scope command description with `Announce`.
2. Validate if `cmd` and `ctrl` are appearing on shortcuts in mac and windows respectively.
3. Change command picker shortcut from `cmd/ctrl + k` to `cmd/ctrl + shift + p`.

krishnanand5:
Greetings,

The shortcut to invoke the command picker is now `Ctrl + Shift + p`.

Professor @jooyoungseo - The updation of `Describe` instances to `Announce` involves changes in core services, Can I take this up as a seperate chore/refactor task as this PR is very UI-centric?

Regards,
Krishna Anandan Ganesan.

ellvix:
Ctrl Shift P conflicts with the default print command (chromium based browsers (Chrome / Edge) + Windows), new incognito (FF + Windows). Is Ctrl Alt P available?

ellvix (changes requested):
Basic functionality works for sighted users, but lots of issues for low vision and screen reader users:

 - dialog needs to be labeled, which it is, but it defaults to role="presentation" which kills the aria-label. Add role="dialog" to the main top parent <Dialog>, around line 160 in CommandPalette.tsx to fix. 
 - Color contrast of highlighted / not is not sufficient. Need a 3:1 contrast ratio at minimum. If you really like that light blue on white, fine, add a darker (3:1 min) border instead. 
 - Main field label is going on the parent parent, just like we found with settings, and the same fix works, our old friend: 
```
slotProps={{
  input: {
    inputProps: {
      'aria-label': 'Search Commands',
    },
  },
}}
```
 - Autocomplete / arrow interaction is pretty bad if you're on a screen reader. First, you'd never know it's there, and second, arrow keys don't seem to do anything audible. 
     - Recommend changing the aria label to include basic instructions like "Search Commands, down arrow to select filtered commands". Hmm... actually that's cumbersome. I'd hate to hear that every time. I'd like Dr Seo's opinion, but I'd rather have those instructions in a visually hidden text node between the modal title and search field. 
     - Need some way to announce what's happening, and aria-live (or better, role="alert") is the way to go. You have that, but then you also have role="presentation" which shuts everything down. We'll probably need to experiment here, but killing all role="presentation" is probably a good starting point. 
 - Oh, and Ctrl Shift P doesn't work on really any browser on Windows. Alt Shift P is available, Ctrl Alt P is available, just Alt P, and probably more.

krishnanand5:
Greetings @ellvix,

Foilowing are the changes I did in the recent patch:
- Fixed dialog role conflict by adding explicit `role="dialog"` 
- improved visual contrast with darker borders for highlighted items. 
- Implemented proper ARIA semantics with `role="listbox"` and `role="option"` for natural keyboard navigation, while adding real-time announcements for search filtering and arrow key navigation. 

Regarding accessibility for the overall command picker, I am open to discussion because I do not think the current implementation should be the final one - you can give it a try and let me know if something is off 

We also have the option of integrating the maidr level aria-level that the user configures via settings with this command picker. Unsure if the functionality of command picker allows us to customize the aria live mode based on the user preference set in settings needs to be investigated.

Regards,
Krishna Anandan Ganesan.

ellvix (approved):
Fantastic. Functionality is all great and fully accessible. I like the instructions, the number of commands found shown, and ease of controls. Very intuitive and well made.

krishnanand5:
Greetings @ellvix,

Thank you for reviewing the changes.

@nk46-cloud, Can you check on why the lint job is detecting new lint errors? I am not able to reproduce the lint errors locally with the latest version of main but they are being detected on the lint action

Regards,
Krishna Anandan Ganesan.

nk1408:
@ellvix  , 

1. Implemented proper ARIA combobox semantics on the search input (role/aria-autocomplete/aria-controls/aria-expanded/aria-activedescendant).
2. Deferred highlighting (highlightedIndex = -1 on load/search) so typed text is announced; first option announced only after Arrow navigation.
3. Guarded aria-activedescendant until an option is highlighted; added id="command-list" on listbox and preserved option roles/ids/aria-selected.
4. Removed selection live-region announcements to avoid duplicates; live region now only announces result counts.

Works fine on Voicever with arrow key navigation, Please let me know if that looks good on your end. Thanks

nk1408:
@jooyoungseo , fixed double sonification, please let me know if this is looking good. Thanks

jooyoungseo (approved):

PR #392, opened 2025-07-16
fix: address subplot announcement message

nk1408:
# Pull Request

## Description

Fix `This is a maidr subplot of type...` message for plots having only a single plot to `This is a maidr plot of type...` 

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas.

### DIFF
--- a/src/model/context.ts
+++ b/src/model/context.ts
@@ -139,10 +139,10 @@ export class Context implements Disposable {
     const clickPrompt = includeClickPrompt ? 'Click to activate.' : Constant.EMPTY;
     switch (state.type) {
       case 'figure':
-        return `This is a MAIDR figure containing ${state.size} subplots. ${clickPrompt} Use arrow keys to navigate subplots and press 'ENTER'.`;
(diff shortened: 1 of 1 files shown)


nk1408:
@jooyoungseo , I went through the logic and it looks like this simple fix would work. Here is my understanding: 

The navigation context in MAIDR determines what the user is currently focused on within a visualization. This context can be at three levels:

Figure: The entire visualization, possibly containing multiple subplots (panels).
Subplot: A single panel within the figure, which may contain multiple layers (traces).
Trace : An individual data layer within a subplot.

The context is managed by the Context class, which sets the initial focus and determines the instruction message shown to the user.

For multi-panel figures, the user is at the figure level when choosing between subplots. The "subplot" context is not used here because:

The user is not yet inside any particular subplot.
Navigation into a subplot (by pressing ENTER) goes directly to the first trace (layer) of that subplot, skipping the "subplot" context.
More detailed navigation announcements (e.g., "Subplot 1 of 3: ...") are handled by the TextService, not by the context's instruction logic.

So almost always, in the when subplot context is chosen it is a single plot with multiple layers. I thought about checking number of plots in figure...but that wouldn't be much helpful. Please let me know if this is fine. Thanks

jooyoungseo (approved):

Issue #361, opened 2025-06-20
feat: support high-Contrast Color Mode Toggle for Low-Vision Users


jooyoungseo:
### Feature Request: 

**Description**
We would like to implement a high-contrast color mode to enhance accessibility for low-vision users. This mode can be toggled on and off using a single key binding.

**Proposed Solution**
- **Toggle Key**: Pressing the 'C' key will enable or disable the high-contrast mode.
- **High-Contrast Mode Settings**:
  - **Background Color**: Black
  - **Plot Color**: White
  - **Highlight Color**: Light Green for any selected elements.
- **Multicolor Plots**: In plots with multiple colors (e.g., multi-class plots), all plot elements will be displayed in white, ignoring the original colors.
- **Aesthetic Consistency**: The background will remain black, and all plot markers will be white, with light green highlights for selected items.

**Benefits**
This feature will improve usability for low-vision users by providing a clear and high-contrast visual mode that can be easily toggled on and off.

jooyoungseo:
## ✅ High-Contrast Mode Implementation Complete

I've successfully implemented the high-contrast color mode feature as requested. Here's what has been delivered:

### 🎯 **Core Features**
- **Toggle Key**: Press `C` to enable/disable high-contrast mode
- **Visual Design**: 
  - Black background (#000000)
  - White plot elements (#ffffff)
  - Light green highlights (#90ee90)
- **Multicolor Support**: All plot elements render in white, ignoring original colors
- **Persistent Settings**: User preference saved and restored automatically

### 🏗️ **Implementation Highlights**
- **Architecture Compliant**: Follows strict UI → ViewModel → Service → Core Model pattern
- **Type-Safe**: Full TypeScript implementation with comprehensive error handling
- **Event-Driven**: Real-time updates across all components
- **Well-Tested**: Unit tests covering all functionality
- **Accessible**: Integrated with help system and settings UI

### 🔧 **Technical Details**
- **New Service**: `HighContrastService` manages color transformations
- **Command Integration**: `ToggleHighContrastCommand` handles key binding
- **Settings UI**: Toggle switch added to settings panel (Ctrl+,)
- **Help Integration**: Updated help documentation includes new feature

### 📋 **Files Modified**
- Core services, commands, and UI components updated
- Settings interface extended with `highContrastMode` boolean
- Comprehensive test suite added

### 🎮 **Usage**
1. **Keyboard**: Press `C` to toggle on/off
2. **Settings**: Use Ctrl+, to access settings panel
[...]

PR #406, opened 2025-08-06
feat: support high contrast mode via C toggle key

ellvix:
# Pull Request

## Description

Adds basic high contrast functionality, including multiple level support in settings (defaults to 2 levels (black and white)). 

## Related Issues

Addresses #361 

## Changes Made

Created a toggle function triggered on 'c'. All colored objects in the standard svg structure are analyzed and converted to their closest grayscale color according to the number of specified contrast levels. EG, for black and white, convert everything to either black or white. 

A few exceptions: Many charts use a color very close to white, so we define a custom "is it actually white, or just close" so the chart items and backgrounds don't end up being the same color in high contrast mode. 

## Additional Notes

Still a minor issue with some heatmap charts, where the numeric text on the cell can be the same as the background. 

Still todo: 
 - Convert to the full available color range, rather than just the closest color.
 - Move from highlightService to displayService.


ellvix:
Issues still pending: 
 - Announcement "High contrast mode enabled." (or disabled) needed
 - Multi layer / figure / etc charts break and don't show some of the items. Cause is that the items are nearly white and get grouped with the background and assigned the same (black) color. This is handled for similar cases like bar, and just needs the same treatment for these multi chart types.
 - Stacked, candle, and other charts that have multiple colors for different types currently use the same high contrast color for all (white). Recommended solution: get access to selectors and use them to determine what should be shown, and then spread their colors across the full grayscale spectrum (depending on number of levels user has available).

nk1408:
@jooyoungseo @ellvix , High contrast mode toggle works fine outside the L scope. However, I see this error on console when I toggle high contrast mode

```
factory.ts:267 Uncaught Error: Invalid command name: MOVE_TO_INDEX
create	@	factory.ts:267
(anonymous)	@	keybinding.ts:281

```

jooyoungseo:
@nk1408 Could you please cross the finish line by fixing Claude suggestions?

nk1408:
@jooyoungseo , addressed the comments and merging it

PR #415, opened 2025-08-13
chore: llm version refactoring

nk1408:


src/service/chat.ts
```diff
@@ -81,15 +81,21 @@ abstract class AbstractLlmModel<T> implements LlmModel {
...
 
       const url = request.clientToken
         ? this.getMaidrUrl()
-        : this.getApiUrl(request.apiKey);
+        : this.getApiUrl(request.apiKey, request.version);
 
-      const headers: Record<string, string> = request.clientToken
-        ? { Authentication: `${request.email} ${request.clientToken}` }
-        : { Authorization: `Bearer ${request.apiKey}` };
+      const isGemini = url.includes('generativelanguage.googleapis.com');
```
saaivenk: Abstract classes are supposed to be generic, if we are introducing type information, this can't be extended to other subtypes. If something is needed specific for a type, then make that logic generic enough and push the implementations to the concrete implementations. I would suggest to introduce a getHeaders abstract method and implement them properly based on the extensions.

src/service/chat.ts
```diff
@@ -127,6 +133,7 @@ abstract class AbstractLlmModel<T> implements LlmModel {
     image: string,
     currentText: string,
     message: string,
+    version?: string,
```
saaivenk: I believe we would have this value at all times. Since the user will be selecting from the drop down to activate the llm, we would be having at all times. So this doesn't needs to be undefined

src/service/chat.ts
```diff
@@ -151,9 +158,10 @@ class Gpt extends AbstractLlmModel<GptResponse> {
     image: string,
     currentPositionText: string,
     message: string,
+    version?: string,
```
saaivenk: same as above?

src/service/chat.ts
```diff
@@ -270,8 +280,9 @@ class Gemini extends AbstractLlmModel<GeminiResponse> {
     super(svg, maidr);
   }
 
-  protected getApiUrl(apiKey?: string): string {
-    return `https://generativelanguage.googleapis.com/v1beta/models/gemini-2.0-flash-exp:generateContent?key=${apiKey}`;
+  protected getApiUrl(apiKey?: string, version?: string): string {
+    const model = version ?? 'gemini-2.0-flash';
+    return `https://generativelanguage.googleapis.com/v1beta/models/${model}:generateContent?key=${apiKey}`;
```
saaivenk: Does this works for all cases? is this the right way to access the api? or have they updated it?

src/service/chat.ts
```diff
@@ -284,6 +295,7 @@ class Gemini extends AbstractLlmModel<GeminiResponse> {
     image: string,
     currentPositionText: string,
     message: string,
+    _version?: string,
```
saaivenk: Same as above

src/type/llm.ts
```diff
@@ -2,14 +2,20 @@ import type { Status } from './event';
...
 
-export interface LlmRequest {
-  message: string;
-  customInstruction: string;
-  expertise: string;
-  apiKey?: string;
-  email?: string;
-  clientToken?: string;
-}
+export const GPT_VERSIONS = ['gpt-4o', 'gpt-4o-mini', 'gpt-4.1', 'o1-mini', 'o3', 'o4-mini'] as const;
```
saaivenk: As @nk46-cloud previously mentioned, fetching the apis from the providers is a better approach. Maybe for now, we can start with this

src/type/llm.ts
```diff
@@ -25,3 +31,16 @@ export interface Message {
   timestamp: string;
   status: Status;
 }
+
+// Base request now always includes provider-specific version
+export interface BaseLlmRequest<T extends Llm = Llm> {
```
saaivenk: We can simplify this for now like below.

```
const LLM_VERSION_MAP = {
  GPT: ['gpt-4o', 'gpt-4-vision'],
  CLAUDE: ['claude-3-opus', 'claude-3-sonnet'],
  GEMINI: ['gemini-2.0-pro', 'gemini-2.0-flash'],
} as const;

export type LlmModel = keyof typeof LLM_VERSION_MAP;
export type LlmVersions = typeof LLM_VERSION_MAP;

export interface LlmRequest<M extends LlmModel = LlmModel> {
  model: M;
  version: LlmVersions[M][number];
  message: string;
  customInstruction: string;
  expertise: string;
  apiKey?: string;
  email?: string;
  clientToken?: string;
}
```

src/ui/component/Settings.tsx
```diff
@@ -46,42 +48,57 @@ interface LlmModelSettingRowProps {
...
   onChangeKey,
-}) => (
-  <SettingRow
-    label={modelSettings.name}
-    input={(
-      <Grid container spacing={1} alignItems="center">
-        <Grid size="auto">
+  onChangeVersion,
+}) => {
+  const options = modelKey === 'GPT' ? GPT_VERSIONS : modelKey === 'CLAUDE' ? CLAUDE_VERSIONS : GEMINI_VERSIONS;
```
saaivenk: If we follow this `export type LlmModel = keyof typeof LLM_VERSION_MAP;
export type LlmVersions = typeof LLM_VERSION_MAP;`
we can directly use LlmVersions[model] to fetch the list of available models. Please don't introduce backend conditional logics in UI.

saaivenk (changes requested):
@nk46-cloud Could you please attach the final screenshot after introducing the dropdown box?

nk1408:
image
@SaaiVenkat , PFA, the settings menu after addition of LLM options

src/service/chat.ts
nk1408: fixed in [fd79930](https://github.com/xability/maidr/pull/415/commits/fd79930d427729c51552679de8526b0471de86e8)

src/service/chat.ts
nk1408: fixed in [fd79930](https://github.com/xability/maidr/pull/415/commits/fd79930d427729c51552679de8526b0471de86e8)

src/service/chat.ts
nk1408: fixed in [fd79930](https://github.com/xability/maidr/pull/415/commits/fd79930d427729c51552679de8526b0471de86e8)

src/service/chat.ts
nk1408: fixed in [fd79930](https://github.com/xability/maidr/pull/415/commits/fd79930d427729c51552679de8526b0471de86e8), added api key with header,looks like gemini does not support adding model into payload unlike others

src/ui/component/Settings.tsx
nk1408: fixed in [fd79930](https://github.com/xability/maidr/pull/415/commits/fd79930d427729c51552679de8526b0471de86e8)

src/type/llm.ts
nk1408: fixed in [fd79930](https://github.com/xability/maidr/pull/415/commits/fd79930d427729c51552679de8526b0471de86e8)

saaivenk:
> image @SaaiVenkat , PFA, the settings menu after addition of LLM options

Toggle and the dropdown are redundant to turn on the model. Instead we can have just a dropdown which is by default set to select version/model. When a model/version is selected, we can animate to introduce the password field to the right. Is it possible to do this?

nk1408:
> > image @SaaiVenkat , PFA, the settings menu after addition of LLM options
> 
> Toggle and the dropdown are redundant to turn on the model. Instead we can have just a dropdown which is by default set to select version/model. When a model/version is selected, we can animate to introduce the password field to the right. Is it possible to do this?

yes feasible, In that case we have to set a default option (first option not selected by default) in select boxes....Also I am not really sure about how aria attributes would work when we animate a password field to appear.

PR #422, opened 2025-08-19
fix: address boundary spatial audio bug

nk1408:
# Pull Request

## Description

Boundary audio (played when navigating beyond chart limits) was incorrectly hardcoded to always play on the left side, regardless of whether the user hit the left or right boundary.

RCA

1. Audio service used hardcoded playBoundaryAudio() with fixed values size=1, index=0
2. Abstract trace used safe fallback indices instead of current position for audio state
3. Spatial calculation always resulted in panning = -1 (left side)

Fix

Removed hardcoded boundary audio - now routes all states through handleTraceState()
Fixed audio state generation - uses current position (this.row, this.col) instead of fallback indices

## Changes Made

src/model/abstract.ts - Fixed audio state generation for boundary conditions
src/service/audio.ts - Removed hardcoded boundary audio, unified audio handling

## Checklist

### DIFF
--- a/src/model/abstract.ts
+++ b/src/model/abstract.ts
@@ -224,17 +224,17 @@ export abstract class AbstractTrace<T> extends AbstractObservableElement<T, Trac
 
   public get state(): TraceState {
     if (this.isOutOfBounds) {
-      // Safety check: ensure we don't access undefined values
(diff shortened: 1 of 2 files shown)


nk1408:
fixing copilot comments

jooyoungseo (approved):

PR #446, opened 2025-09-29
feat: add onHover settings for scatterplots

ellvix:
# Pull Request

## Description

 - Functionality for hover on scatterplot. Currently locked to column movement (like you're doing Left / Right arrow keys), but on request we could add a feature to detect pointer movement and do both horz and vert movement. 
 -  Settings radio for off / hover / click modes.

## Changes Made

Added scatterplot functions for hover feature.

Added radio input in settings to toggle the feature: Off, hover mode (default, better for blind users), or click mode (better for visual users).


nk1408:
The Issue:

  The hover mode setting in the Settings menu wasn't taking effect when saved. Users had to reload the page for
  the new hover mode (off/hover/click) to work.

  Root Cause:

  The Mousebindingservice class was reading the hover mode setting only once during initialization and never
  listening for changes. When settings were updated, the service wasn't notified.

  The Fix Applied:

  Made Mousebindingservice implement the Observer pattern:

  1. Added Observer interface to Mousebindingservice class
  2. Registered as observer of SettingsService in constructor
  3. Added update() method that:
    - Detects when hover mode changes
    - Removes old event listeners
    - Adds new event listeners based on new mode
  4. Added proper cleanup in dispose() method

  Key Code Changes:

  - File: /src/service/keybinding.ts
    - Class now implements Observer<Settings> & Disposable
    - Auto-registers with SettingsService: this.settingsService.addObserver(this)
    - update() method dynamically switches between hover modes

aarthy2000 (approved):
Working as expected

Issue #451, opened 2025-10-14
feat: change boxplot starting position to lower left


ellvix:

ellvix:
fixed with PR #455

PR #455, opened 2025-10-21
feat: starting position is always lower left

ellvix:
# Pull Request

## Description

Changed boxplot to always start at lower left.

## Related Issues

#451 

## Changes Made

Changed and fixed this.boxValues order to match visual order, and align with initial 0,0 coordinates


nk1408:
tested the changes and fixed lint (nav starts from lower left for both the example we have), looking good to me!

nk1408 (approved):

PR #527, opened 2026-02-17
fix(help): show braille mode its own shortcuts, not the trace menu

### LINKED ISSUE #524: feat: redo help shortcut menu to pull automatically
# Feature Request

## Motivation

Currently the shortcut table has to be manually added every time a new key and feature is added, which means that has to be remembered, and is often forgotten until much later. 

## Proposed Solution

Instead of manual addition, create a text description for each key that lives alongside the keystroke in keybinding.ts, and redo the help menu to automatically pull all info and build the table. 

## Additional Context

Will need reference on how to structure this for the codebase.

ellvix:
# Pull Request

## Description

This PR was opened in February to move help text into `keybinding.ts`. That refactor landed independently in #597 (`d207bc0`), and the `AnnouncePositionCommand` this branch also carried landed with tests and follow-ups through #647. Rather than close the branch, it has been merged up to `main` and repointed at two gaps that survived — one #597 left behind, one it exposed.

### 1. Braille mode was shown the trace help menu

`HelpService` handed `Scope.BRAILLE` the menu generated for `Scope.TRACE`:

```ts
[Scope.BRAILLE]: traceHelpMenu,
```

`BRAILLE_KEYMAP` is a strict subset of `TRACE_KEYMAP`, so the help menu opened from inside the braille field advertised four shortcuts that are not bound there:

| Key | Advertised in braille mode | Actually bound in `BRAILLE_KEYMAP`? |
|---|---|---|
| `ctrl + shift + p` | Open Command Palette | no |
| `g` | Go To Extrema | no |
| `alt + L` | Toggle Candlestick Reference Comparison | no |
| `ctrl + shift + L` | Choose Candlestick Reference Line | no |

A braille user pressing `ctrl + /` was offered four dead keys. This is the same drift #524 set out to end, relocated from the keymap to the scope mapping — the descriptions were right, but one scope was reading someone else's.

### 2. Escaping label mode hard-returned to trace scope

Fixing the menu made the second problem reachable: `l` is bound in braille mode too, and it now appears in braille's help. But Escape from a label scope was bound to `ToggleScopeCommand` with a fixed target — `Scope.TRACE` from `TRACE_LABEL`, `Scope.SUBPLOT` from `FIGURE_LABEL` — so a braille reader who pressed `l` and then changed their mind landed in trace scope with the cursor still in the braille field.

The pieces to handle this correctly were already present and already used by the *other* exit from a label scope: `enterLabelScope` deliberately leaves the originating scope on the focus stack, `exitLabelScope` reads it back, and `AnnounceCommand.restoreScope` calls it after announcing a label — so `l` → `x` already returned to braille correctly. Only the Escape path bypassed them.

## Related Issues

fixes #524 — #597 did the bulk of it; this closes the last places the help menu could disagree with the keymaps, and adds tests that keep it that way.

## Changes Made

**`src/service/help.ts`** — generate the braille menu from `BRAILLE_KEYMAP` instead of reusing the trace menu, and register braille's `l` label chord in `NESTED_SCOPE_CONFIG` so the label shortcuts it genuinely does reach stay listed.

**`src/command/toggle.ts`, `src/command/factory.ts`** — add `ExitLabelScopeCommand`, carrying the same guard as `AnnounceCommand.restoreScope`, and bind both `DEACTIVATE_*_LABEL_SCOPE` cases to it. `Focus` excludes only the two label scopes, and `syncFocusStack` runs on the trace/subplot transitions, so the stack holds exactly the scope the old constants named when entering from trace or subplot — those paths are unchanged, and only the braille case moves.

**`src/service/keybinding.ts`** — the `CANDLESTICK_DELTA_KEYMAP` comment justified leaving `l` unbound partly by the hard-return the commit above removed. It now keeps only the reason that still stands (the delta layer already announces the axes) and notes that binding it is a design question, not a hazard. The binding itself is untouched.

**`test/service/help.test.ts`** (new, 20 tests) — cross-checks each menu against the keymaps rather than against the generator, so it catches this class of bug instead of restating the implementation:

- every advertised key is bound in that scope, or behind its `l` chord
- every binding with `showInHelp !== false` is listed
- hidden bindings stay out, unless a visible binding shares the key
- each key is listed only once — `generateNestedScopeHelp` dedupes by keymap property name, so a rename on one side would otherwise quietly emit a duplicate row
- braille gets a menu of its own; the transient label scopes still mirror their parent

Reverting the one-line help fix fails 2 of the 20.

**`test/command/scopeRestore.test.ts`** (+6 tests) — both label scopes take the exit path, the three non-label scopes are no-ops, and an explicit regression test asserts `context.toggleScope` is never called directly, which is what would fail against the old hardcoded-`TRACE` implementation.

**Docs** — `.claude/rules/{command,architecture,accessibility}.md` and two agent briefs still told contributors to "add an entry to `src/service/help.ts`", a step #597 removed. They now point at the keybinding `description`, with `helpKey` and `showInHelp` explained. Copilot mirrors regenerated with `npm run sync:copilot`.

**Merge commit** (`c8f2ac8`) — resolves February's conflicts in favour of `main` throughout, since both of this branch's original changes are already present there in a later form, and drops `.claude/settings.local.json`, which `main` now gitignores. The tree at that commit is identical to `main`; everything after it is the work above.

## Checklist

## Additional Notes

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all clean locally — **1684 unit + 61 ESM tests pass** (26 more than `main`) — and `sync-copilot-instructions.mjs --check` reports 11 files in sync. All 14 CI checks are green on `adbff73`, including the Build, Unit Tests, Type Check, E2E and Docs jobs that were skipped in February when Lint Code failed ahead of them.

The original commit (`95e3fcb`) is preserved in this branch's history rather than force-pushed away, so the February work stays attributable even though `main`'s version of it is what ships.


jooyoungseo:
## Still valid? No — superseded on `main`

Both halves of this PR shipped separately while it sat, and `main` has since moved ~440 commits past its base (`681add6` → `0742033`), which is why it now reads `mergeable_state: dirty`.

**The help refactor** landed in #597 (`d207bc0`, 2026-04-13) — "enable automatic keybinding capture in help menu". It solves #524 the same way this PR does, but one layer lower: instead of a parallel `COMMAND_DESCRIPTIONS` map keyed by command name, the description lives on the binding itself.

```ts
// src/service/keybinding.ts today
TOGGLE_SETTINGS: key(`${Platform.ctrl}+,`, 'Open Settings', { helpKey: `${Platform.ctrl} + ,` }),
ACTIVATE_TRACE_LABEL_SCOPE: key(`l`, 'Access Labels', { showInHelp: false }),
```

`description` is a required parameter of `key()`, so a new binding *cannot* be added without one — a stronger guarantee than this PR's `COMMAND_DESCRIPTIONS`, where a missing entry silently drops the command from help. That failure mode was real here: the bot review flagged that `ANNOUNCE_POSITION` was bound in three keymaps with no `COMMAND_DESCRIPTIONS` entry.

**`AnnouncePositionCommand`** is also on `main`, extending `AnnounceCommand`, with `test/command/announce-position.test.ts` covering it and follow-ups through #647. The `braille.values[0].length` crash the bot flagged is not present in the landed version.

The third review point is resolved too — `.claude/settings.local.json` is in `.gitignore` on `main` (line 185) and untracked.

So there is nothing left to rebase. **Recommend closing this PR**, and closing #524 with it — the issue is still open even though #597 fixed it.

---

## One thing #597 missed, now fixed

Auditing the landed implementation against the keymaps turned up a live bug, and it's the exact drift #524 was about — just relocated from the keymap to the scope mapping:

`HelpService` handed `Scope.BRAILLE` the menu generated for `Scope.TRACE`:

```ts
[Scope.BRAILLE]: traceHelpMenu,
```

`BRAILLE_KEYMAP` is a strict subset of `TRACE_KEYMAP`, so help opened from inside the braille field advertised four shortcuts that are not bound there:

| Key | Advertised in braille mode | Actually bound? |
|---|---|---|
| `ctrl + shift + p` | Open Command Palette | no |
| `g` | Go To Extrema | no |
| `alt + L` | Toggle Candlestick Reference Comparison | no |
| `ctrl + shift + L` | Choose Candlestick Reference Line | no |

A braille user pressing `ctrl + /` got four dead keys. The fix generates the braille menu from `BRAILLE_KEYMAP` and registers its `l` label chord in `NESTED_SCOPE_CONFIG`, so the label shortcuts it *does* reach stay listed.

The accompanying tests cross-check each menu against the keymaps rather than against the generator, so they catch this class of bug rather than restating it: every advertised key must be bound in that scope (or behind its `l` chord), every visible binding must be listed, hidden ones must stay out. Reverting the fix fails 2 of the 16.

Pushed to [`claude/maidr-pr-527-review-iwnk3w`](https://github.com/xability/maidr/compare/main...claude/maidr-pr-527-review-iwnk3w) (`788534b`, `f0f70c0`) — `lint:fix`, `type-check`, `build` clean, 1674 + 61 tests green. Happy to open it as a PR if useful; it's independent of this one.

The second commit is docs: `.claude/rules/{command,architecture,accessibility}.md` and two agent briefs still instructed contributors to "add an entry to `src/service/help.ts`", a step that no longer exists. Copilot mirrors regenerated via `npm run sync:copilot`.

---

jooyoungseo:
Both observations addressed — thanks, the first one was a real bug and worth catching.

## 1. Scope desync on `Escape` after `l` — fixed (`0b6814d`)

Confirmed against the code. The pieces to do this correctly were already in place and already used by the *other* way out of a label scope; only the Escape path bypassed them:

- `enterLabelScope` (`display.ts:164`) deliberately does **not** push to `focusStack`, so the originating scope stays on it
- `exitLabelScope` (`display.ts:177`) reads it back, and `Focus = Exclude<Scope, FIGURE_LABEL | TRACE_LABEL>` — `BRAILLE` is a valid member
- `AnnounceCommand.restoreScope` (`describe.ts:66`) already calls it, so `l` → `x` **did** return to braille correctly

So only the abandon-the-chord path was wrong. Added `ExitLabelScopeCommand` carrying the same guard as `restoreScope`, and bound both `DEACTIVATE_*_LABEL_SCOPE` cases to it.

Worth noting for reviewers: this is not a behaviour change for trace or subplot. `syncFocusStack` is called on both transitions (`move.ts:265`, `move.ts:336`), so `focusStack` holds exactly the scope the old constant named. Only the braille case moves, from wrong to right.

Six tests in `test/command/scopeRestore.test.ts`, alongside the existing `restoreScope` ones: exits from both label scopes, no-ops from `TRACE`/`SUBPLOT`/`BRAILLE`, and one asserting `toggleScope` is never called directly so the hardcoded return cannot come back.

You were right that `CANDLESTICK_DELTA_KEYMAP` avoids binding `l` for exactly this hazard (`keybinding.ts:92-97`). That workaround is now unnecessary, but I've left it alone — the comment there also gives a second, independent reason (the delta layer's announcements already spell out the axes), so removing the binding-avoidance is a judgement call for whoever owns that feature, not a mechanical follow-on.

## 2. Dedup matches property name, not hotkey — asserted (`44b98ab`)

Rather than a comment, this is now enforced: each scope must list every key exactly once. A rename on one side of the parent/nested pair starts emitting a duplicate row, and the test fails instead of the coincidence quietly breaking. The rationale you raised is in the test body, so it stays attached to the check rather than drifting from it.

## Verification

You noted the review environment couldn't run the toolchain. Results from a full local run at `44b98ab`:

- `npm run lint:fix` — clean
- `npm run type-check` — clean (`tsc --noEmit` + `tsconfig.ci.json`)
- `npm run build` — all 12 example builds succeed
- `npm test` — **1684** unit + **61** ESM pass (up from 1674 + 61 on `main`: 4 help tests, 6 scope tests)
- `sync-copilot-instructions.mjs --check` — 11 files in sync

---

jooyoungseo:
Both observations acted on — thanks, the first one was a real bug and worth catching.

**1. `l` → `Escape` in braille mode.** Confirmed and fixed in `0b6814d`, rather than deferred to a follow-up: this PR is what puts the `l` chord in the braille help menu, so shipping the advertisement without the fix would newly point braille users at a path that drops them out of braille mode.

Everything needed was already in place and already used by the *other* way out of a label scope — `enterLabelScope` (`display.ts:164`) deliberately leaves the originating scope on the focus stack, `exitLabelScope` (`display.ts:177`) reads it back, and `AnnounceCommand.restoreScope` (`describe.ts:66-70`) calls it after announcing a label. So `l` → `x` already returned to braille correctly; only the Escape path bypassed it.

Added `ExitLabelScopeCommand` carrying the same guard, and bound both `DEACTIVATE_*_LABEL_SCOPE` cases to it. `Focus` is `Exclude<Scope, FIGURE_LABEL | TRACE_LABEL>`, and `syncFocusStack` is called on the trace/subplot transitions (`move.ts:265,336`), so the stack holds exactly the scope the old constants named when entering from trace or subplot — those paths are unchanged, and only the braille case moves, from wrong to right. Six tests in `test/command/scopeRestore.test.ts` cover it, including that `context.toggleScope` is no longer called directly.

**2. Property-name dedup in `generateNestedScopeHelp`.** Turned it into an assertion rather than a comment (`44b98ab`) — a comment documents the coincidence, a test enforces it:

```ts
it.each(HELP_SCOPES)('$scope lists each key once', ...)
```

A rename on one side of a shared command would now fail the suite instead of quietly emitting a duplicate row. The explanation lives in the test body, where it stays next to the thing that checks it.

**On verification** — this run reported it could not execute the test commands. Locally on `44b98ab`: `lint:fix`, `type-check` and `build` clean, **1684 unit + 61 ESM tests pass** (up from 1674 on `main`; +20 help, +6 scope-restore), and `sync-copilot-instructions.mjs --check` reports 11 files in sync.

---

jooyoungseo:
**1. Stale `CANDLESTICK_DELTA_KEYMAP` comment** — agreed, and it's stale *because of* this PR, so it's fixed here rather than deferred (`adbff73`).

One correction to the reasoning though: the guess that the caution might still hold "for a different reason (the candlestick delta layer isn't pushed onto `focusStack` either)" doesn't survive checking. The delta layer is entered through `this.display.toggleFocus(Scope.CANDLESTICK_DELTA)` (`candlestickDelta.ts:244,300`), and `toggleFocus` pushes onto the stack (`display.ts:207-211`). `CANDLESTICK_DELTA` is also a `Focus` — the type only excludes the two label scopes. So a label scope opened from the delta layer would now return to it correctly.

The comment keeps the reason that still stands (the delta layer already announces the axes, so label mode adds nothing) and now says binding `l` there is a design question rather than a hazard. I've left the binding alone — that's a call for whoever owns the feature.

**2. Shared helper for the guard** — skipping this one deliberately.

Moving the `TRACE_LABEL || FIGURE_LABEL` check into `DisplayService.exitLabelScope()` would make the method a no-op outside label mode, and that changes what the existing tests can observe: `scopeRestore.test.ts` currently asserts `expect(displayService.exitLabelScope).not.toHaveBeenCalled()` for `TRACE` and `SUBPLOT` (the pre-existing `AnnounceCommand.restoreScope` cases, not just my new ones). Those assertions would all have to be rewritten to reach inside the service instead. That's a fair amount of churn to pre-existing tests to deduplicate a four-line guard, and it would make the callers' behaviour less legible at the call site, not more.

The duplication is real but cheap, and both copies are covered. If the guard condition ever needs to change, the tests will point at both. Happy to revisit if you'd rather have it centralised.

**Verification on `adbff73`** (this run again reported it couldn't execute `npm`): `lint:fix`, `type-check` and `build` clean, **1684 unit + 61 ESM tests pass**, `sync-copilot-instructions.mjs --check` reports 11 files in sync.

---

PR #530, opened 2026-02-18
fix: remove JAWS/NVDA special char announcement for spacebar

nk1408:
# Pull Request

## Description

Screen readers (NVDA/JAWS) were spelling out the invisible Unicode character U+2063 (INVISIBLE SEPARATOR) that was used to force re-announcements of identical text in aria-live regions. When a user pressed spacebar to re-read the current data point, the screen reader would announce "invisible separator" instead of silently triggering a re-announcement.

Fix

Replaced the invisible character hack with a revision counter + React key-based re-mount pattern:

1. Added a revision counter to the Redux TextState that increments on every update dispatch — including when the text
is identical
2. In Text.tsx, the role="alert" element uses key={revision}, so React unmounts and re-mounts it on every update.
Inserting a fresh role="alert" element into the DOM is the most reliable way to trigger screen reader re-announcements
(recommended by W3C, MDN, and Adobe's react-aria)
3. Separated the visual text display (<p>) from the SR announcement (role="alert" with visually-hidden styling) so
neither duplicates nor flashes. This eliminates the setTimeout timing dependency and all invisible Unicode characters.

## Changes Made

src/state/viewModel/textViewModel.ts — Added revision counter to state; simplified update() method (removed U+2063 and setTimeout)
src/ui/component/Text.tsx — Separated visual and SR elements; added key={revision} on visually-hidden role="alert" div for reliable re-mount
src/controller.ts — Simplified announceInitialInstruction() to use textViewModel.update() instead of notificationService.notify('\u2063')

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -182,3 +182,4 @@ playwright-report/
 
 # Site
 _site/
+test-react-app/
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -250,14 +250,11 @@ export class Controller implements Disposable {
    * Announces the initial instruction to screen readers using a live region.
    */
   public announceInitialInstruction(): void {
-    // Prime the live region with an invisible separator to force a DOM-change event
-    // U+2063: INVISIBLE SEPARATOR (not trimmed by String.trim())
-    this.notificationService.notify('\u2063');
-    setTimeout(() => {
-      this.notificationService.notify(
-        this.displayService.getInstruction(false),
-      );
-    }, 50);
+    const instruction = this.displayService.getInstruction(false);
+    // Use textViewModel.update() so the revision counter is bumped,
+    // which forces the View to re-mount the role="alert" element and
+    // triggers a screen-reader announcement.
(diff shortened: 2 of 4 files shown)


nk1408:
I think the claude bot's comment is a bit irrelevant here:
 
Point 1: key on conditional element . The three cases (falsy→truthy, truthy+revision change, truthy→falsy→truthy) all work as designed.

Point 2: announceInitialInstruction() semantics change - announceInitialInstruction() is dead code. It is defined as a public method but never called anywhere in the codebase.The main flow uses showInitialInstructionInText() instead.  No race condition exists.

Point 3: update() always dispatches clearMessage() — Existing trade-off, not a regression. The old code already dispatched clearMessage() unconditionally in the else branch (different text). The only case where it wasn't called was during the \u2063 same-text priming phase (between the two dispatches). Now it's always called, which matches the existing different-text behavior.

jooyoungseo (approved):

Issue #535, opened 2026-02-22
feat: support Plotly box plot per-stat highlighting and navigation


jooyoungseo:
## Context

py-maidr now supports Plotly box plots (xability/py-maidr#264). Between-box navigation (ArrowRight/Left for vertical, ArrowUp/Down for horizontal) works correctly with text readout. However, within-box stat navigation and per-stat visual highlighting do not work.

## Problem

Plotly renders each box plot as a **single `<path class="box">`** element that draws the entire box (whiskers, IQ box, median line) in one SVG path command:

```html
<g class="trace boxes">
  <path class="box" d="M...whisker...M...box...M...median..."/>
  <g class="points"><!-- outliers --></g>
</g>
```

By contrast, matplotlib renders each box part as a **separate SVG element** with unique IDs:
- Median line: `<path id="maidr-uuid-1" .../>`
- IQ box: `<path id="maidr-uuid-2" .../>`  
- Min/max whiskers: separate `<path>` elements
- Outlier points: separate `<use>` elements

MAIDR JS's box plot handler expects to individually target and highlight each part via CSS selectors in the structured `selectors` format:
```json
{
  "min": "g[id='maidr-...'] > path",
  "q2": "g[id='maidr-...'] > path", 
  "iq": "g[id='maidr-...'] > path",
  ...
}
```

Since all Plotly box parts are one `<path>`, the selectors all point to the same element, and ArrowDown within a box shows "No plot info to display".

## Current behavior

- **Works**: ArrowRight/Left navigates between boxes with correct text readout ("Group is G1, no lower outlier(s) for Value")
- **Works**: Orientation detection (`"horz"` for horizontal boxes)
- **Broken**: ArrowDown/Up within a box (stat-by-stat navigation) shows "No plot info to display"
- **Broken**: No per-stat visual highlighting (all parts are one path)

## Possible solutions

1. **SVG overlay approach**: MAIDR JS could draw separate transparent overlay elements (rect for IQ box, lines for whiskers/median) on top of the Plotly box path, positioned using the data values
2. **Stat navigation without highlighting**: MAIDR JS could navigate stats based on data alone without requiring matching SVG elements
3. **Plotly-specific rendering**: Use Plotly's JS API to programmatically highlight by changing trace properties

Related: xability/maidr#533 (scatter), xability/maidr#534 (heatmap)

nk1408:
Closed with #577

PR #553, opened 2026-02-24
fix: support per-stat highlighting for Plotly single-path box plots (#535)

jooyoungseo:
Plotly renders each box plot as a single <path> element containing
whiskers, IQ box, and median line, whereas MAIDR expects separate SVG
elements for each statistical component. This caused within-box stat
navigation and per-stat visual highlighting to fail for Plotly box plots.

Add detection for single-path box plots (where all BoxSelector fields
resolve to the same DOM element) and create synthetic SVG line overlays
at data-driven positions using bounding box coordinate mapping. The
existing multi-element code path (matplotlib, etc.) is preserved and
only activates when 2+ non-null selectors resolve to different elements.

Changes:
- Add isSinglePathBox() detection method to BoxTrace
- Add createDataDrivenOverlays() for data-to-SVG coordinate mapping
- Add Svg.createPositionedLineElement() utility for positioned overlays
- Add Plotly-style vertical box plot example

https://claude.ai/code/session_012wiRFDsjMCepsrpfVaPJza

### DIFF
--- a/examples/boxplot-plotly-horizontal.html
+++ b/examples/boxplot-plotly-horizontal.html
@@ -1,0 +1,277 @@
+<!doctype html>
+<html lang="en">
+
+<head>
+  <meta charset="utf-8" />
+  <title>Plotly-style Horizontal Box Plot Example</title>
+  <script>
+    var maidr = {
+  "id": "plotly-boxplot-horizontal",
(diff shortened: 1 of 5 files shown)


nk1408:
Addressed via #591

PR #563, opened 2026-02-25
feat: go to intersection

soundaryakp1999:
# Pull Request

## Description

Implemented intersections in the go to feature, accessible with the g key. In a multiline plot, the g key in addition to the extremas, now lists all the points where the currently selected line intersects another line.
## Changes Made

- Only in multiline plot: when g is pressed on a line, all points are listed where other line(s) intersect.
- This scans and displays floating point values, as the exact intersection was often skipped by just considering integers.
## Checklist


nk1408:
I tested the feature, It found the intersaction points (same x & y value) but the issue is we dont have that exact point as svg element in the image. So goto doesn't exactly land at the intersection point as specified in the options (it lands at the closest available svg point). I think we may have to re-think a bit of how we have to design this. We can discuss further in tomorrow's call.

nk1408:
The navigation model is discrete (index-based: this.col = N), so it can only land on actual data points. Geometric intersections occur between data points and there's no mechanism to represent or navigate to an intermediate position.

We can have two possible solutions:

1. Remove geometric intersections from "Go To" — Only show intersections that fall exactly on data points (use the existing findIntersections() at line.ts:300). This means fewer (possibly zero) intersection targets, but each one would be accurate.
2. Add interpolated navigation support — Create a virtual data point at the intersection coordinates so the cursor can actually land there. This is a larger change requiring updates to the highlight system, text/audio state, and the navigation model.

Issue #529, opened 2026-02-17
feat: grid movement


ellvix:
# Feature Request

For scatterplot only, create a grid cell style movement system. Users will set grid with a tick size and min/max in schema, and then be able to move cell by cell just like heatmap to see data in various areas of the chart. 

## Motivation

There is currently no way for blind users to get a good idea of specific data about various areas of the chart. A grid cell style movement would allow them to explore in a new way and get a better understanding of the data. 

## Proposed Solution

 - Implemented for scatterplot only, though possibly more plot types could follow later. 
 - User will set tick size x and y separately in schema, along with max/min values x and y (check if these exist).
 - A key will toggle grid movement on and off, discussion needed for specific key.
 - Arrow keys will move between grid cells, and play a run of tones for all points that are contained. Movement is similar to heatmap movement, tones are similar to current scatterplot sonification where they can be played all at once or in a run with slight separation.

ellvix:
Change schema, do like: 

Axes.x.min / max / tickStep, and for compatability do both Axes.x.min and Axes.min.x (for all). Not grid as separate / sibling, but part of axes.

Change rotor behavior in scatter: higher / lower value doesn't exist, and 'DATA POINT NAVIGATION' isn't accurate, try 'ROW AND COL NAVIGATION'.

Text to display: 

Verbose: Sepal Width is 2.0 through 3.0, Sepal Length is 2.5 through 3.0, points are: list points
Terse: 2.0 through 3.0, 2.5 through 3.0, points: list points
Fall back to 'x is 2.0 through 3.0' x and y if labels don't exist

Update position feature for scatter when grid mode active, to use 'x is 2.0 through 3.0' etc verbose / terse. Maybe do this as a new issue so it can be done after Position PR is done

PR #570, opened 2026-03-10
feat: support grid based movement for scatterplots

ellvix:
# Pull Request

## Description

Introduces a new navigation mode for scatter plots that divides the plot area into a configurable grid of rectangular cells. Users cycle to "GRID NAVIGATION" via the rotor (Alt+Shift+Up/Down), then use arrow keys to move between cells. Each cell announces its axis ranges and the points it contains.

## Related Issues

#529 

## Changes Made

- Grid config lives in the axes schema (axes.x.min, axes.x.max, axes.x.tickStep, etc.) with dual-format support (axes.min.x also works)
- Rotor modes are now dynamic per trace type — scatter shows "ROW AND COLUMN NAVIGATION" + "GRID NAVIGATION", other plots show "DATA POINT NAVIGATION" + "LOWER/HIGHER VALUE NAVIGATION"
- Text announcements for grid cells show axis ranges and point coordinates in both verbose and terse modes
- Position notification (P key) shows axis ranges when in grid mode
- Visual highlighting works for all points within the active grid cell


soundaryakp1999:
Hi,
Thank you for implementing. The grid navigation works as expected.
There are 2 bugs:
1. When the grid hits the edge, for upper and lower edge the message is unclear.
Left: No grid value found to the left of the current value.
Right: No grid value found to the right of the current value.
Up: No grid value found the current value.
Down: No grid value found the current value. 
2. When the user switches from row and column navigation to grid navigation and vice versa, there is an inconsistency.
i. While switching to grid navigation, there is immediate sonification plus the screen reader first speaks the data point and then it says "Grid navigation". It should either be the opposite, or the data points should not be spoken and the sonification should not be played.
ii. While switching to row and column navigation, it just says "Row and column navigation".
For consistency, Kindly consider making the behavior of both modes similar. Personally I suggest to prefer the behavior of point ii.

ellvix:
Fixed both issues @soundaryakp1999 can you check again? 

Also fixed sync issue and abstract.ts clogging that Claude identified.

nk1408:
@jooyoungseo updated examples for scatter and smooth+kde plots, and resolved merge conflict

soundaryakp1999:
Thank you @ellvix . Both bugs I had mentioned have been fixed. Functionally, the behavior is consistent with the PR description.

nk1408:
Comment 1 — on src/service/rotor.ts line 232 (the setMode method):
snake_case variables (`curr_mode`, `nav_type`) are inconsistent with the camelCase convention used throughout the codebase. Rename to `currMode` and `navType`.

Comment 2 — on src/service/rotor.ts line 341 (the moveGrid method):

Initialize `moved` to `false`:
let moved = false; TypeScript handles the exhaustive switch correctly so this won't cause a compile error, but it's better defensive practice and makes the intent explicit.

Comment 3 — on src/model/scatter.ts line 417 (the moveOnce method, the if (this.isInGridMode) block):

This silently exits grid mode without any signal to developers that an unexpected path was hit. Add a console.warn so this is detectable during development: console.warn('[ScatterTrace] moveOnce() called while in grid mode — exiting grid mode. Rotor index may be out of sync.');

The model can't reset the rotor index directly (that would violate MVVC layer boundaries), so a warning is the right approach here.

Nitpicks

1. **Braille `min: 0, max: 0` for grid cells** — The non-grid scatter braille path already returns `min: 0, max: 0`. This PR is consistent with existing behavior. If this should change, it should be a separate issue for all scatter braille.

2. **Points outside grid bounds silently dropped** — Grid bounds are user-configured via `axes.x.min/max`. Dropping out-of-range points is expected 
  behavior, not data loss.

3. **Hardcoded `'x'`/`'y'` fallbacks** — These match `Constant.X`/`Constant.Y` already. The risk of divergence on single-character axis labels is 
  negligible.

4. **`Math.round` in `computeGridSteps`** — The suggested `Math.ceil` would produce worse results (a near-empty sliver bin). The current code handles the last bin explicitly with `i === numBins - 1 ? axisMax : ...`, which is correct.

PR #592, opened 2026-03-29
feat: description modal with d key

### LINKED ISSUE #566: feat: chart description
# Feature Request

## Motivation

Users need a way to easily find out information, and have information available in multiple pathways.

## Proposed Solution

Create a modal with a text description and data table, triggered by D key. Use headings so users can jump around easily. 

Text description: Describe whatever is objectively possible using schema (no AI), start high level and get more granular:
 - type and title
 - then labels / axes / fill / min max / num data
 - then actual data (where appropriate, like bar is probably useful scatter probably not)

Consider this well for each plot.

Data table will be a literal table of all plot data. Figure out how to do different plot types, multi panel / figure etc.

ellvix:
# Pull Request

## Description

Adds a new Chart Description modal triggered by pressing D while navigating a chart, providing objective, data-driven metadata about the current trace. Description includes chart type, title, axes labels, summary statistics (min/max, counts), and a scrollable data table of the raw values.

## Related Issues

Fixes #566 

## Changes Made

Created DescriptionService and DescriptionState to fetch data from JSON schema, and a new UI component to render data.


soundaryakp1999:
This is an excellent feature. The format of quick description followed by tabular representation is effective. Please find a few specific concerns:
1. Fill inconsistency in dodged and stacked barplots as demonstrated by the following ind dodged:
FILL: unavailable
Fill categories: Below, Above
2. Heatmap does not mention the fill axis.
3. Minor: In histogram, I suggest to first display the min and max bins before the frequency, that is, frequency should be the last column.

soundaryakp1999:
This feature seems to be correctly and consistently implemented. i have a few suggestions, none of them are blockers:
1. Suggestion: write a quick description for facet and multipanel plots. Users might press D the first thing they gain focus of MAIDR, so it might be useful if at least a quick description mentions what are the subplots available in this panel.
2. Suggestion : Its possible to reduce number of rows in multiline plot description table. Let each x value be a single record. Mention all y values of lines in subsequent columns, instead of duplicating the x values for different lines.
3. Suggestion: Add list of intersections to summary in multiline plot. Please note: We consider only point intersections, not slope intersections I.E. both points should be present in the SVG, and the intersection should not be formed by line segments drawn between nearing points.
4. When z label is not defined AKA level, description does not mention the z axis. This seems correct but I wanted to mention it for reference.

ellvix:
Addresses suggestion 1 from Soundarya and Claude review

nk1408:
working very nicely!

PR #593, opened 2026-04-01
fix: escape key works in braille mode for subplot exit

soundaryakp1999:
# Pull Request

## Description

Escape key now exits the currently opened subplot and focuses on the list of subplots in multipanel and facet plots even if braille mode is on.
## Related Issues

Fixes https://github.com/xability/maidr/issues/574
## Changes Made

Earlier behavior: In multipanel and facet plots, the escape key was used to exit the currently opened subplot. This did not work with braille mode on. This PR fixes the issue: Escape exits the subplot and focuses on the subplot list even if braille mode is on. Further, it preserves the state of braille mode along with sonification and text I.E. If the user exits the subplot while braille mode is on and opens a new subplot, braille representation of the new subplot is displayed.
## Checklist

- [x] I have read the [Contributor Guidelines](../CONTRIBUTING.md).
- [x] I have performed a self-review of my own code and ensured it follows the project's coding standards.
- [x] I have tested the changes locally following `ManualTestingProcess.md`, and all tests related to this pull request pass.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation, if applicable.
- [ ] I have added appropriate unit tests, if applicable.

PR #596, opened 2026-04-09
fix: braille cursor issue and extra braille text at beginning in nvda eliminated

soundaryakp1999:
## Description

Moves `role="application"` from the `<textarea>` to the wrapper `<div>` in the Braille component to fix screen reader regressions.

## Related Issues

Follow-up to #593

## Changes Made

The `role="application"` attribute was previously placed directly on the braille `<textarea>`, which caused two regressions:

1. **Screen reader announcement:** NVDA announced "edit braille app" before the braille content instead of treating it as a plain edit field.
2. **Cursor movement:** Arrow keys (e.g. right arrow) did not move the cursor within the braille textarea, because `role="application"` on the textarea itself overrides its native form control behavior.

Moving `role="application"` to the parent `<div>` fixes both issues. The textarea retains its native editing behavior (cursor movement, standard announcements), while the parent's application mode still ensures keys like Escape are passed through to the web app rather than being intercepted by the screen reader.

**Before:** `role="application"` on `<textarea>` → "edit braille app" announced, cursor movement broken
**After:** `role="application"` on wrapper `<div>` → textarea announced normally, cursor works, Escape still passes through in one press

## Checklist


soundaryakp1999:
All major suggestions by claudebot have been implemented. I have tested the PR locally. @nk1408 Could you please test if visually braille mode is as before? @jooyoungseo  @nk1408  I believe its ready to merge.

nk1408:
yes @soundaryakp1999 , looks good to me visually. Merging

PR #609, opened 2026-05-04
fix: harden altair adapter pipeline across line, box, count, dodged plots

nk1408:
# Pull Request

## Description

This change set hardens the Vega-Lite adapter (and its underlying multi-line highlight pipeline) against the full range of Altair-generated specs that py-maidr now ships in its docs. The work fixes five distinct bugs surfaced during py-maidr's Altair-adapter testing cycle and adds one helper plus one new visual-order pass.

## Changes Made

1. mapViaDomElements: pass shouldClone=false

  - const elements = Svg.selectAllElements(selectors[r]);
  + const elements = Svg.selectAllElements(selectors[r], false);

Motive: the default cloning side-effect inserts a hidden clone after every match, shifting :nth-child(N) indices across iterations. A multi-series line chart with sibling <path> elements ended up with iteration r=1 matching the clone of path0 (instead of path1) and r=2 matching the clone-of-clone of path0, etc. — every series' highlight collapsed onto series 0. 

Highlights re-clone via Svg.createHighlightElement anyway, so the pre-cloned hidden elements are unnecessary.

mapViaPathParsing: branch on selector uniqueness

const uniqueSelectors = new Set(selectors).size === selectors.length;
for (let r = 0; r < selectors.length; r++) {
  const lineElement = uniqueSelectors
    ? Svg.selectElement(selectors[r], false)            // standard: 1 unique selector per series
    : Svg.selectNthElement(selectors[0], r, false);     // color-hue: shared selector, walk by index
  ...
}

Motive: matplotlib/seaborn multi-line plots use unique selectors per series (g[id='maidr-<uuid-A>'] path, g[id='maidr-<uuid-B>'] path, …). Vega-Lite's color-encoded multi-series uses identical selectors for every series (each series is a separate .layer_0_marks group with its own single <path>). Using selectNthElement(selectors[r],  r) for both was wrong: for unique selectors with one match each, r ≥ 1 returned null and crashed the highlight pipeline downstream (TypeError: Invalid element provided  for highlight creation in highlight.ts:107). The branch picks the right strategy at runtime.

2. src/util/svg.ts — Add Svg.selectNthElement<T> helper

public static selectNthElement<T extends SVGElement>(
  query: string,
  n: number,
  shouldClone: boolean = false,
): T | null {
  const all = document.querySelectorAll<T>(query);
  return all[n] ?? null;
}

Motive: When a single CSS selector matches multiple sibling elements (e.g. multi-series Vega-Lite line charts where each series renders as a separate <g class="mark-line  role-mark layer_0_marks"> group, all with the same class), :nth-child(N) cannot address the Nth match (each path is its parent's only child) and :nth-of-type(N) would
also count unrelated <g> siblings (axes, legend). This helper does the indexing in JS, returning the live DOM element with no cloning side-effects.

 3. src/adapters/vegalite/selectors.ts — Cover Vega's mark-sugar expansion + drop wrong :nth-child

buildSelector: comma-selector covers both DOM patterns

For non-layered specs, emit g.<class>.role-mark.marks <child>, g.<class>.role-mark.layer_0_marks <child>.

Motive: Vega-Lite expands mark sugar (mark.point: true on lines, mark.line: true on areas, etc.) into additional render-time layers, so even a single user-visible mark may produce a .layer_0_marks group instead of .marks. Trying both patterns in one selector avoids encoding every Vega expansion in the converter.

buildLineSelectors: stop using :nth-child(N), emit one selector per series

For non-layered specs, emit N copies of the same selector and let the caller (LineTrace.mapViaPathParsing) resolve the Nth series via Svg.selectNthElement.

Motive: When the chart has N color-encoded series, Vega creates N sibling groups each with exactly one <path> — :nth-child(2) etc. never match (each path is the only child of its group). Document-order indexing via selectNthElement is the only way to address each series in this layout.

4. src/adapters/vegalite/converters.ts — Five data-extraction fixes

4a. coalesceSiblingLineLayers (new)

Merges runs of consecutive LINE layers with matching axes into a single multi-series LINE layer.
Motive: Altair compiles transform_density + multi-group encodings into separate layer: objects (one mark per group) instead of using a color encoding. Without coalescing, maidr core saw N independent LINE traces and the user could only navigate within one series at a time. SCATTER + LINE pairs (regplot-style) are NOT collapsed because the predicate skips non-LINE marks.

4b. extractBarData / extractSegmentedData: read synthetic __count field

y: Number((row as Record<string, unknown>).__count ?? row[yField] ?? 0)
Motive: Vega-Lite emits a synthetic __count field whenever a channel uses aggregate: 'count' without an explicit field. extractHistogramData already handled this; bar/segmented extraction did not, so count-only bar plots silently extracted 0 for every y.

4c. extractBoxData: fix catField for horizontal orientation

const catField = encoding.x?.type === 'quantitative'
  ? (encoding.y?.field ?? 'y')
  : (encoding.x?.field ?? 'x');
Motive: Previously the code unconditionally took encoding.x.field as the category. Horizontal Iris boxplots (x=quant petalLength, y=nominal species) ended up grouping by 43 unique petalLength values instead of 3 species, producing a BOX BUILD count-mismatch SKIP.

4d. extractBoxData: dedup joinaggregate rows + classify outliers

Replaces the naive rows.map(...) (which produced one BoxPoint per raw row with empty outlier arrays) with a Map<category, BoxAccumulator> that:
- deduplicates rows by category (Vega-Lite's joinaggregate denormalizes the dataset, replicating the same stats on every raw row),
- classifies each raw value as a lower/upper outlier when it falls outside the whisker bounds,
- preserves first-seen category order so highlight indices stay aligned.

Motive: Pre-computed boxplots emitted by Vega had N BoxPoint entries per category (one per raw row), with empty lowerOutliers/upperOutliers. py-maidr's BoxTrace expected one BoxPoint per category and showed wrong outlier counts.

4e. convertLayerSpec (BOX case): set Orientation.HORIZONTAL

if (xIsQuant && !yIsQuant) {
  orientation = Orientation.HORIZONTAL;
}
Motive: Without the orientation field, horizontal boxplots defaulted to vertical-mode navigation (species-to-species via left/right) instead of within-quartile navigation (up/down), and the screen reader announced the wrong axis labels.

5. src/vegalite-entry.ts — reorderDodgedBarsByVisualPosition (new)

Mirror of reorderSegmentedSeriesByVisualBottom but applies a horizontal (leftmost-first) sort for DODGED layers; ensures data[0] is the visually leftmost subgroup.

Motive: Vega-Lite's xOffset encoding draws subgroups left-to-right by xOffset scale-domain order, but the dataset row order driving extractSegmentedData's grouping is independent of that scale order. Without re-sorting, MAIDR's row=0 initial focus landed on whichever fill value happened to appear first in the data rows, not the leftmost-rendered series — confusing for sighted/blind co-exploration where "first" should mean "leftmost".

The pass handles both DOM layouts:
- Subject-major (order: 'column', default for DODGED): the first N DOM elements are the first x-category, one per series;
- Series-major (order: 'row'): each series occupies a contiguous block of size categoryCount.

A try/catch wraps the whole body so a runtime exception here cannot abort the rest of the visual-order pipeline (notably buildBoxPlotSelectorsFromDom).

## Checklist

### DIFF
--- a/examples/vegalite-bindbox-horizontal.html
+++ b/examples/vegalite-bindbox-horizontal.html
@@ -1,0 +1,84 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <title>MAIDR + Vega-Lite Horizontal Box Plot Example</title>
+    <!-- Vega-Lite and its dependencies -->
+    <script src="https://cdn.jsdelivr.net/npm/vega@5"></script>
+    <script src="https://cdn.jsdelivr.net/npm/vega-lite@5"></script>
+    <script src="https://cdn.jsdelivr.net/npm/vega-embed@6"></script>
+    <!-- MAIDR Vega-Lite adapter (bundles MAIDR core) -->
+    <script src="../dist/vegalite.js"></script>
+  </head>
+  <body>
+    <!--
+      This example validates the horizontal orientation fix for boxplots.
+      When x is quantitative and y is nominal, the converter should set
+      orientation = Orientation.HORIZONTAL so that announcements and
+      DOM pairing are correct.
+    -->
+    <h1>MAIDR + Vega-Lite Horizontal Box Plot</h1>
+    <p>Click on the chart and use arrow keys to navigate. Press 'b' for braille, 't' for text descriptions.</p>
+    <div
+      id="box-chart-horizontal"
+      aria-label="Horizontal box plot loading"
+      style="min-height: 400px"
+    ></div>
+
+    <script>
+      // Horizontal box plot: x is quantitative (score), y is nominal (group).
+      // This spec should trigger orientation = Orientation.HORIZONTAL in
+      // convertLayerSpec, ensuring correct axis labels in announcements.
+      const spec = {
+        $schema: 'https://vega.github.io/schema/vega-lite/v5.json',
+        width: 600,
+        height: 400,
+        title: 'Score Distribution by Group (Horizontal)',
+        description: 'Horizontal box plot showing score distributions.',
+        data: {
+          values: [
+            // Group A: tight bulk 60–80 + 2 lower & 2 upper outliers
+            ...Array.from({ length: 28 }, () => ({
+              group: 'A',
+              score: 60 + Math.round(Math.random() * 20),
+            })),
+            { group: 'A', score: 30 },
+            { group: 'A', score: 35 },
+            { group: 'A', score: 95 },
+            { group: 'A', score: 98 },
+            // Group B: tight bulk 65–75 + 1 lower & 2 upper outliers
+            ...Array.from({ length: 28 }, () => ({
+              group: 'B',
+              score: 65 + Math.round(Math.random() * 10),
+            })),
+            { group: 'B', score: 40 },
+            { group: 'B', score: 92 },
+            { group: 'B', score: 96 },
+            // Group C: bulk 50–70 + 2 lower & 1 upper outlier
+            ...Array.from({ length: 28 }, () => ({
+              group: 'C',
+              score: 50 + Math.round(Math.random() * 20),
+            })),
+            { group: 'C', score: 20 },
+            { group: 'C', score: 25 },
+            { group: 'C', score: 100 },
+          ],
+        },
+        mark: { type: 'boxplot', extent: 'min-max' },
+        encoding: {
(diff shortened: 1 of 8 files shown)


nk1408:
@jooyoungseo , After an iteration I think claude's comment is trivial

PR #617, opened 2026-05-19
Feat point nav scatter

### LINKED ISSUE #610: feat: point nav mode in scatter
Duplicate the intersection feature for multiline (hit Enter, then arrows move through points in this intersection, Esc to back out), so that when you're on multiple points in Scatter, hit Enter and then left right moves through individual points. Echo still there, but now you can hear the point you're on.

ellvix:
# Pull Request

## Description

Adds 2 rotor modes: 
 - Point navigation. Arrow keys now move point to point, then wrap to the next row/col until they reach a corner. 
 - Intersection mode. Just like multiline, your current set (of row/col) now is its own little thing that you can explore with left/right arrow keys. 

## Related Issues

Fixes #610


soundaryakp1999:
Hello,
The feature is well implemented. Kindly look at the issues I detected regarding autoplay with ctrl+shift+left and right arrow:
a. In point mode, sometimes there is no autoplay, sometimes there is a repeatition of the same point. Different tunes need to be played based on different points.
b. In intersecting point mode, if at a certain point there is no intersecting point to the left or right, autoplay still plays some tunes. No tune should be played and no movement should happen.
c. I am not sure if this is related to this pr, but if I press enter in any grid and explore points, autoplay there does not work.

jooyoungseo:
## Rebased onto `main` and the reported autoplay issues traced to root cause

This branch was 117 commits behind `main` and no longer merged cleanly. It is now merged up to date, with two commits on top:

- `c0c2bcb` — merge `main`, resolve four conflicts
- `e4a6153` — `fix(scatter): keep the rotor cursor and its announcements in step`

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass — **1978 tests, 151 suites, 0 failures** (12 of them new here).

---

### @soundaryakp1999 — your three reports

**(a) Point mode: "sometimes no autoplay, sometimes a repetition of the same point"** — reproduced and fixed.

`ScatterTrace.moveOnce` tests `isInitialEntry` *before* the mode branches:

```ts
public override moveOnce(direction: MovableDirection): boolean {
  if (this.isInitialEntry) {        // ← taken first, whatever the rotor mode
    this.handleInitialEntry();      // ← resets row/col/mode, NOT pointModeIndex
    this.notifyStateUpdate();       // ← so this re-emits the current point
    return true;
  }
```

Manual arrows in point mode go through `RotorNavigationService.movePoint`, which never touches that flag. So the first autoplay tick took the initial-entry branch: it notified without moving — the same tone and the same announcement a second time — and the step was lost. Over data `[(1,5),(2,5),(3,5),(1,2),(3,8)]`, walking right twice then pressing <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>→</kbd> re-announced `X=3, Y=5` with byte-identical audio state before tick 2 moved on.

Entering a rotor mode seeds the cursor on a real point, so `setPointMode(true)` / `setIntersectionMode(true)` now consume the handshake there. (Assigning the flag directly rather than calling `handleInitialEntry()` — that would reset `row`/`col`/`mode` and re-anchor the stack intersection mode walks.)

**(b) Intersection mode: "no intersecting point left or right, but autoplay still plays tunes"** — reproduced and fixed, and the cause was bigger than intersection mode.

A rotor mode is `rotorIndex` on the service but a **boolean on the trace**, and nothing resynced the two when the trace changed underneath them. After <kbd>PageUp</kbd>/<kbd>PageDown</kbd> to another layer, or a live data update rebuilding the figure:

- the rotor still reported `INTERSECTING POINT NAVIGATION` and `isRotorEnabled()` was still true,
- but the new trace's `isInIntersectionMode` was `false`,
- so right-arrow announced "no intersection to the right" at **every** point, and <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>→</kbd> fell past the intersection branch into plain column navigation — moving and sounding exactly where you said nothing should happen.

Under `POINT NAVIGATION` the same swap was worse: every arrow key returned with no move, no tone, no braille and no message at all — a completely dead, silent keyboard with no way to discover that cycling the rotor would fix it.

Both trace-switch commands and `Controller.updateData` now hand the rotor back to data mode *while the outgoing trace is still active*, so its flag is cleared with it. This is the fix `rotor.ts` already described as needed:

> Known limitation: rotorIndex is not reset when the active plot/trace changes […] Resetting the rotor on context change is a broader UX decision tracked separately from this file.

It was hard to reach before, because `GRID_MODE` needs a grid config and `INTERSECTION_MODE` needs a stack. This PR adds `POINT_MODE` to **every** scatter with data, which is what makes it routine — so it belongs here rather than in a follow-up.

**(c) "Enter in a grid, explore points, autoplay doesn't work"** — confirmed real, but **pre-existing on `main` and not caused by this PR**. `moveOnceInGridMode` moves the *grid* cursor via `moveGridLeft/Right/Up/Down` and never checks `isInGridCellMode`, so autoplay drives the cell selection instead of the points inside the cell. Identical code on `origin/main`:

```ts
// origin/main:src/model/scatter.ts
if (this.isInGridMode && this.gridCells) {
  return this.moveOnceInGridMode(direction);   // grid cursor, cell mode ignored
}
```

`isMovable` has the same gap. I have deliberately left it out of this PR to keep the change one logical thing — happy to open a separate issue for it if you'd like.

---

### Answering the earlier automated review

| Item | Outcome |
|---|---|
| `panY = 0` in COL intersection audio | **Fixed.** `panY = isCol ? idx : this.row`. Announce Position said "row 1 of 3" at every step of a 3-point stack; it now walks `0, 1, 2`. |
| Mode transition enables before clearing | **Fixed.** `setMode` clears all three modes, then enables the active one. The getters resolve intersection before point, so a trace briefly in both would read the point cursor through the intersection branch. |
| Float equality in point lookups | **Addressed structurally.** Each `FlatPoint` now carries `xIndex` / `yIndexInColumn`, resolved once at construction from the same numbers `xPoints` was grouped from. The per-keystroke audio path no longer scans an array or compares two floats at all. |
| No POINT_MODE unit tests | **Added.** Entry handshake, the `columnOrder` (up/down) axis that had no coverage at all, position reporting, boundary panning, rotor `POINT_MODE` dispatch, and the trace-switch reset. Each was checked to fail without its fix. |
| `instanceof` vs interface guard | **Left as is.** `setIntersectionMode` is declared on `AbstractTrace`, so `instanceof` is the correct guard and matches the sibling `moveIntersection` right above it. `isPointNavigable` is right for `PointNavigable`, which is not on the base class. |
| Audit `panning.x` consumers | **Done — and the widening is gone.** `x: number \| number[]` broke code added on `main` since this branch (`playEmptyTone`, the pie dial-pan tests) and forced every single-pan call site through a narrowing helper. Replaced with non-breaking `x: number` plus optional `xPerTone?: number[]`; the chord path falls back to `x` per tone. |

---

### Two more defects found and fixed

**Announce Position could not tell points apart in point mode.** Panning was pinned to `{y: 0, rows: 1}` with `x` indexed over *unique* x values, so `(0,1)` and `(0,2)` both announced "Column 1 of 2, row 1 of 1" — in the one mode whose entire purpose is walking individual points. Each point now reports its place within its own x-column. Stereo placement is untouched; only `panning.x` reaches the oscillator.

**The boundary chime pointed at the wrong edge.** `outOfBoundsState` builds from `row`/`col`, which neither new mode maintains. Walking point mode to the right-hand edge panned every data tone right and then chimed **hard left**, contradicting the spatial cue it had just established. Both modes now reuse their own panning for the bounds state.

---

### Merge conflict resolution

All four were additive — `main` grew rotor filter units, the candlestick zero-click and glide tones, and grid-cell navigation while this branch grew point and intersection modes — so both sides were kept in each. `package-lock.json` was restored to `main`'s copy; this PR adds no dependencies.

---

jooyoungseo:
## Superseded by #619 — closing

While this branch was being brought up to date, #619 merged to `main` (`16446c5`, 02:28 UTC today). It ships the same two rotor modes this PR adds, from the same branch lineage:

| | #617 | #619 (on `main`) |
|---|---|---|
| `POINT_MODE` — left/right reading order, up/down column-major | ✅ | ✅ |
| Scatter `INTERSECTION_MODE` | ✅ | ✅ |
| `flatPoints` / `readingOrder` / `columnOrder` / `stepPoint` / `computeEntryPointIndex` | ✅ | ✅ identical design |
| Per-tone panning for row chords | `xPerTone` | `panX` |
| 3D `z` as an echo train | — | ✅ |
| Autoplay paced against the echo tail | — | ✅ |
| `dispose` releasing the point-navigation arrays | — | ✅ |
| `xValues.indexOf` → map built once | — | ✅ |

#619 is a strict superset, and it reached the same conclusion on the panning API — keeping `panning.x` a plain `number` and putting the per-tone slots in a sibling field rather than widening to `number | number[]`.

Merging this now would put a second copy of both modes into `main`, so I'm closing it rather than merging.

**Issue #610 still lists this PR as its only closing reference.** It needs relinking to #619 or closing directly — flagging rather than doing it, since it isn't mine to close.

### The audit did not go to waste

Auditing this branch surfaced defects that are **live on `main` right now**, shipped by #619. Each verified against `bfb520d7`:

1. **Rotor mode desync on a trace swap.** The mode is `rotorIndex` on the service but a boolean on the trace, and nothing resyncs them. After PageUp/PageDown or a live data update, `POINT NAVIGATION` leaves every arrow key dead and silent — no move, no tone, no braille, no message. `INTERSECTING POINT NAVIGATION` announces "no intersection" at every point while autoplay falls through to plain column navigation. #619 made this routine by offering `POINT_MODE` on every scatter; before, it needed a grid config or a multiline stack.
2. **`isInitialEntry` swallows autoplay's first tick**, replaying the point the user is already on and losing a step.
3. **Intersection COL announces a frozen "row 1 of N"** while walking an N-point stack.
4. **Point mode announces "row 1 of 1" for every point**, so two points sharing an x are indistinguishable — in the mode built to tell points apart.

Those four are fixed with 9 tests in **#823**.

### @soundaryakp1999 — your three reports

- **(a) point mode, repeated point / no autoplay** — root cause is item 2 above; fixed in #823.
- **(b) intersection mode, tones where nothing should happen** — root cause is item 1 above; fixed in #823.
- **(c) Enter in a grid, then autoplay** — confirmed real but predates both PRs. `moveOnceInGridMode` moves the *grid* cursor and never checks `isInGridCellMode`, so autoplay drives the cell selection instead of the points inside it; `isMovable` has the same gap. Left out of #823 to keep that change to one logical thing — worth its own issue.

Thanks for catching all three; (a) and (b) both reproduced exactly as described.

---

jooyoungseo:
Closing the loop on this thread — everything raised here is now resolved somewhere:

| Item | Where it landed |
|---|---|
| The feature itself (POINT_MODE, scatter INTERSECTION_MODE) | #619, merged |
| @soundaryakp1999 (a) — point mode replays a point / no autoplay | #823, merged |
| @soundaryakp1999 (b) — intersection mode plays tones where nothing should happen | #823, merged |
| @soundaryakp1999 (c) — autoplay dead after <kbd>Enter</kbd> in a grid cell | #824 → #825, open |
| Issue #610 | Closed as completed by #619 |

@soundaryakp1999 — (c) turned out to be two defects stacked. The `GRID_CELL` keyboard scope bound only left, right and Esc, so <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>arrow</kbd> was never bound there at all; and underneath that, `moveOnce`/`isMovable` never checked `isInGridCellMode`, so binding the keys alone would have made autoplay walk the cell *selection* while you believed you were inside a cell. Both are fixed in #825. Thanks for the report — all three reproduced.

---

PR #619, opened 2026-06-02
feat(scatter): read a third dimension as an echo train, and walk points one at a time

### LINKED ISSUE #601: feat: 3rd dimension scatter as reverb
# Feature Request

## Motivation

We want to introduce a 3rd dimension to scatterplots, and represent it with audio.

## Proposed Solution

We want the audio representation to be a sort of echo / reverb. Play around with sounds, but we want it to sound like the values are further away or closer. Text is easy as we can just give all 3 dimensions of data.

ellvix:
# Pull Request

## Description

Scatter plots gain an optional `z` per point, sonified as a train of echoes: a point at `zMax` repeats the configured number of times, lower z values proportionally fewer, with a little reverb for colour. Reverb alone was tried first and was simply too hard to hear. Because a whole column or row sounds as one chord, z is hard to attribute in the default row/col navigation — so this also adds a **point navigation** rotor mode that walks one datapoint at a time, and an **intersection** mode that isolates a single point out of a stacked column.

The title is written as a conventional commit because a squashed merge composes the release commit from it.

## Related Issues

Closes #601

## Changes Made

- **z through the whole scatter pipeline** — `AudioState.zIntensity` drives the echo count, `TextState.z` carries the number itself, in every navigation mode.
- **Point navigation mode** (`POINT_MODE` on the rotor) — left/right walk reading order, up/down walk column-major order, no wrap.
- **Intersection mode** for scatter, matching the multiline behaviour.
- **Per-tone stereo panning** for row-mode chords, via a new `AudioState.panX`. Previously a whole row collapsed onto one stereo slot and saturated to one ear; this was an unlogged bug, not a 3D-only problem.
- **Settings**: echo count, echo volume, echo duration.

### Merged with `main` (~130 commits, including 4.0.0)

The branch was ~8 months behind. Five files conflicted; the merge keeps both sides throughout and adapts to the APIs that moved underneath it — `TraceEmptyState`/`NearestPoint` signatures, `noImplicitOverride`, the `MathUtil` helpers that replaced the private ones, and main's chord chain-timer structure so the #623 disposal fix survives.

`AudioState.panning.x` deliberately stays a plain `number`. Widening it to `number | number[]` for the row chord would have forced five scalar readers added on main through a narrowing wrapper, contradicted the position-and-range contract #767 documented for `PieTrace`, and made the position announcement read the first point's column for a whole row. The per-tone slots live in a sibling `panX` instead, absent on every other trace.

The `package-lock.json` churn is reverted — it had stripped the `libc` metadata npm uses to skip incompatible optional deps on musl/Alpine. The `.gitignore` and `settings.ts` changes are reverted as unrelated.

### Autoplay no longer talks over itself

Addressing @soundaryakp1999's report that autoplay jumps to the next point before the previous one finishes. Autoplay's interval was a duration budget divided by a point *count*, with no term for how long a point takes to *sound* — while a max-z echo train runs ~2100 ms against a 200–500 ms step. `AudioService` now publishes an `echoTailDeadline` and autoplay reschedules itself against it.

Cancelling the tail instead was considered and rejected: at typical step lengths the first echo is always cancelled before it fires, so autoplay over a 3D scatter would convey no z at all. A point whose z is 0 waits nothing, and a trace with no z data — every 2D plot — steps at exactly its configured rate.

### Accessibility fixes

- Point mode announces z. It emitted echo *density* but no number, in the one mode built for reading a single 3D point.
- Intersection mode announces and sonifies z, on both channels.
- A point whose y is exactly `0` gets its echoes; that branch plays a zero marker, and the echo train is the only channel left carrying z.
- The point-mode boundary chime pans from the current point, not the stale pre-entry column.
- The three echo settings name themselves — their `aria-labelledby` pointed at ids nothing rendered.
- Two more `z` formatters on main interpolated the raw value, reading an array as `10,20` with no rounding.

### Correctness fixes

- Autoplay in point or intersection mode drove the raw row/col stepper, repeating one tone for the whole run while silently flipping the nav mode underneath the user.
- A pointer hover no longer re-anchors point or intersection mode.
- `getStateAt` reads the position it is asked for rather than the mode cursor, so a live-appended point is announced as itself.
- `dispose` releases the point-navigation arrays, which held live SVG references past teardown.
- `echoCount` is bounded above; the field is free-typed and its value persists.
- `xValues.indexOf` per chord tone becomes a map built once.

### Tests

37 new tests covering point-mode navigation, the z announcements, echo scheduling and cancellation, autoplay pacing, and per-tone panning. Local run: `type-check`, `lint`, `build` clean; 1997 tests pass.

### Known follow-ups (deliberately not in this PR)

- Point mode is a first-class rotor mode rather than a `RotorFilterUnit` (#628): filter units are horizontal-only by design, and point mode needs all four directions. MAIDR now carries three per-trace navigation mechanisms (`GridNavigable`, `PointNavigable`, `RotorFilterUnit`) — worth consolidating separately.
- With OS key-repeat, holding an arrow key cancels each queued echo before it fires, so z is silent while the key is held. That is correct — a user-driven keypress cannot be delayed — but a short fade on cancelled echoes would be gentler than the current drop.
- Grid mode and the two new modes survive `resetToInitialEntry`, matching main's existing grid-mode behaviour. Left as-is for parity.


soundaryakp1999:
Hello,
This feature works well. I have found one challenge: Autoplay does not wait for one tune to complete before jumping to next point. This is a significant challenge, because when the second echo is playing from the first point, I am hearing overlapping sounds from next point.

jooyoungseo:
@soundaryakp1999 the autoplay overlap is fixed — thank you, that report found a real design gap rather than a stray bug.

**Why it happened.** Two clocks that never referenced each other. Autoplay's step interval is `autoplayDuration / pointCount` — a duration budget divided by a point *count*, with no term for how long a point takes to *sound*. At the defaults a max-z point's echo train runs about 2100 ms (five echoes 300 ms apart, plus the last echo's own tone and its reverb decay) against a step of 200–500 ms. So autoplay was starting point N+1 while point N was still on its second echo, exactly as you heard.

`cancelPendingEchoes()` was already running on every navigation, and it was not enough: it clears *queued* echoes but never one already sounding, and the chord's timer chain outlived the step and re-queued echoes after the next point had cancelled them.

**The fix.** `AudioService` now publishes an `echoTailDeadline` — when the point currently sounding falls silent — and autoplay reschedules itself against it rather than running on a fixed interval, so a step is `max(configured rate, time left in the tail)`.

I deliberately did *not* fix this by cancelling the tail. At typical step lengths the first echo is always cancelled before it fires, so that version of autoplay over a 3D scatter would convey no z at all — it would sound like a 2D plot with extra latency. Waiting is slower but keeps the dimension audible.

Two things worth confirming when you next get a chance to listen:

1. **2D plots must be unchanged.** The deadline is only ever non-zero when a point actually scheduled echoes, and `zIntensity` is set only by a scatter trace that has z data — so every other plot autoplays at exactly its old rate. There is a test pinning this, but your ears are the real check.
2. **Pacing on a high-z scatter.** Because the tail scales with z, a point at `zMax` now holds the longest and a point at `zMin` holds not at all — the rhythm itself varies with z. That may read as a useful extra cue or as unevenness; I could not tell from the code alone. If it is distracting, lowering Echo Duration in Settings shortens every tail proportionally.

One case I could not fix, noted so it does not surprise you: while an arrow key is held down, OS key-repeat fires faster than the first echo, so each point's echoes are cancelled before any sound and z goes silent for the duration of the hold. Autoplay can be slowed to fit the audio; a user's own keypress cannot. Deliberate tapping is unaffected.

The branch is also now merged up to `main` (it was about 130 commits behind, including 4.0.0) with a set of other fixes — the details are in the PR description.

---

jooyoungseo:
Thanks — going through the five open items. Three were correct code that kept inviting the question, so I documented the reasoning rather than changing behaviour; one was a real inconsistency and is fixed; two are deliberately left for maintainers.

**Fixed**

- **`notifyIntersectionMode`'s self-contradictory doc** — you're right that it read as nonsense, but the guard is load-bearing rather than defensive: `context.active` returns a `Plot`, which may be the figure or a subplot, neither of which has `setIntersectionMode`. Dropping the `instanceof` would not compile. The comment was the wrong half, and now says both things precisely: the narrowing gets us to a trace, and every trace past it is safe because `AbstractTrace` defaults to a no-op. (`notifyPointMode` can duck-type because `isPointNavigable` is a type guard over an interface — a different tool for a different shape, not a newer style the older code failed to adopt.)

**Documented, not changed**

- **`MAX_ECHO_REVERB_WET = 1.6`** — intentional, and now says why. A `ConvolverNode` normalizes its impulse response by default, so a wet path at 1.0 sits well below the dry tone it is colouring and the tail is inaudible against it; the boost buys that headroom back. It cannot clip: the wet and dry paths meet at a `DynamicsCompressorNode`, which limits rather than wraps, and the echo it applies to is already attenuated by its own `volumeScale`.

- **The `1e-4` ramp target** — I checked this one specifically because "leaves a signal into the compressor indefinitely" would be a real defect, and it does not: `cleanUp` disconnects both the wet gain and the convolver, and for a reverb-active tone `lifetimeMs` is `(duration + tailLen) * 1000 + 50`, i.e. it waits out the whole tail before tearing down. Nothing survives the node. The `1e-4` is only symmetry with the envelope ramps above it — a linear ramp to exactly 0 would be equally legal. Comment added so the next reader doesn't have to re-derive it.

- **`computeEntryPointIndex` float `===`** — agreed with your retraction, and added the comment you suggested. Worth stating the stronger version: a tolerance would be the *more* dangerous choice here, because at a fixed x the candidates are neighbouring y values, and an epsilon could match the wrong point in a dense column. Exact equality holds because both sides are the same source values copied into two groupings, never the result of arithmetic.

**Left for maintainers**

- **Echo settings under General** — agreed they only apply to z-enabled scatter, but where they live is a product call about the Settings dialog's shape, not something to decide inside a feature PR. Happy to move them under an Audio sub-section in a follow-up if you'd like it.

- **`ScatterTrace` constructor length** — real, and I deliberately left it. The constructor is where the merge conflict landed; refactoring it in the same change would have made the resolution unreviewable against either parent. Better as its own commit on top, where the extraction is the whole diff.

- **Scope (point/intersection nav bundled in)** — flagging this for @jooyoungseo and @ellvix rather than deciding it. The two modes are genuinely dependent on the 3D work: a whole column sounds as one chord, so without point-by-point navigation the z cue can be heard but not attributed to a point. Splitting them now would mean landing a feature whose value is hard to demonstrate on its own. The PR description lists the design cost of keeping them (MAIDR now carries three per-trace navigation mechanisms — `GridNavigable`, `PointNavigable`, and `RotorFilterUnit`) as a consolidation worth doing separately.

**One correction on the review's process note:** the full pipeline was run — `lint`, `type-check`, `build`, and `npm test` (1997 passing, 37 of them new) locally, and all 14 CI checks including E2E chromium are green on `438c59b`.

---

PR #620, opened 2026-06-11
feat: add realtime/streaming data support

### LINKED ISSUE #536: feat: add realtime/streaming data support
## Summary

Add support for realtime and streaming data updates to maidr.js, enabling accessible sonification and navigation of live-updating visualizations (e.g., stock tickers, sensor dashboards, live metrics).

## Motivation

[chart2music](https://github.com/julianna-langston/chart2music) already provides production-ready realtime support through:
- **`.appendData()`** — append individual data points in live mode
- **`.setData()`** — replace all chart data with re-calculated axes
- **Monitor mode** — pressing 'M' on a live chart auto-sonifies newly appended data
- **Sliding window** — `maxWidth` option caps visible data for streaming scenarios
- **Axis auto-scaling** — min/max automatically adjust as new data arrives

maidr.js currently has **no mechanism** for updating data after initialization. All data is parsed once from embedded HTML attributes and stored in `readonly` fields. Users navigating live dashboards or streaming data visualizations cannot benefit from maidr's accessible features.

## Current Architectural Barriers

| Barrier | Impact | Details |
|---------|--------|---------|
| **Immutable data model** | High | All trace data is `protected readonly` — no setters or update methods (`src/model/abstract.ts`, `src/model/line.ts`) |
| **One-time SVG binding** | High | Highlight elements (`highlightValues: SVGElement[][]`) are captured from `layer.selectors` during construction and never re-queried |
| **Cached derived state** | High | Min/max values, navigation graph (`MovableGraph`), audio frequency ranges are all computed once at init |
| **No incremental API** | Medium | Only entry point is `new Figure(maidr)` — full reconstruction; no `addDataPoint()` or `replaceData()` |
| **Controller lifecycle** | Medium | Controller is created/disposed on focus changes; no state preservation across data updates |
| **Tight DOM coupling** | Medium | Model expects static `selectors` strings to find SVG elements |

## Proposed API (inspired by chart2music)

```typescript
// Append a single data point to a trace (for streaming)
maidr.appendData(dataPoint, { traceIndex?, groupIndex? });

// Replace all data (for full refresh)
maidr.setData(newMaidrJson);

// Configuration options
{
  live: true,         // Enable live/realtime mode
  maxWidth: 100,      // Sliding window size (optional)
}

// Monitor mode: user presses 'M' to auto-sonify incoming data
```

## Key Design Considerations

1. **Mutable data store:** Refactor trace classes to support data mutations with proper invalidation of derived state (min/max, navigation graph, audio ranges)
2. **Observer pattern reuse:** The existing observer infrastructure (`addObserver`/`notifyStateUpdate`) can be extended to propagate data changes to services (audio, braille, text)
3. **Dynamic SVG handling:** Either re-query selectors on update or accept external SVG mutation notifications
4. **Monitor mode:** Auto-announce and sonify new data points via screen reader and Web Audio API when the user enables monitoring
5. **Performance:** Batch updates, debounce recalculations, and use incremental axis scaling rather than full recomputation
6. **Backward compatibility:** Static charts should continue to work exactly as before; live features should be opt-in

### Phase 1: Data replacement (`setData`)
- Add mutable data support to trace classes
- Implement cache invalidation for min/max and navigation graph
- Add `setData()` public API on Controller/Figure
- Re-query SVG selectors on data replacement

### Phase 2: Data appending (`appendData`)
- Add `appendData()` method with incremental axis scaling
- Implement sliding window (`maxWidth`) support
- Notify observers of data changes

### Phase 3: Monitor mode
- Add `live` configuration option
- Implement 'M' keybinding for monitor mode toggle
- Auto-sonify and announce new data points in monitor mode
- Screen reader announcements for data updates

## References

- chart2music live plot example: [`examples/charts/livePlot.js`](https://github.com/julianna-langston/chart2music/blob/main/examples/charts/livePlot.js)
- chart2music `appendData()` implementation: [`src/c2mChart.ts` L1142-1219](https://github.com/julianna-langston/chart2music/blob/main/src/c2mChart.ts)
- chart2music `setData()` implementation: [`src/c2mChart.ts` L941-974](https://github.com/julianna-langston/chart2music/blob/main/src/c2mChart.ts)
- chart2music live tests: [`test/chart_live.test.ts`](https://github.com/julianna-langston/chart2music/blob/main/test/chart_live.test.ts)

jooyoungseo:
## Summary

Implements realtime/streaming data support as proposed in #536, covering all three suggested phases plus user documentation.

Closes #536

### Phase 1 — `setData` (full data replacement)
- **Script tag:** `window.maidrLive.setData(maidrJson)`
- **npm/React:** `setMaidrData()` from `maidr/react`, or simply pass a new `data` prop on charts configured with `live: true`
- Core mechanism: `Context.replaceFigure()` swaps the Figure/Subplot/Trace model **in place**. The old figure is disposed first (removing stale SVG highlight clones before the new model re-queries selectors), the navigation stack is rebuilt at the same depth, and the user's position is silently restored with clamping (including across layer-type changes, which reset navigation via the shape signature). Services, view models, and keybindings are untouched, so updates are cheap enough for streaming.

### Phase 2 — `appendData` (streaming)
- `window.maidrLive.appendData(point, { id?, layerId?, layerIndex?, groupIndex?, subplotRow?, subplotCol? })` / `appendMaidrData()`
- New `LiveDataManager` registry routes updates to mounted chart instances and merges points immutably — flat layers (bar, scatter, histogram, **candlestick**, box) and nested group layers (line, smooth, segmented) are both supported, including initially empty layers and creating a new series at `groupIndex === count`; heatmaps are rejected with a warning
- New `Maidr.maxWidth` option implements the sliding window: oldest points are trimmed on append, and the cursor follows the same data point as the window slides (section-aware for candlesticks)
- Axis ranges, navigation graphs, braille, and text recompute naturally since the model is rebuilt from the merged data
- Multi-layer charts (e.g. py-maidr's candlestick + volume + moving-average) stream per layer via `layerId`, preserving the user's active layer and x-position across ticks

### Phase 3 — Monitor mode
- New `Maidr.live: true` opt-in flag
- Pressing **M** (TRACE, BRAILLE, and SUBPLOT scopes) toggles `MonitorService`, announcing "Monitoring on/off" (or explaining monitoring is only for live charts)
- While on, each appended point is auto-sonified (AudioService) and announced (TextService) **without moving the user's cursor**, mirroring chart2music's behavior — via `AbstractTrace.getStateAt(row, col)`, which computes state at a position without notifying observers (now structurally enforced: `notifyStateUpdate` throws during computation). Candlesticks announce the new candle's **close** price (with trend), overriding `getStateAt` to map their internal section/point state
- Wired through the standard Command pattern (`ToggleMonitorCommand`, `CommandContext`, keymaps)

### AI chat freshness
`ChatService.updateData()` stores the latest data and lazily re-serializes on the first LLM request after a change (cached via `getDataJson()`), so AI answers reflect the data currently on screen with zero serialization cost on streaming appends.

### Documentation & examples
- New user manual: `docs/LIVE_DATA.md` (Live & Streaming Data) with full API reference, script-tag and React examples, a candlestick/stock-ticker section, a real-data-feed integration guide, monitor mode, sliding window, performance guidance, and behavior details; registered in the docs site build
- Cross-links and sections added to `README.md`, `docs/SCHEMA.md` (`live`/`maxWidth` properties), `docs/CONTROLS.md` (Monitor Mode key), `docs/react.md` (Live & Streaming Data section), and `docs/llms.txt`
- Runnable streaming demos: `examples/live-line.html` (simulated sensor stream), `examples/live-candlestick.html` (simulated multi-layer stock ticker: candlestick + volume + moving average), and `examples/live-coinbase.html` (**real** live market data — BTC-USD trades from Coinbase's keyless public WebSocket feed, aggregated into 10-second candles; requires network)

## Design rationale

Rather than making all 15 trace classes mutable with cache invalidation (the high-risk path flagged in the issue's barrier table), updates rebuild the model layer through the existing, tested constructors and restore navigation state at the Context level. This works uniformly for every chart type, reuses the Observer wiring as suggested in the issue, and leaves static charts completely unaffected (live features are opt-in).

## Testing

- **TDD:** 60+ new unit tests (written red-first) — `test/service/liveData.test.ts` (data merging incl. candlestick/empty-layer/new-series edges, sliding window, clone semantics, manager registry), `test/service/monitor.test.ts`, `test/model/contextReplaceFigure.test.ts` (position preservation/clamping/shape & type changes/multi-layer mixed types/figure-level depth-1/observer cleanup), `test/model/traceStateAt.test.ts` + `test/model/candlestickStateAt.test.ts` (cursor-free state computation incl. throw-path restore and the structural no-notify guard), `test/service/chat.test.ts` (lazy serialization), `test/command/toggleMonitor.test.ts`
- Full Jest suite: 262 passing; the 3 failures in `display.test.ts`/`braille-escape.test.ts` pre-exist on `main`
- **End-to-end (Playwright, real browser, production bundle): 20/20 checks** — API exposure, navigation, append + reachability, monitor on/off announcements, auto-announcement of appended points, cursor preservation across appends, `setData` with position clamping, the `maxWidth` sliding window, new-series creation, and the full multi-layer candlestick ticker flow (layer switching with x-sync, per-layer appends, monitor announcing the new candle's close price). The Coinbase real-data demo was additionally verified live in a browser against the actual feed.
- `tsc --noEmit`, repo-wide ESLint, full 12-bundle production build, and the docs site build all pass

https://claude.ai/code/session_01KuiTyhDAEWtQX9oTKGatzN


jooyoungseo:
Addressed both review rounds in 7fa8177 (and the earlier Copilot threads in 7d8668d). Summary:

**Fixed**
- `announceAppendedPoint` now restores the trace cursor in a `finally` block, so a throwing state getter can't leave the cursor mutated.
- The orphaned `LiveDataPoint` JSDoc is re-attached to its declaration.
- `Context.replaceFigure` accepts only a factory (tests updated), keeping the dispose-before-construct ordering explicit.
- `resolveActiveColShift`: documented the layer-index ↔ trace-row 1:1 mapping (subplots build `traces[layerIndex][0]` by construction) and the catch now logs via `console.warn`.
- Fields that intentionally dropped `readonly` are marked mutable-by-design; `resolveId` uses a non-null assertion; the `liveDataManager` singleton docs tell tests to construct their own instance.
- Added `cloneMaidrData` unit tests (deep-clone semantics, `onNavigate` preserved by reference, absent when omitted).
- Docs: the declarative React streaming example now trims its own window with a note that `maxWidth` applies only to `appendData`; `react.md`/`LIVE_DATA.md` steer streaming users to `appendMaidrData`; the demo explains its parallel SVG point copy; added the data-ownership comment in `useMaidrController`.

**Not changed, with reasons**
- *`layerId` typed `string` but `layer.id` optional*: `MaidrLayer.id` is declared required (`id: string`) in `src/type/grammar.ts`, so there is no optionality bug here.
- *`maxWidth` on the prop/`setData` path*: intentional — `setData` replaces data verbatim (matching chart2music's `setData`). Now documented explicitly in both docs and the example.
- *M key announcing on static charts*: kept as a short notification ("Monitoring is available only for live charts"). For screen-reader users, silent dead keys are harder to debug than a one-line explanation; happy to gate the keybinding on `live: true` instead if maintainers prefer.
- *`chat.test.ts` private-field access*: kept the narrow cast for now; asserting on a stubbed `fetch` payload would couple the test to provider payload shapes.

Verification: 243 unit tests passing (3 failures pre-exist on `main`), type-check/lint clean, and the 13-check Playwright run against the rebuilt production bundle passes.

---

jooyoungseo:
Third review round addressed in 5c98acb:

**Fixed**
- **#1 (temporary trace mutation):** introduced `AbstractTrace.getStateAt(row, col)` — the compute-state-without-moving pattern now lives in one unit-tested method on the trace (cursor restored in `finally`, no observer notifications), and `announceAppendedPoint` just calls it. Covered by `test/model/traceStateAt.test.ts`.
- **#2 (`describeShape` ignores layer types):** the shape signature now uses `subplot.traceTypes`, so a same-count type change (bar → line) resets navigation instead of restoring onto an incompatible structure. Regression test added.
- **#3 (depth-1 ambiguity):** the depth-1 rebuild now discriminates with the exact predicate `initializePlotContext` uses (`figureState.size !== 1`) instead of a parallel heuristic, and the same-shape branch documents that scope is intentionally preserved.
- **#4 (serialization per append):** went with the suggested lazy approach — `ChatService` stores the latest `Maidr` and serializes on the first LLM request after a change (cached via `getDataJson()`); models receive a shared supplier. Streaming appends now do zero serialization work.
- **#5 (brittle test):** the chat tests now assert the public `getDataJson()` surface (including laziness and caching) instead of reaching into private model fields.
- **#6 (`setData` reference ownership):** documented that the manager stores the caller's reference and it must not be mutated after handoff.
- **#7 (scope preservation):** comment added.
- Nits: VIOLIN_KDE nested-data clarification added; redundant `aria-hidden="false"` removed from the demo.

**Not changed**
- `useMaidrController` integration test: would need jsdom + React Testing Library, neither of which is in the toolchain; the 13-check Playwright suite exercises this exact path (registration, live updates while focused, monitor announcements) against the production bundle.
- `layerId` optionality (re-raised in Minor): `MaidrLayer.id` is required (`id: string`) in `src/type/grammar.ts`.

Verification: 250 unit tests passing (3 pre-existing failures on `main`), type-check/lint clean, 13/13 Playwright checks pass against the rebuilt bundle.

---

jooyoungseo:
Fourth review round addressed in 522098c:

**Fixed**
- **#1:** added `Subplot.activeLayerIndex` — the `traces[layerIndex][0]` invariant is now owned and documented by `Subplot` itself; `resolveActiveColShift` uses the accessor.
- **#2:** confirmed the dispose chain clears observers (`AbstractPlot.dispose()` does `this.observers.length = 0`, and Figure → Subplot → Trace all run through it), and added a regression test so streaming updates can never silently accumulate stale observers.
- **#7:** registered the **M** toggle in SUBPLOT scope as well, so multi-panel users at subplot level get feedback instead of a silent dead key.
- **#6:** `updateStoredData` now carries the same reference-ownership caveat as `setData`.
- **#8:** the out-of-range group warning now includes the layer's group count.
- **Docs:** added selector-stability guidance for highlight rebinding (stable classes for keyed D3/Vega re-renders, and the graceful count-mismatch fallback) plus performance guidance — per-update cost scales with chart size, so `maxWidth` is the essential bound for fast streams.

**No change needed**
- **#3 (depth-1 cross-reference):** the comment already states "Discriminate exactly like initializePlotContext does" and the branch uses that method's exact predicate as of the previous round.
- **#5 (`getDataJson` visibility):** keeping it public — the previous review round specifically criticized tests reaching into private state, and this read-only accessor is the agreed public surface; making it private would reintroduce that problem.
- **Test gaps:** `announceAppendedPoint` unit tests remain impractical at the Controller level (its core logic now lives in the unit-tested `Trace.getStateAt`); the React hook path stays covered by the 13-check Playwright suite since jsdom/RTL aren't in the toolchain.

Verification: 251 unit tests passing (3 pre-existing failures on `main`), type-check/lint clean, 13/13 Playwright checks pass against the rebuilt bundle.

---

jooyoungseo:
Fifth review round addressed in 572f23c:

**Fixed (documentation items)**
- **#3:** `getStateAt` JSDoc now warns about the re-entrancy hazard — the pattern is safe only while state getters stay synchronous and notification-free.
- **#6:** `describeShape` now documents why per-layer series counts are deliberately excluded: `appendData` can start a new series (`groupIndex === count`) and that must *preserve* the user's position; restored indices are clamped, so a differing series count can never land out of range.
- **#7:** `AppendDataOptions` JSDoc notes that per-update cost scales with total chart size and that `Maidr.maxWidth` is the mitigation for high-frequency streams (this also lives in `docs/LIVE_DATA.md`).

**Not changed, with verification**
- **#1 (`layerId` type bug):** verified again — `MaidrLayer.id` is declared **required** (`id: string`) at `src/type/grammar.ts:403`. There is no `string | undefined` assignment; this claim has now been checked three times.
- **#5 (`Observer<PlotState>` vs `Observer<TraceState>`):** both `AudioService` and `TextService` declare `implements Observer<PlotState>` with `update(state: PlotState)` (`audio.ts:71/189`, `text.ts:38/622`). `MonitorService`'s parameter type matches them exactly — no bivariance is involved; it passes a `TraceState`, which is a member of the `PlotState` union, soundly.
- **#4 (SSR guard on `window.maidrLive`):** `src/index.tsx` is the browser-only script-tag bundle and already accesses `window`/`document` unconditionally at module top level (e.g. `window.disconnectMaidrObservers`, the DOMContentLoaded wiring). SSR/npm consumers use `maidr/react`, which never imports this entry. Adding a guard to one line would be inconsistent with the file's established conventions without enabling any new environment.
- **#2 (`resolveId` non-null assertion):** the previous review round specifically requested this exact `!` form (replacing an `as string` cast); `Array.from(keys())[0]` types as `string | undefined` too, so it wouldn't remove the assertion, just hide it.

Verification: 251 unit tests passing (3 pre-existing failures on `main`), type-check/lint clean.

---

jooyoungseo:
Sixth review round addressed in 033fab3:

**Confirmed (no change needed)**
- **`announceAppendedPoint` indexing:** `traces[layerIndex][0]` is correct by construction — `Subplot` builds `this.traces = layers.map(layer => [TraceFactory.create(layer)])` (`plot.ts:346`), i.e. one single-trace row per layer. The `Subplot.activeLayerIndex` accessor added in an earlier round exists precisely to own and document this invariant.

**Fixed**
- **#7:** `MonitorService` now types its collaborators as `Observer<TraceState>`, matching exactly what `handleNewPoint` feeds them. (`PlotState`-accepting services remain assignable — a function accepting the wider union is soundly contravariant — so no call sites changed.)
- **#3:** the `getStateAt` re-entrancy note now states explicitly that safety depends on the entire call chain being synchronous, so timers such as autoplay ticks cannot interleave before the `finally`-restore.
- **#4:** `LIVE_DATA.md` documents the violin KDE caveat — appends are structurally accepted but do not recompute the density distribution; `setData` with freshly computed densities is recommended.
- **#8:** `LIVE_DATA.md` Behavior Details now covers series-count changes: a `setData` removing the user's series clamps the cursor to the nearest remaining series rather than resetting navigation.

**Not changed, with reasons**
- **#1 (SSR guard)** and **#2 (`layerId` type):** both re-raise claims verified false in the previous round — `src/index.tsx` is the browser-only script-tag bundle that accesses `window`/`document` unconditionally throughout (npm/SSR consumers import `maidr/react`, which never loads it), and `MaidrLayer.id` is required (`id: string`, `grammar.ts:403`), so `layerId: layer.id` can never be `undefined`.
- **#5 (observer re-registration cost):** re-registering observers is a handful of array pushes per trace — orders of magnitude below the model rebuild it accompanies; `maxWidth` bounds both.
- **#6 (announce on update failure):** intentionally left as console-level. The failure path disposes the controller, which also tears down the live region that would carry the announcement; the next focus-in rebuilds from stored data, restoring function when the data is valid. When the data itself is invalid, this is an integrator error surfaced to the developer console — happy to revisit the recovery UX if maintainers prefer a different behavior.

Verification: 251 unit tests passing (3 pre-existing failures on `main`), type-check/lint clean.

---

jooyoungseo:
Seventh review round addressed in 1f8a7e5:

**Fixed**
- **#5:** `window.maidrLive` assignment now warns when it would clobber an existing global (script loaded twice / user-defined global). Latest bundle still wins so its own components stay wired to the matching manager instance.
- **#8:** added a `ToggleMonitorCommand` unit test asserting `execute()` delegates to `MonitorService.toggle()`.
- **#2, #4, #6, #7 (comments):** inline note at the monitor announce site documenting the `traces[layerIndex][0]` invariant (cross-referencing `Subplot.activeLayerIndex`); rationale comment in `MonitorService.handleNewPoint` explaining the deliberate direct `update()` calls (announce without moving the cursor or fanning out to highlight/braille); `getDataJson` JSDoc marks the caching strategy as not part of the stable API; the monitor test's partial `TraceState` stub documents its intent.

**Not changed, with reasons**
- **#1 (`layerId` type):** fifth occurrence of this claim. `MaidrLayer.id` is **required** — `id: string` at `src/type/grammar.ts:403` — so `layerId: layer.id` cannot be `undefined`.
- **#3 (formatter refresh per append):** `refresh` re-resolves a handful of format functions per layer — constant-time work orders of magnitude below the model rebuild it accompanies. Diffing the format config would cost about as much as rebuilding it.
- **Minor (`resolveId` assertion):** this line has now received three mutually contradictory style requests across rounds (`!` over `as string` in round 3; `Array.from(...)[0]` in round 5; `as string` in round 7). Keeping the round-3 form, which the codebase already uses elsewhere (e.g. `plotContext.peek()!`).

Verification: 252 unit tests passing (3 pre-existing failures on `main`), type-check/lint clean, 13/13 Playwright checks pass against the rebuilt bundle.

---

jooyoungseo:
Eighth review round addressed in 9380ec0:

**Fixed (documentation)**
- Multi-panel cost now called out explicitly: per-update cost covers *all* subplots/layers of a multi-panel figure, not just the layer receiving the point (`LIVE_DATA.md`).
- The no-mutation-after-handoff contract is now stated on the public `MaidrLiveApi.setData` JSDoc, where script-tag consumers actually read it.
- The demo comment now draws the boundary explicitly: mutating its local SVG-redraw array is fine; never mutate data already handed to MAIDR.

**Not changed, with reasons**
- **`getStateAt` → parameter-threading refactor:** the suggestion ("`get state()` delegating to `computeStateAt(row, col)`") would require rewriting the `audio`/`braille`/`text`/`highlight` getters across all 15 trace classes to take positions as parameters — exactly the high-risk refactor this PR's design deliberately avoids (per the issue's barrier analysis). The cursor-swap is contained in one unit-tested method whose synchronous/no-notification constraint is documented at the definition; a future async refactor of state getters would be a deliberate architectural change that revisits this method by name.
- **`layerId` optionality:** sixth occurrence. `MaidrLayer.id` is **required** (`id: string`, `src/type/grammar.ts:403`) — "verify layer.id is guaranteed non-optional" has been verified in each of the last four rounds. Adding a runtime assert for a state the type system already excludes would be dead code.
- **`TOGGLE_MONITOR` at SUBPLOT scope:** added at review round four's explicit request. Toggling there announces "Monitoring on", so the user is informed before any later announcements arrive; monitoring is chart-wide by design (chart2music behaves the same).
- **Failed-update focus loss:** acknowledged trade-off, already documented at the recovery site in `useMaidrController` ("drop the controller so the next focus-in rebuilds cleanly").

Verification: 252 unit tests passing (3 pre-existing failures on `main`), type-check/lint clean.

---

jooyoungseo:
Tenth review round — two items taken in a7550a9, the rest previously settled:

**Fixed**
- **`MaidrLiveApi` not reachable from `maidr/react`:** good catch — the interface moved into the liveData service and is now re-exported from `maidr/react`, so TypeScript consumers can type `window.maidrLive` without importing the side-effectful script-tag entry.
- **Depth-1 discrimination duplication:** extracted `Context.isFigureLevel()` as the single source of truth used at construction, instruction resolution, and depth-1 stack restoration (this concern had now come up twice).

**Previously settled (no change)**
- `getStateAt` guard flag / async hazard (rounds 6, 8, 9): contract documented at the definition, throw-path restore now unit-tested (`traceStateAt.test.ts`).
- `replaceFigure` dispose-before-factory failure path (rounds 7, 9): recovery is documented and implemented at the hook level; the reviewer's own analysis confirms it is the correct path.
- VIOLIN_KDE append semantics (round 6): documented per that round's offered alternative; a per-append runtime warn would be noise at streaming rates.
- `window.maidrLive ??=` (rounds 7, 9): deliberate latest-bundle-wins with a clobber warning — preserving a pre-load shim with `??=` would leave real charts wired to a different manager instance than the API the page sees.
- Singleton test-isolation Jest hook: tests construct their own `LiveDataManager`; the constraint is documented on the singleton.

Verification: 259 unit tests passing (3 pre-existing on `main`), type-check/lint clean, 17/17 Playwright checks against the rebuilt bundle.

---

jooyoungseo:
Eleventh round triaged — fixed in d1dca55/99240db:

**Fixed**
- **#3 (stale JSDoc):** genuine catch — the `isFigureLevel` extraction had orphaned `initializePlotContext`'s doc block above the new helper. Reattached.
- **Minor (candlestick `finally` ordering):** added the inside-out restore-order note.
- Also caught and fixed a strict-mode type error in the new multilayer test data (`CandlestickPoint` requires `trend`/`volatility`).

**Verified, no change**
- **#1/#8 (SSR guard/test):** fourth occurrence — `src/index.tsx` is the browser-only script bundle (touches `document`/`window` unconditionally at module load throughout); React/SSR consumers use `maidr/react`, which never imports it.
- **#2 (zombie figure):** third occurrence — the throw propagates synchronously to the hook's catch (no observers can fire in between on the same stack), which disposes the controller; documented at the recovery site.
- **#4 (singleton comment "truncated"):** the full JSDoc reads coherently — "Tests should construct their own `new LiveDataManager()` instead of using this instance…"; the review snippet was cut mid-sentence by the reviewing tool, not the source.
- **#5 (naming):** backwards — 16 toggle commands in `toggle.ts` use the `Command` suffix; `ToggleHighContrast` is the lone pre-existing exception.
- **#6 (VIOLIN_KDE), #7 (rate limiting), #9 (singleton contamination), `disposeController` identity (stable: `useCallback` with empty deps):** previously settled/documented; a `batchAppend` variant is reasonable follow-up territory if high-frequency multi-point producers appear.

Verification: 260 unit tests passing (3 pre-existing on `main`), type-check/lint clean.

---

jooyoungseo:
Thirteenth round addressed in 0761eba:

**Fixed**
- **`getStateAt` guard (raised across five rounds; this round's structural proposal sealed it):** `notifyStateUpdate()` now throws if fired while `getStateAt` computes state at a temporary cursor, making the side-effect-free contract structural rather than documentary — a future getter that notifies observers fails loudly in the first test run instead of silently corrupting the cursor. Chose throw over the suggested silent `return` so the regression is visible, not masked. Unit-tested (guard fires, cursor restored, observers untouched).
- **Depth-1 figure-level test gap:** valid catch — added a test pinning multi-panel figure-position restoration at depth 1.
- **Performance note visibility:** the `maxWidth` guidance now also appears on `appendMaidrData()` and `MaidrLiveApi.appendData` JSDoc, where script-tag/npm consumers actually look.
- **PR description:** updated to include `examples/live-coinbase.html`.

**No change, with reasons**
- **Coinbase example disclaimers:** already present — the file header states it needs network access and points to the offline simulated demo, and the page shows live connection status with an offline fallback message. The example exists at the repo owner's request (issue #536's motivating use case is stock tickers) and uses a keyless public market-data feed precisely because equity APIs require keys an example can't ship; whether it stays is the maintainers' call.
- **`resolveId`:** this single line has now received a fifth distinct style demand across rounds (`as string` → `!` → `Array.from()[0]` → `as string` → destructure-with-`??`). Keeping the round-3 form.
- **"announce coordinates" typo:** not a typo — for flat layers those coordinates exist specifically for monitor announcements (e.g. the candle's close section).
- **SSR `@remarks` on the singleton:** `liveData.ts` has no DOM access, and registration only happens in a client-side `useEffect`, so server-side import is inert; the existing JSDoc covers the real (test-isolation) caveat.

Verification: 262 unit tests passing (3 pre-existing on `main`), type-check/lint clean, 20/20 offline Playwright checks against the rebuilt bundle.

---

jooyoungseo:
Fourteenth round addressed in d888328:

**Fixed**
- **#2 (pre-super guard window):** valid refinement of the new structural guard — `Candlestick.getStateAt` now raises `isComputingStateAt` before its pre-super section/index mapping and clears it in its own `finally`, so the whole override is covered including pre-super throws.
- **#3 (Coinbase demo `JSON.parse`):** wrapped defensively; non-JSON frames are ignored.

**No change, with reasons**
- **#1 (`layerId` type):** ninth occurrence. `MaidrLayer.id` is required (`id: string`, `grammar.ts:403`).
- **#6 (`isComputingStateAt` should be private):** directly contradicts this same review's #2, which requires the `Candlestick` subclass to write the flag. It stays `protected`.
- **#4 (dual-ref comment):** already present — the "Data ownership chain" comment in `useMaidrController` (added in an earlier round) explains exactly who writes `latestDataRef` and when.
- **#5 (rate limiting), #7 (`resolveId`, sixth style demand), #8 (concurrent-mode speculation):** previously settled / follow-up territory.

Verification: 262 unit tests passing (3 pre-existing on `main`), type-check/lint clean, 20/20 offline Playwright checks against the rebuilt bundle.

---

jooyoungseo:
Fifteenth round — one valid item, taken in c3a4192:

**Fixed**
- **Test gap #2 (registration dispose guard):** genuine — added a test pinning that disposing a stale registration does not evict a newer one for the same chart id (the listener-identity guard).

**Verified false / already covered**
- **`layerId` optionality:** tenth occurrence. `MaidrLayer.id` is required (`id: string`, `grammar.ts:403`), not `id?: string` as claimed.
- **Warning prints "index -1":** false — the message is `options.layerId ?? \`index ${layerIndex}\``, so when a `layerId` lookup fails it prints the id, never `-1`; the unknown-layerId unit test exercises exactly this path.
- **Test gap #1 (`setLive(false)` while enabled):** already covered — `monitor.test.ts` has had "setLive(false) turns monitoring off" since the second review round.
- **SSR guard (fifth occurrence), re-entrant `getStateAt` (nested calls are safe by stack discipline of save/restore, and the no-notify invariant is now code-enforced), `activeLayerIndex` assertion (invariant holds by construction), React hook integration test (no jsdom/RTL in toolchain; covered by the 20-check Playwright suite):** previously settled.

Verification: 263 unit tests passing (3 pre-existing on `main`), lint clean.

---

PR #622, opened 2026-07-02
fix: monitor only the focused layer and stop live-region chatter in live demo

jooyoungseo:
## Summary

Fixes two user-reported issues in the live/streaming data feature shipped in #620.

## 1. Monitor mode: announce only the focused layer (and series)

**Reported:** on multi-layer candlestick charts (candle + volume + moving average), every streamed tick produced multiple overlapping sonifications — one per appended layer. Users want to hear only the new point in the layer they are currently navigating.

**Fix:** monitor announcements are gated by a new shared focus predicate, `isAppendedPointFocused(figure, appended)` in the liveData service: an append is sonified/announced only when it targets the active subplot, the active layer (`Subplot.activeLayerIndex`), and — for nested multi-series layers such as multiline — the active series (otherwise a multi-series line ticker would replay the same bug at series granularity). Appends elsewhere stay silent, and switching layers with PageUp/PageDown switches which metric is monitored. The same predicate now also drives the sliding-window cursor shift, deduplicating the previously parallel logic in `resolveActiveColShift`. Single-layer, single-series charts are unaffected.

Docs (`LIVE_DATA.md` monitor-mode and multi-layer sections) and both ticker demos' instructions updated to describe the focused-layer behavior.

## 2. Coinbase demo: continuous announcements with monitor mode off

**Reported:** in `examples/live-coinbase.html`, screen readers kept announcing automatically even with monitor mode off, while sitting on a data point.

**Root cause:** not the library — appends are verifiably silent with monitor off (covered by a committed e2e test, below). The demo's status element was a `role="status"` **live region updated on every trade tick** (several times per second), so assistive tech announced each price update.

**Fix:** the per-tick price now renders in a plain (non-live) element; the `role="status"` region is reserved for rare connection-state transitions (connecting → connected → disconnected/error). Added a "never put per-tick values in an `aria-live` region" note to the real-feeds section of `LIVE_DATA.md`, since this is an easy footgun for anyone copying the integration pattern.

## 3. Rider: repair the broken e2e runner config

`e2e_tests/config/test-config.ts` used CommonJS `__dirname` in an ESM package, so `npm run e2e` threw `ReferenceError: __dirname is not defined in ES module scope` before running anything — meaning no Playwright coverage (including the new spec below) could execute in the scheduled e2e workflows. Fixed by deriving the config directory from `import.meta.url`; the config now loads and lists all tests.

## 4. Rider: retry transient Pages deployment failures in docs.yml

The docs deploy on main (run 28695930130, commit 984a7a4e) failed with the Pages backend error "Deployment failed, try again later." seconds after the deployment was created, with a healthy 53MB artifact — the same transient failure mode as the July 2 GitHub Pages incident. `actions/deploy-pages` does not retry a hard deployment failure, so the deploy job now retries once after a 30s backoff before failing. (The two earlier docs-workflow failures were build-side and are already fixed on main: typedoc tsc errors by #623, the react-example `@adapters` alias by #624.) Merging this PR also re-triggers the docs deploy, which should publish the site that main's last run built successfully.

## Testing (committed in this PR)

- **Unit (Jest, CI-enforced):** seven new tests in `test/service/liveData.test.ts` — six exercise `isAppendedPointFocused` against real `Figure` instances (focused-layer match, unfocused-layer rejection, layer-switch flipping, nested-series gating, flat-layer announce coordinates, and subplot mismatch), and one pins that a new-group append never trims the sliding window (the invariant that makes evaluating the focus predicate against the pre-swap figure safe). Full suite: 361 passing; the 3 failures (`display`/`braille-escape` jsdom issues) and 3 `pointerGuidance.test.ts` `tsc` errors pre-exist on clean `main`.
- **E2E (Playwright): new `e2e_tests/specs/liveData.spec.ts`** with four tests against the offline demos and the built bundle: (1) monitor-off appends are completely silent, (2) a sliding-window trim keeps the cursor on the same data point (the positive `resolveActiveColShift` path), (3) nested layers announce only the focused series, (4) the multi-layer ticker announces only the focused layer, including after PageUp/PageDown focus switches. The focus-gating and cursor-shift tests were each verified regression-sensitive by temporarily disabling the gate/shift, which makes them fail; in the multi-layer test, the focused layer's append lands first so an ungated announcement would overwrite it (last write wins). All 4 pass through the repo's `npm run e2e` config (chromium) after the ESM fix above.
- Repo-wide ESLint clean (also fixes a leftover Ollama JSDoc param from the #620/#621 merge); core bundle builds.
- Branch is merged up to date with `main` (post-#623/#624); all of the above re-verified on the merged tree.

https://claude.ai/code/session_01KuiTyhDAEWtQX9oTKGatzN


jooyoungseo:
Both review points addressed in 31fc7f74:

**Test coverage (the blocker) — valid catch, now committed.** The "22/22 Playwright checks" had been run from a session-local harness that was never committed; the criticism stands. This commit lands the coverage properly:
- `e2e_tests/specs/liveData.spec.ts` — three Playwright tests against the offline demos: monitor-off appends are completely silent, nested layers announce only the focused series, and the multi-layer ticker announces only the focused layer across PageUp/PageDown focus switches. (They run under the repo's e2e setup with a built dist; note `e2e_tests/config/test-config.ts` currently fails to load under ESM due to `__dirname` — pre-existing on `main`, verified with an equivalent config.)
- Six Jest unit tests exercising the new `isAppendedPointFocused` predicate against real `Figure` instances: focused-layer match, unfocused-layer rejection, layer-switch flipping, nested-series gating, flat-layer announce coordinates (candlestick sections are not series), and subplot mismatch.
- PR description corrected to describe exactly what is committed.

**Nested-series asymmetry — good catch, implemented rather than just confirmed.** You're right that a multi-series line ticker would have replayed the layer-level bug at series granularity. The focus decision is now extracted into `isAppendedPointFocused(figure, appended)`, which requires the active series for `nested` appends (flat layers keep ignoring `row`, since there it is an announce coordinate like a candlestick OHLC section). Both `announceAppendedPoint` and `resolveActiveColShift` now share this single predicate, eliminating the asymmetry you noted between the two code paths.

Verification: 360 unit tests passing (pre-existing `main` failures unchanged), lint clean, 3/3 new Playwright tests, plus the full local regression harness at 22/22 against the rebuilt bundle.

---

jooyoungseo:
Thanks — both optional points handled in 694325f9:

- **Old/new figure ordering:** agreed it's load-bearing; `updateData` now carries an explicit comment that the sliding-window shift resolves against the OLD figure before the swap while the announcement runs against the NEW one.
- **Dropped `if (!trace)` null check:** intentional — `isAppendedPointFocused` guarantees the subplot/layer indices are in range, and the surrounding `try/catch` turns any residual access failure into a `console.warn` rather than a silent return, which is the more diagnosable of the two behaviors. Noting it here per the suggestion.

CI is re-running on the new commit; all local gates (lint, 360 unit tests, the 3 new Playwright tests, 22-check regression harness) are green.

---

jooyoungseo:
Third round addressed in 0c618ac0:

- **Stale `#status` CSS selector:** genuine catch — the status split left the rule targeting a now-nonexistent id. It styles `#conn, #price` now.
- **Null-guard readability:** added the suggested inline comment in `announceAppendedPoint` explaining that the focus predicate guarantees in-range indices, so residual failures intentionally land in the catch — closing the question for future readers.
- **`waitForTimeout` in the e2e spec:** acknowledged; kept consistent with the existing specs' style as you noted, and a repo-wide move to condition-based waits would be its own change.

All local gates remain green.

---

jooyoungseo:
Fourth round addressed in 49357b62:

- **#2 (fixed sleeps):** took the `waitForFunction` suggestion — positive announcement assertions now poll the aria region for the expected value (with a 5s ceiling); only silence assertions keep a short settle wait, since the absence of an announcement can't be polled. Side benefit: the spec runs ~3× faster (6s vs 21s).
- **#1 (page-object convention):** documented as a deliberate exception in the spec header — these tests target the live-data demo pages and the public `window.maidrLive` API surface rather than the standard plot fixtures the `BasePage` objects model. If live-data e2e coverage grows, a `LiveDataPage` object is the right follow-up.
- **#3 (`'Connected.'` transition):** intended UX, not a side effect — "Connected — waiting for trades…" is accurate only until the first trade arrives, so the one-time transition keeps the live region truthful without per-tick chatter.

3/3 spec tests green against the rebuilt bundle; lint clean.

---

jooyoungseo:
Fifth round addressed in 69a6081b:

- **#2 (remaining fixed nav waits):** converted — every setup/navigation step now polls a deterministic aria marker (the focus-in instruction, the landed point's value, `Layer N of 3` announcements). The only fixed waits left are the silence assertions, where absence can't be polled. The spec now runs in ~3.4s total, down from 20.7s in its first version.

**Not changed, with reasons**
- **#1 (Controller-level Jest test):** constructing a real `Controller` under the node test environment isn't feasible (AudioContext, DOM focus plumbing, and ~15 service dependencies) — the established reason the codebase pins Controller wiring via e2e. The committed spec asserts exactly the wiring outcome you describe (unfocused-layer append → no announcement reaches the aria region; focused → it does), just observed end-to-end rather than via a `handleNewPoint` spy.
- **#3 (`'Connected.'` boolean flag):** keeping the string-compare guard — it's stateless and self-evidently idempotent; as noted, harmless either way.

3/3 spec tests green; lint clean.

---

jooyoungseo:
Sixth round — both minor items taken in 5cb5213c:

- **#1 (`col` unchecked):** documented on the predicate — `col` is the *new point's position*, not a focus requirement; checking it would defeat monitoring, since users hear new points on their focused layer/series wherever their cursor sits along it.
- **#2 (silence settles under CI load):** widened from 300ms to 600ms with an inline note; the positive assertions all poll, so total spec runtime stays ~4s.

34/34 unit tests and 3/3 spec tests green; lint clean. Thanks for the invariant-tracing verification — nothing else outstanding on my side.

---

PR #629, opened 2026-07-05
fix(rotor): re-announce compare-mode boundary messages to screen readers

jooyoungseo:
## Description

Fixes a screen-reader accessibility gap in rotor **compare mode** (lower/higher value navigation): boundary messages ("No higher value found…") were dispatched only to the rotor `aria-live` area, which does **not** re-announce identical content on repeat key presses. So a screen-reader user holding/repeating a direction at a compare boundary heard it once, then silence.

Fixing this surfaced a deeper root cause (see Changes Made #2): the `notify` reducer never bumped the `revision` counter that the alert region is keyed on, so routing a message through `notification.notify()` did **not** reliably force a re-announcement on its own. Compare mode happened to still re-announce because each boundary also emits a model *warning* state that flows through the `update` reducer (which does bump `revision`) — but paths that `notify()` without a warning (e.g. intersection mode's *vertical-unavailable* message) would go silent on repeat. This PR fixes both the routing and the underlying reducer, so re-announcement is robust and self-contained for **all** rotor boundary paths.

Split out of #628 (candlestick rotor trend units) because it affects **all** trace types with compare navigation (bar, line, heatmap, candlestick, box…), not just candlesticks.

## Related Issues

Follow-up to #628 — surfaced during that PR's review. Deferred rotor items tracked in #630.

## Changes Made

1. **`src/service/rotor.ts`** — Route the five compare-mode boundary returns (`moveUp`, `moveDown`, `moveLeft`, `moveRight`, and the shared `callMoveToNextCompareMethod`) through the notification helper so they re-announce. Generalized the helper name `announceIntersectionMessage → announceRotorMessage` now that it serves the intersection, compare, and filter-unit paths; updated the related doc comments.
2. **`src/state/viewModel/textViewModel.ts`** — Root-cause fix: the `notify` reducer now bumps `revision` (previously only `update` did). The `role="alert"` region in `Text.tsx` is keyed by `revision`, so this is what actually forces a re-mount / re-announcement on repeat presses. Makes the "route through `notify()` to re-announce" contract true for **every** `notify()` caller, and closes the intersection *vertical-unavailable* repeat gap.
3. **Tests** — `test/service/rotor.test.ts`: three structurally distinct wrapped call sites (the `callMoveToNextCompareMethod` fallback via `moveRight`/higher, the left/lower fallback, and a heatmap `moveUp` **inline** compare branch) plus text-off suppression. `test/state/viewModel/textViewModel.test.ts`: reducer-level test asserting `notify` advances `revision` on identical repeats (and that `update` still does).

N/A — screen-reader announcement behavior.

## Checklist

- [x] I have tested the changes locally, and all tests related to this pull request pass (61 suites / 793 tests green; production build passes).

## Additional Notes

- Behavior is now consistent across all rotor boundary paths (compare / intersection / filter): every repeat press at a boundary re-announces, and text-off mode suppresses the message.
- **Known pre-existing trade-off (not changed here):** rotor boundary messages are written to two live regions (`rotor_value` and the `role="alert"` region), so the *first* hit may announce twice. This predates #628/#629 (it already applied to intersection mode). Consolidating onto a single announcement path is tracked as item 3 in #630, gated on a screen-reader pass.
- #628 (now merged) performed the same `announceIntersectionMessage → announceRotorMessage` rename; this branch has been rebased onto it.

https://claude.ai/code/session_01ENgT6oNeYux73K1A699ky6

### DIFF
--- a/src/service/rotor.ts
+++ b/src/service/rotor.ts
@@ -573,9 +573,9 @@ export class RotorNavigationService {
    * screen readers to re-announce on every keystroke. Without this, repeated
    * identical messages dispatched only to the rotor area announce once and
    * then stay silent. Used by the intersection, compare-mode, and filter-unit
-   * boundary paths. Returns the message unchanged so callers can chain. Empty strings
-   * (text-off mode) are passed through; NotificationService already no-ops on
-   * empty input.
+   * boundary paths. Returns the message unchanged so callers can chain. Empty
+   * strings (text-off mode) are passed through; NotificationService already
+   * no-ops on empty input.
    * @param message The text to announce; returned unchanged.
    */
   private announceRotorMessage(message: string): string {
--- a/src/service/rotor.ts
+++ b/src/service/rotor.ts
@@ -50,11 +50,12 @@ export class RotorNavigationService {
    * Creates a new RotorNavigationService instance.
    * @param context - The context providing access to the active trace
    * @param text - Provides terse/verbose/off mode for message formatting
-   * @param notification - Used to push intersection-mode messages through the
-   *   text alert region. The alert region re-mounts on every dispatch (keyed
-   *   by a revision counter), which is what causes screen readers to
-   *   re-announce on repeat key presses. Without this, identical messages
-   *   dispatched only to the rotor area do not re-announce.
+   * @param notification - Used to push rotor boundary messages (intersection
+   *   and compare mode) through the text alert region. The alert region
+   *   re-mounts on every dispatch (keyed by a revision counter), which is what
+   *   causes screen readers to re-announce on repeat key presses. Without
+   *   this, identical messages dispatched only to the rotor area do not
+   *   re-announce.
    */
   public constructor(context: Context, text: TextService, notification: NotificationService) {
     this.context = context;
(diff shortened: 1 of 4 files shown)


jooyoungseo:
Thanks for the deep trace on the `revision` mechanism — verified against the reducer/service/component. One correction and one fix:

**Compare mode already re-announces on repeat presses today.** There's a second path the trace missed: every compare boundary calls `notifyRotorBounds()`, which emits a `warning` state. `TextService.update()` does *not* early-return on it (`!state.warning` is `false` at `text.ts:562`), so it reaches `format(state)` → `"No plot info to display"` (`text.ts:229`) → `onChangeEmitter.fire` → the `update` reducer → `revision += 1` on **every** press. So `revision` does change each press, the `role="alert"` div re-mounts, and the SR re-announces; `notify(msg)` then supplies the specific "No higher value found…" text.

**But the underlying point is valid and now fixed.** `notify()` itself didn't bump `revision`, so the mechanism wasn't self-sufficient — it only worked here by coincidence of the parallel warning-state path. The intersection **vertical-unavailable** path (`moveUp`/`moveDown` in intersection mode) intentionally does *not* call `notifyRotorBounds()`, so it would announce once then go silent on repeat. Commit ab8cb45 makes the `notify` reducer bump `revision`, so re-announcement is robust and self-contained across compare / intersection / filter modes, and the doc comments are now accurate. Added a reducer test asserting `notify` advances `revision` on identical repeats.

The possible double-announce across the two live regions is unchanged pre-existing intersection-mode behavior. The `moveGrid` unsupported-branch gap is tracked in #630.

---

PR #634, opened 2026-07-06
fix(text): restore edge boundary alert for single-panel navigation

jooyoungseo:
## Description

Navigating to a trace **edge** (far left / far right / far top / far bottom) within a MAIDR single panel produced **no text alert**. It previously announced something like _"No plot info to display"_ — this was a silent regression.

**Root cause.** When a move fails at a boundary, `AbstractTrace.notifyOutOfBounds()` (`src/model/abstract.ts`) pushes an out-of-bounds trace state: `{ empty: true, type: 'trace', … }` with **no `warning` field**. A guard added to `TextService.update()` in #557 (the AnyChart adapter PR) turned that branch into a bare `return`, short-circuiting **before** `format()` / `onChangeEmitter.fire()`. So the boundary alert never reached the info/alert region. This is most visible in single-panel plots, whose navigation is entirely trace-level (subplot- and figure-level edges were never affected, since the guard is scoped to `type === 'trace'`).

**Fix.** Restore the alert while keeping the guard's intent, with **mode-appropriate wording** (the maintainer asked for terse ≠ verbose):

| Text mode | Behavior at edge |
|-----------|------------------|
| OFF       | silent (no announcement) |
| TERSE     | **"No more data"** |
| VERBOSE   | **"No more data to display"** |

Both modes use data-oriented phrasing and signal the boundary ("no more …") explicitly. The terse/verbose strings are kept **local to the out-of-bounds branch**, so they apply only to edge navigation — the rotor/section-bounds path (`warning: true`, excluded by `!state.warning`) keeps its original mode-independent `"No plot info to display"` via `format()`.

The out-of-bounds branch still `return`s **before** the `currentState` bookkeeping (so the AI chat keeps the user's last valid position) and **before** the `first_navigation` gate (so screen-reader announce-gating is unaffected).

This is a deliberate **hybrid, not a pure revert**: pre-#557 also overwrote `currentState` with the empty state; we intentionally preserve it, which is strictly better for the AI chat's "current position".

## Related Issues

Regression introduced in #557 (the out-of-bounds edge alert was silently swallowed).

## Changes Made

- **`src/service/text.ts`**
  - In the out-of-bounds guard of `update()`, emit the boundary alert via `onChangeEmitter` when text mode is not OFF, before returning. The terse/verbose strings (`"No more data"` / `"No more data to display"`) are literals **local to this branch**, so the wording is scoped to edge navigation only.
  - `format()` is left unchanged, so the rotor/section-bounds message (a `warning: true` empty trace state routed through the same `format()` call) keeps its original `"No plot info to display"` in both modes.
  - No model or ViewModel changes needed — every trace type funnels through the same out-of-bounds state shape, so the alert is restored uniformly (bar, line, scatter, heatmap, histogram, boxplot, candlestick, violin, segmented, …).
- **`e2e_tests/utils/constants.ts`** — updated `PLOT_EXTREME_VERIFICATION` to `"No more data to display"` (asserted in the default/verbose mode across the plot specs, which all reference the constant).
- **`test/service/textFirstNavigation.test.ts`** — regression coverage: out-of-bounds emission in VERBOSE (`"No more data to display"`) and TERSE (`"No more data"`), silence in OFF, re-announcement on repeated edge presses, `currentState` preservation, a model-level proof that the first key press un-gates announcements even in a boundary direction, and a scoping test that the rotor-bounds/warning path stays `"No plot info to display"` in both modes.

N/A — screen-reader / text-alert behavior; verified via unit tests and the e2e extreme-navigation specs.

## Checklist

- [x] I have tested the changes locally, and all tests related to this pull request pass (`npm run type-check`, targeted Jest suites, `npm run build`, ESLint — all green).
- [ ] I have updated the documentation, if applicable (no doc changes needed).

## Additional Notes

- **Verification:** `npm run type-check` clean; `test/service` + `test/state` → 198 passed; `npm run build` succeeds; ESLint clean.
- **Wording:** terse `"No more data"` / verbose `"No more data to display"` per maintainer preference; both use data-oriented phrasing.
- **Announce-gating (deliberate):** the boundary alert follows the same `announce` gating as normal navigation text — always shown visually, announced to screen readers only when `announce` is `true`. It's unreachable as a *first* interaction (the initial-entry guard makes the first move succeed, un-gating announcements; covered by a test), and during autoplay the boundary text stays visually present but unannounced, consistent with per-point nav text under `announce: false`.

https://claude.ai/code/session_01V6gyjDvAqC2TUpvrds69dU

### DIFF
--- a/e2e_tests/utils/constants.ts
+++ b/e2e_tests/utils/constants.ts
@@ -42,7 +42,7 @@ export abstract class TestConstants {
   /**
    * MAIDR plot identifiers
    */
-  static readonly PLOT_EXTREME_VERIFICATION = 'No plot info to display';
+  static readonly PLOT_EXTREME_VERIFICATION = 'No more data to display';
   /**
    * MAIDR component identifiers
    */
--- a/src/service/text.ts
+++ b/src/service/text.ts
@@ -556,20 +556,17 @@ export class TextService implements Observer<PlotState>, Disposable {
    * @param state - The new plot state to process
    */
   public update(state: PlotState): void {
-    // Pure out-of-bounds events (empty trace state with no `warning` field) are
-    // fired by AbstractTrace.notifyOutOfBounds() when navigation hits a boundary
-    // (far left/right/top/bottom of a trace). The user stays located at the last
-    // valid data point, so we must NOT overwrite `currentState` (used by the AI
-    // chat) below — hence the early return before that bookkeeping.
+    // Out-of-bounds events (empty trace state, no `warning`) are fired by
+    // AbstractTrace.notifyOutOfBounds() when navigation hits a boundary. The
+    // user stays at the last valid data point, so we must NOT overwrite
+    // `currentState` (used by the AI chat) — hence the early return before that
+    // bookkeeping below.
     //
-    // We DO, however, announce a boundary alert so reaching an edge is not
-    // silent. Respect the text mode: OFF stays silent, while TERSE ("No info")
-    // and VERBOSE ("No plot info to display") get mode-appropriate wording from
+    // We still announce a boundary alert so reaching an edge is not silent,
+    // respecting text mode: OFF stays silent, while TERSE ("No info") and
+    // VERBOSE ("No plot info to display") get mode-appropriate wording from
     // format(). Returning early also avoids firing `first_navigation` for an
     // empty state, keeping announce-gating intact.
-    //
-    // (A regression from #557 turned this into a bare `return`, silencing the
-    // edge alert entirely — this restores it without clobbering `currentState`.)
     if (
       state
       && state.empty
--- a/src/service/text.ts
+++ b/src/service/text.ts
@@ -557,13 +557,14 @@ export class TextService implements Observer<PlotState>, Disposable {
     // bookkeeping below.
     //
     // We still announce a boundary alert so reaching an edge is not silent,
-    // respecting text mode: OFF stays silent, TERSE gets a short "No info" cue,
-    // and VERBOSE reuses format()'s full "No plot info to display". The wording
-    // split is kept LOCAL to this branch so it applies only to edge navigation
-    // — the warning/rotor-bounds path (excluded by `!state.warning`) still flows
-    // through format() and keeps its original, mode-independent wording.
-    // Returning early also avoids firing `first_navigation` for an empty state,
-    // keeping announce-gating intact.
+    // respecting text mode: OFF stays silent, TERSE gets the short "No more
+    // data" cue, and VERBOSE gets "No more data to display". These literals are
+    // LOCAL to this edge branch (not from format()) so the wording applies only
+    // to edge navigation — the warning/rotor-bounds path (excluded by
+    // `!state.warning`) still flows through format() and keeps its original,
+    // mode-independent "No plot info to display" wording. Returning early also
+    // avoids firing `first_navigation` for an empty state, keeping
+    // announce-gating intact.
     if (
       state
       && state.empty
@@ -571,7 +572,7 @@ export class TextService implements Observer<PlotState>, Disposable {
       && !state.warning
     ) {
       if (this.mode !== TextMode.OFF) {
-        const text = this.mode === TextMode.TERSE ? 'No info' : this.format(state);
+        const text = this.mode === TextMode.TERSE ? 'No more data' : 'No more data to display';
         this.onChangeEmitter.fire({ value: text });
       }
       return;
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Thanks for the careful review. Digging into the "first-navigation boundary is silent to screen readers" concern — I don't think it's reachable in practice, because of the initial-entry guard:

- `MovableGrid.moveOnce` / `moveToExtreme` (`src/model/movable.ts:48-52`, `78-81`), `MovableGraph.moveOnce` (`:151-155`), and `LineTrace.moveOnce` (`src/model/line.ts:326-332`) all **unconditionally return `true` while `isInitialEntry`** — they call `handleInitialEntry()` (clamp + clear the flag) and emit a normal, non-empty state.
- So a user's very first key press — including `Up`/`Down` on a single-row trace, or `Home`/`End` at an extreme on load — always lands on a valid point and fires a non-empty update, which is exactly what trips `first_navigation` → `setAnnounce(true)`.
- The first *out-of-bounds* update can therefore only happen after ≥1 successful navigation, by which point `announce` is already `true` and the `role="alert"` region speaks. The new unit tests exercise `announce`-independent behavior at the `TextService` layer precisely because the announce flip is already covered/guaranteed upstream.

The one path without an initial-entry guard is `moveToIndex` (`movable.ts:106-115`) — i.e. a "go to index" command targeting an out-of-range index *as the literal first interaction*. That's a corner case unrelated to edge navigation and out of scope for this regression fix; happy to file a follow-up if you'd like it hardened.

On the minor points:

- **OFF re-check** (`text.ts`): this one is structurally required rather than redundant. The boundary block must return *before* the `currentState` bookkeeping (to preserve the AI-chat position), and the general OFF gate deliberately sits *after* that bookkeeping (so `currentState` stays fresh even in OFF mode). Since the boundary path returns before reaching that gate, it needs its own `mode !== OFF` check.
- **Comment length**: fair — trimmed in `609886a`. Dropped the #557 regression narrative (that's commit/PR history) and kept the load-bearing WHY (`currentState` preservation + mode-aware wording).

Let me know if you'd still like `first_navigation` to also fire on the very first boundary attempt as a belt-and-suspenders measure — it'd mean revisiting the `does not fire first_navigation for an out-of-bounds empty trace state` assertion from #633, so I left that design as-is.

---

jooyoungseo:
This is the same first-navigation concern raised in the earlier review, and it isn't reachable — the key detail is the model's **initial-entry guard**. `MovableGrid`/`MovableGraph`/`LineTrace` `moveOnce` all `return true` unconditionally while `isInitialEntry` (`movable.ts:48-52`, `line.ts:326-332`), calling `handleInitialEntry()` and emitting a **non-empty** state.

So the premise "pressing Left while sitting on the first bar as the first move" can't occur: before the first key press the cursor is in initial-entry state, not "on the first bar." That first press — in *any* direction, including a would-be-edge one — lands on a valid point and emits a non-empty state, which is exactly what fires `first_navigation` → `setAnnounce(true)`. A genuine out-of-bounds (empty) event can only follow ≥1 successful move, when `announce` is already `true`.

I added a model-level regression test in `2f0f042` that drives `UPWARD` (a boundary direction for a single-row bar) as the very first press and asserts `first_navigation` still fires — turning the refutation into executable proof.

Minor notes:
- The `!state` / `state.type` dereference in `format()` is pre-existing and untouched by this diff (as you note) — leaving it out of scope.
- Terse `"No info"` wording is deliberately a shortening of the verbose `"No plot info to display"` (kept identical semantics for consistency); flagged to the author as an accessibility-wording judgment call.

---

jooyoungseo:
Good catch — this one is real. You're right that `format()`'s empty-trace branch is shared: the warning/rotor-bounds state (`notifyRotorBounds()` → `{ empty: true, type: 'trace', warning: true }`) is excluded from the out-of-bounds guard by `!state.warning` but still hits the same `format()` call at the bottom of `update()` (confirmed `isLayerSwitchTraceState` requires `!state.empty`, so it doesn't divert). So my mode-aware wording was leaking into the rotor-bounds message in terse mode — out of scope, since this PR is only about edge navigation.

Fixed in `e18a33a`: moved the terse/verbose split **into the out-of-bounds guard block** and restored `format()` to its original mode-independent wording. Now:

- **Edge navigation** (`!state.warning`): TERSE → `"No info"`, VERBOSE → `"No plot info to display"`.
- **Rotor/section bounds** (`warning: true`): flows through unchanged `format()` → `"No plot info to display"` in **both** modes, exactly as before.

Added a regression test (`keeps the warning/rotor-bounds wording mode-independent in terse mode`) that drives a warning state in terse mode and asserts it still emits `"No plot info to display"`, pinning the scope down so this can't drift. Full suite green (198 passed), type-check and lint clean.

---

PR #635, opened 2026-07-07
feat(candlestick): rework delta layer with Ctrl+L toggle, OHLC nav, and directional sonification

jooyoungseo:
Replace the F7 reference-comparison dialog with a two-key model:
- Ctrl+Shift+L opens an accessible listbox of the reference (moving-average)
  lines in the candlestick subplot; Enter remembers the chosen line and
  activates the comparison.
- Ctrl+L toggles the invisible delta layer on/off using the remembered
  reference, preserving the candle x position. The first Ctrl+L with no
  reference chosen warns and auto-opens the picker; the reference persists
  across toggles until re-selected via Ctrl+Shift+L.

Make the virtual delta layer navigable like a real candlestick: left/right
between candles and up/down between OHLC fields (value-sorted per candle),
recomputing the signed delta per field. Volatility is omitted.

Add an "on line" rotor unit that jumps to points sitting exactly on the
reference line, alongside the existing above/below-line units.

Sonify direction: below-line points play a falling pitch glide and
above-line points a rising glide (a new AudioState `glide` field), while
pitch still encodes the delta magnitude; on-line points keep the click.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01BoRu6Lav5mD8mjsJbWCsw7

### DIFF
--- a/src/command/candlestickDelta.ts
+++ b/src/command/candlestickDelta.ts
@@ -3,7 +3,7 @@ import type { CandlestickDeltaViewModel } from '@state/viewModel/candlestickDelt
 import type { Command } from './command';
 
 /**
- * Command bound to Ctrl+L: toggles the virtual candlestick delta layer on or
+ * Command bound to Alt+L: toggles the virtual candlestick delta layer on or
  * off using the remembered reference line. With no reference chosen yet it
  * warns and opens the reference picker.
  */
--- a/src/command/candlestickDelta.ts
+++ b/src/command/candlestickDelta.ts
@@ -3,25 +3,87 @@ import type { CandlestickDeltaViewModel } from '@state/viewModel/candlestickDelt
 import type { Command } from './command';
 
 /**
- * Command bound to F7: opens (or closes) the candlestick reference
- * comparison dialog where the user picks a reference line and OHLC field.
+ * Command bound to Ctrl+L: toggles the virtual candlestick delta layer on or
+ * off using the remembered reference line. With no reference chosen yet it
+ * warns and opens the reference picker.
  */
-export class ToggleCandlestickDeltaCommand implements Command {
+export class ToggleCandlestickDeltaLayerCommand implements Command {
   private readonly candlestickDeltaViewModel: CandlestickDeltaViewModel;
 
-  /**
-   * Creates an instance of ToggleCandlestickDeltaCommand.
-   * @param {CandlestickDeltaViewModel} candlestickDeltaViewModel - The candlestick delta view model.
-   */
   public constructor(candlestickDeltaViewModel: CandlestickDeltaViewModel) {
     this.candlestickDeltaViewModel = candlestickDeltaViewModel;
   }
 
-  /**
-   * Toggles the candlestick delta settings dialog.
-   */
   public execute(): void {
-    this.candlestickDeltaViewModel.toggle();
+    this.candlestickDeltaViewModel.toggleLayer();
+  }
+}
+
+/**
+ * Command bound to Ctrl+Shift+L: opens the reference picker so the user can
+ * choose (or change) the moving-average line to compare against.
+ */
+export class SelectCandlestickDeltaReferenceCommand implements Command {
+  private readonly candlestickDeltaViewModel: CandlestickDeltaViewModel;
+
+  public constructor(candlestickDeltaViewModel: CandlestickDeltaViewModel) {
+    this.candlestickDeltaViewModel = candlestickDeltaViewModel;
+  }
+
+  public execute(): void {
+    this.candlestickDeltaViewModel.openReferencePicker();
+  }
+}
+
+/** Moves the reference-picker selection up one option. */
+export class CandlestickDeltaRefMoveUpCommand implements Command {
+  private readonly candlestickDeltaViewModel: CandlestickDeltaViewModel;
+
+  public constructor(candlestickDeltaViewModel: CandlestickDeltaViewModel) {
+    this.candlestickDeltaViewModel = candlestickDeltaViewModel;
+  }
+
+  public execute(): void {
+    this.candlestickDeltaViewModel.moveSelectionUp();
+  }
+}
+
+/** Moves the reference-picker selection down one option. */
+export class CandlestickDeltaRefMoveDownCommand implements Command {
+  private readonly candlestickDeltaViewModel: CandlestickDeltaViewModel;
+
+  public constructor(candlestickDeltaViewModel: CandlestickDeltaViewModel) {
+    this.candlestickDeltaViewModel = candlestickDeltaViewModel;
+  }
+
+  public execute(): void {
+    this.candlestickDeltaViewModel.moveSelectionDown();
+  }
+}
+
+/** Confirms the highlighted reference line and activates the comparison. */
+export class CandlestickDeltaRefSelectCommand implements Command {
+  private readonly candlestickDeltaViewModel: CandlestickDeltaViewModel;
+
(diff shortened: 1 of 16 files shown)


jooyoungseo:
Thanks for the review — addressed in `0a952c4`:

- **🔴 Ctrl+L is browser-reserved** — correct, and confirmed with the maintainer. The toggle moved from `Ctrl+L` to **`Alt+L`** (not a reserved browser shortcut, keeps the "L" mnemonic, and sits in the app's existing `Alt+Shift` rotor-shortcut family). The reference picker stays on **`Ctrl+Shift+L`**, which isn't reserved. Announcements and the on-screen picker instructions were updated to say "Alt L".
- **🟡 ViewModel test coverage** — added `test/state/viewModel/candlestickDeltaViewModel.test.ts` covering the `toggleLayer` three-way branch (deactivate / re-activate remembered reference / first-use warn-and-open-picker), picker open/preselect/cancel, selection clamping, and confirm/cancel.
- **aria nit** — dropped `aria-activedescendant` from the listbox container; focus is moved onto the highlighted `role="option"` via roving tabindex (matching the `GoToExtrema` pattern), so the two focus patterns no longer overlap.
- **`handleInitialEntry` latent inconsistency** — `moveToNextCompareValue` / `moveToRotorFilter` now settle the cursor on a first-action rotor jump, matching `Candlestick`.

All 840 unit tests, type-check, lint, and build pass locally.

---

jooyoungseo:
Second review addressed in `22b9e95`:

- **Stale F7 references** — updated all three doc comments (`keybinding.ts`, `type/event.ts`, `type/grammar.ts`) to the current Alt+L / Ctrl+Shift+L bindings.
- **Regression test for the negative-frequency clamp** — added `test/service/audio.glide.test.ts`. The mock is deliberately spec-faithful: its oscillator `frequency.exponentialRampToValueAtTime` throws on a non-positive target, so the `safeBase`/`safeTarget` clamps are now a genuine guard (remove them and the test throws exactly as a browser would). It asserts a negative Min Frequency no longer throws out of `update()`, and that above-line points rise / below-line fall.

Two I intentionally left as-is:

- **`min: 0` on the frequency inputs is cosmetic** — agreed. The `playGlideTone` clamp is the real fix, so it's fully covered functionally; a negative Min Frequency simply can't crash anything now. I left the `onChange` handlers unclamped to avoid janky mid-typing behavior (`type="number"` with intermediate `-`/empty states), rather than partially fix input validation.
- **`getStateAt` raw-vs-clamped `row`** — pre-existing pattern shared verbatim with `Candlestick.getStateAt`, and only reachable via `announceAppendedPoint` with real in-range indices, so out of scope for this PR (would be a separate change touching both traces).

All 842 unit tests, type-check, lint, and build pass locally.

---

jooyoungseo:
Great catch on the rotor bug — confirmed and fixed in `c766690`.

**Bug: up/down jumped candles instead of switching fields in above/below-line rotor modes.** Exactly as described: `CandlestickDeltaTrace` did vertical nav via `moveOnce` but never overrode `moveUpRotor`/`moveDownRotor`, so in a compare mode those calls hit the throwing base impl and got silently caught → mirror right/left. Fixed by delegating both to `moveOnce` (mirroring `Candlestick`), and added a `RotorNavigationService`-driven regression test (the model-only `moveToNextCompareValue('up'/'down')` test didn't exercise that path). Left/right still fall through to `moveToNextCompareValue`, unchanged.

The three minor notes — evaluated, leaving as-is:

- **`playGlideTone` duplication** — agreed it parallels `playOscillator`, but as you noted the ADSR/harmonic-palette machinery doesn't apply to a two-point frequency ramp. Folding a glide/ramp branch into `playOscillator` would add conditional complexity to the hot path for every tone; I'd rather keep the glide path self-contained and readable than couple them. Noting as a possible future consolidation.
- **Double focus in `CandlestickDeltaSettings`** — the `modalRef.focus()` is an intentional fallback for the frame before `selectedItemRef` attaches (or if the selected index isn't rendered yet), so the scope keybindings always have a non-input focus target. Harmless and superseded by the option focus in the normal case, as you said; keeping it as the safety net.
- **`dispose()` asymmetry** — not a leak (the per-field caches are GC'd with the instance); the derived caches are `readonly` and several are `Record`s rather than arrays, so clearing them would be non-uniform for no real benefit. `this.candles.length = 0` stays as the one explicit release.

All 844 unit tests, type-check, lint, and build pass locally.

---

jooyoungseo:
Thanks for the thorough pass. Addressed in `424b885`:

**1. Focus containment (real gap) — fixed.** Added a small Tab trap on the picker's modal container that cycles focus between the close button and the highlighted option and prevents the default traversal, so keyboard focus can't escape behind the backdrop. Only Tab is intercepted — arrows/Enter/Escape still flow to the hotkeys-js scope (which is why the modal stays hand-rolled rather than MUI `Modal`, whose keydown plumbing swallowed Escape in the original F7 dialog). The backdrop is a full-screen fixed layer at a higher z-index than the page, so pointer clicks were already blocked (and close the modal); the Tab escape was the actual hole.

**2. On-line unit always advertised — intentional, maintainer-directed.** This is the behavior the maintainer explicitly asked for this session: gating it on "current field has an exact zero" made the unit flicker in/out as you move between OHLC fields, and gating on "any field" hid the whole above/on/below category for real MA data (exact coincidence is rare). Always-on keeps the trichotomy stable and symmetric with the always-present above/below compare units; an empty unit announces "no point on the line". Crossing/tolerance-based variants were offered and can be a follow-up if desired.

**3. `activate()` default param — leaving as-is** (documented + covered by tests, as you noted).

Also confirming your "verified, no issues" spot-checks match my understanding (glide clamp regression-tested, up/down-vs-left/right dispatch, `getStateAt` save/restore parity, braille array identity).

All 844 unit tests, type-check, lint, and build pass locally.

---

jooyoungseo:
Good catch — confirmed and fixed in `655b4ad`.

**Silent x-desync on reconfiguration** — the prior commit gated only fresh activations (`!wasActive`), so an in-place reconfigure (Ctrl+Shift+L while active) to a reference whose domain doesn't include the current delta-layer x still teleported the cursor to index 0. Fixed by extending the same x-sync gate to the reconfigure path: when the newly chosen reference doesn't reach the current candle, MAIDR now **keeps the current comparison and position** and alerts *"Keeping the current comparison: … does not reach …"* instead of jumping. The previously working reference stays remembered. This matches your suggested option (a) — reject the picker selection, keep the old reference/position — and honors the maintainer's stated invariant (the layer is only available where the current x has a delta).

Went with refuse-and-keep rather than announce-the-jump because it's consistent with the fresh-activation behavior and the strict x-sync principle; a teleport (even announced) would still break that invariant.

**Test coverage** — added a service test that reconfigures to a reference with a shorter domain than the current candle and asserts the old comparison/position/remembered-reference are all preserved and the alert fires.

Also, while here, I tightened one thing an adversarial review pass flagged: the remembered reference is now committed only **after** the trace swap succeeds (previously it could be set before the `swapActiveTrace` null-guard).

On the two minor notes: the `maxAbsDelta`-global vs `brailleMaxByField`-per-field asymmetry is intentional (documented + tested), and the first-run "Alt+L at the default leftmost candle refuses because the MA doesn't cover it yet" is the spec'd behavior — the user navigates into the covered range first. Happy to revisit that default-cursor UX if you'd prefer a nudge there.

All 847 unit tests, type-check, lint, and build pass locally.

---

PR #642, opened 2026-07-17
build: parallelized the vite library build to speed up by 3x

ellvix:
# Pull Request

## Parallelize the library build

Reduced build time from nearly 5 min to 100 sec by running build steps in parallel. 

### Problem
`npm run build` produces 12 independent bundles, but ran them one at a time. On a 12-core machine that meant 11 cores sat idle for the entire ~4.5 minute build. The 7 React-based bundles (chartjs, vegalite, amcharts, recharts, react, core, victory) were the bulk of the cost at ~217s combined.

The build comment blamed a `vite-plugin-dts` limitation for the serial approach. The actual cause was narrower: every bundle writes intermediate TypeScript declaration (`.d.ts`) files into the **shared `dist/` folder**, so two builds running at once would clobber each other's files. It's a shared-directory conflict, not something inherent to building in parallel.

This was all done on Windows, but should be safe for Mac/Linux, and probably will work even better. 

### Fix
The build orchestrator (`scripts/build.js`) now runs bundles concurrently:

- One child process per bundle, several at a time. Each worker builds into its own isolated temp dir (`dist/.tmp/<name>`), then the parent moves the finished files into `dist`. Because each bundle's output filenames are unique, the merges never collide, and the `.d.ts` conflict disappears entirely.
- Memory-aware default concurrency: My machine has 12 cores, but what if we have a dev with an old clunky laptop? We don't want an OOM error. So we do `min(⅔ of CPU cores, a RAM-based cap, 8)`. Each heavy build holds ~1.75GB during type rollup, so the RAM cap keeps low-memory laptops (e.g. 8GB) from running out of memory. Override with `--jobs=N` or `MAIDR_BUILD_JOBS`.
- Windows-safe `dist` cleanup: empties the folder's *contents* instead of deleting the folder, avoiding an `EPERM` error on Windows.
- Escape hatches preserved: `--sequential` restores the old behavior, and single-bundle builds (`node scripts/build.js d3`) stay in-process and don't wipe `dist`.

### Results

| Config | Time |
|---|---|
| Before (serial) | 282s |
| After (default, 8 jobs on an 12 core machine) | **~106s** (2.6× faster) |

The default scales per machine: e.g. ~3 jobs on an 8GB laptop, up to the cap of 8 on beefier hardware. After 8 simultaneous jobs it started to get slower, but we could theoretically test with a really beefy machine and see if running all 12 at once is better at the high end. 

### Verification
- Parallel `dist` output is byte-identical to the serial build (all 52 artifacts, spot-checked `maidr.js`, `chartjs.mjs`, `react.d.mts`, etc.).
- Temp dirs cleaned up after build; single-bundle build and lint pass.

### Not included
Each React bundle still re-bundles React + MUI + emotion + the full MAIDR UI independently. Deduplicating that shared code is a larger, riskier restructure left for a follow-up.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -119,7 +119,9 @@ function defaultJobs() {
  * report success having built nothing.
  */
 function parseJobs(raw, source) {
-  const jobs = Number.parseInt(raw, 10);
+  // Number(), not parseInt(): parseInt stops at the first non-digit, so a
+  // value like "4x" would silently parse as 4 instead of being rejected.
+  const jobs = Number(raw);
   if (!Number.isInteger(jobs) || jobs < 1) {
     console.error(`Invalid ${source} value: "${raw}" — expected a positive integer.`);
     process.exit(1);
@@ -534,7 +536,7 @@ async function main() {
     : builds;
 
   // Resolve the job cap early: before the sequential-path branch (a cap of 1
-  // routes there) and before touching dist, so a malformed
+  // routes there) and before emptying dist, so a malformed
   // --jobs/MAIDR_BUILD_JOBS aborts without wiping previous build output.
   const jobs = jobsArg
     ? parseJobs(jobsArg.slice('--jobs='.length), '--jobs')
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -90,6 +90,9 @@ function defaultJobs() {
   // reports the host's cores (Node >= 18.14; fall back for older runtimes).
   const cores = os.availableParallelism?.() ?? os.cpus().length;
   const cpuCap = Math.round(cores * 2 / 3);
+  // Unlike availableParallelism, totalmem() is NOT cgroup-aware: it reports
+  // the host's physical RAM, so a memory-limited container can still
+  // over-provision. Set MAIDR_BUILD_JOBS/--jobs explicitly in that case.
   const memGb = os.totalmem() / 1024 ** 3;
   const memCap = Math.floor((memGb - 1.5) / 1.75);
   return Math.max(1, Math.min(8, cpuCap, memCap));
@@ -358,8 +361,10 @@ async function runParallel(selected, jobs, outDir) {
   const scriptPath = fileURLToPath(import.meta.url);
   const total = selected.length;
   let started = 0;
-  let done = 0;
   const activeChildren = new Set();
+  // Filenames merged into dist this run, keyed to the bundle that emitted
+  // them — lets us detect two bundles emitting the same name (see below).
+  const mergedBy = new Map();
 
   const runWorker = config => new Promise((resolve, reject) => {
     const step = `[${++started}/${total}]`;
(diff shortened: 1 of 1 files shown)


jooyoungseo:
Addressed the review feedback in `c3aee4f`:

- **#1 (merge collision guard)** — fixed. `fs.rename` does silently replace an existing destination file, so the merge now tracks the filenames emitted during the run and fails the build if two bundles ever produce the same name. Stale artifacts from earlier selective builds (which never empty `dist`) are still replaced as before — verified by running a two-bundle selective build twice in a row.
- **#4 (log counters)** — fixed. The `Done` line now reuses the same `[n/total]` index the bundle was announced with, so Building/Done lines correlate under concurrency.
- **#2 (`os.totalmem()` not cgroup-aware)** — documented in `defaultJobs()` with a pointer to `--jobs`/`MAIDR_BUILD_JOBS` for memory-limited containers.
- **#3 (sequential-vs-parallel CI comparison)** — not included here; it needs a CI workflow change and feels like a separate follow-up if the maintainers want it.

---

jooyoungseo:
The CI failures on `c3aee4f` (Build Project / Docs Build / Build example preview) were caused by the merge-collision guard itself: every React-based bundle emits an identical shared `maidr.css`, which the serial build had always handled by silently overwriting in `dist`. Fixed in `94ae69e` — merges are now serialized and a colliding filename is byte-compared against the already-merged copy: identical duplicates keep the first copy, only genuinely different contents fail the build. Verified locally with `core react` (the exact failing pair) plus repeated selective builds.

Also picked up the latest review round:
- **Stale `dist/.tmp` sweep** — now runs on every non-worker invocation (sequential and single-bundle paths included), so a crashed build can't leave temp output for `npm pack` to ship. One caveat vs. the suggestion: the sweep must be skipped in worker processes, since `dist/.tmp` holds sibling workers' in-progress output.
- **Signal handling** — SIGINT/SIGTERM now kill in-flight workers before exiting.
- **Dead `config.outDir` fallback** — removed.

---

jooyoungseo:
Responding to the latest review round: the reviewer's own verdict is that nothing blocks merging, and after seven hardening commits the remaining items are polish, so I'm consciously accepting these rather than pushing another round. Rationale for the record:

1. **`os.freemem()` in `defaultJobs()`** — declining on technical grounds. On Linux, `freemem()` excludes reclaimable page cache, so a healthy machine with a warm cache reports very little "free" memory; sizing off it would collapse concurrency (e.g. a 32 GB box throttled to 1 job) far more often than `totalmem()` over-provisions. This is why `make -j`-style tooling doesn't use free-memory snapshots either. The limitation is documented in the code and `--jobs`/`MAIDR_BUILD_JOBS` remain the correct tools for constrained environments.
2. **Per-bundle log prefixing/buffering** — agreed it would help failure attribution in CI, but it's a self-contained follow-up (piping and re-framing child stdio) rather than part of this PR's correctness story.
3. **`--sequential` hint in the merge-collision error / all-or-nothing merges** — both benign today (no bundle emits directories; a collision fails the whole build non-zero either way, and the next full build rebuilds dist from scratch). Fine additions for whoever touches the merge logic next.
4. **CI parity step (`--sequential` vs default checksums)** — still the most valuable follow-up on this list, but it's a workflow change beyond this PR's scope; flagging it for the maintainers as a separate issue/PR.

@ellvix from my side the PR is ready: all CI checks are green on `fd8980f` and every substantive finding from the review rounds has been fixed and verified.

---

PR #643, opened 2026-07-23
fix: address audiocontext suspension on menu tones

### LINKED ISSUE #641: AudioContext may be suspended: first navigation/menu sound cue can be silently dropped
## Summary

`AudioService`'s menu-tone engine (`playMenuTone`, in `src/service/audio.ts`) short-circuits when the Web Audio `AudioContext` is not in the `running` state (the `state !== 'running'` guard). There is currently no `audioContext.resume()` call anywhere in `src/`. As a result, if a user focuses the plot and immediately triggers a menu tone **before any interaction has resumed the context**, that cue is skipped silently while the action still happens.

This affects **all** menu-tone cues, including:
- the new multi-panel subplot **enter/exit navigation arpeggios** (`playSubplotEnterTone` / `playSubplotExitTone`),
- mode-toggle confirmation tones,
- warning tones.

By contrast, the data-tone path (`playOscillator`) has no such guard.

## Why it matters

MAIDR is an accessibility tool, and a silently-dropped audio cue is exactly the kind of gap that hurts screen-reader / blind users who rely on sonification for feedback. The transition still occurs, but the audible confirmation is lost.

## Severity / likelihood

Low in practice: by the time a user reaches a subplot they have usually already pressed keys, so the context is typically already `running`. This is an **edge case**, and it is **pre-existing** — not introduced by any single feature. It was surfaced during review of #639 (multi-panel figure lobby cues), which reuses the existing `playMenuTone` engine.

## Suggested fix (separate PR)

Resume the `AudioContext` on a user gesture (or lazily, right before playing a menu tone when the context is `suspended`) so the first cue is not dropped. This is a small, self-contained audio-lifecycle change that should be its own PR rather than bundled into a feature.

## References

- Raised in the review of #639.
- Code: `src/service/audio.ts` — `playMenuTone` (the `state !== 'running'` guard), and the public `playSubplotEnterTone` / `playSubplotExitTone` that reuse it.

ellvix:
# Pull Request

## Description

Fixed playMenuTone so it no longer drops a cue when AudioContext is suspended (calls resume() via promise once context is actually running). Should fix all affected menu tones: subplot enter/exit arpeggios, mode toggle confirmations, and warning ticks. 

## Related Issues

Fixes #641 

## Changes Made

audio.ts: 
 - Rewired playMenuTone to call resume() then schedule via .then() once call back is in running state and currentTime is advancing normally.
 - moved scheduling into scheduleMenuTone helper
 - deffered path rechecks mode / vol / state because the async gap means user could theoretically toggle sound off or dispose and kill context before it resolves. Reject on a closed context. 
 - test: resume() added, flipping state to running and resolving
 - test: minor rewrite of the menu tone suspended test

## Additional notes

Major reliance on claude for unit test and general testing as this issue couldn't be verified manually.

### DIFF
--- a/src/service/audio.ts
+++ b/src/service/audio.ts
@@ -824,8 +824,36 @@ export class AudioService implements Observer<PlotState>, Disposable {
     if (this.volume <= 0) {
       return;
     }
-    // A suspended context reports currentTime === 0; scheduling start(0)/stop()
-    // now would fire an unexpected beep the instant the context resumes.
+
+    // A suspended context (before the first user gesture reaches the audio
+    // graph) reports currentTime === 0; scheduling start(0)/stop() against it
+    // would collapse the whole arpeggio to the instant it later resumes. Resume
+    // it first and play the notes once it is actually running, so the first cue
+    // after focus is heard rather than dropped.
+    if (this.audioContext.state !== 'running') {
+      void this.audioContext.resume()
+        .then(() => this.scheduleMenuTone(frequencies))
+        .catch(() => {
+          // resume() rejects on a closed context (dispose raced the resume) or
+          // when the browser still blocks playback; nothing to schedule then.
+        });
+      return;
+    }
+
+    this.scheduleMenuTone(frequencies);
+  }
+
+  /**
+   * Schedules a menu arpeggio's beeps back-to-back from the current time.
+   * Re-checks mode/volume/context because it can run after an async
+   * {@link AudioContext.resume}, by which point the user may have turned sound
+   * off or to zero, or the context may have been closed on disposal.
+   * @param frequencies - Beep frequencies in play order.
+   */
+  private scheduleMenuTone(frequencies: number[]): void {
+    if (this.mode === AudioMode.OFF || this.volume <= 0) {
+      return;
+    }
     if (this.audioContext.state !== 'running') {
       return;
     }
--- a/src/service/audio.ts
+++ b/src/service/audio.ts
@@ -765,6 +765,12 @@ export class AudioService implements Observer<PlotState>, Disposable {
    * instead of collapsing to a stray beep when the context later resumes.
    */
   public playWarningTone(): void {
+    // At volume 0 the beeps are skipped anyway (see playOneWarningBeep); bail
+    // early so a silent cue neither triggers resume() nor claims the deferred
+    // cue slot (mirroring playMenuTone's outer guard).
+    if (this.volume <= 0) {
+      return;
+    }
     this.scheduleWhenRunning(() => this.scheduleWarningTone());
   }
 
@@ -795,6 +801,10 @@ export class AudioService implements Observer<PlotState>, Disposable {
    */
   public playWarningToneIfEnabled(): void {
     if (this.mode === AudioMode.OFF) {
+      return;
+    }
+    // See playWarningTone: a silent cue must not resume() or claim the slot.
+    if (this.volume <= 0) {
       return;
     }
     this.scheduleWhenRunning(() => {
--- a/src/service/audio.ts
+++ b/src/service/audio.ts
@@ -847,17 +847,21 @@ export class AudioService implements Observer<PlotState>, Disposable {
       // The already in-flight resume() below will pick up this newer cue.
       return;
     }
-    void this.audioContext.resume()
-      .then(() => {
+    // Two-argument then(): the rejection handler must cover only resume()
+    // itself. A bug thrown by the cue callback should surface as an unhandled
+    // rejection, not be silently absorbed as if resume() had failed.
+    void this.audioContext.resume().then(
+      () => {
         const pending = this.pendingCue;
         this.pendingCue = null;
         pending?.();
-      })
-      .catch(() => {
+      },
+      () => {
         // resume() rejects on a closed context (dispose raced the resume) or
         // when the browser still blocks playback; nothing to schedule then.
         this.pendingCue = null;
-      });
+      },
+    );
   }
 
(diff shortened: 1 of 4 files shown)


jooyoungseo:
Reviewed the PR and pushed follow-up improvements directly to this branch in 80f2ae7. The `playMenuTone` resume-then-schedule fix itself is sound; the commit closes the remaining gaps:

- **Warning tones now get the same fix.** Issue #641 and the PR description also cover warning ticks and mode-toggle confirmations, but those go through `playWarningTone`, which never checked the context state — while suspended it scheduled `start(0)` beeps that fired at an arbitrary instant after resume. It now routes through the same deferral idiom.
- **Only the latest cue deferred behind `resume()` plays (last one wins).** Previously each call made while suspended chained its own arpeggio off `resume().then(...)`, so all of them replayed stacked at the same start time on resume. The shared `scheduleWhenRunning` helper keeps a single pending cue.
- **`dispose()` cancels the pending cue**, so the resume callback can't schedule audio after disposal even where the context's `state` lags `close()`.
- **Tests for the async-gap re-checks** the PR description promised but didn't exercise: disposal during the gap, audio toggled OFF during the gap, `resume()` settling without the context running, cue stacking, plus a new warning-cue suite.

All 957 tests, type-check, lint, and the production build pass. Two pre-existing points left alone deliberately: `playPointerGuidanceBeep` still skips (rather than resumes) on a suspended context, which seems right for a high-frequency hover cue, and `playOneWarningBeep` doesn't register its nodes in `activeAudioIds` — worth a separate cleanup if desired.

---

jooyoungseo:
Addressed the latest review in cce646b:

- **Cross-cue-type last-wins is now an explicit, verified contract**: the `scheduleWhenRunning` doc comment states the slot is deliberately shared across cue types, and a new test pins a warning superseding a pending menu cue (only the newest cue plays on resume; replaying both would stack them at the same start time — worse for comprehension than dropping the stale cue in this pre-first-gesture window).
- Added the two suggested async-edge tests: `resume()` rejecting outright (swallowed without an unhandled rejection, service recovers on the next cue) and multiple deferred calls kicking off exactly one `resume()`.
- Left the mock extraction into a shared test helper for a future change, as suggested — it only becomes worthwhile if a fourth cue suite appears.

959 tests, type-check, and lint all pass.

---

jooyoungseo:
Addressed the mode re-check finding in b679b57, with one deliberate deviation from the suggested fix:

- **The race is real and is now fixed** — a warning queued behind `resume()` via `playWarningToneIfEnabled` played even if the user toggled sound OFF during the async gap. The deferred callback now re-checks the mode.
- **But the check lives at the `playWarningToneIfEnabled` entry point, not inside `scheduleWarningTone`.** `playWarningTone` is the deliberately unconditional variant: `ToggleBrailleCommand` (`src/command/toggle.ts:61`) and `TextViewModel.warnIfTextOff` (`src/state/viewModel/textViewModel.ts:207`) call it precisely so the alert sounds even when audio mode is OFF. Moving `mode === OFF` into `scheduleWarningTone` would have silenced those callers on every path, not just the deferred one. The asymmetry is now documented in both doc comments and pinned by two tests: `IfEnabled` drops a deferred warning after toggling OFF, while the unconditional variant still plays it.
- The `state === 'closed'` early bail-out and shared mock extraction remain deferred as stylistic/non-blocking, per the review.

961 tests, type-check, and lint pass.

---

jooyoungseo:
Applied the two optional nits in e94ed84: `playWarningTone`/`playWarningToneIfEnabled` now short-circuit on `volume <= 0` before deferring (with a resume()-call-count test), and `audio.subplotCue.test.ts` gained an explicit suspended-context test for the subplot enter cue. On point 2 — yes, the shared `pendingCue` slot across cue types is a conscious decision, not an implementation accident: within the pre-first-gesture window, playing only the newest cue beats replaying stale ones stacked together, and the cross-cue supersession test exists precisely to keep that choice visible. 963 tests, type-check, and lint pass.

---

jooyoungseo:
Applied in 89ac413: `scheduleWhenRunning` now uses the two-argument `then()` form so the rejection handler covers only `resume()` itself — a bug thrown by a cue callback surfaces as an unhandled rejection instead of vanishing as an autoplay failure. Added the missing volume-drops-to-0-during-the-gap test (extending the settings stub to fire a live change through the constructor listener), and the test mocks now note that their `resume()` flips state synchronously unlike a real `AudioContext`, so ordering-sensitive tests must override it with a deferred flip.

Deliberately not changed: logging in the reject branch (both rejection causes — dispose race and autoplay block — are expected and benign; logging them from an embedded library would be noise), and the shared-mock extraction stays a follow-up as before. On the shared-slot tradeoff: the window closes once the first `resume()` settles, i.e. within the first user gesture, so a boundary-warning colliding with a menu cue there means two cues from effectively one interaction — playing the newer one remains the intended behavior.

964 tests, type-check, and lint pass.

---

jooyoungseo:
Confirmed the scope question against issue #641: the report covers exactly the menu-tone engine cues (subplot enter/exit arpeggios, mode-toggle confirmations, warning tones) and explicitly contrasts the data-tone path as having no suspended-state guard at all — so `playEmptyTone`/`playOscillator` staying untouched matches the report's scope. One nuance: `playEmptyTone` doesn't silently drop on a suspended context — like the pre-fix warning path it has no state guard, so it schedules at `currentTime === 0` and fires on a later resume. If the maintainers want the deferral idiom there too, that's a clean follow-up issue rather than part of #641. No code change this round.

---

PR #651, opened 2026-07-27
ci: skip commitlint for bot dependency bump commits

jooyoungseo:
## Description

Dependabot security-update PRs currently never get built or tested.

Dependabot lists every bumped package in the commit body, and those lines routinely exceed commitlint's `body-max-line-length: 100`. The `commitlint` job therefore fails on every Dependabot PR — and because `.github/workflows/ci.yml` declares `needs: [commitlint, lint]` on **`build`, `unit-test`, `type-check`, `e2e-smoke`, and `docs-build`**, all five are SKIPPED. Dependency updates, including security fixes, ship completely untested.

This PR unblocks those five jobs by making commitlint skip bot dependency-bump commits, while leaving the rules fully enforced for human-authored commits.

## Related Issues

Observed on #650, where the commitlint job failed with:

```
✖   body's lines must not be longer than 100 characters [body-max-line-length]
✖   found 1 problems, 0 warnings
```

Note: the separate vite/esbuild `ERESOLVE` failure on #650 is tracked independently and is **not** addressed here.

## Changes Made

**`commitlint.ignores.ts`** (new) — the predicate, deliberately import-free:

```ts
export function isBotDependencyBump(message: string): boolean {
  return /^(?:build|chore)\(deps(?:-dev)?\)/.test(message);
}
```

**`commitlint.config.ts`** — wires it in as `ignores: [isBotDependencyBump]`.

**`test/config/commitlint.test.ts`** (new) — 11 unit tests covering the predicate.

`ignores` was preferred over relaxing `body-max-line-length` globally, because it keeps the rule active for humans. The anchored regex matches only the commit *subject*, so it targets Dependabot/Renovate subjects (`build(deps):`, `build(deps-dev):`, `chore(deps):`, `chore(deps-dev):`) and nothing else.

`.husky/commit-msg` uses the same config, so local commits inherit this behavior.

### Why the predicate lives in its own file

So it can be unit tested. Importing `commitlint.config.ts` from Jest fails with `SyntaxError: Cannot use import statement outside a module`, because `@commitlint/config-conventional` is ESM while this repo's Jest setup is CommonJS. The alternatives were modifying the shared `jest.config.ts` `transformIgnorePatterns` (invasive for a CI-config fix, and it would add transform cost to all 88 suites) or re-declaring the regex in the test — which is worse than no test, since editing the config wouldn't fail it.

## Verification

Verified locally with `@commitlint/cli` 19.8 against the **real** commit message from #650, using the exact CI invocation (`npm run commitlint`, i.e. `commitlint --from=HEAD~1 --to=HEAD`) as well as the husky `--edit` path. Re-run after the predicate was extracted, since that could have broken commitlint's TypeScript config loading — it didn't.

| # | Case | Path | Result |
|---|------|------|--------|
| 0 | Dependabot message, **before** the fix | `--from/--to` | ❌ fails with `body-max-line-length` (reproduces #650) |
| 1 | Dependabot message, **after** the fix | `--from/--to` | ✅ passes |
| 2 | Human `fix(audio):` with a >100-char body line | `--from/--to` | ❌ still fails — rule not globally disabled |
| 3 | Dependabot message | `--edit` (husky) | ✅ passes |
| 4 | Human over-long body | `--edit` (husky) | ❌ still fails |
| 5 | `chore(deps-dev):` / `chore(deps):` with over-long body | stdin | ✅ ignored |
| 6 | `feat(deps):` with over-long body | stdin | ❌ still linted — only `build`/`chore` bypass |
| 7 | `build(deps)` appearing only in the **body** of a `fix:` commit | stdin | ❌ still linted — `^` anchor holds |

Full suite green: **88 suites / 1059 tests**, plus `npx eslint` and `npx tsc --noEmit` on the changed files.

**Most importantly:** the five jobs this PR unblocks are demonstrably running on this very PR — Unit Tests, Type Check, Build Project, E2E Smoke and Docs Build all executed rather than being skipped.

N/A — CI configuration change.

## Checklist

## Additional Notes

Manually authored dependency bumps that use a `build(deps):` / `chore(deps):` subject will also bypass commitlint. That is inherent to matching on the subject — commit content cannot distinguish bot from human — and it is intentional here, matching the semantic-release configuration, which already treats `build`/`chore` as non-releasing types. Bot-only enforcement would require a CI-level actor check (`github.actor == 'dependabot[bot]'`), which would not cover the local husky hook; happy to add that as a follow-up if preferred.

https://claude.ai/code/session_01XpUcHHmMhRLPWdmZPiHhs2

### DIFF
--- a/commitlint.config.ts
+++ b/commitlint.config.ts
@@ -1,15 +1,13 @@
 import type { UserConfig } from '@commitlint/types';
 import conventional from '@commitlint/config-conventional';
+import { isBotDependencyBump } from './commitlint.ignores';
 
 const config: UserConfig = {
   extends: ['@commitlint/config-conventional'],
   rules: conventional.rules,
-  // Bot dependency bumps (Dependabot, Renovate) list every bumped package in
-  // the commit body, which routinely exceeds `body-max-line-length`. Failing
-  // commitlint on those skips every job that declares `needs: [commitlint]`,
-  // so security updates would ship untested. Skip linting for those commits
-  // only; the rules stay fully enforced for human-authored commits.
-  ignores: [message => /^(?:build|chore)\(deps(?:-dev)?\)/.test(message)],
+  // Let bot dependency bumps through; the rules stay fully enforced for
+  // human-authored commits. See commitlint.ignores.ts for why.
+  ignores: [isBotDependencyBump],
   helpUrl: 'https://github.com/conventional-changelog/commitlint/#what-is-commitlint',
 };
 
--- a/commitlint.config.ts
+++ b/commitlint.config.ts
@@ -4,6 +4,12 @@ import conventional from '@commitlint/config-conventional';
 const config: UserConfig = {
   extends: ['@commitlint/config-conventional'],
   rules: conventional.rules,
+  // Bot dependency bumps (Dependabot, Renovate) list every bumped package in
+  // the commit body, which routinely exceeds `body-max-line-length`. Failing
+  // commitlint on those skips every job that declares `needs: [commitlint]`,
+  // so security updates would ship untested. Skip linting for those commits
+  // only; the rules stay fully enforced for human-authored commits.
+  ignores: [message => /^(?:build|chore)\(deps(?:-dev)?\)/.test(message)],
   helpUrl: 'https://github.com/conventional-changelog/commitlint/#what-is-commitlint',
 };
 
--- a/commitlint.ignores.ts
+++ b/commitlint.ignores.ts
@@ -1,0 +1,21 @@
+/**
+ * Matches dependency-bump commits authored by bots (Dependabot, Renovate).
+ *
+ * Those commits list every bumped package in the body, which routinely
+ * exceeds `body-max-line-length`. A failing commitlint job skips every job
+ * declaring `needs: [commitlint, lint]` in ci.yml -- build, unit-test,
+ * type-check, e2e-smoke and docs-build -- so dependency updates would merge
+ * without ever being built or tested.
+ *
+ * Anchored on the subject so a `build(deps)` mention inside the body of an
+ * ordinary commit does not silently bypass linting.
+ *
+ * Kept free of imports so it can be unit tested; commitlint's own config
+ * packages are ESM and cannot be loaded by this repo's CommonJS Jest setup.
+ *
+ * @param message - The raw commit message, subject line first.
+ * @returns Whether commitlint should skip this commit.
+ */
+export function isBotDependencyBump(message: string): boolean {
+  return /^(?:build|chore)\(deps(?:-dev)?\)/.test(message);
+}
--- a/test/config/commitlint.test.ts
+++ b/test/config/commitlint.test.ts
@@ -1,0 +1,47 @@
+import { isBotDependencyBump } from '../../commitlint.ignores';
+
+/**
+ * Guards the `ignores` predicate that lets bot dependency bumps bypass
+ * commitlint. Without it, Dependabot's package-listing body trips
+ * `body-max-line-length`, the commitlint job fails, and every job declaring
+ * `needs: [commitlint, lint]` in ci.yml is skipped -- so dependency updates
+ * would merge without ever being built or tested.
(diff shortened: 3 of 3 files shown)


jooyoungseo:
Thanks — addressed the test-coverage point in `171f3e4`, and confirmed the fix works in this PR's own CI.

**Regression protection added.** The predicate now lives in `commitlint.ignores.ts` as `isBotDependencyBump()`, with 11 unit tests in `test/config/commitlint.test.ts` covering the bot subjects (including #650's verbatim subject), human-authored commits, `feat(deps):`, a `chore(depsomething):` near-miss, and the body-anchor case.

The split into a second file is deliberate, not incidental: importing `commitlint.config.ts` from Jest fails with `SyntaxError: Cannot use import statement outside a module`, because `@commitlint/config-conventional` is ESM and this repo's Jest setup is CommonJS. The options were to modify the shared `jest.config.ts` `transformIgnorePatterns` (invasive for a CI-config fix, and it would add transform cost to all 88 suites), or to re-declare the regex in the test — which would be worse than no test, since editing the config wouldn't fail it. An import-free module that both the config and the test consume avoids both.

I re-ran the end-to-end verification after the split, since moving the predicate could have broken commitlint's TypeScript config loading. It didn't: Dependabot's message still passes via `npm run commitlint` (`--from=HEAD~1 --to=HEAD`) and via the husky `--edit` path, and a human commit with a >100-char body line still fails. Full suite is 1059 tests / 88 suites green, plus eslint and `tsc --noEmit`.

**On the other two points:**

- *Human `build(deps):` bypass* — agreed it's a real if low-risk loosening, and I've left it called out in the PR description rather than silently narrowing it. It's inherent to matching on the subject, since commit content can't distinguish bot from human. If you'd prefer bot-only enforcement, that needs a different mechanism (a CI-level actor check on `github.actor == 'dependabot[bot]'`), which I'm happy to do as a follow-up — but it wouldn't cover the local husky hook, so the config-level predicate would likely stay regardless.
- *`chore(deps)(security)` nit* — I checked, and that isn't valid conventional-commit syntax (a commit has at most one scope), so commitlint's parser wouldn't produce it and Dependabot doesn't emit it. Agreed it's not worth guarding against.

Worth noting for reviewers: the five jobs this PR unblocks are now demonstrably running on this very PR — Unit Tests, Type Check, Build Project, E2E Smoke and Docs Build all executed rather than being skipped.

---

jooyoungseo:
Thanks for the second pass. No code changes from this round — reasoning below.

**On the JSDoc nit** (`@param`/`@returns` being redundant): I checked this against the codebase rather than assuming, and the premise doesn't hold. 119 of 209 `.ts`/`.tsx` files under `src/` use `@param`, and 105 use `@returns`. The prevailing house style is exactly what `commitlint.ignores.ts` does — "why" prose first, then the tags. For example, `src/model/context.ts:107-114`:

```ts
 * scope, and returns the element to use as the instruction context.
 *
 * @param figure - The figure to initialize the stack from
 * @returns The instruction context element
 */
private initializePlotContext(figure: Figure): Plot {
```

Removing the tags would make this file the outlier, so I've left them. Happy to drop them if a maintainer prefers otherwise — it's cosmetic either way.

**On `ignores` skipping all rules rather than just `body-max-line-length`:** agreed, and agreed it shouldn't change. Worth adding one point in its favor that reinforces your reasoning — the alternative (`'body-max-line-length': [0, 'always']`) would disable that rule for *every* commit in the repo, including human ones, whereas this scopes the bypass to a narrow subject pattern. Broader bypass, far narrower blast radius. That's the trade I made deliberately.

Also confirming your `tsconfig.build.json` observation independently: `commitlint.ignores.ts` sits outside `src/**/*`, so it is tooling-only and never enters the published bundle.

**CI status on `171f3e4`:** Lint Commit Messages ✅, Lint Code ✅, Unit Tests ✅, Type Check ✅, CodeQL ✅, Analyze (actions + javascript-typescript) ✅. Build Project, E2E Smoke and Docs Build still running — I'm watching and will push a fix if any of them go red.

---

PR #652, opened 2026-07-28
docs(claude): restructure .claude/ into path-scoped rules and mirror them to Copilot

jooyoungseo:
## Description

Reorganises `.claude/` to follow [Anthropic's documented `.claude/` layout](https://code.claude.com/docs/en/claude-directory), then makes the repository's GitHub Copilot configuration a **generated mirror** of it so the two toolchains cannot drift.

---

## 1. Restructure `.claude/` into path-scoped rules

The three reference docs under `.claude/` were pulled in by `@`-imports from `CLAUDE.md`. Imports are expanded at launch, so `ARCHITECTURE.md`, `PATTERNS.md`, and `DEBUGGING.md` loaded **in full on every session** — roughly 60KB (~15k tokens) regardless of the task. The docs are explicit that this doesn't help: *"Splitting into `@path` imports helps organization but doesn't reduce context, since imported files load at launch."*

`.claude/rules/` is the mechanism designed for this. Rules with `paths:` frontmatter load **only when Claude opens a matching file**.

| | Loaded every session | Worst case (all rules matched) |
|---|---|---|
| Before | ~60 KB (~15k tokens) | — |
| After | ~4.5 KB (~1.1k tokens) | ~25 KB (~6.3k tokens) |

**93% smaller baseline.**

### Do `agents/` and `rules/` both belong here?

Yes — this was the open question behind the change. They're orthogonal, not alternatives, and Anthropic's `.claude/` reference lists both as first-class directories. Rules are *instructions* that shape how work is done; agents are *workers* with their own context window. They compose: a subagent inherits the CLAUDE.md hierarchy **including project rules**, so a rule written once applies in the main conversation and inside every subagent. That's what lets the agent definitions get shorter here — they now describe a role, not the codebase.

### Rules added

| Rule | Scope |
|---|---|
| `architecture.md` | `src/**` — MVVC layer contracts, canonical data flow |
| `model.md` | `src/model/**` — traces, `notifyStateUpdate()`, highlight ownership |
| `service.md` | `src/service/**`, `controller.ts` — Observer, Emitter, Disposable, secrets |
| `viewmodel.md` | `src/state/**` — the Redux bridge |
| `ui.md` | `src/ui/**` — React rendering rules, no untrusted markup |
| `command.md` | `src/command/**`, keybinding services — commands and scopes |
| `accessibility.md` | UI + the four modality services |
| `typescript.md` | all `.ts`/`.tsx` — types, style, error handling |
| `testing.md` | `test/**`, `e2e_tests/**` — Jest and Playwright |
| `git-workflow.md` | *unscoped* — Conventional Commits and semantic-release |

`git-workflow.md` is deliberately unscoped: commit conventions are needed even in sessions that never open a source file.

### Also here

- **Debugging playbook → skill** (`.claude/skills/debug-maidr/`). Per the docs, a multi-step procedure belongs in a skill rather than always-on context. Named `debug-maidr` so it doesn't shadow the bundled `/debug`.
- **`CLAUDE.md` trimmed** to ~95 lines, no heavy imports, and corrected commands that don't exist — there is no `npm run dev` or `npm run preview`; the real ones are `dev:recharts` / `dev:victory` and `build:preview`.
- **Subagents consolidated.** Merged `react-ts-code-reviewer` → `code-reviewer` and `react-ts-engineer` → `implementer`; overlapping descriptions make delegation ambiguous, and both were ~1500 chars of inline examples loaded on *every* request. The surviving eight match `.github/agents/` exactly, which already had the clean set.
- `feature-builder` used the legacy `Task(...)` alias, whose agent-type allowlist is [ignored inside a subagent definition](https://code.claude.com/docs/en/sub-agents), so the restriction was a no-op → switched to `Agent`, with the delegation list stated in the body where it can actually be read.
- `debugger` preloads the skill via `skills:`; `test-runner` gained the `Skill` tool. Neither could reach a skill before.

### Skills: two of three removed

`code` and `code-review` were ~90% restatements of what the new rules now deliver — TypeScript standards, React rules, MVVC layer contracts, accessibility — with the remainder duplicating the `implementer` and `code-reviewer` subagents. Two sources for one convention is exactly the drift risk the docs warn about, and the project `code-review` skill *shadowed Claude Code's bundled skill of the same name*.

Deleting both means the bundled `/code-review` is reachable again. Their one piece of unique content — the XSS/secrets checklist — moved into `rules/ui.md` and `rules/service.md`, where it now loads at *write* time rather than only when a reviewer invoked a checklist.

### Copilot: generated, not duplicated

The repo configures Copilot under `.github/`, and the two tools have the same four concepts:

| Concept | Claude Code | GitHub Copilot |
|---|---|---|
| Repo-wide, always loaded | `CLAUDE.md` | `.github/copilot-instructions.md` |
| Path-scoped instructions | `.claude/rules/*.md` (`paths:`) | `.github/instructions/*.instructions.md` (`applyTo:`) |
| Isolated worker | `.claude/agents/*.md` | `.github/agents/*.agent.md` |
| On-demand workflow | `.claude/skills/*/SKILL.md` | `.github/prompts/*.prompt.md` |

The path-scoped pair matches almost exactly, so rather than hand-maintain a second copy that's guaranteed to drift, `scripts/sync-copilot-instructions.mjs` generates that side:

```
CLAUDE.md          →  .github/copilot-instructions.md
.claude/rules/*.md →  .github/instructions/*.instructions.md
```

Transformations: `paths:` list → comma-separated `applyTo:`; unscoped rule → `applyTo: "**"`; brace groups expanded and deduplicated (since `applyTo` documents commas but not braces); cross-references repointed (`rules/model.md` → `model.instructions.md`) so a Copilot-only reader never follows a pointer to a file that doesn't exist on their side. Subdirectories are mirrored.

- `npm run sync:copilot` regenerates; `npm run sync:copilot:check` fails if stale.
- New `copilot-sync` CI job runs the check. Pure Node with no dependencies, so it skips `npm ci`; no job has `needs:` on it, so it gates nothing. **Suggest adding it as a required check** — drift nobody notices defeats the purpose, and it's invisible in review because generated files look plausible. That's branch-protection config, so it's a maintainer call rather than something this PR changes.
- `style-guide.instructions.md` replaced by the generated `typescript.instructions.md`. Its one accurate claim not already covered — barrel `index.ts` at module boundaries only — was folded into the rule first. Its "80% coverage" and "100 character line length" rules were **dropped as false**: neither is configured in `jest.config.ts` or `eslint.config.ts`.
- Fixed relative link paths in `.github/agents/`, which resolved to `.github/agents/.github/...` and were **already broken before this PR**.

**Deliberately not generated:** agents and prompts. The agent frontmatter schemas genuinely differ (`handoffs` and a display `name` on Copilot's side; `permissionMode`, `memory`, `skills`, `maxTurns` on Claude's), and `.github/prompts/` holds GitHub chores — filing an issue from a template, replying to a PR review — that Claude Code does through its GitHub tools instead. `agentParity.test.ts` guards the drift that *is* cheap to catch: a role added, removed, or renamed on one side only.

## Related Issues

None.

N/A — configuration and tooling only.

## Checklist

## Additional Notes

**On testing.** No application source is touched, so there is no runtime behaviour to test manually — and there is no `ManualTestingProcess.md` in the repo to follow, despite the template referencing it. Three test files cover the tooling:

| File | Cases | What it guards |
|---|---|---|
| `test/scripts/syncCopilotInstructions.test.ts` | 26 | Drives the generator as a CLI against fixture trees via `SYNC_COPILOT_ROOT`, asserting exit codes and generated output |
| `test/scripts/claudeRuleGlobs.test.ts` | 58 | Every generated `applyTo` pattern matches ≥1 tracked file via `git ls-files` and has no unexpanded brace; also asserts the committed mirror is in sync |
| `test/scripts/agentParity.test.ts` | 4 | Both tools define the same agent roles; names match filenames; descriptions stay short; `skills:` entries exist |

Run locally with dependencies installed: **lint clean, type-check clean, build succeeds, 1136 tests across 90 suites pass.**

**Seven defects were found in the sync script during review.** Five were the same shape — wrong output, exit code 0:

| Bug | Effect |
|---|---|
| Non-recursive `readdirSync` | A rule in a subdirectory was dropped; `--check` still said "in sync" |
| `paths: [...]` flow sequence | Parsed to zero globs → scoped rule silently became `applyTo: "**"` |
| Glob scan not key-scoped | A sibling `tags:` list folded into `applyTo` |
| Regex quote-stripping | A trailing YAML comment survived into the glob, matching nothing |
| `readTitle` not fence-aware | A leading code block donated its `# comment` as the description |

Plus two lesser ones: `--check` created the output directory despite being documented read-only, and nested brace groups expanded correctly but emitted a duplicate pattern.

Each is now pinned by a regression case, and the cases were verified to **fail against the commit that had the bug** — a suite that passes against buggy code would be worse than none.

**Reviewer note:** `scripts/sync-copilot-instructions.mjs` and its three test files are the real code here. The files under `.github/instructions/` are mechanical output — reviewing `.claude/rules/*.md` covers their content.

**Sources:** [CLAUDE.md and `.claude/rules/`](https://code.claude.com/docs/en/memory) · [Subagents](https://code.claude.com/docs/en/sub-agents) · [Copilot custom instructions](https://docs.github.com/en/copilot/how-tos/configure-custom-instructions) · [Copilot custom agents](https://docs.github.com/en/copilot/reference/custom-agents-configuration) · [VS Code custom instructions](https://code.visualstudio.com/docs/agent-customization/custom-instructions)

---


jooyoungseo:
Thanks — the subdirectory finding was a real bug and I've fixed it in `b58f35d`. Responses to each point:

### 1. Nested rules silently dropped — confirmed and fixed

You were right, and it was worse than a missing feature: it was silent. I reproduced it before fixing:

```
$ mkdir -p .claude/rules/frontend && echo '...' > .claude/rules/frontend/react.md
$ node scripts/sync-copilot-instructions.mjs --check
Copilot instructions are in sync (11 files).      # exit 0
$ ls .github/instructions | grep react
(nothing)
```

A rule that Claude Code loads, that the mirror never generates, and a check that passes anyway — exactly the drift this script exists to catch, made worse by `.claude/README.md` explicitly telling contributors nested rules work.

I went with mirroring the structure rather than flattening, since [Copilot supports subdirectories of `.github/instructions/`](https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions) too — so `.claude/rules/frontend/react.md` → `.github/instructions/frontend/react.instructions.md`. Flattening to `frontend-react.instructions.md` would have introduced a collision case (`frontend/react.md` vs `frontend-react.md`) for no benefit. Orphan detection now walks recursively as well, which the flat version would also have missed.

### 2. Claude-only references leaking into Copilot's copy — fixed, at two different levels

Split this by cause:

- **Cross-references between rules** (`See \`rules/model.md\``) are now rewritten by the generator to `model.instructions.md`. This is mechanical and belongs in the script.
- **`/debug-maidr` in `copilot-instructions.md`** I fixed at the source instead. Rather than adding a substitution rule to the script, `CLAUDE.md` now names the real file first — "the workflow and per-layer checklists are in `.claude/skills/debug-maidr/SKILL.md`; in Claude Code, run `/debug-maidr`" — which reads correctly on both sides and points somewhere real either way. Same treatment for the `.claude/README.md` pointer, which now describes `.claude/` and `.github/` together.

I deliberately didn't build a general templating/conditional-block mechanism. Two dead pointers didn't justify a second syntax in these files, and rewording is harder to get subtly wrong.

### 3. Missing `debugger` row in the agents table

Correct, and it was a genuine inconsistency with the "same eight roles" sentence a few lines below. Added. I also reworded `feature-builder`'s row, which said "orchestrates the six above" — adding a seventh row above it would have quietly made that wrong.

### 4. Orphan handling asymmetry

Fair. Kept the behaviour (plain sync warns, `--check` fails) since surprise deletes are worse, but it's now documented in `.claude/README.md` so it isn't discovered via a CI failure.

### On the earlier review's point about rules reaching subagents

Worth recording since it's load-bearing for deleting the old docs — it's explicit in the [subagents documentation](https://code.claude.com/docs/en/sub-agents), under what a subagent's initial context contains:

> **CLAUDE.md files**: every level of the CLAUDE.md hierarchy the main conversation loads, including `~/.claude/CLAUDE.md`, **project rules**, `CLAUDE.local.md`, and managed policy files.

So project rules are injected at subagent startup, not conditionally on that subagent's own file reads.

### Verification

I installed dependencies this round rather than reasoning about what CI would say. Full suite green locally — lint, type-check, build, 1048/1048 unit tests — plus an 8-case regression pass over the sync script: clean state, edited rule, nested rule, nested output path, nested orphan, flat orphan, malformed frontmatter, restored clean. The nested cases fail against the previous commit and pass against this one.

---

jooyoungseo:
All three addressed in `8e917c2`. Findings 1 and 2 were real; I reproduced both before fixing.

### 1. `readPaths` silently widening scope — fixed

Good catch, and the failure direction was the bad one. Reproduced:

```
$ cat .claude/rules/zz-flow.md
---
paths: ["src/model/**"]
---
$ node scripts/sync-copilot-instructions.mjs && grep applyTo .github/instructions/zz-flow.instructions.md
applyTo: "**"
```

A model-layer rule mirroring as "applies to every file in the repo." The guard was checking for the `paths:` *key* rather than the parsed *result*, so the flow sequence sailed through it and then hit the `globs.length > 0 ? … : '**'` fallback.

A `paths:` key that yields no globs is now an error naming the supported syntax:

```
sync-copilot-instructions: .claude/rules/zz-flow.md: has a `paths:` key but no
globs were parsed from it. Use a block list, one glob per line (`- "src/**"`);
flow sequences (`paths: [...]`) are not supported
```

A rule with *no frontmatter at all* still maps to `applyTo: "**"` — that's intentional, since it's what an unscoped rule means in Claude Code. There's now a regression case pinning both behaviours apart so a future "fix" can't collapse them. Also documented the block-list requirement in `.claude/README.md`'s "adding a rule" section, so authors hit it before the script does.

### 2. `/debug-maidr` inside a mirrored rule body — fixed

Right, and it was a miss on my part rather than a deliberate limitation: I reworded this exact phrasing in `CLAUDE.md` last commit and didn't grep for other instances. `rules/architecture.md` had the same line. Now reads "the full workflow and the per-layer checklists are in `.claude/skills/debug-maidr/SKILL.md`; in Claude Code, run `/debug-maidr`" — points at a real file from either side.

I kept this as source rewording rather than adding a skill-invocation rule to `rewriteRuleLinks`. Cross-references between rules are a closed, mechanical set; "any Claude-only concept in prose" isn't, and a substitution table would be a quiet source of wrong rewrites later.

### 3. Node version floor — noted, but not via `engines`

Checked this one because the version looked off: the `recursive` option for `fs.readdirSync` landed in **v18.17.0**, not v20.1.0 ([PR #41439](https://github.com/nodejs/node/pull/41439), shipped in the 18.17.0 release). So the floor is a release lower than stated and 18.x contributors are mostly fine. There *was* a real bug in that option between v18.17.0 and v18.19.0, but only when combined with `withFileTypes`, which this script doesn't use.

I documented the requirement in the script header rather than adding `engines.node` to `package.json`. `maidr` is a published package, so an `engines` field changes install-time behaviour for every consumer — that's a project-wide decision, not something to slip into a config PR. Happy to add it if a maintainer wants it, as its own change.

### Also verified from your list

Confirming the two you flagged as correct, since both were things I'd want caught if wrong: `readTitle`'s `/^#\s+(\S.*)$/m` does reject `##` headings (the second `#` fails `\s+`), and `copilot-sync` has no job depending on it, so it stays non-gating.

Regression suite is now 11 cases — clean, edited rule, nested rule, nested output path, nested orphan, flat orphan, missing `paths:` key, flow sequence, empty `paths:`, unscoped-still-maps-to-`**`, restored clean. Lint clean.

---

jooyoungseo:
Findings 1 and 2 are addressed in `5533222`. Declining 3, with reasoning.

### 1. `readPaths` scanning the whole frontmatter — confirmed and fixed

Real, and reproduced before fixing:

```
$ cat .claude/rules/zz-leak.md
---
paths:
  - "src/model/**"
tags:
  - "leaked-value"
---
$ node scripts/sync-copilot-instructions.mjs && grep applyTo .github/instructions/zz-leak.instructions.md
applyTo: "src/model/**,leaked-value"
```

Your read of *why* it evades the guard is exactly right: `globs.length` is non-empty, so the "fail instead of widening" check added last commit never fires. Third bug in this script of the same shape — wrong output, no error. The scan now stops at the next top-level (`^\S`) key.

### 2. No automated test — you're right, and it was my own rule I was failing

This is the finding that should have blocked. `.claude/rules/testing.md`, added by this very PR, says "new behaviour ships with a test that would have failed before it," and I was running an 11-case suite by hand and not checking it in. Three regressions have now reached review in this one file. Manual diligence demonstrably wasn't holding.

Added `test/scripts/syncCopilotInstructions.test.ts` — 18 cases driving the CLI end to end. Each builds a fixture tree in a temp directory, points the generator at it via a new `SYNC_COPILOT_ROOT` env var (defaults to the repo root, so normal use is unchanged), and asserts on exit code and generated output.

I then checked the tests actually catch what they claim, by running them against the earlier commits:

| Script version | Result |
|---|---|
| `5533222` (this commit) | 18 passed |
| `8e917c2` (before key scoping) | **1 failed** — `should ignore block lists belonging to other frontmatter keys` |
| `009c958` (before recursive + flow-seq) | **7 failed** — nested rule, nested orphan, flow sequence, empty `paths:`, cross-references |

A suite that passes against the buggy code would have been worse than none, so that seemed worth confirming rather than asserting.

Full run: lint clean, **1066 tests across 88 suites** pass.

### 3. Pre-commit hook — declining, but happy to be overruled

Skipping this one deliberately rather than by oversight. Husky currently runs only `commit-msg`, which is a validation on text the contributor just typed. A `pre-commit` hook is a different thing: it adds latency to every commit for every contributor, including ones who never touch `.claude/`. The `copilot-sync` CI job already catches this, and it's non-gating and finishes in ~5 seconds, so the feedback loop is short.

Changing everyone's commit flow feels like a maintainer call rather than something to fold into a config PR. If you'd like it, a `pre-commit` that early-exits unless `CLAUDE.md` or `.claude/rules/` is staged would be the cheap version — say the word and I'll add it as its own change.

---

jooyoungseo:
Thanks. CI is green across all 13 checks on `5533222`. Taking the three nits in turn — one acted on, two declined:

### 2. Orphan callout — acted on (`1df96e4`)

This was already documented, but you're right that it wasn't where you'd look: the note lived in the Copilot section, ~80 lines below the "Adding a rule" heading someone actually reads when changing rules. Orphan handling has now come up in two consecutive review rounds, which is decent evidence the placement was the problem rather than the wording. Retitled the section to "Adding **or removing** a rule" and put the extra step there:

> **Removing or renaming a rule takes one extra step.** The generator will not delete the instruction file left behind, to avoid surprise deletions — it only prints a note naming it. Delete that file yourself, or `sync:copilot:check` keeps failing in CI after you have "already run the generator."

Which is exactly the stuck state you described.

### 1. Pre-commit hook — declining, [already answered last round](https://github.com/xability/maidr/pull/652#issuecomment-5110487449)

Short version: `commit-msg` validates text the contributor just typed; `pre-commit` taxes every commit from every contributor, including those who never touch `.claude/`. The `copilot-sync` job already catches it, is non-gating, and finishes in ~8s. Changing everyone's commit flow reads as a maintainer decision rather than something to fold into a config PR — happy to add it as its own change if a maintainer wants it.

### 3. `splitFrontmatter` comment — declining

The `readTitle` comment earns its place because the regex looks needlessly convoluted; without the note someone would "simplify" `(\S.*)` back to `(.+)` and reintroduce the backtracking. `indexOf('\n---\n', 3)` has no such trap — it's the obvious way to find a delimiter, and a comment saying so would just be a line explaining that straightforward code is straightforward. I'd rather comments stay load-bearing so the ones that exist keep getting read.

---

Nothing outstanding on my side. Green and ready for a maintainer look.

---

jooyoungseo:
Both acted on in `1ad1980`. Both reproduced first.

### 1. Inline YAML comments corrupting a glob — fixed

Real, and it produced exactly the silent-wrong output you described:

```
$ cat .claude/rules/zz-c.md
---
paths:
  - "src/model/**"  # only the model layer
---
$ node scripts/sync-copilot-instructions.mjs && grep applyTo .github/instructions/zz-c.instructions.md
applyTo: "src/model/**\"  # only the model layer"
```

Your diagnosis of the mechanism is precise — `replace(/^["']|["']$/g, '')` only works when the quote is the final character, so the leading one goes and the trailing one doesn't, taking the comment along with it.

I went with **supporting comments** rather than documenting them as unsupported. They're ordinary YAML, a contributor writing one is doing nothing wrong, and "documented footgun" still leaves a rule silently dead. `readGlob` now reads a quoted scalar to its closing quote, permits a trailing comment, and strips ` #` from unquoted values. Input it *can't* read — unterminated quote, stray text after the closing quote — errors rather than guesses, matching how the parser already handles a flow-sequence `paths:`.

### 2. No check that a glob matches anything — fixed

This was the better of the two findings, and it's the same shape as everything else that's gone wrong in this file: parses fine, generates plausible output, exits 0, rule silently never loads. Manual verification doesn't survive contact with future edits — that's precisely how the last three bugs got in.

`test/scripts/claudeRuleGlobs.test.ts` asserts every generated `applyTo` pattern matches at least one tracked file, and carries no unexpanded brace group.

Two design notes:

- **Reads the patterns from the generated instruction files, not by re-parsing the rules.** Re-parsing would duplicate the generator's parser, and a bug in that copy could mask the very thing being checked. The `copilot-sync` job already guarantees the generated files agree with the rules, so they're a safe proxy.
- **`git ls-files` does the matching.** It's the authority on what's tracked, it handles every glob style in use here (verified `src/**`, `src/**/*.ts`, `src/model/**`, bare file paths, and a nonexistent path returning 0), and it means no new dependency for glob matching.

I put it in the test suite rather than in the sync script, since the script has to stay runnable against the fixture trees its own tests build, which aren't git repositories.

### Verification

Both additions were checked against the code they describe, since a test that passes on the buggy version is worse than none:

| | Result |
|---|---|
| Four comment cases vs `5533222` | **4 failed** |
| Glob check with `src/model/**` → `src/modle/**` | **1 failed**, naming the pattern |
| `1ad1980` | 79 passed in `test/scripts` |

Full run: lint clean, type-check clean, **1127 tests across 89 suites**.

---

Worth noting where this landed: four review rounds, four bugs in this one script, all the same shape — wrong output, exit 0. The mechanism is now guarded by tests at both levels: the generator's behaviour, and the rules' own globs.

---

jooyoungseo:
Both addressed in `15f0331`.

### `readTitle` and fenced code blocks — fixed rather than commented

You suggested a one-line comment noting the assumption. I fixed it instead, because skipping fences costs about what the comment would have. Reproduced first:

```
$ cat .claude/rules/zz-t.md
---
paths:
  - "src/model/**"
---

```bash
# install deps first
npm install
```

# Real heading
$ node scripts/sync-copilot-instructions.mjs && grep description .github/instructions/zz-t.instructions.md
description: "install deps first"
```

You're right that it's low impact — the description is hover text, not scope — but it's the same shape as the four parser bugs already fixed here: a plausible wrong value, substituted silently. Documenting a known-wrong behaviour when the fix is a fence-state boolean seemed like the wrong trade, especially in a file whose entire theme has been "fails loudly instead of guessing."

The ReDoS-safe `(\S.*)` capture is preserved; it just runs per line now instead of with `/m`.

### Orphan note — louder, but no `--prune`

Took the first option. The note now reads:

```
WARNING: .github/instructions/foo.instructions.md has no matching rule. Nothing
was deleted — remove it yourself, or `sync:copilot:check` will keep failing in CI.
```

That names the exact stuck state you described — generator looks successful, CI still red for a file you didn't know about.

Declining `--prune`. Deleting files the caller didn't ask about is the surprising behaviour, and a flag people have to know to pass doesn't help the contributor who doesn't know the file exists. The warning plus the README's "Adding **or removing** a rule" section reach them without a flag.

### On the coupling you flagged

Confirming it's intentional, since you noted it as such: `claudeRuleGlobs.test.ts` reading the committed `.github/instructions/` output rather than regenerating is the point. Re-parsing the rules would duplicate the generator's parser, and a bug in that copy could mask the thing being checked. `copilot-sync` already guarantees the output is current, so the coupling buys independence from the code under test.

### Verification

Fence case fails against `1ad1980`, passes here. Full run: lint clean, **1129 tests across 89 suites**, mirror in sync.

---

Five rounds, five findings in this script — every one the same failure mode, wrong output with a zero exit. Both layers are now test-guarded, and I don't have anything outstanding. Green and ready for a maintainer.

---

**Comment by jooyoungseo** [comment]
Item 2 was a real gap — fixed in `bef0b77`. Declining item 1, with reasoning.

### 2. Did the deleted skills' security content survive? Partly, in the wrong place

Good question to ask, and worth checking rather than assuming. Grepping `.claude/` for XSS-related terms:

```
$ grep -rn -i "innerhtml\|xss\|sanitiz" .claude/rules/
(nothing)
$ grep -rn -i "innerhtml\|xss\|sanitiz" .claude/agents/
.claude/agents/code-reviewer.md:54:  ... No `innerHTML` with data that could come from a user.
```

So it survived only in the reviewer agent. That means the guidance reached someone *reviewing* a change and nobody *writing* one — rules load at write time, and none of them said anything about it. Backwards for a preventable bug class, and a real regression from deleting the skill even though the words existed somewhere.

Rather than add a broad `security.md` that loads across `src/**`, I scoped each part to where it can actually go wrong:

- **`rules/ui.md`** — no `dangerouslySetInnerHTML` or `innerHTML` with data arriving from chart input, an LLM response, or a live feed; render as text and let React escape it. Confirmed nothing in `src/` uses it today, so this is preventive rather than a fix.
- **`rules/service.md`** — no API keys in source, and treat chart input, LLM responses, and live data as untrusted at the boundary. Names `chat.ts` and the LLM services specifically, since that's where it would slip.

This is arguably a better home than the skill had: it now loads automatically when someone opens a UI or service file, instead of only when a reviewer remembered to invoke a checklist.

### 1. `/debug-maidr` pointer in `copilot-instructions.md` — declining

I'd already changed this line in an earlier round, and I think it's landed where it should. It currently reads:

> The full workflow and the per-layer checklists are in `.claude/skills/debug-maidr/SKILL.md`; in Claude Code, run `/debug-maidr`.

The real file path comes first and is readable from either side; the slash command is explicitly qualified as Claude Code-specific. A Copilot reader isn't sent somewhere that doesn't exist — they're told where the content lives and that the shortcut belongs to the other tool.

Adding a Copilot-side pointer to `.github/agents/debugger.agent.md` would mean either putting Copilot routing into `CLAUDE.md` (the Claude-facing source, which would then carry instructions for a tool it isn't for), or teaching the generator to inject tool-specific text — a templating layer I declined earlier and would rather not add for one qualified sentence.

### Verification

Lint clean, **1129 tests across 89 suites**, mirror in sync — and the new cross-reference rewrote correctly, so `service.instructions.md` points at `ui.instructions.md` rather than `rules/ui.md`.

---

**Comment by jooyoungseo** [comment]
Item 1 fixed in `ef13c6a`. Item 2 is a maintainer call — details below.

### 1. `debugger.md` duplicating the skill it preloads

You're right, and it's the sharpest version of the point: `skills:` injects the skill **in full**, so that agent was starting every run with two copies of the same layer guide and symptom table. Exactly the drift risk this PR exists to remove, and directly against what `.claude/README.md` says agent bodies are for. A leftover from trimming the *other* agents and not re-reading this one after adding `skills:` to it.

Before deleting, I diffed the agent's content against the skill's and found two things that weren't in the skill, so they moved rather than went away:

- **A chat row for the symptom table** — `chat.ts` / `llmValidation.ts`, unset or rejected API key, or a promise that never settles. The skill's table had no chat entry at all.
- **The `Movable` strategies in the model checklist** — `MovableGrid`, `MovableGraph`, `MovablePlane`. A trace moving in the wrong direction is often wired to the wrong one.

Both verified against `src/` first (`src/model/movable.ts` exports all three; `heatmap.ts` uses `MovableGrid`) rather than propagated on the old file's say-so — the old docs in this repo had a track record of stale claims.

Agent is now role-only: how to work, what to report, how to finish. 79 → 59 lines, reference material in one place.

### 2. `copilot-sync` and branch protection

Splitting this one, because half is mine and half isn't:

**In `ci.yml`, which this PR controls:** no job has `needs: [copilot-sync]`, so it can't block the pipeline. That's deliberate and matches the description — the check is cheap and dependency-free, and gating `Build`/`Unit Tests` behind a docs-mirror check would be the wrong trade.

**Whether it's a *required* status check** is branch-protection configuration on `main`, which lives in repo settings, not in this diff — I can't read or change it from here, and wouldn't want to alter branch protection as a side effect of a config PR.

My recommendation: **add it as required.** Your reasoning is right that drift nobody notices defeats the purpose. The failure mode it catches — Copilot silently working from stale conventions — is invisible in review precisely because the generated files look plausible. But that's a maintainer decision, so flagging rather than acting.

### Verification

Lint clean, **1129 tests across 89 suites**, mirror in sync.

---

**Comment by jooyoungseo** [comment]
Items 1 and 3 acted on in `be8ac4b`. Item 2 was verified against the docs; item 4 is a fair limitation, now written down.

### 1. `implementer` lost research access — you were right, and it wasn't intentional

Checked, and it was silently dropped rather than cut:

```
$ git show origin/main:.claude/agents/react-ts-engineer.md | grep -E "^tools:|^model:"
model: inherit
```

No `tools:` key, so it inherited everything — including `WebFetch` and `WebSearch` — and its body had a "Research When Uncertain" step. The merge kept `implementer`'s existing narrow allowlist and dropped both. I merged the bodies and didn't audit the frontmatter difference. Thanks for catching it.

Restored deliberately, with a guardrail rather than just re-adding the tools. For a library built on Web Audio, ARIA, and SVG, looking up the spec beats guessing — but the risk is importing a web pattern that fights the project's conventions, so the new step reads:

> **Look it up rather than guess** when the uncertainty is external — a Web Audio behaviour, an ARIA pattern, an SVG detail, a chart library's API. The codebase is the authority on convention; the specification is the authority on the platform. Never import a pattern found online that contradicts the project rules.

### 2. How the `Task(...)` allowlist claim was verified

Reasonable thing to want evidence for, since if the allowlist *had* been live I'd have deleted a guardrail. It's stated directly in the [subagents documentation](https://code.claude.com/docs/en/sub-agents):

> The `Agent(agent_type)` allowlist syntax applies only to an agent running as the main thread with `claude --agent`. In a subagent definition, listing `Agent` in `tools` lets that subagent spawn subagents of its own while the depth limit allows it, **but any type list inside the parentheses is ignored**.

`feature-builder` is a subagent definition, so the parenthesised list was inert. The same page notes `Task(...)` still resolves as an alias for `Agent`, so the old form wasn't broken — it just wasn't restricting anything. The "delegate only to these six" sentence in the body is the only mechanism available either way, which is why it's now stated explicitly instead of implied by syntax that does nothing.

### 3. Hand-maintained agent mirror — parity now checked

Holding the position that agents shouldn't be *generated*: the schemas genuinely disagree (`permissionMode`/`memory`/`skills`/`maxTurns` vs `handoffs`/display `name`/a different tool vocabulary), and a mapping table would be its own bug source — this PR has already produced five bugs in one generator, which is evidence enough about mapping layers.

But you're right that the drift risk is the same class, and the *realistic* drift is cheap to catch: someone adds, removes, or renames an agent on one side only. `test/scripts/agentParity.test.ts` asserts the two directories describe the same roles, plus three things worth pinning while I was there — `name` matches the filename, descriptions stay short (they load on every request), and a `skills:` entry names a skill that exists. Both drift cases were simulated to confirm it fails on them.

**It immediately caught a false claim of mine.** I'd written that all descriptions are "under 250 characters"; `feature-builder` was exactly 250. Reworded to 201 — so the statement is now true rather than the test being loosened to match it.

### 4. `git ls-files` vs runtime glob semantics

Agreed, and you characterised it correctly. Exact parity would mean reimplementing both tools' matchers and then trusting the reimplementation — more surface than it buys. It catches the realistic failure (a typo'd path matching nothing anywhere) and doesn't prove byte-for-byte agreement. That caveat is now in the test file's header comment rather than only in this thread.

### Verification

Lint clean, **1133 tests across 90 suites**, mirror in sync.

---

**Comment by jooyoungseo** [comment]
All four addressed in `1061d05`. Two needed correcting, one needed measuring, one you diagnosed slightly differently than it turned out.

### 1. `mkdirSync` under `--check` — real, fixed

You're right that a documented read-only mode shouldn't write. It was also redundant: `listMarkdown` already tolerates a missing directory, and each write creates its own parent. So the call was pure side effect and is simply gone, with a test asserting a check run leaves no directory behind.

### 2. Nested braces — measured, and it's a duplicate rather than a failure

Checked before changing anything, and the behaviour is slightly better than described:

```
a/*.{ts,tsx}        -> ["a/*.ts","a/*.tsx"]
a/*.{ts,{jsx,tsx}}  -> ["a/*.ts","a/*.jsx","a/*.ts","a/*.tsx"]
{a,b}/{c,d}/*.ts    -> ["a/c/*.ts","a/d/*.ts","b/c/*.ts","b/d/*.ts"]
```

Nesting *does* expand — the regex `\{([^{}]*)\}` matches the innermost group, so recursion unwinds correctly. What it produces is a **duplicate** (`a/*.ts` twice, once from each branch of the outer group), not a confusing expansion. Harmless in `applyTo`, but noise, so it's deduplicated now and pinned by a test.

Worth noting because it changes the risk: there was never a case here that would silently mis-scope a rule.

### 3. `no-console` — verified rather than assumed, and the claim was imprecise

Good instinct to flag an unverifiable claim in agent context; that's exactly the stale-docs problem this PR exists to fix. I could check, since I have dependencies installed. Dropped a probe file into `src/`:

```
$ npx eslint src/zz-probe.ts
  2:3  error  Unexpected console statement. Only these console methods are allowed: warn, error  no-console
```

So the rule is real — but the wording was imprecise in a way that could mislead. "`no-console` is enforced by lint" reads as a blanket ban, while `console.warn` and `console.error` are allowed, and `rules/typescript.md` elsewhere tells people to log errors with context. An agent could reasonably have concluded it should strip a legitimate `console.error` from a catch block. Both the skill and the debugger agent now name the exemptions.

### 4. No local test that the real repo is in sync

Fair, and cheap to close. `claudeRuleGlobs.test.ts` now runs the generator with `--check` against the actual repo, so a rule edited without regenerating fails on `npm test` instead of after a push. The CI job still backstops it; this just shortens the loop.

### On `copilot-sync` gating nothing

Flagging again since you raised it: within `ci.yml` no job has `needs:` on it, which is deliberate. Making it a **required status check** is branch-protection config on `main` — outside this diff, and not something I'll change as a side effect. I'd recommend adding it, for the reason you give.

### Verification

Both new cases fail against `be8ac4b` and pass here. Lint clean, type-check clean, **1136 tests across 90 suites**.

---

PR #654, opened 2026-07-29
feat(settings): report maidr.js version, bundle source and browser in the settings dialog

jooyoungseo:
## Description

The settings dialog (`Ctrl` + `,` / `Cmd` + `,`) now ends with an **About** section reporting everything a maintainer needs to reproduce a bug, plus a button that copies it as a plain-text block:

| Row | Example |
| --- | --- |
| maidr.js Version | `3.74.0` |
| Loaded From | `CDN` — `https://cdn.jsdelivr.net/npm/maidr@latest/dist/maidr.js` |
| Browser | `Chrome 141` |
| Operating System | `Linux` |
| Diagnostics | **Copy diagnostics** |

"Loaded From" is the field that is hardest to get any other way and the one most likely to explain a mismatch: py-maidr and r-maidr can load maidr.js either from jsDelivr at `@latest` or from the copy they ship, so the running bundle is not always the version the binding expects.

## Related Issues

> [!IMPORTANT]
> **Read before merging — #655 makes half of this feature's accessibility work inert on arrival.**
>
> Every MAIDR dialog is currently hidden from the accessibility tree while open: MUI's `ModalManager` marks the MAIDR wrapper `aria-hidden="true"`, and because the dialogs use `disablePortal` they sit inside that wrapper, so each one hides itself.
>
> The consequence for *this* PR: the copy-status live region **will not be announced to a screen reader user today**. The button works, the clipboard write works, the visible text updates — the confirmation is silent.
>
> This is pre-existing and repo-wide, not introduced here. The Volume slider, Reset, Close and Save & Close are equally unreachable today, so merging this does not regress anything, and holding it does not help anyone. But the live region, `aria-describedby` wiring and WCAG 2.5.3 label-in-name work in this PR are unobservable in practice until #655 lands, so the two are best landed close together.

Also filed while working on this: #656 (<kbd>Esc</kbd> does not close the settings dialog) and #657 (no component-test harness — jest is node-only and skips `.tsx`).

## Changes Made

**`src/util/version.ts` (new)** — exports `MAIDR_VERSION`, read from the `version` field of `package.json`. semantic-release already owns that field, so there is no second source of truth to keep in sync, and bundlers inline the value at build time.

**`src/util/diagnostics.ts` (new)** — browser and OS names parsed from the user agent, plus bundle-source detection:

- `document.currentScript` is snapshotted at module evaluation. For a classic script that is the tag which loaded the bundle — the path both bindings use, CDN or bundled copy alike. A `<script>` with no `src` means the bundle is embedded in the page.
- Module scripts leave `currentScript` null, so a scan of maidr-shaped `script[src]` URLs backs it up. That scan takes the first match in document order, so the patterns are deliberately narrow: a bare `/maidr/` segment counts only directly under `/npm/` or `/node_modules/`, because the project's own docs are served from `xability.github.io/maidr/` where every asset URL contains it.
- Anything still unattributed is reported as `Unknown` rather than guessed at. An unversioned `unpkg.com/maidr/dist/…` falls in here — declining to answer beats naming the wrong script in a field that exists to be trusted.
- CDN vs local is decided by comparing origins, with a protocol check first because every `file://` URL reports its origin as the opaque string `"null"` — without that, any two `file://` URLs would compare as same-origin.

Only what the user agent can actually support is reported: Windows 10 and 11 are indistinguishable there, Safari/Chrome freeze the macOS version at 10.15.7, and iPadOS Safari claims `Macintosh` with no iPad token. Edge, Opera and Samsung Internet are matched before Chrome, whose token they all carry, and Opera is matched under all three of its per-platform tokens (`OPR`, `OPiOS`, `OPT`).

**`src/util/clipboard.ts` (new)** — `navigator.clipboard` where available, falling back to `execCommand` otherwise, since the async API is unavailable on the `file://` pages py-maidr and r-maidr write charts to. The fallback restores focus to the copy button afterwards, and deliberately does not mark the throwaway textarea `aria-hidden`, since `select()` focuses it and WAI-ARIA forbids that combination.

**`src/ui/component/Settings.tsx`** — the About section, built from the existing `SettingRow` layout. The copy status is a `role="status" aria-live="polite"` region rendered while empty, keyed by an attempt counter so a repeat copy still mutates the DOM — identical text alone produces no mutation, and a live region announces on the mutation, not on the state update.

**Tests (new)** — `version.test.ts`, `diagnostics.test.ts`, `clipboard.test.ts`, 44 cases over the pure logic: browser and OS families including the ones that shadow another's token, script-URL matching including the false positives that would shadow the real bundle, `detectMaidrSource`'s document scan, origin classification, URL redaction, the clipboard fallback chain, and the focus restore.

### Privacy

The copied block is written to be pasted into a public issue, so two things are held back — and the dialog applies the same rule to what it displays, since a screenshot travels just as far as pasted text:

- **The page URL is omitted entirely.** It is the field most likely to carry private paths or credentials in a query string.
- **The script URL is redacted.** A `file://` bundle sits wherever the reporter saved it, so its path carries their OS username; only the protocol and filename survive. Hosted URLs keep origin and path — that is what exposes a jsDelivr `@latest` against a pinned local copy — and lose only the query and fragment, where a signed asset URL would carry a token.

Bottom of the dialog, from `examples/barplot.html` after clicking **Copy diagnostics**:

```
About

maidr.js Version     3.74.0
Loaded From          Local assets
                     file:///.../maidr.js
Browser              Chrome 141
Operating System     Linux
Diagnostics          [ COPY DIAGNOSTICS ]  Copied to clipboard
```

The clipboard block, read back after the copy:

```
MAIDR diagnostics
maidr.js version: 3.74.0
Loaded from: CDN (http://127.0.0.1:8002/maidr.js)
Browser: Chrome 141
Operating system: Linux
User agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/141.0.0.0 Safari/537.36
```

## Checklist

### Verification

- `npm run lint` over `src` and `test` — clean
- `npm test` — 94 suites, 1195 tests, all passing
- `npm run build` — all 12 bundle targets built

**Bundle content**, checked across every artifact rather than one: 11 artifacts across the 7 React-bundling targets embed `"3.74.0"`; **zero** artifacts contain `devDependencies` or `"semantic-release"`, including `chartjs.js` and `amcharts.js` with their differing externals. So the tree-shaking holds across the differing plugin configs, not just the core bundle.

**Manual browser pass** against the built bundle, since the React glue has no automated coverage — 19 checks, all passing: the four rows render with real values, the live region exists and is empty before the first click, three clicks produce three DOM mutations, the button is keyboard-focusable and activates on <kbd>Enter</kbd>, the clipboard payload is complete and carries no home directory or page URL, and on the `execCommand` fallback the copy succeeds, focus returns to the button, and no textarea is left behind.

### Known gaps

No component-level test for the `Settings.tsx` wiring and no e2e assertion for the new rows. jest here runs `testEnvironment: 'node'` with `testMatch: ['**/test/**/*.test.ts']`, so a component test needs `jest-environment-jsdom` + `@testing-library/react` and an environment switch — repo-wide infrastructure, tracked as #657. The e2e page object is shared across all 12 specs, which is wider than this change needs.

Docs are unchanged: `docs/CONTROLS.md` lists the shortcut but does not describe the dialog's contents.

### DIFF
--- a/src/ui/component/Settings.tsx
+++ b/src/ui/component/Settings.tsx
@@ -52,11 +52,24 @@ import {
   selectBraillePreset,
   SINGLE_LINE_BRAILLE_PRESETS,
 } from '@util/braillePreset';
+import { copyToClipboard } from '@util/clipboard';
+import {
+  collectDiagnostics,
+  describeMaidrSource,
+  formatDiagnostics,
+} from '@util/diagnostics';
 import { resolveVersionOptions } from '@util/llm';
-import { MAIDR_VERSION } from '@util/version';
-import React, { useCallback, useEffect, useId, useState } from 'react';
+import React, { useCallback, useEffect, useId, useMemo, useState } from 'react';
 
 const MIN_CUSTOM_INSTRUCTION_LENGTH = 10;
+
+type CopyStatus = 'idle' | 'copied' | 'failed';
+
+const COPY_STATUS_MESSAGE: Record<CopyStatus, string> = {
+  idle: '',
+  copied: 'Copied to clipboard',
+  failed: 'Could not copy — select the values above and copy them manually',
+};
 
 // Letter portion of the dialog accelerator keys. Shared between the
 // keydown handler and the aria-keyshortcuts attributes so the two
@@ -407,6 +420,22 @@ const Settings: React.FC = () => {
   // arrives in a coherent shape and does not need to re-normalize here.
   const [generalSettings, setGeneralSettings] = useState<GeneralSettings>(general);
   const [llmSettings, setLlmSettings] = useState<LlmSettings>(llm);
+
+  const [copyStatus, setCopyStatus] = useState<CopyStatus>('idle');
+  const copyStatusId = `${id}-copy-status`;
+  // The bundle source and the browser cannot change while the dialog is open,
+  // so the DOM scan behind this runs once per mount rather than per render.
+  const diagnostics = useMemo(() => collectDiagnostics(), []);
+
+  const handleCopyDiagnostics = useCallback(async (): Promise<void> => {
+    try {
+      await copyToClipboard(formatDiagnostics(diagnostics));
+      setCopyStatus('copied');
+    } catch (error) {
+      console.error('[Settings] Failed to copy diagnostics', error);
+      setCopyStatus('failed');
+    }
+  }, [diagnostics]);
 
   useEffect(() => {
     viewModel.load();
@@ -1171,7 +1200,85 @@ const Settings: React.FC = () => {
             <SettingRow
               label="maidr.js Version"
               input={(
-                <Typography variant="body2">{MAIDR_VERSION}</Typography>
+                <Typography variant="body2">{diagnostics.version}</Typography>
+              )}
+            />
+          </Grid>
+          <Grid size={12}>
+            <SettingRow
+              label="Loaded From"
+              alignLabel={diagnostics.source.url ? 'flex-start' : 'center'}
+              input={(
+                <>
+                  <Typography variant="body2">
+                    {describeMaidrSource(diagnostics.source)}
+                  </Typography>
+                  {diagnostics.source.url && (
+                    <Typography
+                      variant="caption"
+                      sx={{ color: 'text.secondary', wordBreak: 'break-all' }}
+                    >
+                      {diagnostics.source.url}
+                    </Typography>
+                  )}
+                </>
+              )}
+            />
+          </Grid>
+          <Grid size={12}>
+            <SettingRow
+              label="Browser"
+              input={(
+                <Typography variant="body2">{diagnostics.browser}</Typography>
+              )}
+            />
+          </Grid>
+          <Grid size={12}>
+            <SettingRow
+              label="Operating System"
+              input={(
+                <Typography variant="body2">
+                  {diagnostics.operatingSystem}
+                </Typography>
+              )}
+            />
+          </Grid>
+          <Grid size={12}>
+            <SettingRow
+              label="Diagnostics"
+              input={(
+                <Grid container spacing={1} alignItems="center">
+                  <Grid size="auto">
+                    <Button
+                      variant="outlined"
+                      color="inherit"
+                      size="small"
+                      onClick={handleCopyDiagnostics}
+                      aria-label="Copy diagnostics to clipboard"
+                      aria-describedby={copyStatusId}
+                    >
+                      Copy diagnostics
+                    </Button>
+                  </Grid>
+                  <Grid size="auto">
+                    {/* Rendered even while empty: a live region has to be in
+                        the DOM before its text changes for the update to be
+                        announced. */}
+                    <Typography
+                      id={copyStatusId}
+                      variant="caption"
+                      role="status"
+                      aria-live="polite"
+                      sx={{
+                        color: copyStatus === 'failed'
+                          ? 'error.main'
+                          : 'text.secondary',
+                      }}
+                    >
+                      {COPY_STATUS_MESSAGE[copyStatus]}
+                    </Typography>
+                  </Grid>
+                </Grid>
               )}
             />
           </Grid>
--- a/src/ui/component/Settings.tsx
+++ b/src/ui/component/Settings.tsx
@@ -53,6 +53,7 @@ import {
   SINGLE_LINE_BRAILLE_PRESETS,
 } from '@util/braillePreset';
 import { resolveVersionOptions } from '@util/llm';
+import { MAIDR_VERSION } from '@util/version';
 import React, { useCallback, useEffect, useId, useState } from 'react';
 
 const MIN_CUSTOM_INSTRUCTION_LENGTH = 10;
@@ -1148,6 +1149,32 @@ const Settings: React.FC = () => {
               </Grid>
             </Grid>
           )}
+        </Grid>
+
+        <Grid size={12}>
+          <Divider className="settings-divider" />
+        </Grid>
+
+        {/* About */}
+        <Grid container spacing={0.5} className="settings-section">
+          <Grid size={12}>
+            <Typography
+              variant="h6"
+              fontWeight="bold"
+              gutterBottom
+              className="settings-section-title"
+            >
+              About
+            </Typography>
+          </Grid>
+          <Grid size={12}>
+            <SettingRow
+              label="maidr.js Version"
+              input={(
+                <Typography variant="body2">{MAIDR_VERSION}</Typography>
+              )}
+            />
+          </Grid>
         </Grid>
 
         <Grid size={12}>
--- a/src/ui/component/Settings.tsx
+++ b/src/ui/component/Settings.tsx
@@ -57,6 +57,7 @@ import {
   collectDiagnostics,
   describeMaidrSource,
   formatDiagnostics,
+  redactScriptUrl,
 } from '@util/diagnostics';
 import { resolveVersionOptions } from '@util/llm';
 import React, { useCallback, useEffect, useId, useMemo, useState } from 'react';
@@ -426,6 +427,13 @@ const Settings: React.FC = () => {
   // The bundle source and the browser cannot change while the dialog is open,
   // so the DOM scan behind this runs once per mount rather than per render.
   const diagnostics = useMemo(() => collectDiagnostics(), []);
+  // Displayed with the same redaction the copied block uses: a screenshot of
+  // this dialog is handed to a maintainer just as readily as the pasted text,
+  // so both have to drop the OS username and any signed-URL token.
+  const sourceUrl = useMemo(
+    () => (diagnostics.source.url ? redactScriptUrl(diagnostics.source.url) : null),
+    [diagnostics],
+  );
 
   const handleCopyDiagnostics = useCallback(async (): Promise<void> => {
     try {
@@ -1207,18 +1215,18 @@ const Settings: React.FC = () => {
           <Grid size={12}>
             <SettingRow
               label="Loaded From"
-              alignLabel={diagnostics.source.url ? 'flex-start' : 'center'}
+              alignLabel={sourceUrl ? 'flex-start' : 'center'}
               input={(
                 <>
                   <Typography variant="body2">
                     {describeMaidrSource(diagnostics.source)}
                   </Typography>
-                  {diagnostics.source.url && (
+                  {sourceUrl && (
                     <Typography
                       variant="caption"
                       sx={{ color: 'text.secondary', wordBreak: 'break-all' }}
                     >
-                      {diagnostics.source.url}
+                      {sourceUrl}
                     </Typography>
                   )}
                 </>
--- a/src/ui/component/Settings.tsx
+++ b/src/ui/component/Settings.tsx
@@ -72,6 +72,17 @@ const COPY_STATUS_MESSAGE: Record<CopyStatus, string> = {
   failed: 'Could not copy — select the values above and copy them manually',
 };
 
+interface CopyState {
+  readonly status: CopyStatus;
+  /**
+   * Counts attempts so a repeat copy still reaches the user. Setting the same
+   * status twice changes no text, and unchanged text is no DOM mutation — a
+   * live region announces on the mutation, not on the state update, so without
+   * this a second successful copy would be silent to a screen reader.
+   */
+  readonly attempt: number;
+}
+
 // Letter portion of the dialog accelerator keys. Shared between the
 // keydown handler and the aria-keyshortcuts attributes so the two
 // cannot drift apart and announce a shortcut that no longer fires.
@@ -422,7 +433,7 @@ const Settings: React.FC = () => {
   const [generalSettings, setGeneralSettings] = useState<GeneralSettings>(general);
   const [llmSettings, setLlmSettings] = useState<LlmSettings>(llm);
 
-  const [copyStatus, setCopyStatus] = useState<CopyStatus>('idle');
+  const [copyState, setCopyState] = useState<CopyState>({ status: 'idle', attempt: 0 });
   const copyStatusId = `${id}-copy-status`;
   // The bundle source and the browser cannot change while the dialog is open,
   // so the DOM scan behind this runs once per mount rather than per render.
@@ -436,13 +447,16 @@ const Settings: React.FC = () => {
   );
 
   const handleCopyDiagnostics = useCallback(async (): Promise<void> => {
+    let status: CopyStatus;
     try {
       await copyToClipboard(formatDiagnostics(diagnostics));
-      setCopyStatus('copied');
+      status = 'copied';
     } catch (error) {
       console.error('[Settings] Failed to copy diagnostics', error);
-      setCopyStatus('failed');
+      status = 'failed';
     }
+    // Always a fresh object, so an unchanged status still re-renders.
+    setCopyState(prev => ({ status, attempt: prev.attempt + 1 }));
   }, [diagnostics]);
 
   useEffect(() => {
@@ -1278,12 +1292,18 @@ const Settings: React.FC = () => {
                       role="status"
                       aria-live="polite"
                       sx={{
-                        color: copyStatus === 'failed'
+                        color: copyState.status === 'failed'
                           ? 'error.main'
                           : 'text.secondary',
                       }}
                     >
-                      {COPY_STATUS_MESSAGE[copyStatus]}
+                      {/* Keyed by attempt so React swaps the node on every
+                          copy. Re-rendering the same text would leave the
+                          region untouched, and an untouched live region is
+                          never announced. */}
+                      <span key={copyState.attempt}>
+                        {COPY_STATUS_MESSAGE[copyState.status]}
+                      </span>
                     </Typography>
                   </Grid>
                 </Grid>
--- a/src/util/clipboard.ts
+++ b/src/util/clipboard.ts
@@ -45,9 +45,13 @@ function copyWithExecCommand(text: string): void {
   const textarea = document.createElement('textarea');
   textarea.value = text;
   textarea.setAttribute('readonly', '');
-  // Hidden, but not via `display`/`visibility`/`hidden` — those make the
-  // selection empty and turn the copy into a silent no-op.
-  textarea.setAttribute('aria-hidden', 'true');
+  // Kept off-screen rather than hidden: `display`, `visibility` and the
+  // `hidden` attribute all empty the selection and turn the copy into a silent
+  // no-op. `aria-hidden` is deliberately NOT set either — `select()` below puts
+  // focus on this element, and WAI-ARIA forbids `aria-hidden` on focused
+  // content, which leaves assistive technology with undefined behaviour at the
+  // exact moment the user asked for something to happen. Off-screen and
+  // short-lived is the honest trade.
   textarea.style.position = 'fixed';
   textarea.style.top = '-1000px';
   textarea.style.opacity = '0';
--- a/src/util/clipboard.ts
+++ b/src/util/clipboard.ts
@@ -1,0 +1,65 @@
+/**
+ * Writes text to the system clipboard.
+ *
+ * `navigator.clipboard` is unavailable on insecure origins and on `file://`
+ * pages — both of which MAIDR runs on routinely, since py-maidr and r-maidr
+ * write charts to standalone HTML files — so a `execCommand` fallback covers
+ * the cases the async API refuses.
+ * @param text - The text to place on the clipboard.
+ * @throws Error if neither the async API nor the fallback can copy.
+ */
+export async function copyToClipboard(text: string): Promise<void> {
+  if (typeof navigator !== 'undefined' && navigator.clipboard) {
+    try {
+      await navigator.clipboard.writeText(text);
+      return;
+    } catch (error) {
+      // Not fatal: the async API also rejects when the document is not
+      // focused or the permission was denied, and the fallback still works
+      // in both cases. Logged rather than swallowed so a genuine failure of
+      // both paths can be traced.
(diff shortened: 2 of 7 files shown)


jooyoungseo:
Addressed the review in `6f4bcbc`.

**Clipboard tests** — added `test/util/clipboard.test.ts`, 9 cases: async API success, async rejection handing over to `execCommand`, async API absent, `execCommand` refusing the copy, focus restoration on both the success and failure paths, nothing focused beforehand, the no-`document` guard, and that the textarea is hidden without `display`/`visibility`/`hidden` (any of which would empty the selection and make the copy a silent no-op).

`jest-environment-jsdom` is not a dependency here — jest runs `testEnvironment: 'node'` — so the DOM surfaces are stubbed rather than adding one for a single file. Mutation-checked to confirm they are not vacuous: deleting `previouslyFocused?.focus()` fails exactly the two focus-restore cases and nothing else.

**Script pattern false positives** — real bug, fixed. Both of the examples matched: `/maidr-analytics/tracker.js` via the directory alternative and `maidrical.js` via the filename one. A version separator now has to look like a version (`@` takes any npm spec, `-` must be followed by a digit) and a filename has to continue at a separator. Since `findMaidrScriptUrl` returns the *first* match, a false positive would shadow the real bundle, so this is covered by tests now — including the two names above.

The first attempt at that regex tripped `regexp/no-super-linear-backtracking`: `-` was in both the separator and segment classes, letting the repetition reach itself. The classes are disjoint now.

**Inlined module scripts reported as `unknown`** — correct, and it is inherent rather than incidental: `document.currentScript` is null while a module evaluates, so at that point nothing distinguishes an inlined `<script type="module">` from a bundle rolled into a host application's chunk. Reporting `unknown` is the honest answer, and the reasoning is now recorded at the branch.

**UMD spot-check** (from the first review) — `dist/maidr.js` *is* the UMD build; it carries the `define.amd` wrapper, contains `"3.74.0"`, and greps clean for `devDependencies`/`semantic-release`. The other 11 bundles do not include `Settings.tsx` at all.

**Not changed:** `expect(MAIDR_VERSION).toBe(packageVersion)`. It is thin, but it is the thing that fails if `version.ts` ever stops being a straight re-export — cheap enough to keep. The odd `describe('mAIDR_VERSION')` title (an eslint autofix artefact) is renamed.

Re-verified after the change: `type-check`, `eslint src test`, `npm test` (94 suites / 1178 tests), `npm run build` all clean, plus a browser re-run of the settings dialog confirming detection and the copy still behave as before.

---

jooyoungseo:
Confirmed and fixed in `4ee30b9`. The analysis was right, including the GitHub Pages scenario — reproduced before touching anything:

```
gh-pages docs bundle : true    <- false positive
gh-pages real bundle : true
```

The optional suffix did collapse the alternative to "contains `/maidr/`", and since `findMaidrScriptUrl` takes the first match in document order, a docs asset emitted before the chart script would have been reported as the running bundle — wrong in exactly the field this feature exists to make trustworthy.

**Fix.** A bare `/maidr/` segment now counts only directly under `/npm/` or `/node_modules/`; anywhere else the version has to be present. Split into two patterns (`MAIDR_PACKAGE_DIR_PATTERN`, `MAIDR_FILENAME_PATTERN`) so each carries its own reasoning instead of one dense literal.

Of the two suggested approaches I took both, split by case, rather than either alone: requiring `/npm/`-or-`/node_modules/` *and* making the version non-optional elsewhere. Requiring the markers alone would have dropped `unpkg.com/maidr@3.74.0/dist/recharts.mjs`, which has neither.

**Accepted trade-off.** An unversioned `unpkg.com/maidr/dist/recharts.mjs` now reports `unknown`. That is the right way to be wrong here — declining to answer beats naming the wrong script — and it is called out in the pattern's doc comment and pinned by a test so it reads as a decision rather than an oversight.

**Coverage gap closed.** `detectMaidrSource` is now tested directly, which was the better half of the suggestion: `document.currentScript` is null under `testEnvironment: 'node'`, and that is precisely the module-script path that falls back to scanning the document — so a stubbed `document`/`window` exercises the real fallback rather than a mock of it. Four cases, including the two-script GitHub Pages ordering. Mutation-checked against the old pattern: exactly 3 of the new tests fail, and they pass with the fix.

**iPadOS.** Agreed it is not this PR's to solve, but I pinned it with a test and a line in the JSDoc so it is a documented answer rather than a surprise in a bug report.

**Settings.tsx component/e2e coverage** — still out of scope, noted so it does not get lost.

Full gate re-run: `eslint src test` (exit 0), `type-check`, `npm test` (94 suites / **1185** tests), `npm run build` (12 bundles), plus a browser re-run of the dialog confirming detection and the copy still behave.

---

jooyoungseo:
**1. `file://` path leak — agreed and fixed in `7fb2e00`.** You're right that my own reasoning for dropping the page URL applies here, and I'd left the more sensitive of the two in: a `file://` bundle sits wherever the reporter saved it, so the path carries their OS username.

I took the suggested split — redact the copied text, leave the dialog caption alone — but scoped the redaction by URL kind rather than blanket-stripping the path, because the path is not uniformly noise:

- `file://` → protocol and filename only. `file:///Users/jane.doe/reports/dist/maidr.js` becomes `file:///.../maidr.js`. The filename still says which bundle; the label already says it was local.
- Hosted → origin and path kept. `https://cdn.jsdelivr.net/npm/maidr@latest/dist/maidr.js` is exactly the mismatch this field exists to expose, and stripping the path would throw away the `@latest` that makes it actionable.
- Query and fragment dropped everywhere, which the original report did not cover — a signed asset URL can carry a token there.

Verified end-to-end in the browser rather than only in unit tests:

```
dialog caption  : file:///home/user/maidr/dist/maidr.js      <- stays on screen
clipboard block : Loaded from: Local assets (file:///.../maidr.js)
leaks home directory? false
```

**2. Component-level RTL test — declining, with the reason.** Not just scope: `jest.config.ts` is `testEnvironment: 'node'` with `testMatch: ['**/test/**/*.test.ts']`, so `.tsx` is not collected and there is no DOM. Landing this would mean adding `jest-environment-jsdom` and `@testing-library/react`, widening `testMatch`, and switching the environment — repo-wide test infrastructure in a feature PR, and it would be the first component test in the codebase. Worth doing deliberately as its own change; happy to open an issue for it if that's wanted.

**3. The missing `default` — I checked, and it works the opposite way round.** Adding a fifth member to `MaidrSourceKind` errors *at the switch*, not at the call site:

```
src/util/diagnostics.ts(226,59): error TS2366: Function lacks ending return
  statement and return type does not include 'undefined'.
```

The explicit `string` return type plus no `default` is what makes it exhaustive — a `default` would satisfy the return-path check and turn a compile error into a silent fallthrough. Leaving it as is, deliberately.

Re-run after the change: `eslint src test` (exit 0), `type-check`, `npm test` (94 suites / **1188** tests), `npm run build` (12 bundles), plus the browser check above.

---

jooyoungseo:
**1. Unredacted URL in the dialog — you're right, fixed in `41a9c77`.**

My "stays on the reporter's own screen" justification doesn't survive the observation that the screenshots on this very PR *are* dialog captures — I published `file:///home/user/maidr/dist/maidr.js` in the description while arguing the same value was too sensitive to paste. Drawing the line between copy and display was drawing it in the wrong place.

Both paths now go through `redactScriptUrl`, verified against the built bundle:

```
dialog caption  : file:///.../maidr.js
clipboard block : Loaded from: Local assets (file:///.../maidr.js)
leaks home directory? false
```

The visible cost is smaller than it looks: for a hosted bundle the redaction only strips query and fragment, so the origin and `@latest` that answer "which copy is running" stay on screen. Only a `file://` path loses its directory, and the `Local assets` label plus the filename already carry what that row is asked to answer.

I also updated the PR description, which was still showing the leaking path in its screenshot block.

**2. Build assertion for the tree-shaking invariant — declining, and I don't think the risk is what it looks like.** `package.json` ships to npm verbatim, so a regression here discloses nothing that isn't already public; the real cost would be a few KB of bundle size. Guarding it means a post-build content assertion, which is new CI machinery (jest never builds) for a non-secret with no correctness impact. If bundle size becomes something the repo wants to guard, that check belongs in a size budget covering all 12 bundles rather than a grep for one string — happy to open an issue if that's wanted.

**On the notes:** agreed on all, and thanks for independently confirming the ReDoS analysis on the two patterns — that one had already bitten me once (`regexp/no-super-linear-backtracking` caught an earlier revision where `-` appeared in both the separator and segment classes).

Re-run: `eslint src test` (exit 0), `type-check`, `npm test` (94 suites / **1192** tests), `npm run build` (12 bundles), plus the browser check above.

---

jooyoungseo:
**Finding confirmed and fixed in `99bd46c`.** I verified it against the built bundle rather than by reading, by patching `document.execCommand` to capture focus state at the instant it runs:

```
before:  { activeTag: "textarea", activeAriaHidden: "true", selectionLength: 257 }
after:   { activeTag: "textarea", activeAriaHidden: null,   selectionLength: 257 }
```

So focus really is on the textarea while `aria-hidden="true"` is set — the spec violation is real, and on the branch that matters most here, since `file://` pages are how py-maidr and r-maidr ship saved charts.

You're right that the attribute was never what kept it out of sight: `position`/`top`/`opacity` do that, and unlike `display`/`visibility`/`hidden` they leave the selection intact — `selectionLength` is unchanged at 257, so the copy still works. Full fallback path re-verified end to end:

```
status             : Copied to clipboard
focus restored     : true
stray textarea left: 0
```

One correction to the write-up, for anyone reading later: no Chrome console warning fires for this shape. Chrome's "Blocked aria-hidden on an element because its descendant retained focus" covers the *ancestor* case — `aria-hidden` on the focused element itself is silent, which is presumably why this survived until now. The violation stands regardless; it just won't announce itself in DevTools.

Tests updated to pin the reasoning rather than the old behaviour: the off-screen assertions now check `position`/`opacity` explicitly, and a separate case asserts `aria-hidden` is absent with the reason attached, so re-adding it fails.

**On the minor notes:**

- **Stale screenshot** — already fixed. The PR description now reads `file:///.../maidr.js`; your snapshot predates that edit by a few minutes.
- **`@util` reaching into `document`/`navigator`** — agreed it's worth a maintainer's eye, and thanks for framing it as a question rather than a verdict. My read matches yours: `svg.ts` and `platform.ts` set the precedent, and `platform.ts` in particular reads `navigator` at module scope for exactly this kind of one-shot environment fact.
- **Component-level test gap** — confirmed, and the constraint is concrete: `testEnvironment: 'node'` with `testMatch: ['**/test/**/*.test.ts']`, so it needs `jest-environment-jsdom` + `@testing-library/react` and an environment switch. Repo-wide infrastructure, worth its own change.

Re-run: `eslint src test` (exit 0), `type-check`, `npm test` (94 suites / **1193** tests), `npm run build` (12 bundles).

Noted on not being able to run the toolchain in your sandbox — CI has been green on every commit in this series, and the numbers above are from a local full run.

---

jooyoungseo:
**Repeat-click announcement gap — confirmed and fixed in `d0a594d`.** Measured with a `MutationObserver` on the live region, which is what the browser exposes to assistive technology:

```
before   click 1 -> ["childList"]        <- first copy announces
         click 2 -> ["childList"]        <- nothing new
         click 3 -> ["childList"]        <- nothing new

after    click 1 -> 1 mutation batch
         click 2 -> 2 mutation batches
         click 3 -> 3 mutation batches
```

One mutation total across three clicks, exactly as described. And it is the worst possible shape for this control: pressing Copy again is precisely what someone does when they are not sure it worked, so silence is the one response that cannot be right.

**Fix.** The status now carries an attempt counter, so every press produces a fresh state object and React cannot bail out, and the message sits in a node keyed by that counter so the node is swapped even when the text is word-for-word identical.

I went with the keyed-node approach over the two suggestions for specific reasons. Appending a visible counter changes what the user reads ("Copied to clipboard (2)") to work around a rendering detail. Clearing to `'idle'` first is the more common idiom, but here it is not deterministic: `setCopyState('idle')` sits immediately before an `await`, so whether React commits that render before the promise resolves depends on batching — it would work most of the time and fail unpredictably, which is worse than not doing it. Keying the node needs no timing assumption.

**On the minor notes:**

- **`redactScriptUrl` trailing-slash `→ null`** — intended, and now pinned by a test with the reasoning. It is unreachable from a script that actually executed, but the important property is that the redaction *fails closed*: it degrades to the bare `Local assets` label rather than falling through to the unredacted URL.
- **`aria-live="polite"` redundant with `role="status"`** — correct per spec, but keeping it. This file already pairs the two on the API-key validation status (`role="status" aria-live="polite"`), so dropping it here would make the newer code the odd one out, and the repo guidance is explicit about preserving surrounding patterns. The redundancy is also the conventional belt-and-braces for older AT/browser pairings.
- **No e2e for the About section** — confirmed, same reasoning as before; noted rather than raised as new.

Re-run: `eslint src test` (exit 0), `type-check`, `npm test` (94 suites / **1194** tests), `npm run build` (12 bundles), plus the mutation measurement above and a re-check that the `execCommand` fallback still reports `Copied to clipboard`, restores focus to the button, and leaves no textarea behind.

---

jooyoungseo:
Thanks for re-reading the diff independently rather than taking the write-ups at face value — that is the check that actually matters after this many rounds.

**ChromeOS comment — you're right, corrected in `56e4e85`.** Verified before changing it:

```
ChromeOS  contains "Linux": false  | contains "CrOS": true
Android   contains "Linux": true   | contains "CrOS": false
```

The order was right and the reason given for it was not, which is the worse kind of wrong in a comment — it would leave the next reader thinking the `CrOS` branch guards a collision that never happens. It now says Android is the one carrying `Linux`, and that ChromeOS sits there for grouping rather than dependency. Worth the commit precisely because it is load-bearing for whether someone later feels safe reordering these branches.

**Component-test gap — filed as #657.** You are the second reviewer to ask, so it should be tracked rather than left in a PR body. The issue argues it from this PR's own evidence: the two bugs found here (`aria-hidden` on a focused element, the silent repeat-click) were both caught by hand-driving a browser with a `MutationObserver`, because there was nowhere to write them as tests — and both are exactly the shape a component test catches and a pure unit test cannot.

**Real screen reader vs. `MutationObserver` — agreed, and I want to be precise about what I did and did not establish.** I measured that the DOM mutates on every click. I did not verify how NVDA or JAWS treat a node replacement versus a text-content change, and I have no way to here. Node replacement inside a live region is what React does on any keyed re-render and is well-trodden, so I am reasonably confident — but "reasonably confident" is the honest ceiling, and #657 notes that a jsdom harness would not raise it either. That one needs a human with a screen reader.

**#655 linkage** — already bidirectional: #655 calls out that this PR's copy-status region stays silent until it lands, and the PR body links back. Agreed they should not land far apart, and I have added #657 to #655's related list so the three stay findable together.

Re-run after the comment fix: `eslint src test` (exit 0), `type-check`, `npm test` (94 suites / 1194 tests), `npm run build` (12 bundles).

---

jooyoungseo:
**Opera's mobile tokens — taking this one rather than leaving it, fixed in `ffe926d`.** The pattern is now `/(?:OPR|OPiOS|OPT)\/(\d+)/`, with tests for Opera iOS and Opera Touch that fail against the old pattern.

I don't think it's the same category as the iPadOS limitation, and the difference is worth naming: iPadOS Safari genuinely cannot be distinguished from desktop Safari, so reporting `macOS` is the honest answer. Opera on iOS *is* detectable — it was just missed. And the failure mode is worse than an unknown: Opera Touch carries a `Chrome/` token, so it was being reported as **Chrome**, and a maintainer reading that has no reason to doubt it. One alternation removes it, so the known-limitations list now holds only things the user agent cannot express.

**`describeMaidrSource`'s missing `default` — I'd push back, and I checked before saying so.** You're right that `noImplicitReturns` isn't in `tsconfig.json`, but the exhaustiveness here doesn't come from that flag. It comes from the explicit `: string` return type: a fall-through path would make the function return `undefined`, which isn't assignable. Adding a fifth member to the union today:

```
src/util/diagnostics.ts(228,59): error TS2366: Function lacks ending return
  statement and return type does not include 'undefined'.
```

That's a compile error at the switch, not `undefined` reaching the DOM. So `default: return 'Unknown'` would *remove* the guarantee — it satisfies the return-path check and converts a build failure into a silently wrong label, which is the outcome the suggestion is trying to prevent. Leaving it as is, deliberately. (Flagging that this is the second review to read it this way — worth a comment in the code if it keeps coming up, though I'd rather not add one that just explains a TypeScript rule.)

**`redactScriptUrl` computed twice** — correct that it's intentional. It's a cheap pure function, and threading a precomputed string through `formatDiagnostics` would make it possible to pass an *un*redacted one, which is the failure mode worth designing against here.

Thanks for the independent check on the two fragile spots — the disjoint-class backtracking argument and the `file://` protocol comparison are exactly the two I'd want a second pair of eyes on.

Gate after the change: `eslint src test` (exit 0), `type-check`, `npm test` (94 suites / **1195** tests), and the new Opera cases mutation-checked against the old pattern.

---

jooyoungseo:
**Tree-shaking across all bundles — ran it, and it turned up a correction I owe on an earlier comment of mine.** I said further up this thread that "the other 11 bundles do not include `Settings.tsx` at all." That was wrong. Every artifact in `dist/`:

```
bundle                version  devDeps  scripts
amcharts.js                 1        0        0
anychart.js                 0        0        0
chartjs.js                  1        0        0
d3.js                       0        0        0
frappe.js                   0        0        0
google-charts.js            0        0        0
highcharts.js               0        0        0
maidr.js                    1        0        0
vegalite.js                 1        0        0
amcharts.mjs                1        0        0
anychart.mjs                0        0        0
chartjs.mjs                 1        0        0
d3.mjs                      0        0        0
frappe.mjs                  0        0        0
google-charts.mjs           0        0        0
highcharts.mjs              0        0        0
react.mjs                   1        0        0
recharts.mjs                1        0        0
vegalite.mjs                1        0        0
victory.mjs                 1        0        0
```

Eleven artifacts across seven targets embed the version — exactly the seven built with `useReact: true` in `scripts/build.js` (core, react, recharts, vegalite, chartjs, amcharts, victory), since those are the ones that bundle the UI. The five without React (anychart, d3, frappe, google-charts, highcharts) never reach `Settings.tsx`, which is why they show `0`.

The thing you actually asked about holds everywhere: **zero** artifacts contain `devDependencies` or `"semantic-release"`, including `chartjs.js` and `amcharts.js` with their different externals. So the tree-shaking generalizes across the differing plugin/external configs, not just the one artifact I'd checked.

**`currentScript` snapshot timing — good catch, documented in `d14d567`.** You're right that it's an implicit assumption. It holds today because no build target code-splits, and the failure mode if that changed is exactly what you describe: the snapshot goes `null` on classic-script loads too, every such page silently drops to the DOM scan, and the result stays *correct* while getting weaker. Silent degradation is the part worth writing down, so the assumption and its consequence are now stated at the declaration. The same edit fixes a stale `findMaidrScript` reference — the function is `findMaidrScriptUrl`.

I didn't add a regression test for it: I can't see a way to assert "this module was evaluated during the initial synchronous run" from inside the module without reproducing the assumption in the test. If you have a shape in mind I'm happy to take it.

**Component-test gap — already tracked, filed as #657** after the previous review raised the same point. It argues from this PR's evidence: the `aria-hidden` and repeat-click bugs were both found by hand-driving a browser because there was nowhere to write them as tests.

Gate: `eslint src test` (exit 0), `type-check`, `npm test` (94 suites / 1195 tests).

---

jooyoungseo:
**#655 visibility — agreed it was too easy to miss, PR description restructured.** The Related Issues section now leads with a callout stating the consequence in plain terms ("the copy-status live region will not be announced to a screen reader user today") instead of pointing at a note further down.

On making it *blocking* rather than related, I'd argue against, and the reason is the same fact that makes it serious: #655 hides the whole dialog, so Volume, Reset, Close and Save & Close are equally unreachable right now. This PR doesn't regress anything, and blocking it wouldn't get a screen reader user any closer to reading the dialog — it would just hold a working keyboard/visual feature behind an unrelated repo-wide fix. What the callout does say is that the two want to land close together, since the accessibility work here is unobservable until #655 does.

**Manual browser pass — ran a full one against the current head.** Fair ask given the React glue has no automated coverage. 19 checks, all passing:

```
✓ About heading rendered              ✓ live region present before click
✓ version row — 3.74.0                ✓ live region starts empty
✓ loaded-from row                     ✓ three clicks -> three announcements
✓ url redacted in dialog              ✓ status text after clicks
✓ browser row — Chrome 141            ✓ button is keyboard focusable
✓ os row — Linux                      ✓ Enter activates the button
✓ clipboard payload complete          ✓ fallback reports success
✓ clipboard redacted                  ✓ fallback restores focus to the button
✓ clipboard omits page URL            ✓ fallback leaves no textarea behind
✓ button accessible name contains its visible text (WCAG 2.5.3)
```

Keyboard operability was the one I hadn't explicitly checked before — the button is focusable and <kbd>Enter</kbd> activates it, producing a fourth announcement.

**Tree-shaking across bundles** — covered in my previous comment with the full per-artifact table. Short version: 11 artifacts across the 7 React-bundling targets embed `"3.74.0"`, and **zero** contain `devDependencies` or `"semantic-release"`, `chartjs.js` and `amcharts.js` included. It generalizes across the differing externals, not just the core bundle.

**`labelId` on the four static rows — I'd leave it off, and I think adding it would be a mistake.** `labelId` exists in this file to let an *input* use `aria-labelledby` instead of `aria-label`, avoiding the double announcement of the same text; its two users are the braille `Select` and the braille `RadioGroup`, both controls. The four new rows are plain text with nothing to label, and `aria-labelledby` on a role-less element has no defined mapping — so it would add markup that does nothing at best. Document order already reads "maidr.js Version, 3.74.0", which is the correct behaviour rather than a fallback.

Gate: `eslint src test` (exit 0), `type-check`, `npm test` (94 suites / 1195 tests), `npm run build` (12 targets).

---

PR #658, opened 2026-07-29
fix(settings): close the settings dialog on escape

### LINKED ISSUE #656: fix: Esc does not close the settings dialog
# Bug Report

## Description

<kbd>Esc</kbd> does not close the settings dialog, even though `SETTINGS_KEYMAP` binds it (`src/service/keybinding.ts:275`):

```ts
const SETTINGS_KEYMAP = {
  // Misc
  TOGGLE_SETTINGS: key(`esc`, 'Close Settings', { showInHelp: false }),
} as const;
```

The binding can never fire. MUI's Modal swallows the Escape keydown before it reaches the document-level `hotkeys-js` listener — `node_modules/@mui/material/Modal/useModal.js:114`:

```js
if (event.key !== 'Escape' || event.which === 229 || !isTopModal()) {
  return;
}
if (!disableEscapeKeyDown) {
  // Swallow the event, in case someone is listening for the escape key on the body.
  event.stopPropagation();
  if (onClose) {
    onClose(event, 'escapeKeyDown');
  }
}
```

`stopPropagation()` runs unconditionally whenever `disableEscapeKeyDown` is false — which is the default — and `onClose` is only called if one was passed. `Settings.tsx` passes no `onClose`, so the event is swallowed and nothing closes. `Help.tsx:53` and `Chat.tsx:212` both pass `onClose={handleClose}`, which is why <kbd>Esc</kbd> works there and only Settings is affected.

This breaks the project's own accessibility rule that anything which opens must close with <kbd>Esc</kbd>. The dialog is still closable with its Close button, so this is an escape-hatch failure rather than a trap.

## Steps to Reproduce

1. `npm run build`
3. Press <kbd>Ctrl</kbd> + <kbd>,</kbd> to open settings.
4. Press <kbd>Esc</kbd>.

## Actual Behavior

The dialog stays open. Reproduced with focus on the dialog's default target and on a plain `<button>` inside it, so this is not the `hotkeys.filter` editable-field guard in `keybinding.ts:452` — that guard returns `true` for buttons.

Side-by-side against Help, which uses the same `esc` binding pattern:

```
HELP     open  : 1 dialog
HELP     escape: 0 dialogs   <- closes
SETTINGS open  : 1 dialog
SETTINGS escape: 1 dialog    <- stays open
```

## Expected Behavior

<kbd>Esc</kbd> closes the settings dialog and returns focus to the chart, discarding unsaved changes exactly as the Close button does.

## Screenshots

Not applicable — the dialog renders correctly, it just does not respond to the key.

## Environment

- Operating System: Linux (the cause is in MUI's shared modal logic, so not platform-specific)
- Browser: Chromium 141

## Additional Information

**Suggested fix.** Pass `onClose={handleClose}` to the settings `<Dialog>`, matching `Help.tsx` and `Chat.tsx`. `handleClose` already calls `viewModel.toggle()` — the same discard-and-close path as the Close button — so the semantics line up with what the <kbd>Esc</kbd> binding intends.

That would leave `TOGGLE_SETTINGS: key('esc', …)` in `SETTINGS_KEYMAP` dead, since MUI would handle the key before `hotkeys-js` sees it. Worth deciding deliberately: either drop the entry so the keymap stops advertising a binding that never runs, or set `disableEscapeKeyDown` on the dialog so MUI stops swallowing the event and the keymap becomes the real handler. The first matches how Help and Chat already work.

**Worth checking while in here.** `Description.tsx`, `CommandPalette.tsx` and `ModelSelection.tsx` also render MUI dialogs; whichever of them relies on a keymap `esc` entry rather than `onClose` will have the same problem.

**Worth a regression test.** No current e2e spec presses <kbd>Esc</kbd> to dismiss the settings dialog — `showSettingsMenu()` in `e2e_tests/page-objects/base-page.ts` closes it another way — which is why this went unnoticed.

Found while verifying #654. Unrelated to that PR's changes, and distinct from #655 (which is about dialogs being hidden from the accessibility tree).

jooyoungseo:
# Pull Request

## Description

Escape now closes the settings dialog, discarding unsaved changes and returning focus to the chart — the same path the Close button takes.

## Related Issues

Fixes #656.

## Changes Made

**Root cause (confirmed by reproduction, not just by reading).** MUI's `Modal` calls `event.stopPropagation()` on the Escape keydown whenever `disableEscapeKeyDown` is false, which is the default. That happens before the event can reach the document-level `hotkeys-js` listener, so `SETTINGS_KEYMAP`'s `esc` binding could never fire. `Settings.tsx` passed no `onClose`, so nothing handled the key at all. Reproduced against `examples/barplot.html` on the current `main`, side by side with Help:

```
HELP     open  : 1 dialog
HELP     escape: 0 dialogs   <- closes
SETTINGS open  : 1 dialog
SETTINGS escape: 1 dialog    <- stays open
```

**`src/ui/component/Settings.tsx`** — the `<Dialog>` now takes an `onClose` that calls the existing `handleClose()`, matching `Help.tsx`, `Chat.tsx`, `Description.tsx` and `CommandPalette.tsx`. Handling it here rather than in the keymap also means Escape works from inside the dialog's text fields (API keys, custom instruction, braille size), where `hotkeys.filter` would have suppressed a keymap binding — the same reason `GoToExtrema.tsx` handles Escape at the dialog level.

One deliberate deviation from the suggested fix in the issue: `onClose` fires for `backdropClick` as well as `escapeKeyDown`, so a bare `onClose={handleClose}` would have made a click outside the dialog discard unsaved edits. The settings dialog is a form and currently ignores backdrop clicks, so the handler is guarded on `reason === 'escapeKeyDown'` to keep that behaviour unchanged. Verified: backdrop click leaves the dialog open, the Close button still closes it.

**`src/service/keybinding.ts`** — dropped the dead `TOGGLE_SETTINGS: key('esc', …)` entry from `SETTINGS_KEYMAP`, taking the first of the two options the issue lays out. I checked whether the binding could still fire in some focus state before removing it: with focus moved out of the dialog, MAIDR tears the controller down rather than leaving the dialog open, and with `disableEnforceFocus` set the keydown target still stays inside the dialog — the binding is unreachable in practice. The scope entry itself is kept (it is what stops chart shortcuts firing while the dialog is open) and now carries a comment recording why MUI owns Escape here.

**Other dialogs.** As the issue suggested, I checked the rest: `Description.tsx` and `CommandPalette.tsx` already pass `onClose` and are unaffected; `ModelSelection.tsx` is not a dialog (it renders a `Select` inside the chat bubble); `CandlestickDeltaSettings.tsx` and `GoToExtrema.tsx` are hand-rolled modals with their own Escape handling. Settings was the only one broken. The now-dead `esc` entries in `HELP_KEYMAP`, `CHAT_KEYMAP`, `DESCRIPTION_KEYMAP` and `COMMAND_PALETTE_KEYMAP` are left alone — those dialogs work, and cleaning them up is a separate change.

**Regression test** — `closeSettingsMenuWithEscape()` in `e2e_tests/page-objects/base-page.ts` opens the settings dialog and dismisses it with Escape, wired up as a spec in `dodgedBarplot.spec.ts`. It fails on the unfixed build and passes with the fix:

```
unfixed: ✘ should close settings menu with escape (3.4s)
fixed:   ✓ should close settings menu with escape (1.3s)
```

Not applicable — the dialog rendered correctly before and after; only the key handling changed.

## Checklist

## Additional Notes

**Verification run locally:**

- `npm run lint:fix`, `npm run type-check`, `npm run build` — all clean.
- `npm test` — 1195 passed, 94 suites, 0 failed.
- `npx playwright test e2e_tests/specs/dodgedBarplot.spec.ts --project=chromium` — 24 passed.
- Manual browser check on `examples/barplot.html`: Escape closes from the dialog's default focus target and from inside a text field; backdrop click leaves it open; Close button still works.

**One pre-existing failure worth flagging, unrelated to this PR.** `barplot.spec.ts` was the natural home for the regression test, but its `test.beforeAll` fails on `main` before any test body runs:

```
Error: maidr-data attribute not found on SVG element
```

The hook reads a `maidr-data` attribute off `svg#bar`, but `examples/barplot.html` uses the `var maidr = {…}` global form and that attribute is only ever set by the AnyChart adapter, so the hook cannot succeed. The same beforeAll failure takes out `boxplotHorizontal`, `boxplotVertical`, `heatmap`, `multiLayer`, `multiLineplot` and `violinPlot` — 7 specs in total, all red on a clean checkout of `main`. I put the regression test in `dodgedBarplot.spec.ts` instead so that it actually runs today. Since the e2e workflows are scheduled and use `continue-on-error`, this has not been surfacing on PRs; it looks worth a separate issue.

---
_Generated by [Claude Code](https://claude.ai/code/session_017MMPC3fcSzWtsUoekmbfrX)_

### DIFF
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -224,7 +224,7 @@ export class BasePage {
   ): Promise<void> {
     const _modal = this.page.locator(modalSelector);
     const closeButton = this.page.locator('//button[text()=\'Close\']');
-    closeButton.click();
+    await closeButton.click();
     await expect(this.page.locator(modalSelector)).not.toBeVisible({ timeout });
   }
 
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -363,6 +363,44 @@ export class BasePage {
   }
 
   /**
+   * Shows the Settings menu and verifies a backdrop click leaves it open
+   *
+   * The settings dialog is a form holding unsaved edits, so unlike the other
+   * dialogs it must not discard them on a stray click outside itself — only
+   * Escape and the Close button close it.
+   * @throws AssertionError if Settings menu does not appear or the backdrop click closes it
+   */
+  public async verifySettingsMenuIgnoresBackdropClick(): Promise<void> {
+    try {
+      await this.pressKeyCombination(
+        TestConstants.COMMAND_KEY,
+        TestConstants.COMMA_KEY,
+        'show settings menu',
+        100,
+      );
+
+      await this.verifyModal(
+        this.selectors.settingsModal,
+        TestConstants.SETTINGS_MENU_TITLE,
+      );
+
+      await this.page
+        .locator(TestConstants.MAIDR_MODAL_BACKDROP)
+        .first()
+        .click({ position: { x: 5, y: 5 }, force: true });
+
+      await expect(this.page.locator(this.selectors.settingsModal)).toBeVisible();
+
+      await this.closeSettingsMenu(this.selectors.settingsModal);
+    } catch (error) {
+      const errorMessage = error instanceof Error ? error.message : String(error);
+      throw new AssertionError(
+        `Settings menu did not survive a backdrop click: ${errorMessage}`,
+      );
+    }
+  }
+
+  /**
    * Shows the Chat dialog and verifies it appears correctly
    * @throws AssertionError if Chat Dialog does not appear or doesn't have expected content
    */
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -336,6 +336,33 @@ export class BasePage {
   }
 
   /**
+   * Shows the Settings menu and verifies the Escape key closes it
+   * @throws AssertionError if Settings menu does not appear or Escape does not close it
+   */
+  public async closeSettingsMenuWithEscape(): Promise<void> {
+    try {
+      await this.pressKeyCombination(
+        TestConstants.COMMAND_KEY,
+        TestConstants.COMMA_KEY,
+        'show settings menu',
+        100,
+      );
+
+      await this.verifyModal(
+        this.selectors.settingsModal,
+        TestConstants.SETTINGS_MENU_TITLE,
+      );
+
+      await this.closeModal(this.selectors.settingsModal);
+    } catch (error) {
+      const errorMessage = error instanceof Error ? error.message : String(error);
+      throw new AssertionError(
+        `Failed to close settings menu with Escape: ${errorMessage}`,
+      );
+    }
+  }
+
+  /**
    * Shows the Chat dialog and verifies it appears correctly
    * @throws AssertionError if Chat Dialog does not appear or doesn't have expected content
    */
--- a/e2e_tests/specs/dodgedBarplot.spec.ts
+++ b/e2e_tests/specs/dodgedBarplot.spec.ts
@@ -279,6 +279,11 @@ test.describe('Dodged Barplot', () => {
       await dodgedBarplotPage.closeSettingsMenuWithEscape();
     });
 
+    test('should keep settings menu open on backdrop click', async ({ page }) => {
+      const dodgedBarplotPage = await setupDodgedBarplotPage(page);
+      await dodgedBarplotPage.verifySettingsMenuIgnoresBackdropClick();
+    });
+
     test('should show chat dialog', async ({ page }) => {
       const dodgedBarplotPage = await setupDodgedBarplotPage(page);
       await dodgedBarplotPage.showChatDialog();
--- a/e2e_tests/specs/dodgedBarplot.spec.ts
+++ b/e2e_tests/specs/dodgedBarplot.spec.ts
@@ -274,6 +274,11 @@ test.describe('Dodged Barplot', () => {
       await dodgedBarplotPage.showSettingsMenu();
     });
 
+    test('should close settings menu with escape', async ({ page }) => {
+      const dodgedBarplotPage = await setupDodgedBarplotPage(page);
+      await dodgedBarplotPage.closeSettingsMenuWithEscape();
+    });
+
     test('should show chat dialog', async ({ page }) => {
       const dodgedBarplotPage = await setupDodgedBarplotPage(page);
       await dodgedBarplotPage.showChatDialog();
--- a/e2e_tests/utils/constants.ts
+++ b/e2e_tests/utils/constants.ts
@@ -53,6 +53,7 @@ export abstract class TestConstants {
   static readonly MAIDR_HELP_MODAL = '.MuiDialog-container';
   static readonly MAIDR_HELP_MODAL_TITLE = '.MuiDialogTitle-root h6';
   static readonly MAIDR_SETTINGS_MODAL = '.MuiDialog-container div[role="dialog"]';
+  static readonly MAIDR_MODAL_BACKDROP = '.MuiBackdrop-root';
   static readonly MAIDR_CHAT_MODAL = '.MuiDialog-container div[role="dialog"]';
 
   /**
--- a/src/service/keybinding.ts
+++ b/src/service/keybinding.ts
@@ -269,11 +269,15 @@ const REVIEW_KEYMAP = {
 
 /**
  * Keymap configuration for settings interface interactions.
+ *
+ * Deliberately empty. The settings dialog is a MUI Modal, and MUI calls
+ * `stopPropagation()` on the Escape keydown before it reaches the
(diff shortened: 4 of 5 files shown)


jooyoungseo:
Addressed the coverage gap in `1ff9c10`, though not with a component unit test.

A `Settings.test.tsx` isn't reachable from the current setup: `jest.config.ts` runs `testEnvironment: 'node'` with `testMatch: ['**/test/**/*.test.ts']`, and neither `jest-environment-jsdom` nor a testing library is in `package.json` — there are no React component tests in the repo at all. Pinning the reason-guard that way would mean adding a rendering stack and widening `testMatch` for one assertion, which is a bigger change than the bug fix it protects.

Covered the same invariant with the harness that already exists instead: `verifySettingsMenuIgnoresBackdropClick()` in `base-page.ts`, wired up as `should keep settings menu open on backdrop click`. It opens the dialog, clicks the backdrop, asserts the dialog is still visible, then closes it via the Close button.

Confirmed it's a real guard rather than a test that passes either way — with `onClose` simplified back to `handleClose`, exactly the regression you flagged:

```
✓ should show settings menu
✓ should close settings menu with escape
✘ should keep settings menu open on backdrop click (6.5s)
    Settings menu did not survive a backdrop click: expect(locator).toBeVisible() failed
```

and with the guard restored: `25 passed` for the full `dodgedBarplot.spec.ts`. `lint:fix`, `type-check`, `build` clean; `npm test` 1195 passed.

On the nit — agreed that `handleDialogKeyDown`'s Alt+C path and `handleDialogClose` are worth keeping in sync if cancel ever grows a confirm-discard step. Leaving both calling `handleClose()` directly for now, since a shared wrapper with one behavior would be abstraction without a second caller.

---

jooyoungseo:
Both points checked. One fixed in `3c2afb3`, one confirmed pre-existing.

**Unawaited `closeButton.click()` — fixed.** You're right that it only passed because the following `expect(...).not.toBeVisible()` auto-retries, and the new backdrop-click test leans on this helper to clean up after itself. Added the `await`. Verified nothing depended on the old fire-and-forget timing: `dodgedBarplot.spec.ts` 25 passed, and `histogram` + `lineplot` + `stackedBarplot` (the other specs reaching this helper via `showSettingsMenu`) 69 passed.

**`Keys = keyof Keymap[Scope]` — pre-existing, confirmed empirically.** Your reasoning is right, and I checked it rather than inferring it. Dropped a probe into `src/`:

```ts
import type { Keys } from '@type/event';
type IsNever<T> = [T] extends [never] ? true : false;
export const probe: IsNever<Keys> = true;
```

It type-checks clean on this branch *and* against `origin/main`'s `keybinding.ts` with the `esc` entry still present — so `Keys` was already `never` before this PR, exactly as you suspected. `keyof` over the union of keymap branches computes the intersection, and `CHAT_KEYMAP`/`HELP_KEYMAP` had already emptied it. Emptying `SETTINGS_KEYMAP` changes nothing there. Probe removed, not committed.

Worth someone's attention as a separate issue — `Keys` being `never` means the `commandName: Keys` annotation in `KeybindingService.register()` is checking nothing — but it's not this PR's to fix.

Thanks for tracing through the empty-keymap runtime path and the `Scope.SETTINGS` retention rationale; that matched what I verified when deciding to drop the entry rather than keep it as a dead binding.

---

jooyoungseo:
The `Keys`/`never` point is already settled — see [my earlier comment](https://github.com/xability/maidr/pull/658#issuecomment-5117856006), which answered the same question from the previous review round with a compile-time probe. Short version: `Keys` was **already** `never` on `origin/main` before this PR, so nothing here collapsed it.

Since you couldn't run `tsc` in your sandbox, the probe was:

```ts
import type { Keys } from '@type/event';
type IsNever<T> = [T] extends [never] ? true : false;
export const probe: IsNever<Keys> = true;
```

It type-checks clean against `origin/main`'s `keybinding.ts` with the `esc` entry still present, and against this branch. So no — `Keys` does not narrow as a literal union in an editor today, and it didn't before this PR either.

I also tested your suggested alternative rather than assuming: with `SETTINGS_KEYMAP = {} satisfies Record<string, KeybindingEntry>`, the probe still type-checks clean, i.e. `Keys` is still `never`. That's expected — `satisfies` doesn't widen the inferred type, so `keyof` is unchanged, and the intersection is forced to `never` by `HELP_KEYMAP` (`{TOGGLE_HELP}`) ∩ `CHAT_KEYMAP` (`{TOGGLE_CHAT}`) regardless of what `SETTINGS_KEYMAP` holds. Making the change would add noise without altering the type, so I've left `{} as const` in place. Probe removed, not committed.

The underlying quirk is real but pre-existing and wider than this PR: because `Keys` is `never`, the `commandName: Keys` annotation in `KeybindingService.register()` isn't constraining anything. Worth its own issue; not this one's to fix.

Nothing blocking outstanding on my side — no code change from this round.

---

PR #659, opened 2026-07-29
fix(ui): scope dialog aria-hidden to the modal's own parent

### LINKED ISSUE #655: fix: every MAIDR dialog hides itself from the accessibility tree while open
# Bug Report

## Description

While a MAIDR dialog is open, the dialog — and the chart behind it — are absent from the accessibility tree, so a screen reader has nothing to read.

MUI's `ModalManager` marks the modal's sibling elements `aria-hidden="true"` when a modal opens. On a MAIDR page `document.body` has a single child: the wrapper holding the MAIDR article. Because the dialogs are rendered with `disablePortal`, they live *inside* that wrapper rather than beside it, so the element MUI hides is an **ancestor** of the dialog. The dialog hides itself, along with the chart it was opened from.

The state is correct again once the dialog closes through its own control, so this is confined to the window in which a dialog is open — but that is exactly the window in which the user needs to read it.

Found while adding an About section to the settings dialog in #654. It is not caused by that PR: the untouched Help dialog behaves identically, and the pre-existing Reset / Close / Save & Close buttons are equally unreachable.

## Steps to Reproduce

1. `npm run build`
2. Open `examples/barplot.html` in a browser and focus the chart.
3. Press <kbd>Ctrl</kbd> + <kbd>,</kbd> to open the settings dialog (<kbd>Ctrl</kbd> + <kbd>/</kbd> for Help reproduces it too).
4. Inspect `document.body.firstElementChild`, or query the accessibility tree for the dialog and its buttons.

Probe output across the dialog's lifecycle:

```
                           wrapper aria-hidden | dialogs in DOM | chart in a11y tree
before any modal                          null                0                    2
settings open                             true                1                    0
closed via Close button                   null                0                    2
```

The ancestor chain, with the hidden element marked:

```
html
body
div                                                  <== aria-hidden="true"
article#maidr-article-bar
figure#maidr-figure-bar
div#react-container-bar
div.MuiDialog-root.settings-dialog                       role=presentation
div.MuiDialog-container.MuiDialog-scrollPaper            role=presentation
div.MuiPaper-root                                        role=dialog
…
button[aria-label="Copy diagnostics to clipboard"]
```

Querying by role returns nothing for any of the dialog's controls, while the same elements resolve fine by CSS selector:

```
getByRole button "Reset Settings"                 -> 0
getByRole button "Close Settings with no changes" -> 0
getByRole button "Save & Close Settings"          -> 0
getByRole dialog                                  -> 0
button[aria-label="Reset Settings"] (CSS)         -> 1, visible
```

## Actual Behavior

Opening any dialog sets `aria-hidden="true"` on the MAIDR wrapper, removing the dialog, its controls, its live regions, and the chart itself from the accessibility tree. Nothing in the dialog is announced or reachable by screen reader while it is open.

## Expected Behavior

An open dialog and its controls are present in the accessibility tree and announced normally. Content *behind* the dialog may be hidden — that is the point of the pattern — but the dialog must not be inside the hidden subtree.

## Screenshots

Not applicable — the dialog renders correctly on screen. The defect is only visible in the accessibility tree, captured in the probe output above.

## Environment

- Browser: Chromium 141

## Additional Information

**Scope.** Verified on Settings and Help. `disablePortal` is used the same way by `Chat.tsx`, `Description.tsx`, `CommandPalette.tsx` and `ModelSelection.tsx`, so the same structure applies to those dialogs.

**Why this matters beyond the usual.** MAIDR exists so blind and low-vision users can read statistical graphics. A dialog that is invisible to screen readers is not a cosmetic defect here — and any `aria-live` region inside a dialog is silent for as long as this holds, which is why #654's copy-confirmation status will not be announced until this is resolved.

**Likely fix.** Pass MUI's `container` prop pointing at the dialog's own parent, so `ModalManager` hides that parent's *other* children instead of walking up to an ancestor of the modal. `disablePortal` itself looks deliberate — MAIDR renders inside the host page's chart container so it survives notebooks and iframes — so removing it is probably not the answer. Whatever the approach, it changes `aria-hidden` and focus handling for Settings, Help, Chat, Description and the Command Palette at once, so it wants its own change and a pass over each dialog.

**Worth a regression test.** The existing e2e specs assert on modal text via CSS selectors, which is why this went unnoticed; a role-based assertion (`getByRole('dialog')`) would have caught it and would keep it caught.

**Separate observation, not part of this report.** <kbd>Esc</kbd> does not close the settings dialog, even though `SETTINGS_KEYMAP` binds `esc` to `TOGGLE_SETTINGS` — reproduced with focus on the dialog's default target and on a plain button, so it is not the `hotkeys.filter` input guard. Happy to file that separately if it is not already known.

jooyoungseo:
> [!NOTE]
> This PR was opened by Claude Code on behalf of a maintainer.

# Pull Request

## Description

While a MAIDR dialog was open, the dialog — and the chart behind it — were absent from the accessibility tree, so a screen reader had nothing to read.

Every MAIDR dialog is rendered with `disablePortal`, so it mounts *inside* the host page's chart wrapper rather than beside it. MUI's `ModalManager` marks the modal container's other children `aria-hidden="true"` while a modal is open, and that container defaults to `document.body` — whose only child on a MAIDR page is the wrapper holding both the chart and the dialog. The element MUI hid was therefore an **ancestor** of the dialog, and the dialog hid itself.

The fix passes MUI's `container` prop, pointing at the modal root's own parent, so `ModalManager` hides the modal's *siblings* instead of walking up to an ancestor of the modal. `disablePortal` is left in place — MAIDR renders inside the host page's chart container so it survives notebooks and iframes.

## Related Issues

Fixes #655

## Changes Made

**`src/state/hook/useModalContainer.ts` (new)** — returns the ref to attach to a modal root and the `container` value derived from that root's parent. Both identities are stable across renders; MUI resolves `container` in an effect, by which point React has attached the ref for the commit that mounted the modal root.

**Dialogs wired up** — `Settings`, `Help`, `Chat`, `Description`, `CommandPalette`.

**Select menus wired up** — a MUI `Menu` is a modal too, and the `disablePortal` menus nested inside these dialogs re-hid the whole page the moment a dropdown opened. This got *more* visible once the dialogs themselves were fixed, so the same scoping is applied to the three Selects in `Settings` and the chat model picker in `ModelSelection`. `ModelSelection`'s dropdown moves into a small `ModelVersionSelect` component so each menu in the mapped list owns its own ref.

**`e2e_tests/page-objects/base-page.ts`** — adds `openHelpMenu()` / `openSettingsMenu()`. The existing `showHelpMenu()` / `showSettingsMenu()` open *and* close, so they cannot be used to inspect an open dialog.

Measured on `examples/barplot.html` in Chromium, before and after:

```
                              wrapper aria-hidden | getByRole dialog | getByRole button | chart in a11y tree
before, settings open                        true                  0                  0                    0
after,  settings open                        null                  1                  4                    1
after,  + select menu open                   null                  1                  4                    1   (listbox 1, option 4)
after,  settings closed                      null                  0                  0                    1
```

Not applicable — the dialogs render identically on screen. The defect and the fix are only visible in the accessibility tree, captured in the table above.

## Checklist

### Commits

| Commit | What |
| --- | --- |
| `fix(ui):` | The fix itself, plus the unit and e2e regression tests |
| `ci:` | Adds the e2e spec to the PR-gating smoke allowlist |
| `build:` | Declares `jsdom`, which the tests relied on only transitively |
| `test(e2e):` | Extends role-based coverage to Description, Command Palette and Chat |
| `test:` | Drops a throwaway JSDOM from the unit-test harness |

The last three came out of review rounds on this PR.

### Regression coverage

`test/state/hook/useModalContainer.test.ts` mounts a real MUI dialog into the real MAIDR DOM shape in jsdom. The first case deliberately pins the defect — a `disablePortal` dialog with the default container *does* hide the wrapper — so the suite fails if the scoping is ever dropped. The rest assert the dialog stays reachable, the chart stays reachable, the modal's own siblings still get hidden, and the tree is restored on close. This runs in the `Unit Tests` job.

`e2e_tests/specs/dialogAccessibility.spec.ts` asserts by ARIA role, as suggested in the issue. The existing specs reach the dialogs through CSS selectors, which resolve fine inside an `aria-hidden` subtree — which is why this went unnoticed. All five rewired dialogs are covered, plus a Settings select menu and the chart behind an open dialog; the hook is per-component, so a missed `container` on one dialog is invisible to the others and to the isolated hook test.

That spec is in the `E2E Smoke (chromium)` allowlist in `ci.yml`, which the workflow comment invites extending as specs are verified healthy — a regression test that only runs on the biweekly schedule cannot fail the PR that reintroduces the defect. **Note for maintainers: this does widen what blocks merges.** The specs are deterministic (no audio, autoplay, network or timing dependence) and were checked for flakiness before being added: three repeats of the full smoke list, 33 passed.

### Verification

Local: `npm run lint`, `npm run type-check`, `npm run build` clean; `npm test` 1200 passed / 95 suites; smoke spec list (`liveData.spec.ts` + `dialogAccessibility.spec.ts`, 11 tests) passed, and passed 3x on repeat.

CI on `6c832c6`: all twelve checks green, including `Unit Tests`, `Type Check`, `Build Project` and `E2E Smoke (chromium)`.

One pre-existing e2e failure is unrelated to this change and is not in the smoke set: `barplot.spec.ts`'s `beforeAll` reads a `maidr-data` attribute off `svg#bar`, but `examples/barplot.html` binds through the `var maidr = {...}` global and never carries that attribute (`grep -c maidr-data examples/barplot.html` → 0), so the whole `Bar Plot` block errors out before any test body runs. Untouched here.

### Not addressed

- The issue's closing note that `Esc` does not close the settings dialog is explicitly out of scope for this report and is left for a separate change.
- The committed `package-lock.json` records `"version": "3.68.0"` against a 3.74.0 `package.json`. Any `npm install` therefore rewrites unrelated lockfile content, which is why the `jsdom` entry was added by hand and validated with `npm ci --dry-run`. Worth a separate look.
- Reviewers suggested replacing the unit test's hand-built jsdom globals with `jest-environment-jsdom`. That needs a new devDependency with its own dependency tree — unavoidable lockfile churn given the point above — and `test/adapters/recharts/panelDom.test.ts` uses the same pattern, so converting one test would leave two conventions. Better as its own test-infrastructure change.


jooyoungseo:
Thanks both — two of the suggestions are now addressed in `af3210e` and `96cb916`, and one is declined with a reason.

**e2e covered only 2 of the 5 rewired dialogs — fixed in `96cb916`.** Correct, and the isolated hook test could not have caught a per-component miss either. Description (`d`), the Command Palette (`ctrl+shift+p`) and Chat (`shift+/`) now each assert their dialog resolves by role with no `aria-hidden` ancestor, so all five dialogs plus the Settings select menu are covered. All three run in well under a second with no network wait, and I checked for flakiness before adding them to a gating job: three repeats of the full smoke list, 33 passed. The palette shortcut uses `TestConstants.COMMAND_KEY` since its binding is `Platform.ctrl`, which is Command on macOS.

**Undeclared `jsdom` — fixed in `af3210e`.** Verified: `npm ls jsdom` shows it only as `rehype-mathjax@5.0.0 → jsdom@22.1.0`, with nothing in `package.json`. Declared as `^22.1.0`, matching what is already resolved so nothing installed today changes.

Two notes on how. I edited `package.json` and the lockfile's root entry by hand rather than running `npm install`: the committed lockfile still records `"version": "3.68.0"` against a 3.74.0 `package.json`, so a full rewrite pulled in that bump plus removal of `libc` metadata from ~20 optional deps — 65 deletions of churn unrelated to this PR, and the `libc` loss is a regression in its own right. The two-line result validates with `npm ci --dry-run`. That lockfile staleness looks worth a separate look, but not from this branch.

I also skipped `@types/jsdom` deliberately. jsdom ships no bundled types, so adding them would make the `@ts-expect-error` directives on its imports unused — a type error in `test/adapters/recharts/panelDom.test.ts`, which this PR does not otherwise touch.

**Duplicated `hiddenAncestorsOf` — not acting.** The e2e copy is serialized into the page by `page.evaluate` and cannot close over an import, so the two cannot share an implementation without a build step for the browser-side helper. Worth revisiting only if a third caller appears.

On the MUI-internals coupling: agreed it is not part of MUI's public contract. There is no way to express "hide siblings of *this* modal" in the current API, so the mitigation is that the first unit case fails loudly if a future MUI version resolves `container` differently, rather than silently reintroducing the defect. The `keepMounted` caveat is a good flag — none of these dialogs set it today, and a future one that does would want re-verifying against the same test.

---

jooyoungseo:
Took note 2, declined 1 and 3 — reasons below.

**Throwaway JSDOM in `mount()` — fixed in `6c832c6`.** Fair, and it was obscure as well as wasteful: building a second document per call just to read its body markup back out. The markup is now its own `MAIDR_BODY` constant assigned directly, with the full-document string the harness needs composed from it. Suite still 1200/95 green.

**`jest-environment-jsdom` instead of the hand-rolled harness — not in this PR.** The fragility you describe is real: a future MUI/Emotion bump reading an unlisted global fails deep in third-party code rather than saying "missing browser API." Two reasons to leave it:

- It needs a new devDependency with its own dependency tree, which I cannot add by hand the way I added `jsdom`. Running `npm install` here rewrites the lockfile with 65 lines of unrelated churn — a stale `3.68.0 → 3.74.0` version bump and removal of `libc` metadata from ~20 optional deps — which I have been keeping out of this branch.
- `test/adapters/recharts/panelDom.test.ts` already uses the same hand-rolled pattern, so switching only the new test would leave two conventions in the suite. Converting both, with the dependency and lockfile churn that implies, is a test-infrastructure change that deserves its own PR rather than riding along with an accessibility fix.

The mitigation meanwhile is that a missing global fails loudly in CI and the fix is one entry in `BROWSER_GLOBALS`.

**Silent `null` fallback in `container()` — leaving as is.** I considered logging when the modal root has no parent at resolution time, since a silent fallback means the defect returns invisibly, which is the worst failure mode for this particular bug. Decided against it: `container()` is only called from MUI's open effect, by which point the root is mounted, so a `null` there is unreachable rather than merely unlikely — and a warning for an unreachable branch is noise in a library that runs inside other people's pages. The coupling to the host DOM shape is real, but it is the same coupling `disablePortal` already carries.

On not re-running the suite: reasonable to flag. For what it is worth, the reported runs are visible in CI rather than only in the description — `Unit Tests`, `Type Check`, `Build Project` and `E2E Smoke (chromium)` all passed on `96cb916`, and the smoke job now includes `dialogAccessibility.spec.ts` covering all five dialogs.

---

jooyoungseo:
Took the nit; the lockfile concern is already covered by existing CI. Details, plus two corrections.

**`readonly` on `ModalContainer` — fixed in `8e54f6d`.** Fair catch against `.claude/rules/typescript.md`; both members are stable function identities callers only read.

**`npm ci` sanity check on the hand-edited lockfile — already runs, and has passed.** Reasonable thing to want given the change bypasses the normal generation path, so worth pinning down rather than leaving as a to-do: every job in `ci.yml` installs with `npm ci`, not `npm install` — `commitlint`, `lint`, `e2e-smoke`, `build`, `unit-test`, `type-check` and `docs-build`, seven jobs. All of them passed on `6c832c6`, which is the commit containing the hand-edited lockfile. I also ran `npm ci --dry-run` locally before committing it. So the lockfile does get validated end-to-end on every push here; it is not resting on my say-so.

**Site count.** Minor, and it does not change your conclusion that nothing was missed — but it is 9 sites, not 7: five dialogs plus four Selects (`Settings.tsx` has three, `ModelSelection.tsx` has one). `grep -c disablePortal` per file gives Chat 1, CommandPalette 1, Description 1, Help 1, Settings 4, ModelSelection 1. All nine are wired.

**Two things landed after your snapshot.**

`main` moved while this was open — #658 fixed the Escape-does-not-close-Settings bug (#656), touching `Settings.tsx` and `base-page.ts`, both of which this PR also touches. That produced a real conflict in `base-page.ts`, resolved in `271d4f6` by keeping both sides' page-object methods; `Settings.tsx` auto-merged, so upstream's `onClose={handleDialogClose}` now sits alongside this PR's `ref`/`container`. I ran #658's own Escape and backdrop-click specs against the merged tree — 3 passed — since a change to modal `aria-hidden` scoping landing next to a change in modal close handling is exactly where an interaction would hide.

The lockfile/`package.json` version mismatch that forced the hand-edit now has a root cause and an issue: #662. `.releaserc.json`'s `@semantic-release/git` `assets` omits `package-lock.json`, so `@semantic-release/npm` bumps it in the runner's tree and the change is discarded — the drift equals the eight releases since it was last committed by hand. `npm ci` ignores that field, which is why it never surfaced.

---

jooyoungseo:
Point 4 taken in `d3c4449` — the only one here not already covered.

I checked it before documenting it, and it is slightly sharper than "MUI may not re-invoke it": `getContainer(container)` runs once in `handleOpen` (`Modal/useModal.js:63`), and `ModalManager` stores the resolved element, reusing *that* element to restore `aria-hidden` on close (`Modal/ModalManager.js:196`). So re-parenting an open modal would not just miss the new parent — it would leave the original parent's children hidden. The JSDoc now says so, and names the assumption the current callers satisfy.

Points 1–3 are all already tracked: the jsdom harness and the smoke-gate widening are in the PR description, and the lockfile drift you suggested deserves its own fix now has one — #662, with the root cause: `.releaserc.json`'s `@semantic-release/git` `assets` omits `package-lock.json`, so the release bump is made in the runner's tree and discarded.

---

jooyoungseo:
Taken in `9141771` — and it was worse than "silently check the wrong element", which is why it was worth fixing rather than noting.

The assertion on that helper is `toBeNull()`. If `body.firstElementChild` stopped being the wrapper, the helper returns `null` for whatever it found instead, the assertion passes, and the test stays green while verifying nothing. Confirmed both halves against a page with no MAIDR article:

```
NEW helper threw: page.evaluate: Error: no maidr article in the document
OLD helper returned null -> passes toBeNull() vacuously
```

It now walks up from `article[id^="maidr-article"]` to the body-level ancestor and throws if the article is missing or not under `document.body`. Spec still 7/7.

Everything else in the review is already tracked — the deferred jsdom harness and the smoke-gate widening in the PR description, the lockfile drift in #662.

---

PR #660, opened 2026-07-29
test: add a jsdom component-test harness and cover the settings copy-diagnostics control

### LINKED ISSUE #657: test: no way to test React components — jest runs node-only and skips .tsx
# Feature Request

## Description

There is currently no way to unit-test a React component in this repo. `jest.config.ts` sets:

```ts
testEnvironment: 'node',
testMatch: ['**/test/**/*.test.ts'],
```

so `.tsx` files are never collected and there is no DOM to render into. Every component in `src/ui/` is therefore covered only by Playwright e2e, which is slow, runs against built examples, and is shared across 12 specs — so a small interaction is expensive to assert.

The gap has been raised independently on two reviews of #654, where the untested surface is the About section's copy button: the click handler, the `role="status"` live region it drives, and the `aria-describedby` wiring between them. The pure logic underneath is well covered (`test/util/diagnostics.test.ts`, `test/util/clipboard.test.ts`), but nothing exercises the wiring.

## Why it matters here specifically

This is an accessibility library, and the parts of a component that are hardest to eyeball are exactly the parts that carry the accessibility contract: whether a live region actually mutates, whether an accessible name still contains its visible label, whether focus lands where it should after an interaction.

#654 turned up two bugs of precisely that shape — `aria-hidden` on an element that receives focus, and a live region that announced the first copy but silently ignored every one after it. Both were found by driving a real browser by hand with a `MutationObserver`, because there was no place to write them as tests. Neither would have survived a component test; neither was catchable by the unit tests that do exist.

## Proposal

Add the missing infrastructure so component tests are possible:

- `jest-environment-jsdom` and `@testing-library/react` (plus `@testing-library/jest-dom` for the matchers) as devDependencies.
- Widen `testMatch` to include `.tsx`.
- Choose the environment per file via the `@jest-environment jsdom` docblock rather than switching the global default, so the ~1200 existing node-environment tests keep their current speed and behaviour.

Then backfill, starting where the accessibility contract is densest rather than aiming for uniform coverage — the settings dialog's status regions and the modal components' focus handling are the useful first targets.

## Additional Information

Deliberately not folded into #654: it is repo-wide test infrastructure and would be the first component test in the codebase, which deserves its own review rather than riding along with a feature.

Related: #655 (dialogs hidden from the accessibility tree) is the other half of this — a component test harness would make that class of regression assertable too, though a jsdom-based test cannot substitute for verification with a real screen reader.

jooyoungseo:
# Pull Request

## Description

Jest ran node-only and collected `.test.ts`, so a React component could not be unit-tested at all — every component in `src/ui/` was covered by Playwright alone. This adds the missing infrastructure and the first component test on it.

The environment default stays `node`, so the ~1200 existing tests keep their current start-up cost; a component test opts into a DOM per file with an `@jest-environment jsdom` docblock, as proposed in the issue.

## Related Issues

Closes #657. Pins #663 with an executable case (see below); does not fix it. Now that `main` carries #659, the harness also guards the #655 regression rather than pinning it.

## Changes Made

**Infrastructure**

- devDependencies: `jest-environment-jsdom`, `@testing-library/react`, `@testing-library/dom` (its peer), `@testing-library/jest-dom`, and `@types/jsdom`.
- `jest.config.ts`: `testMatch` widened to `.test.tsx`; `collectCoverageFrom` widened to `src/**/*.tsx`, which is what makes the untested component surface visible in the report rather than absent from it. The report gains a row per component — happy to drop that half if the noise is not wanted.
- With `@types/jsdom` resolvable, the `@ts-expect-error` directives guarding the transitive `jsdom` imports in 19 test files became unused and failed `tsc --noEmit`. They are removed along with the comments explaining them.
- `.claude/rules/testing.md` gains a Component tests section and records the `it.failing` convention; `CLAUDE.md` updated; `.github/` mirror regenerated with `npm run sync:copilot`.
- `package-lock.json` also picks up npm's own sync of its `version` field to `package.json` (3.68.0 → 3.74.0) — semantic-release commits the manifest but not the lock, so the two had drifted.

**Why three jsdom versions appear in the lockfile**

Worth spelling out, since no two of them match:

| Package | Version | Who uses it |
| --- | --- | --- |
| `jsdom` | `^22.1.0` | imported directly by the node-environment suites (declared on `main` by #659) |
| `@types/jsdom` | `^21.1.7` | types those imports |
| `jest-environment-jsdom` | `29.7.0` | bundles its own `jsdom@20` + `@types/jsdom@20`, nested, to build the DOM for `.test.tsx` files |

`@types/jsdom` has no `22` to pin: DefinitelyTyped published majors 20, 21, then 27 and 28, so `21.1.7` is the newest release below the jsdom 24+ typings and the closest match to the `22` runtime — 27/28 would put the types two-plus majors ahead of it. The environment's nested `jsdom@20` never meets the other two: it provides the global `document` for component tests, while `import { JSDOM } from 'jsdom'` in the node-environment suites resolves to the hoisted `22`. Moving those two together would mean Jest 30, which ships jsdom 26.

**First test — `test/ui/settings.copyDiagnostics.test.tsx`**

Renders the real `Settings` through `MaidrContext.Provider` with a real `ViewModelRegistry` holding stub view models, so the component reaches its state the same way it does at runtime and no service has to be constructed. Eight cases covering the About section's copy control:

- the status region is in the DOM before the first copy, with `role="status"` / `aria-live="polite"` and no text;
- the button's `aria-describedby` resolves to that region (looked up the way assistive technology resolves it, so a dangling reference fails here);
- the accessible name still contains the visible label;
- a copy writes the formatted diagnostics and reports success;
- a **repeat** copy still mutates the region — asserted with a `MutationObserver`, not on text, because a live region announces on the mutation. Reverting the `key={copyState.attempt}` fix from #654 turns this case red, which is the bug that motivated the issue;
- both clipboard paths failing produce the actionable failure message;
- focus is back on the button after the fallback path runs;
- the dialog and its controls are exposed by role — the #655 regression guard.

**What the harness turned up**

On its first run, before `main` carried the fix, it reproduced #655 in jsdom: `getByRole` found nothing inside the dialog. That is now merged as #659, so the case has become a plain assertion and the role queries have dropped the `hidden: true` they needed to see past it.

Making the dialog reachable exposed the next layer: **the settings dialog has no accessible name**. `aria-label="Settings"` is spread onto the `role="presentation"` modal root, while the `role="dialog"` paper carries the `aria-labelledby` MUI generates for a `DialogTitle` this dialog never renders — so the reference dangles and a screen reader announces the dialog unnamed. Filed as #663 and pinned here as an `it.failing` case. The pin matches any non-empty accessible name rather than the literal "Settings", so any fix that names the dialog turns it red, which is the signal to delete the marker.

Not applicable — no visual change.

## Checklist

## Additional Notes

Verified locally after merging `main`: `npm run lint`, `npm run type-check`, and `npm test` (96 suites, 1208 tests, all passing). The new suite adds ~14s, most of it ts-jest compiling the `Settings` import graph — the renders themselves are ~0.5s each.

Not done here, deliberately: the modal focus-handling tests the issue lists as the other first target. #659 has just changed how every dialog is mounted, so those are better written against the new structure in their own change than guessed at here.

---
_Generated by [Claude Code](https://claude.ai/code/session_0115MPwR3LxFbABik2k76VKh)_


jooyoungseo:
Thanks — acted on two of the four, and here is where the other two stand.

**Fixed in 4133739:**

1. **Table alignment.** Widened the Runner column instead of shrinking the label, so all three rows line up in the raw source. `.claude/` is in the eslint ignore list, so no formatter would have caught it — regenerated the `.github/` mirror alongside.
3. **Lockfile churn.** Confirmed and fixed: the `libc` arrays on the optional `@esbuild/*` platform packages were dropped by npm 10 writing a lockfile that npm 11 had produced — 33 lines of unrelated diff. Regenerated with `npx npm@11 install --package-lock-only` from `main`'s lockfile, so the diff is now purely additive apart from npm syncing its own `version` field to `package.json` (3.68.0 → 3.74.0; semantic-release commits the manifest but not the lock, so the two had drifted on `main`). `npm ci --dry-run` is clean against it, and `lockfileVersion` stays 3.

**Not changed, deliberately:**

2. **Shared jsdom setup file.** Agreed, and agreed on the timing — one caller, so the side-effect import stays local to the file. Whoever writes the second component test is the right person to lift it into `setupFilesAfterEnv`, and by then it will be clear whether the shared setup wants anything besides the matchers.
4. **`@types/jsdom@20` against `jsdom@22`.** Real but latent, and it predates this PR — the adapter tests were already importing a transitive `jsdom@22` and had suppressed the types entirely, so the drift risk trades an unchecked import for a slightly-stale-checked one. Fixing it properly means declaring `jsdom` and `@types/jsdom` as direct devDependencies at matching majors, which is its own change; those tests only touch `new JSDOM()` and the document surface, which is stable across 20 → 22.

One correction for the record: the `@ts-expect-error` removal spans 18 files, not 15 — 17 test files plus `test/adapters/highcharts/helpers.ts`, which is a helper rather than a suite.

---

jooyoungseo:
Thanks — this round took three of the four. Force-pushed, so the branch is now two commits instead of four.

**1. `@types/jsdom` was transitive.** Fixed: declared directly as a devDependency at `^21.1.7`. That also closes the version-drift point from the earlier review — the 18 files import the `jsdom@22` that `rehype-mathjax` hoists, and `@types/jsdom@21` is the range that types it, where the transitive `20.0.1` from `jest-environment-jsdom` was a major behind. `tsc --noEmit` is clean against it.

**2. Import style.** Fixed: the file now imports `describe`/`it`/`expect`/`jest` from `@jest/globals`, matching 73 of the 95 existing test files. That in turn required the jest-dom import to become `@testing-library/jest-dom/jest-globals` — the bare entry point only augments the ambient global, so the matchers type-check against the imported `expect`, not the other one. Worth knowing for the next component test, so the rules snippet now shows both imports.

**3. Commit split.** Rebuilt rather than re-split: the two follow-up commits were fixes to the first commit, not changes of their own, so they are folded into it. Two commits now — the harness, and the first test on it.

**4. Coverage widening.** Left as-is for maintainer preference, as you read it. One line to drop if the 0% rows are unwanted.

One correction: the directive removal spans 18 files, not 15 — 17 suites plus `test/adapters/highcharts/helpers.ts`, which is a helper rather than a suite.

Re-verified locally after the rewrite: `npm test` (95 suites, 1202 tests), `npm run lint`, `npm run type-check` all clean.

---

jooyoungseo:
Took the third point in 4339ef6.

**3. `it.failing` convention documented.** Written into "Expectations for a change" rather than the Component tests section — it applies to any suite, not just component tests, and the situation it covers (a change surfaces a bug that belongs to a different fix) shows up most often in node-environment tests. Says what the pattern buys, and that the case turning red is the cue to delete it along with any workaround written around the bug. Mirror regenerated.

**2. `packageManager` pin.** Agreed on the diagnosis — nothing pins an npm version, so the lockfile will drift again on the next contributor with a different one, and this PR only re-syncs it incidentally. Deliberately not doing it here: pinning `packageManager` changes what every contributor and every CI job runs, which is a repo-wide tooling decision that should not ride in on a test-infrastructure PR. Happy to open a separate issue for it if that is wanted.

**1. Coverage widening.** Still a one-line revert if the 0% rows are unwanted; leaving the call to a maintainer.

---

jooyoungseo:
Two of these landed in 268388a; the version-skew one turns out to be a numbering artifact.

**MutationObserver timing.** Good catch, and hardened rather than commented — the assertion no longer depends on when the queue drains. Records come from both ends now: the callback takes what the microtask checkpoint delivers, and `observer.takeRecords()` picks up anything still queued behind it. Re-verified the guard still bites after the change: dropping `key={copyState.attempt}` from the status span turns the case red.

**`viewModel.state` vs `useViewModelState`.** Agreed on both halves — pre-existing, and the stub pattern should not generalize. Written into the render helper's doc comment: a static `state` works here because `Settings` reads a getter once per render, and a component subscribing through `useViewModelState` reads the Redux slice, which a stub like this would never re-render.

**`@types/jsdom@21` against `jsdom@22`.** Checked, and it is the right pin rather than a compromise. DefinitelyTyped's published majors for this package are 20, 21, then 27 and 28 — nothing in between. `21.1.7` is the newest release that predates the jsdom 24+ typings, so it is the closest match to the 22.1.0 runtime; 27/28 would put the types two-plus majors *ahead* of it. `tsc --noEmit` is clean across all 18 files, which between them touch only `new JSDOM()`, `.window.document`, and the Element surface.

**Coverage gate.** Confirmed there is none: no `coverageThreshold` in `jest.config.ts`, and the Unit Tests job runs plain `npm test`. Coverage is reported, never enforced, so the newly visible `.tsx` rows cannot fail a build.

---

jooyoungseo:
No changes from this round — here is why for each.

**`@types/jsdom` skew / TODO to bump at v22.** There will not be a v22 to bump to: DefinitelyTyped's published majors for this package are 20, 21, then 27 and 28, with nothing in between. `21.1.7` is the newest release predating the jsdom 24+ typings, so it is the closest available match to the 22.1.0 runtime — 27/28 would put the types two-plus majors *ahead* of it, which is the worse trade. A TODO pointing at a release that will never ship would age into noise, so I would rather leave the pin unannotated and correct. (Same ground as the previous round's first point.)

**`jest-environment-jsdom` bundling jsdom 20 while the adapter tests import 22.** Agreed on all counts, including that it is fine today. Worth adding: the two copies never meet — the environment's jsdom provides the global document for `.test.tsx` files, and the direct `import { JSDOM } from 'jsdom'` in the node-environment suites resolves to the hoisted 22. A component test that needs a Web API jsdom 20 lacks would be a reason to move to Jest 30, which ships jsdom 26.

**Module-scope console spies.** The failure mode you describe is narrower than it looks: `beforeEach` calls `mockClear()` on both, so a case added to this file starts with empty call history and can assert on it safely. What a future case cannot do is observe the console *unspied* — inherent to silencing at file scope, and the alternative (installing and restoring per test) buys that back at the cost of the two expected-failure cases printing stack traces on every run. Leaving it as is.

CI is green on 268388a — 13/13, including the run that produced this review.

---

jooyoungseo:
Merged `main` (8de8048). It brought #659 in, which fixes the #655 containment bug this PR was pinning — so the pin has been retargeted, and the harness immediately earned its keep a second time.

**#655 is fixed, so the pin becomes a guard.** `it.failing` is gone from that case; it now asserts plainly that the dialog and its controls are exposed by role, and the other queries have dropped the `hidden: true` they needed to see past the `aria-hidden` ancestor.

**What the fix uncovered: the settings dialog has no accessible name.** `aria-label="Settings"` is spread onto the `role="presentation"` modal root, where it names nothing, while the `role="dialog"` paper carries the `aria-labelledby` MUI generates for a `DialogTitle` that `Settings.tsx` never renders — so it points at an id that does not exist:

```
div.MuiDialog-root      role="presentation"  aria-label="Settings"     <== the label is here
  div.MuiPaper-root     role="dialog"        aria-labelledby="_r_1_"   <== the role is here
                                             ^ document.getElementById → null
```

`queryByRole('dialog')` finds it; `queryByRole('dialog', { name: 'Settings' })` does not. This was unobservable while #655 stood — a missing name inside a subtree that was itself hidden makes no difference. Filed as #663 with the reproduction and a survey of the other four dialogs (Description, CommandPalette, Help and Chat are all named; Settings is the only one with this shape), and pinned here the same way #655 was.

**Also merged in:** #659's own `test/state/hook/useModalContainer.test.ts` arrived with the jsdom `@ts-expect-error` directive, now unused for the same reason as the other 18 — removed, so the count in this PR is 19.

Conflicts were confined to `package.json` and `package-lock.json`, both from #659 declaring `jsdom` as a direct devDependency alongside this PR's additions: kept both, and regenerated the lock from `main`'s with npm 11.

Re-verified on the merge: `npm test` (96 suites, 1208 tests), `npm run lint`, `npm run type-check` all clean.

---

jooyoungseo:
Both actionable points taken.

**`it.failing` signal precision.** Good catch — pinned on the literal `'Settings'`, a fix that named the dialog anything else would have left the case red, and red-for-wrong-name is indistinguishable from red-for-no-name. The contract is that the dialog is named at all, so it now matches any non-empty accessible name (`{ name: /\S/ }`). Verified the matcher discriminates rather than matching everything: a regex name query finds a dialog with an `aria-label` and skips one without. Pushed as 83660ad.

**Three jsdom lines.** Agreed the reasoning belonged somewhere durable — the PR body now carries a table of which package uses which version and why no two match. Short version: `@types/jsdom` has no `22` to pin (DefinitelyTyped published majors 20, 21, then 27 and 28), so `21.1.7` is the newest release below the jsdom 24+ typings and the closest match to the `22` runtime. The environment's nested `jsdom@20` never meets the other two — it builds the global `document` for `.test.tsx` files, while `import { JSDOM } from 'jsdom'` in the node-environment suites resolves to the hoisted `22`. Moving those two together would mean Jest 30, which ships jsdom 26.

**`Object.defineProperty` stubbing.** Noted, no change — as you say, it holds because jsdom environments are per-file.

---

jooyoungseo:
Three of the four turned into changes (8f3d20e).

**Stub view models drifting behind `as unknown as`.** Fixed rather than noted — this one had teeth, since the whole point of the file is to be copied. Each stub is now typed as a `Pick` of the real class listing the members `Settings` calls, and only that narrower type is cast. The members the stub defines are checked against production signatures; only the ones it never touches are bridged. Verified it bites: renaming a member inside the `Pick` fails `tsc` in three places, where the old form compiled clean.

**Console spies as an exception to the `afterEach` convention.** Agreed it reads as unusual against the rule as written, so the rule now carries the exception: spy at file scope, `mockClear()` per test, restore in `afterAll`, when the noise itself is what needs suppressing.

**`jsdom` / `@types/jsdom` coupling.** Also written into the rules doc rather than left to `git blame` — bumping one means checking `npm run type-check` still passes, since DefinitelyTyped has no release matching every jsdom major. That is the file a contributor is in when they touch this, which `package.json` cannot be (no comments in JSON).

**`Object.defineProperty` stubs never deleted.** No change — as you note, jsdom builds a fresh window per test file, so there is nothing to leak into.

`npm test` (96 suites, 1208 tests), `npm run lint` and `npm run type-check` are clean on the result.

---

PR #664, opened 2026-07-29
fix(ui): name the settings dialog with a rendered title

### LINKED ISSUE #663: fix: the settings dialog has no accessible name
# Bug Report

## Description

The settings dialog is announced without a name. `aria-label="Settings"` never reaches the element that carries `role="dialog"`.

MUI's `Dialog` treats the two labelling props differently. `aria-labelledby` is a first-class prop applied to the paper — the `role="dialog"` element. `aria-label` is not: it lands in the props spread onto the modal **root**, which is `role="presentation"`, where it names nothing. Meanwhile MUI generates an `aria-labelledby` for the `DialogTitle` it provides through `DialogContext` and puts it on the paper whether or not a `DialogTitle` is ever rendered. `Settings.tsx` renders none, so the reference dangles.

The result is a dialog whose accessible name comes from a pointer to an element that does not exist:

```
div.MuiDialog-root   role="presentation"  aria-label="Settings"      <== the label is here
  div.MuiDialog-container role="presentation"
    div.MuiPaper-root  role="dialog"  aria-labelledby="_r_1_"        <== the role is here
                                      ^ resolves to nothing
```

This was not observable until #659. While #655 stood, the whole dialog sat outside the accessibility tree, so a missing name inside it made no difference. Fixing the containment made the dialog reachable and left the naming gap as what remains.

## Steps to Reproduce

1. `npm run build`
3. Press Ctrl + , to open the settings dialog.
4. Query the accessibility tree for the dialog by name.

```js
document.querySelectorAll('[role="dialog"]').length                      // 1
document.querySelector('[role="dialog"]').getAttribute('aria-labelledby')// "_r_1_"
document.getElementById('_r_1_')                                         // null
```

Or in the component test added by #660, which pins this as an `it.failing` case:

```js
screen.queryByRole('dialog', { name: 'Settings' })   // null
screen.queryByRole('dialog')                         // found
```

## Actual Behavior

The dialog is exposed with role `dialog` and no accessible name. A screen reader announces "dialog" — nothing distinguishes it from the help, chat, description or command palette dialogs.

## Expected Behavior

The dialog is announced with its name, "Settings".

## Screenshots

Not applicable — the dialog renders correctly on screen. The defect is only in the accessibility tree.

## Environment

- Browser: reproduced in jsdom via the component test; the markup is browser-independent

## Additional Information

**Scope.** Verified for `Settings.tsx`, which is the only dialog with this shape — it is the only one that both passes `aria-label` to `<Dialog>` and renders no `DialogTitle`. The others are named:

| Dialog | How it is named | Named? |
| --- | --- | --- |
| `Settings.tsx` | `aria-label` on the modal root, dangling `aria-labelledby` on the paper | no |
| `Description.tsx` | `aria-labelledby` pointing at its own title element | yes |
| `CommandPalette.tsx` | `aria-labelledby="command-palette-title"`, rendered at line 109 | yes |
| `Help.tsx` | renders a `DialogTitle`, so MUI's generated id resolves | yes |
| `Chat.tsx` | renders a `DialogTitle`, so MUI's generated id resolves | yes |

**Likely fix.** Give the dialog a real title element and point `aria-labelledby` at it, the way `Description.tsx` does — that also puts a heading in the dialog for screen reader users navigating by heading, which the settings dialog currently lacks at the top level. Passing the label through `slotProps={{ paper: { 'aria-label': 'Settings' } }}` would work too, but leaves the dangling `aria-labelledby` in place, and an `aria-labelledby` that resolves to nothing beats an `aria-label` in the name computation order — so the label has to displace it, not sit beside it.

**Worth checking with the fix.** Whether MUI still emits `aria-labelledby` when no `DialogTitle` is rendered. A dangling reference is a defect in its own right even once the dialog has a name by another route.

jooyoungseo:
Closes #663.

## Root cause

Confirmed against `@mui/material` v7 source rather than inferred:

- `Dialog/Dialog.js:346-352` applies `role`, `aria-labelledby`, `aria-describedby` and `aria-modal` to the **paper** — the `role="dialog"` element — and spreads `...other`, where `aria-label` lands, onto the **root**, which is `role="presentation"` and names nothing.
- `Dialog/Dialog.js:264`, `useId(ariaLabelledbyProp)`, generates an id when none is passed and puts it on the paper unconditionally, whether or not a `DialogTitle` is ever rendered.

`Settings.tsx` passed `aria-label` and rendered no `DialogTitle`, so its name resolved through a reference to an element that did not exist. It was the only dialog with that shape. Invisible until #659 put the dialog back in the accessibility tree.

## The fix

Render a `DialogTitle` and point `aria-labelledby` at it, the way `Description.tsx` does.

- `aria-label="Settings"` → `aria-labelledby={titleId}`.
- Added `<DialogTitle id={titleId}>Settings</DialogTitle>`. Passing the id explicitly also settles the reference MUI would otherwise generate, which answers the "worth checking with the fix" item on the issue: nothing dangles.
- Section headings (General Settings, LLM Settings, About) move from `h6` to `h3`. `DialogTitle` renders as `h2`, and adding it above bare `h6`s would have introduced a level skip in a dialog whose new heading is meant to help heading navigation.

The id sits on the `DialogTitle` itself rather than on a nested `Typography`. `DialogTitle.js:57` resolves `id: idProp ?? titleId`, so the wrapper picks up the context id regardless — the nested-`Typography` pattern gives two elements the same id.

## Verification

- `npm run type-check`, `eslint`, `npm run build` — clean.
- `npm test` — 96 suites, 1209 tests, all pass.
- The `it.failing` pin from #660 becomes two real cases: one asserting the name, one resolving `aria-labelledby` through the DOM so a dangling reference fails there rather than passing against a name that happens to compute.
- `dialogAccessibility.spec.ts` now asserts the settings dialog by name rather than by role alone; all 7 tests pass in chromium.
- Reproduced the issue's exact repro steps in a real browser: 1 `role="dialog"`, `aria-labelledby` resolving to an existing `H2` "Settings", exactly one element carrying that id, root no longer carrying the stray `aria-label`, outline reading H2 → H3/H3/H3, and Playwright's accessible-name computation resolving `getByRole('dialog', { name: 'Settings' })`.

## Two things to flag

**Firefox and WebKit are unverified, and this PR's CI will not cover them either.** The `e2e-smoke` job in `ci.yml` runs `dialogAccessibility.spec.ts` on every PR but is chromium-only by design; the full three-browser suite lives in the scheduled workflows (`e2e.yml`, every 2 days). So the first firefox/webkit run against this change is the next scheduled one, after merge. Locally I could only run chromium as well — the box has chromium rev 1194 against a `^1.52` pin with a 1.59 CLI, so the other two projects fail on a missing browser. The fix is a static markup change with no browser-specific API, so the risk is low, but it is untested there and worth saying plainly.

Separately: root `playwright.config.ts` is broken on `main` (`__dirname` in an ESM package, `testDir` pointing at a nonexistent `test/e2e/specs`). Nothing uses it — every workflow passes `--config=e2e_tests/config/test-config.ts` — but `npx playwright test` with no `--config` fails at load. Predates this branch; not touched here.

**A pre-existing duplicate-id bug in `Description.tsx`,** not touched here. Lines 136-142 pass `aria-labelledby={`${id}-title`}` to `Dialog` and render `<Grid component={DialogTitle}>` with an inner `<Typography id={`${id}-title`}>` — so both the `DialogTitle` wrapper (via `idProp ?? titleId`) and the `Typography` carry that id. The dialog is still announced correctly, so this is cosmetic rather than a naming failure, but it is worth a follow-up.

## Out-of-scope edit worth a look

`.claude/rules/testing.md` cited this test file as the live `it.failing` example for #663. Fixing the bug makes that citation stale, so I replaced it with generic guidance and regenerated the `.github/` mirror via `npm run sync:copilot`. No `it.failing` pins remain in the repo.

https://claude.ai/code/session_01VH49ExPkrEGyEWZJj1EFmm

### DIFF
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -104,5 +104,6 @@ When a change surfaces a bug that belongs to a different fix, pin it with
 `it.failing` and the issue number rather than skipping it or leaving a comment.
 The case keeps running, the suite stays green while the bug stands, and the day
 the fix lands the case turns red — which is the reminder to delete it and
-anything written to work around the bug. `test/ui/settings.copyDiagnostics.test.tsx`
-does this for #663.
+anything written to work around the bug. Replace it then with a case that
+asserts the fixed behaviour, rather than dropping the marker and leaving the
+weaker assertion the pin was written to tolerate.
--- a/.github/instructions/testing.instructions.md
+++ b/.github/instructions/testing.instructions.md
@@ -103,5 +103,6 @@ When a change surfaces a bug that belongs to a different fix, pin it with
 `it.failing` and the issue number rather than skipping it or leaving a comment.
 The case keeps running, the suite stays green while the bug stands, and the day
 the fix lands the case turns red — which is the reminder to delete it and
-anything written to work around the bug. `test/ui/settings.copyDiagnostics.test.tsx`
-does this for #663.
+anything written to work around the bug. Replace it then with a case that
+asserts the fixed behaviour, rather than dropping the marker and leaving the
+weaker assertion the pin was written to tolerate.
--- a/e2e_tests/specs/dialogAccessibility.spec.ts
+++ b/e2e_tests/specs/dialogAccessibility.spec.ts
@@ -96,10 +96,13 @@ test.describe('dialog accessibility tree', () => {
     const barPlotPage = await setupBarPlotPage(page);
     await barPlotPage.openSettingsMenu();
 
-    await expect(page.getByRole('dialog')).toBeVisible();
+    // By name, not by role alone (#663): the dialog used to resolve here with
+    // no accessible name, indistinguishable from the four other dialogs.
+    await expect(page.getByRole('dialog', { name: 'Settings' })).toBeVisible();
     expect(await hiddenAncestorsOf(page, '.settings-dialog')).toEqual([]);
     expect(await wrapperAriaHidden(page)).toBeNull();
 
+    await expect(page.getByRole('heading', { name: 'Settings', level: 2 })).toBeVisible();
     await expect(page.getByRole('button', { name: 'Reset Settings' })).toBeVisible();
     await expect(
       page.getByRole('button', { name: 'Close Settings with no changes' }),
--- a/src/ui/component/Settings.tsx
+++ b/src/ui/component/Settings.tsx
@@ -18,6 +18,7 @@ import {
   Dialog,
   DialogActions,
   DialogContent,
+  DialogTitle,
   Divider,
   FormControl,
   FormControlLabel,
@@ -441,6 +442,7 @@ const Settings: React.FC = () => {
   const [llmSettings, setLlmSettings] = useState<LlmSettings>(llm);
 
   const [copyState, setCopyState] = useState<CopyState>({ status: 'idle', attempt: 0 });
+  const titleId = `${id}-title`;
   const copyStatusId = `${id}-copy-status`;
   // The bundle source and the browser cannot change while the dialog is open,
   // so the DOM scan behind this runs once per mount rather than per render.
@@ -631,7 +633,14 @@ const Settings: React.FC = () => {
     <Dialog
       id={id}
       role="dialog"
-      aria-label="Settings"
+      // Names the dialog. `aria-label` cannot do it here: MUI applies
+      // `role` and `aria-labelledby` to the paper — the `role="dialog"`
+      // element — but spreads everything else, `aria-label` included, onto
+      // the modal root, which is `role="presentation"` and names nothing.
+      // Passing the id also settles the reference MUI derives from it for
+      // `DialogContext`; left to generate its own, it points the paper at a
+      // `DialogTitle` that need not exist.
+      aria-labelledby={titleId}
       open={true}
       onClose={handleDialogClose}
       maxWidth="sm"
@@ -644,9 +653,16 @@ const Settings: React.FC = () => {
       onKeyDown={handleDialogKeyDown}
       className="settings-dialog"
     >
+      {/* Renders as an `h2`, so the dialog also gains the top-level heading
+          it lacked — the section headings below were the only ones in it,
+          leaving nothing to land on when navigating by heading. */}
+      <DialogTitle id={titleId} className="settings-dialog-title">
+        Settings
+      </DialogTitle>
+
       <DialogContent className="settings-dialog-content">
         <Grid size="grow">
-          <Typography variant="h6" fontWeight="bold" gutterBottom>
+          <Typography variant="h6" component="h3" fontWeight="bold" gutterBottom>
             General Settings
           </Typography>
         </Grid>
@@ -1114,6 +1130,7 @@ const Settings: React.FC = () => {
           <Grid size={12}>
             <Typography
               variant="h6"
+              component="h3"
               fontWeight="bold"
               gutterBottom
               className="settings-section-title"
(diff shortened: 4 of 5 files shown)


jooyoungseo:
Thanks — two of the three notes need an answer, and the closing line needs a correction.

### CI will not confirm Firefox/WebKit on this PR

> Good to merge pending CI confirming Firefox/WebKit.

That gate does not exist, so waiting on it would mean waiting indefinitely. `E2E Smoke (chromium)` in `ci.yml` is the only Playwright job wired to `pull_request`, and it is chromium-only by design — the comment in the workflow says so directly: *"Fast Playwright smoke on every PR: chromium only, core bundle only... The full 3-browser suite stays on the scheduled e2e workflows."*

Firefox and WebKit live in `e2e.yml` / `e2e_tests.yml`, both `schedule` + `workflow_dispatch` only (`e2e.yml` runs every 2 days at 02:00 UTC). So the first three-browser run against this change is a scheduled one *after* merge.

That smoke job has since gone green, which covers `dialogAccessibility.spec.ts` on chromium — the same single browser I verified locally, and the honest extent of the cross-browser coverage here. If confirming the other two before merge matters, the way to get it is a manual `workflow_dispatch` on `e2e.yml` against this branch, not waiting on the PR checks. My read is that it is not necessary: the change is static JSX/ARIA markup with no browser-specific API, and the accessible-name computation it depends on is spec-defined. But that is a judgement call, not a verified result, which is why it is flagged rather than glossed.

### `settings-dialog-title` is the established pattern, not a stray

Checked this rather than assuming. There are **no stylesheets in the repo at all** — no `.css`/`.scss` outside `coverage/`. Every `settings-*` class in this component has zero in-repo consumers except `settings-dialog`, which `dialogAccessibility.spec.ts` uses as a selector:

| class | consumers outside `Settings.tsx` |
| --- | --- |
| `settings-dialog` | 2 (e2e spec) |
| `settings-dialog-content` | 0 |
| `settings-section-title` | 0 |
| `settings-footer` | 0 |
| `settings-divider` | 0 |
| `settings-dialog-title` (new) | 0 |

So the new class is consistent with five siblings on the same component. These are the stable theming/selector surface for a library whose emotion class names are hashed and unstable — `.settings-dialog` is already used exactly that way. Leaving the one new structural node as the only unclassed one seemed worse than matching the convention. Happy to drop it if you would rather not grow that surface.

### `Description.tsx` duplicate id

Agreed it is out of scope here. I have not filed a tracking issue — that is the maintainers' call on triage. Say the word and I will open one with the specifics (`Dialog` gets `aria-labelledby={`${id}-title`}` while `DialogTitle` resolves the same id via `idProp ?? titleId`, so the wrapper and the inner `Typography` both carry it; the name still computes correctly, so it is cosmetic).

---

jooyoungseo:
Tracking issue filed: #665.

Scoped wider than the note here. Checking the other dialogs before writing it up turned out to be worth doing — the duplicate id is one symptom of a shared pattern, and the more user-visible one is different: `Description.tsx`, `Help.tsx` and `Chat.tsx` all nest a heading inside `DialogTitle` (itself an `h2`), so each dialog's title lands **twice** in the heading outline. Description additionally carries the duplicate id.

Verified in chromium against a real build:

| Dialog | duplicate id | nested heading | title in outline |
| --- | --- | --- | --- |
| `Description.tsx` | **yes (2)** | `h6` in `h2` | **twice** |
| `Help.tsx` | no | `h6` in `h2` | **twice** |
| `Chat.tsx` | no | `h2` in `h2` | **twice** |
| `CommandPalette.tsx` | no | none | once |
| `Settings.tsx` (this PR) | no | none | once |

The last row is the useful one for reviewing this PR: putting the id on `DialogTitle` rather than on a nested `Typography` is what keeps Settings clear of both defects, and #665 proposes the same shape for the other three.

---

PR #666, opened 2026-07-29
fix(ui): give each dialog exactly one title heading

### LINKED ISSUE #665: fix: three dialogs nest a heading inside DialogTitle, announcing the title twice
# Bug Report

## Description

Found while fixing #663 (PR #664), which flagged a narrower version of this — the duplicate id in `Description.tsx`. Checking the other dialogs turned up a shared root cause with a more user-visible symptom, so this covers the pattern rather than the one file.

Three dialogs render a heading *inside* `DialogTitle`, which is itself a heading (`component = 'h2'` by default, `DialogTitle.js:52`). The result is `<h2><h6>…</h6></h2>` — invalid HTML, since `h2`'s content model is phrasing content and `h6` is flow content — and, more importantly, **the dialog's title appears twice in the heading outline**. A screen reader user navigating by heading lands on the same title on two consecutive presses.

`Description.tsx` has an additional defect on top of that. It passes `aria-labelledby={`${id}-title`}` to `Dialog` *and* puts that same id on the inner `Typography`. MUI resolves `DialogTitle`'s own id as `idProp ?? titleId` (`DialogTitle.js:57`), where `titleId` comes from `DialogContext` — so the wrapper claims that id too, and **two elements end up carrying it**.

The accessible name still computes correctly in every case (`getElementById` returns the first match, which has the right text), so this is not a naming failure like #663 was. It is a heading-navigation and validity defect.

## Steps to Reproduce

1. `npm run build`
3. Press <kbd>d</kbd> for Chart Description (or <kbd>Ctrl</kbd>+<kbd>/</kbd> for Help, <kbd>Shift</kbd>+<kbd>/</kbd> for Chat).
4. Enumerate the dialog's headings, and count the elements carrying the id the dialog points at.

```js
const d = document.querySelector('[role="dialog"]');

// Title appears twice in the outline.
[...d.querySelectorAll('h1,h2,h3,h4,h5,h6')].map(h => `${h.tagName}: ${h.textContent.trim()}`)
// ["H2: Chart Description", "H6: Chart Description", "H6: Axes", "H6: Summary"]

// A heading whose ancestor is also a heading.
[...d.querySelectorAll('h1,h2,h3,h4,h5,h6')].filter(h => h.parentElement.closest('h1,h2,h3,h4,h5,h6'))
// [h6]

// Description only: two elements carry the referenced id.
const lb = d.getAttribute('aria-labelledby');
document.querySelectorAll(`[id="${CSS.escape(lb)}"]`).length   // 2
```

## Actual Behavior

Verified in chromium against a real build, all five dialogs:

| Dialog | Elements with the referenced id | Nested heading | Title in the outline |
| --- | --- | --- | --- |
| `Description.tsx` | **2 — duplicate** | `h6` inside `h2` | **twice** |
| `Help.tsx` | 1 | `h6` inside `h2` | **twice** |
| `Chat.tsx` | 1 | `h2` inside `h2` | **twice** |
| `CommandPalette.tsx` | 1 | none | once |
| `Settings.tsx` (after #664) | 1 | none | once |

Concretely, Description's outline is `H2: Chart Description | H6: Chart Description | H6: Axes | H6: Summary`, and both of those first two elements carry `id="_barr_0_-title"`.

## Expected Behavior

Each dialog contributes exactly one top-level heading, and each id is unique in the document. Heading navigation moves from the dialog's title straight to its first section.

## Screenshots

Not applicable — all five dialogs render correctly on screen. The defect is in the accessibility tree and the heading outline.

## Environment

- Browser: reproduced in chromium; the markup is browser-independent

## Additional Information

**Root cause.** The `<Grid container component={DialogTitle}>` wrapper with an inner heading `Typography` (`Description.tsx:138-144`, `Help.tsx:63-69`) and Chat's explicit `<DialogTitle component="h2">` around a `<Typography component="h2">` (`Chat.tsx:227-236`). In each case the outer element is already the heading; the inner one duplicates it.

**Likely fix.** Let `DialogTitle` be the heading and drop the nested `Typography`, or keep the `Typography` for styling with `component="span"` so it contributes no heading. #664 does the former for `Settings.tsx`:

```tsx
<DialogTitle id={titleId}>Settings</DialogTitle>
```

That also removes the need for a separate id on an inner element, which is what causes Description's duplicate — with the id on the `DialogTitle` itself, `idProp ?? titleId` resolves to the same value and only one element carries it.

**Worth deciding with the fix.** Whether the dialog title should stay `h2`. `Description.tsx` and `Help.tsx` use `h6` for their section headings, so once the nested duplicate is removed the outline jumps `h2` → `h6`. #664 moved Settings' sections to `h3` for that reason; the same treatment would suit these two.

**Not covered here.** `Chat.tsx:235` also puts `aria-label="Chart Assistant - AI Chat Interface"` on the heading `Typography`, which overrides its visible text "Chart Assistant" for assistive technology. That is a separate question about whether the accessible name should diverge from the visible label (WCAG 2.5.3 Label in Name), and is not part of this report.

jooyoungseo:
Closes #665. Follow-up to #664, which fixed the same class of defect in `Settings.tsx`.

## Root cause

`DialogTitle` is a heading — `component = 'h2'` by default (`DialogTitle.js:52`). Three dialogs rendered *another* heading inside it, so each dialog's title landed in the heading outline twice. Navigating by heading hit the same title on two consecutive presses. The markup is also invalid: `h2` takes phrasing content, and `h6`/`h2` is flow content.

`Description.tsx` compounded it. It passed `aria-labelledby={`${id}-title`}` to `Dialog` and put that id on an inner `Typography` — but MUI resolves the title's own id as `idProp ?? titleId`, feeding `titleId` from `DialogContext` (`DialogTitle.js:49,57`). The wrapper claimed the id too, so **two elements carried it**.

## One defect worse than the issue reported

Capturing a baseline before editing turned up something #665 did not record. **Chat's accessible name was `"Chart Assistant - AI Chat Interface Close chat dialog"`** — its `DialogTitle` holds the close button as well as the title, and MUI's generated `aria-labelledby` pointed at that whole row, so the button's label was concatenated into the dialog's name. Same root cause, so this fixes it; the name is now the heading alone.

## The fix

Let `DialogTitle` be the heading and carry its text directly, the shape #664 established.

- **`Help.tsx`** — `Grid container component={DialogTitle}` + heading `Typography` → a plain `DialogTitle`.
- **`Description.tsx`** — same, with the id on `DialogTitle` itself. That is what resolves the duplicate: with `idProp` set to the same value the context supplies, only one element renders it. Sections move `subtitle2` (which maps to `h6`) → `h3`, so the outline no longer skips `h2` to `h6`.
- **`Chat.tsx`** — the title row genuinely needs to hold the close button, so it stays for layout but as a `div`, with the heading inside it named explicitly via `aria-labelledby` and given an id of its own so it cannot claim the title's.

## Verification

Chromium, all five dialogs, before and after:

| Dialog | before | after |
| --- | --- | --- |
| `Description.tsx` | duplicate id, title ×2 | clean, `H2 → H3/H3/H3` |
| `Help.tsx` | title ×2 | clean, single `H2` |
| `Chat.tsx` | title ×2, name included "Close chat dialog" | clean, named after the heading alone |
| `CommandPalette.tsx` | clean | clean |
| `Settings.tsx` | clean (#664) | clean |

No nested headings, no duplicated ids, and every accessible name preserved or improved.

- **Both directions.** 7/7 `dialogAccessibility` pass with the fix; reverting *only* the three source files fails exactly the three affected dialogs and leaves the two already-clean ones green.
- `npm test` — 107 suites, 1350 tests, all pass.
- `npm run type-check`, `eslint`, `npm run build` — clean.
- Full chromium e2e suite — 308/308.

All of the above re-run on the merged tree at each base refresh (see below), not just against the original base.

### On the tests

`dialogStructure` resolves the reference through the DOM — counting the elements actually claiming the referenced id — so a dangling reference (0) or a duplicated one (2+) fails even when the name still computes correctly. Asserting on the markup that produces the name would not catch either.

The revert run earned its keep by catching a flaw in the tests themselves: **Playwright matches `getByRole` `name` as a substring by default**, so the Chat assertion passed against the polluted name it was written to reject. All five dialog-name assertions now use `exact: true`, with a comment saying why so it does not get dropped as noise later.

### Fast-loop coverage, and why it is not a superset

`test/ui/dialogTitles.test.tsx` (added after review) asserts the same contract in jsdom, so this class of regression is caught in the ordinary jest pass rather than only in the chromium-only e2e job. 8 cases across the three dialogs.

Writing it turned up something worth recording. The natural assertion for the Chat defect — `toHaveAccessibleName('Chart Assistant - AI Chat Interface')` — **passes against the reverted source**:

```
Expected element to have accessible name:
  __force_failure__
Received:
  Chart Assistant - AI Chat Interface     ← with the fix REVERTED
```

`dom-accessibility-api` does not fold a descendant button's `aria-label` into the row's name during name-from-content the way a browser does, so the polluted name never appears in jsdom. A name assertion there would have passed against the bug it was written to reject — the same failure mode as the substring-match issue above, reached from a different direction.

So the Chat case asserts **where the dialog points** — the element behind `aria-labelledby` holds no interactive control — which is the structure that produced the name and does discriminate. The computed name stays asserted in `dialogAccessibility.spec.ts`, and a comment there now records that it is the only assertion anywhere covering the concatenation.

Bidirectional, which is how the above surfaced: with the three source files reverted, 5 of the 8 fail. The 3 that still pass are name characterisations for Help and Description, whose names were never broken — only their outline and duplicate id were.

## Merged `main` in three times (`c139853`, `d0ecbd9`, `ca482f3`)

None is a routine refresh. GitHub does not re-run a PR's checks when the base moves, so each time `main` advanced, the green board here was left attesting to a base that no longer existed.

**`c139853`** — #679 (*make the WebKit project pass*) edits **this same spec**, on a line adjacent to one of mine, and rewrites the page objects it depends on (`base-page.ts`, `barplot-page.ts`, `constants.ts`, plus a new `utils/platform.ts`). #684 and a dependency bump also landed. The two changes compose correctly: the command-palette test now has main's `await modifierKey(page)` alongside this PR's named-dialog and structure assertions.

**`d0ecbd9`** — six commits, three of which reach this PR directly:

- **#689** rewrites `barplot-page.ts` again, moving autoplay up to the base class and adding `override` throughout. Every test in this spec drives the page through `BarPlotPage`.
- **#695** and **#699** change how chat messages render — sanitisation now lets headings through, and `TypingEffect` gained an `h2` override. That is heading markup landing *inside* the chat dialog whose outline this PR asserts on.
- **#690** turns on `noImplicitOverride` repo-wide, so this branch had never been type-checked under the flag.

**`ca482f3`** — #700 and #701. #700 adds a `no-restricted-syntax` selector banning `aria-label` built from `children`; checked against this diff rather than assumed irrelevant — every `aria-label` here is a string literal or a template-literal ternary, and the only occurrence of "children" is inside a prose comment. #701 is the one that reaches this PR: it splits Jest into `unit` and `esm` projects and moves `npm test` behind `scripts/test.js`, which is the harness running the new `dialogTitles.test.tsx`. Confirmed it lands in the `unit` project (`PASS unit test/ui/dialogTitles.test.tsx`) rather than being swept into the ESM one, and that the `jest.mock('@ui/components/MessageBubble')` workaround is still the right call under #701's updated rule — "reach for [ESM] only when the CommonJS project cannot do the job", and this one can.

Re-verified on each merged tree rather than reasoned about: `tsc --noEmit` clean, `eslint` clean, `npm run build` clean, jest green through the new runner, and `dialogAccessibility` 7/7.

## To flag

**Firefox and WebKit are still not covered by any check on this PR.** The e2e job here is chromium-only by design — `ci.yml` is explicit that "the other two browsers stay on the scheduled e2e workflows" — so the first Firefox/WebKit run against this change is a scheduled one, after merge. #679 is actively making the WebKit project pass, which narrows the risk considerably compared with when this PR opened, but it does not make this change verified there. Static JSX/ARIA markup with no browser-specific API, so I read the risk as low — judgement, not a result.

Worth noting the PR e2e job also changed shape in #679: it now runs the **whole** chromium suite rather than a two-spec smoke subset, because per its own comment "a subset let the specs rot for two months while PRs stayed green (see issue #626)". So the gate on this PR is meaningfully broader than it was.

**Deliberately out of scope:** `Chat.tsx`'s heading reads "Chart Assistant" visually while its `aria-label` makes its accessible name "Chart Assistant - AI Chat Interface". Pre-existing, and not a WCAG 2.5.3 failure — the visible text is a prefix of the accessible name. But this PR is what makes that element the dialog's canonical name source, so it is worth its own issue: whether the suffix should be heard by someone navigating by heading, or belong to the dialog alone. Changing it here would change the dialog's accessible name, which is beyond "give each dialog exactly one title heading".

https://claude.ai/code/session_01VH49ExPkrEGyEWZJj1EFmm


jooyoungseo:
Thanks — the diagnosis read matches mine on all three dialogs. One answer, on the CSS nitpick.

### The spacing does not change

The premise is right and the conclusion does not follow from it. There is indeed no `CssBaseline`/`ThemeProvider` and no global stylesheet, so UA heading defaults *are* live on the page — but MUI's `Typography` sets `margin: 0` in its own root styles (`Typography.js:59`), which overrides them regardless of which element `component` resolves to.

Measured rather than reasoned about, in chromium against the built bundle. Probe elements injected into the same dialog confirm the UA defaults are genuinely active there, so this is not a case of the reset being absent:

| Element | margin-top | margin-bottom | font-size |
| --- | --- | --- | --- |
| `Axes` (`subtitle2` → `h3`, `sx={{ mt: 1 }}`) | 8px | **0px** | 14px |
| `Summary` (same) | 8px | **0px** | 14px |
| `Data (4 rows)` (`sx={{ mt: 1, mb: 1 }}`) | 8px | 8px | 14px |
| *bare `<h3>` probe, same container* | 18.72px | 18.72px | 18.72px |
| *bare `<h6>` probe, same container* | 24.98px | 24.98px | 10.72px |

The rendered headings sit at `0px` bottom margin, not the 18.72px a UA-default `h3` would take — so the `h6` → `h3` swap moves nothing. `font-size` stays 14px and weight 700 because `variant="subtitle2"` still supplies the typography; only the element changed. The `mt`/`mb` asymmetry between the section headings and `Data` predates this PR and is unchanged by it.

So: purely semantic, no visual delta, nothing to eyeball. No change made.

### On the unrun commands

Understood, and worth noting the PR checks now cover most of it independently: `Unit Tests`, `Type Check`, `Lint Code`, `Build Project`, `Docs Build` and `E2E Smoke (chromium)` are all green on `26b09db`. The e2e smoke job runs `dialogAccessibility.spec.ts`, so the `dialogStructure` assertions and the `exact: true` names are confirmed in CI's environment rather than only mine.

The one thing no check on this PR covers is Firefox and WebKit — `e2e-smoke` is chromium-only by design and the three-browser suite is scheduled, so a green board here is not cross-browser coverage. Flagged in the description for the same reason.

---

jooyoungseo:
Thanks — point 1 is a fair gap and is now closed in `d905668`. Points 2 and 3 I'm deliberately not changing; reasoning below.

## 1. Unit-level coverage — added, and it did not go the way it looked like it would

`test/ui/dialogTitles.test.tsx` covers all three dialogs in jsdom, so this class of regression is caught in the ordinary jest pass rather than only in the chromium-only e2e job.

The interesting part is that **the obvious version of that test does not work**, and it fails in exactly the direction that would have been worst: silently passing.

The natural assertion for the Chat defect is `toHaveAccessibleName('Chart Assistant - AI Chat Interface')` — the name should be the heading alone, without `"Close chat dialog"` appended. Running it against the reverted source shows why that is not a test:

```
Expected element to have accessible name:
  __force_failure__
Received:
  Chart Assistant - AI Chat Interface     ← with the fix REVERTED
```

`dom-accessibility-api` does not fold a descendant button's `aria-label` into the row's name during name-from-content the way a browser does, so in jsdom the polluted name never appears. A name assertion there passes against the bug it was written to reject — the same failure mode as the Playwright substring-match issue this PR already fixed, arrived at from a different direction.

So the Chat case asserts **where the dialog points**, which is the structure that produced the name and is checkable in jsdom: the element behind `aria-labelledby` contains no interactive control. Reverted, that element is the row holding the close button, and it fails. The computed name stays asserted in `dialogAccessibility.spec.ts`, where a real accessibility tree computes it.

Verified in both directions, which is how the above surfaced:

| | with fix | fix reverted |
| --- | --- | --- |
| `dialogTitles.test.tsx` | 8/8 pass | **5 fail**, 3 pass |

The three that still pass are name characterisations for Help and Description — those names were correct before this PR too, since only the outline and the duplicate id were broken there. Worth stating plainly rather than padding the count.

Full suite after the change: 105 suites, 1336 tests, green; `tsc --noEmit`, `eslint`, `npm run build` clean.

One implementation note: the test mocks `@ui/components/MessageBubble`, following `test/adapters/recharts/panelDom.test.ts`. Chat opens with no messages so it never mounts, and it pulls in the ESM-only `react-markdown` chain that the ts-jest CJS transform does not cover.

## 2. `Chat.tsx`'s `aria-label` diverging from its visible text

Correct, and deliberately left alone. `aria-label="Chart Assistant - AI Chat Interface"` on a heading reading "Chart Assistant" predates this PR, and changing it would change the dialog's accessible name — a behaviour change beyond "give each dialog exactly one title heading."

You are right that this PR makes it *more* visible: the dialog is now named after that heading specifically, so the divergence became load-bearing for the name in a way it was not before. That is an argument for a follow-up issue, not for widening this diff. Happy to file one — the substantive question is whether the "- AI Chat Interface" suffix earns its keep at all, given it is inaudible to a sighted user and the dialog role already announces "dialog".

## 3. Implicit `h6` variant on `DialogTitle`

Agreed on the reading, and taking you at your word that it is not a change request. Adding `variant="h6"` back would restate MUI's own default, which is the kind of line that later gets "cleaned up" by someone who checks that it is redundant. The rendered output is identical.

---

jooyoungseo:
Both observations taken — `49deeb3`.

**The unscoped ancestor walk was a real defect in my helper**, not just a theoretical one, and worth more than the "not worth guarding against now" you gave it. `closest` walks to the document root, and MAIDR embeds in host pages it does not control, so the false positive would have come from *someone else's* markup and pointed at this dialog. Both copies of the helper now check `dialog.contains(ancestor)`.

Confirmed it still discriminates rather than just going quiet — with the three source files reverted, `dialogTitles.test.tsx` is still 5 red / 3 green, the same split as before the change, and `dialogAccessibility.spec.ts` is 7/7 with the fix in place.

**On the division of labor** — agreed it is easy to assume the jsdom suite is a superset, so it is now written down at the assertion that carries the difference:

```ts
// This is also the only assertion anywhere that covers the concatenation.
// `test/ui/dialogTitles.test.tsx` checks the same dialog's structure, but
// cannot check its name: jsdom does not fold a descendant button's
// `aria-label` into the row's computed name, so the polluted name never
// appears there. The jsdom suite is not a superset of this one.
```

The jsdom side already carried the same note from its end; what was missing was the pointer in the direction a reader is more likely to travel.

**On the Chat heading's `aria-label`** — same item as point 2 of the earlier review, answered [above](https://github.com/xability/maidr/pull/666#issuecomment-5153352436): pre-existing, deliberately out of scope here because changing it changes the dialog's accessible name, and worth its own issue rather than a wider diff.

---

jooyoungseo:
The `CSS.escape` note was worth pushing on — **my comment was wrong, and I had checked it against the wrong install.** Corrected in `a55ef47`.

I verified it originally by constructing a `JSDOM` from the `jsdom` in devDependencies (22.x). But that is not the environment these tests run in: `jest-environment-jsdom@29.7.0` bundles its own nested `jsdom@20.0.3`, and `.claude/rules/testing.md` even warns that the two are separate installs on separate major lines. Probing inside the jest environment instead:

```
typeof CSS:        undefined
typeof CSS.escape: undefined
```

So the whole `CSS` object is absent there — `CSS.escape(...)` would be a `ReferenceError`, not a missing method. The property match was already the right implementation; only the reason given for it was inaccurate, which is the kind of comment that sends the next reader to a polyfill check that cannot work. It now says which install it refers to and why.

On the other two:

- **Duplicated `dialogStructure`** — agreed, and I'd rather keep it duplicated. The Playwright copy is serialized into the browser context by `page.evaluate` and the jsdom copy closes over `screen`; sharing them would mean a helper written to survive serialization, which costs more clarity than the ~15 duplicated lines. Noting your point in case the count grows.
- **Chat's `aria-label`** — third review to raise it, so the signal is clear. Still declining to fold it into this diff for the reason given [earlier](https://github.com/xability/maidr/pull/666#issuecomment-5153352436) — it changes the dialog's accessible name, which is beyond "give each dialog exactly one title heading" — but you are right that this PR is what makes it load-bearing. Flagged with the repo owner for a follow-up issue.

---

jooyoungseo:
Point 1 taken — `512ae14`. "PR descriptions don't travel with the code" is the right argument, and it is the one thing five reviews in a row have raised, which is itself evidence that the reasoning needs to live next to the line rather than in a description nobody will have open.

The comment now sits directly above the `Typography`:

```tsx
{/* The `aria-label` below predates #665 and is left alone on
    purpose. It makes the accessible name longer than the visible
    text, and since the dialog is now named after this heading,
    editing it would change the dialog's name too — a behaviour
    change beyond giving each dialog one title heading. It is not a
    2.5.3 failure either: the visible text is a prefix of the name.
    Whether the suffix should be heard when navigating by heading,
    rather than only when the dialog opens, is its own question. */}
```

I did not use the issue reference from your suggested wording — #665 is the issue this PR closes, not one about the `aria-label`, so citing it as the explanation would point the next reader somewhere that does not discuss it. The comment states the reasoning directly instead, and names the open question rather than implying it is settled.

Comment only; no behaviour change. `tsc`, `eslint` and `dialogTitles.test.tsx` (8/8) all still clean.

On the other two:

- **Duplicated `dialogStructure`** — agreed it stays duplicated, for the reason you identified: one runs inside `page.evaluate` and the other against jsdom directly. Sharing them would mean a helper written to survive serialisation, which costs more clarity than the ~30 lines. Third review to note it, so it is well and truly on record if the contract changes again.
- **Firefox/WebKit** — right, and worth being concrete: the cross-browser suite is the scheduled `e2e_tests.yml` (`cron: 37 2 */2 * *`), so the first run covering this change is a scheduled one after merge, not anything on this PR. Confirming green there rather than inferring it from the chromium job is the correct standard.

Base note: this branch has since merged `main` up to `625c6cf`, which includes #708's `chatViewModel` / `src/type/llm.ts` refactor. That one mattered — `dialogTitles.test.tsx` stubs the view model as `Pick<ChatViewModel, 'toggle' | 'sendMessage' | 'canSend'>` specifically so a signature change breaks the test, and `Chat.tsx` imports `Message` from `@type/llm`. Re-verified on the merged tree: `tsc` clean, unit 107 suites / 1341 tests, esm 5 / 57, `dialogAccessibility` 7/7 on chromium.

---

PR #667, opened 2026-07-29
fix(build): stop emitting a core ES bundle that UMD overwrites

jooyoungseo:
## Problem

The `core` bundle was declared with two output formats that collided on one filename. In `scripts/build.js`:

```js
{
  name: 'core',
  entry: 'src/index.tsx',
  libName: 'maidr',
  formats: ['es', 'umd'],
  fileName: () => 'maidr.js',   // ignores its `format` argument
  ...
}
```

The `fileName` callback ignores its `format` argument and returns `'maidr.js'` for **both** formats. Vite wrote the ES output first and the UMD output then silently overwrote it, so roughly half of the core build's work was discarded on every build.

Evidence from the build log on `main` — two `dist/maidr.js` lines with different sizes, but only one file surviving in `dist/`:

```
dist/maidr.css  1,458.77 kB │ gzip: 945.07 kB
dist/maidr.js   3,224.37 kB │ gzip: 757.80 kB │ map: 9,020.07 kB   <- ES, written first
dist/maidr.css  1,458.77 kB │ gzip: 945.07 kB
dist/maidr.js   1,890.62 kB │ gzip: 553.21 kB │ map: 8,684.68 kB   <- UMD, overwrites it
✓ built in 52.40s
```

The single surviving `dist/maidr.js` begins with a UMD wrapper, confirming the ES output lost the race:

```
(function(Xu){typeof define=="function"&&define.amd?define(Xu):Xu()})(...
```

The root `vite.config.ts` declared the same `formats: ['es', 'umd']` + `fileName: () => 'maidr.js'` combination and had the same problem.

## Fix

Stop building the `es` format for `core` rather than emitting a second file.

`src/index.tsx` contains **no `export` statements at all** (verified — the only `export` match in the file is the word inside a comment). It is a pure side-effect entry point: it assigns `window.maidrLive` and `window.disconnectMaidrObservers`, registers DOM listeners, and auto-initialises charts on `DOMContentLoaded`. An ES-module build of an entry with no exports has no consumer value, and its output never survived anyway.

So `formats` becomes `['umd']` in both `scripts/build.js` (the `core` entry) and `vite.config.ts`. `fileName` is left as-is — it is now correct, and matches how the other single-format entries are written.

### Scope

- **Only the `core` entry changed.** Every other entry in the `builds` array already handled this correctly and is untouched: multi-format adapters use `fileName: format => format === 'es' ? 'x.mjs' : 'x.js'`. The full build confirms this still holds — e.g. `anychart` emits both `anychart.mjs` (39.57 kB) and `anychart.js` (28.62 kB).
- **`package.json` needed no change**, confirmed rather than assumed:
  - `main` is `dist/maidr.js`
  - `exports["."]` is `{"default": "./dist/maidr.js"}` — a single condition, with no `import` key
  - there is no `module`, `types`, or `browser` field at the root

  Both already pointed at the UMD file that is actually produced, so they stay correct. Nothing in the repo referenced a core ES build; the only `maidr.mjs` string outside `dist/` is in `test/util/diagnostics.test.ts`, exercising the permissive `isMaidrScriptUrl()` URL matcher, which does not depend on the build emitting that file.
- `dist/` is gitignored and untouched.

## Verification

**1. Core build now prints one `dist/maidr.js` line instead of two**

```
[1/1] Building core...
dist/maidr.css  1,458.77 kB │ gzip: 945.07 kB
dist/maidr.js   1,890.62 kB │ gzip: 553.21 kB │ map: 8,684.83 kB
✓ built in 37.31s
[1/1] Done (37.4s)
```

Line count of `dist/…maidr.js ` in the log: **2 before → 1 after**.

Core bundle build time: **52.5s → 37.4s** (~29% faster, 15.1s saved).

**2. The surviving artifact is unchanged** — `dist/maidr.js` is 1,890,618 bytes (~1.88 MB), byte-for-byte the same size as the UMD output produced before this change, and still starts with `(function(Xu){typeof define=="function"&&define.amd?define(Xu):Xu()})(`.

**3. No stray outputs** — no `dist/maidr.mjs`. The core build emits exactly `maidr.css`, `maidr.js`, `maidr.js.map`.

**4. Runtime smoke test (real browser, via Playwright)** — served a page loading the built `dist/maidr.js` plus an inline SVG bar chart carrying a `maidr` attribute with a valid spec (4 bars: Sat/87, Sun/76, Thur/62, Fri/19). Results:
- `window.maidrLive` is present, so the UMD side effects ran
- `<figure id="maidr-figure-testbar">` mounted, with a focusable `div[tabindex="0"][role="application"]`
- arrow-key navigation works and announces data points: first `ArrowRight` → **"Day is Sat, Count is 87"**, second `ArrowRight` → **"Day is Sun, Count is 76"**, matching the spec
- `HighlightService` created a `data-maidr-owned="true"` highlight rect on the navigated bar
- no console errors other than an unrelated `favicon.ico` 404

**5. Full build** — `node scripts/build.js` succeeds (all 12 bundles, 284.8s, exit 0) with **no "Merge collision" error** and no build errors. `dist/.tmp` is swept.

**6. Type-check and tests**
- `npm run type-check` — clean, no errors
- `npx jest --coverage=false` — **96 suites passed / 96 total, 1209 tests passed / 1209 total, 0 failures**
- The anticipated pre-existing `test/adapters/recharts/panelDom.test.ts` failure (`Cannot find module 'recharts'`) did **not** occur here, because `npm ci` installed the `recharts` devDependency in this tree; that suite passes (4/4).

**7. Lint** — `npx eslint scripts/build.js vite.config.ts` clean.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -144,7 +144,10 @@ const builds = [
     name: 'core',
     entry: 'src/index.tsx',
     libName: 'maidr',
-    formats: ['es', 'umd'],
+    // UMD only: src/index.tsx is a pure side-effect entry with no exports, so
+    // an ES build has no consumer value. Adding 'es' back here would also make
+    // both formats resolve to the same fileName and silently overwrite.
+    formats: ['umd'],
     fileName: () => 'maidr.js',
     emptyOutDir: true,
     external: [],
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -138,8 +138,11 @@ function onWarn(warning, warn) {
 
 /**
  * Build configurations
+ *
+ * Exported so the build-config test can assert against the real array rather
+ * than a fixture that could drift away from it.
  */
-const builds = [
+export const builds = [
   {
     name: 'core',
     entry: 'src/index.tsx',
@@ -499,8 +502,71 @@ async function runParallel(selected, jobs, outDir) {
     throw firstError;
 }
 
+/**
+ * Reject any entry whose formats would fight over one output filename.
+ *
+ * Vite resolves `build.lib.fileName` once per format and writes the outputs in
+ * order, overwriting silently when two formats resolve to the same name: the
+ * earlier format's work is simply lost, and the only hint is the same filename
+ * appearing twice in the build log. That is exactly how the core bundle spent
+ * every build producing an ES output that the UMD output immediately replaced,
+ * so catch the whole class here instead of shipping half a build again.
+ *
+ * `fileName` is invoked the way Vite invokes it — `(format, entryName)` — so an
+ * implementation that ignores its `format` argument (and therefore collides) is
+ * caught behaving exactly as it would at build time.
+ *
+ * @param {typeof builds} configs Entries to check.
+ * @throws {Error} If any entry maps two formats onto the same filename.
+ */
+export function assertUniqueOutputFilenames(configs) {
+  for (const config of configs) {
+    // Vite's lib-mode default when `name` is set. Every entry declares
+    // `formats` today; defaulting keeps an omission from skipping the check.
+    const formats = config.formats ?? ['es', 'umd'];
+    // A string `fileName` cannot collide — Vite appends a per-format extension.
+    if (typeof config.fileName !== 'function')
+      continue;
+    // Vite derives the entry name from the entry file's base name.
+    const entryName = path.basename(config.entry, path.extname(config.entry));
+
+    /** @type {Map<string, string>} filename -> the format that claimed it */
+    const claimedBy = new Map();
+    for (const format of formats) {
+      let fileName;
+      try {
+        fileName = config.fileName(format, entryName);
+      } catch (cause) {
+        throw new Error(
+          `Build config error: bundle "${config.name}" threw while resolving its `
+          + `fileName for the "${format}" format: ${cause.message}`,
+          { cause },
+        );
+      }
+
+      const previous = claimedBy.get(fileName);
+      if (previous !== undefined) {
+        throw new Error(
+          `Build config error: bundle "${config.name}" emits "${fileName}" for both the `
+          + `"${previous}" and "${format}" formats. Vite writes them in order, so `
+          + `"${format}" would silently overwrite "${previous}" and that output would be `
+          + `lost. Give each format its own filename — the adapter entries use `
+          + `\`fileName: format => format === 'es' ? '<name>.mjs' : '<name>.js'\` — or drop `
+          + `the redundant format.`,
+        );
+      }
+      claimedBy.set(fileName, format);
+    }
+  }
+}
+
 async function main() {
   const startTime = Date.now();
+
+  // Fail fast, before emptying dist or forking anything. Checked across every
+  // entry rather than just the selected ones: a collision in a bundle nobody
+  // asked for today is still a latent bug, and the check costs nothing.
+  assertUniqueOutputFilenames(builds);
 
   const argv = process.argv.slice(2);
   const sequential = argv.includes('--sequential');
@@ -592,7 +658,14 @@ async function main() {
   console.log(`\nAll builds complete in ${((Date.now() - startTime) / 1000).toFixed(1)}s`);
 }
 
-main().catch((err) => {
-  console.error('Build failed:', err);
-  process.exit(1);
-});
+// Orchestrate only when this file is the process entry point — which includes
+// each forked worker, whose argv[1] is this same path. Importing the module,
+// as the build-config test does to reach the real `builds` array, must never
+// kick off a build.
+const invokedAs = process.argv[1] ? path.resolve(process.argv[1]) : '';
+if (invokedAs === fileURLToPath(import.meta.url)) {
+  main().catch((err) => {
+    console.error('Build failed:', err);
+    process.exit(1);
+  });
+}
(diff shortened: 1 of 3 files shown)


jooyoungseo:
Implemented the review suggestion in a67cd8e1 — a guard so this bug class fails loudly instead of silently losing output.

## What changed

`scripts/build.js` gained `assertUniqueOutputFilenames()`, which resolves `fileName` for every format of every entry and throws when two formats claim the same name. It runs as the **first statement of `main()`** — the single entry point for the parent orchestrator, the sequential path, and each forked worker — so no path can slip a bad config through. It validates the **whole `builds` array**, not just the bundles named on the command line, and runs before dist is emptied or any worker is forked.

Details that make the check reflect reality rather than an idealised config:

- `fileName` is invoked the way Vite invokes it, `(format, entryName)`, with `entryName` derived from the entry file's base name — so a callback that ignores its `format` argument is caught behaving exactly as it would at build time.
- A throwing `fileName` is re-thrown naming the bundle and format, with the original attached as `cause`.
- An entry omitting `formats` is checked against Vite's lib-mode default of `['es', 'umd']`, so an omission isn't a hole.
- A string `fileName` is skipped — Vite appends a per-format extension, so it cannot collide.

## Evidence it actually fires

Temporarily restoring `formats: ['es', 'umd']` on `core` aborted every path with exit 1:

```
$ node scripts/build.js core
Build failed: Error: Build config error: bundle "core" emits "maidr.js" for both the
"es" and "umd" formats. Vite writes them in order, so "umd" would silently overwrite
"es" and that output would be lost. Give each format its own filename — the adapter
entries use `fileName: format => format === 'es' ? '<name>.mjs' : '<name>.js'` — or
drop the redundant format.
    at assertUniqueOutputFilenames (scripts/build.js:549:15)
    at main (scripts/build.js:569:3)
EXIT: 1
```

- `node scripts/build.js core` → exit 1
- `node scripts/build.js react` → exit 1 as well, proving a collision in an **unselected** entry is still caught
- `node scripts/build.js` (full parallel) → exit 1
- `dist/` stayed at 52 files across all three, confirming it aborts before emptying anything

That experiment was then reverted and the file diffed byte-for-byte against its pre-experiment state.

## Test

New `test/scripts/buildOutputFilenames.test.ts` (11 cases). `scripts/build.js` is plain ESM under `allowJs: false`, so it drives the module through `node --input-type=module` subprocesses, following the existing `execFileSync` pattern in `test/scripts/`.

To let the test assert against the **real `builds` array** rather than a fixture that could drift, `builds` and the validator are now exported and `main()` runs only when the file is the process entry point — which still includes each forked worker, whose `argv[1]` is that same path. The full 12-bundle parallel build passing is the proof that workers still run. Nothing else in the repo imports `scripts/build.js` (only `package.json` and `ci.yml` invoke it as a command).

Covered: the real `builds` array passes; importing the module starts no build; a format-ignoring `fileName` is rejected with a message naming the bundle and filename; per-format naming and single-format entries are accepted; a three-format collision is caught; a throwing `fileName` is surfaced; and the omitted-`formats` / string-`fileName` cases behave as described.

I also mutation-checked the tests — replacing the validator body with a no-op failed exactly the 4 rejection cases, so they aren't passing vacuously.

## Verification

- `node scripts/build.js core` — exit 0, 37.7s, still one `dist/maidr.js` line
- `node scripts/build.js` (full) — exit 0, 173.1s, 52 files, no collision error; `dist/maidr.js` still 1,890,618 bytes, no `dist/maidr.mjs`
- `npx jest --coverage=false` — **97 suites passed / 97, 1220 tests passed / 1220** (up from 96 / 1209; +1 suite, +11 tests)
- `CI=true npx eslint scripts/build.js vite.config.ts test/scripts/buildOutputFilenames.test.ts` — clean, exit 0 (ran with `CI=true` so antfu's in-editor rule relaxation was not in effect)

jooyoungseo:
Confirming CI, since the review sandbox couldn't run it: **all 13 checks pass** on `a67cd8e1` — Unit Tests (97 suites / 1220 tests), Type Check, Build Project, E2E Smoke (chromium), Lint Code, Lint Commit Messages, CodeQL, Docs Build.

No code changes from this round. Taking the three observations in turn:

**`formats` default in the guard** — agreed it's dead code today; every entry sets both `formats` and `libName`. It's there so a future entry that omits `formats` isn't silently skipped by the very check meant to catch it, which felt like the safer failure direction for a guard. Happy to drop it if you'd rather not carry an unexercised branch.

**Subprocess overhead per test case** — noted. Left as-is at this scale; if a third `test/scripts/*` file needs the same trick, that's the point to extract a shared helper rather than now.

**Brace style** — `Lint Code` passes, including under `CI=true` (which re-enables the rules `@antfu/eslint-config` relaxes in editor mode), so it isn't a `curly` violation. No change.

One thing worth recording, because CI can't catch it: the guard moves `main()` behind an `invokedAs === fileURLToPath(import.meta.url)` check so the test can import `builds` without triggering a build. If that comparison ever evaluated false, `npm run build` would become a silent no-op — the same class of failure this PR exists to fix. CI only runs Linux, and this comparison is a classic place for Windows drive-letter casing to bite, so it was checked on Windows across four invocation styles: relative path, absolute path, lowercase drive letter (`c:/…`), and via `npm run`. All four match — Node normalizes `import.meta.url` to the same casing as `argv[1]`.

PR #668, opened 2026-07-29
perf(css): ship only woff2 for inlined KaTeX fonts

jooyoungseo:
Two related bundle-size fixes.

## 1. `dist/maidr.css`: 1,425 KB → 359 KB

`src/ui/components/TypingEffect.tsx` imports `katex/dist/katex.min.css` so LaTeX in AI chat responses renders, and Vite's library mode inlines the fonts that stylesheet references as base64 data URIs. KaTeX declares each of its 20 faces with three alternatives:

```css
src:url(fonts/KaTeX_AMS-Regular.woff2) format("woff2"),
    url(fonts/KaTeX_AMS-Regular.woff)  format("woff"),
    url(fonts/KaTeX_AMS-Regular.ttf)   format("truetype")
```

So every glyph set shipped three times. 1,402 KB of the 1,425 KB file was font data — MAIDR's own styling is only ~23 KB of it:

| inlined format | before |
|---|---|
| `font/ttf` | 669 KB |
| `font/woff` | 395 KB |
| `font/woff2` | 338 KB |

A browser picks exactly one alternative per face, and woff2 has been baseline since Chrome 36, Safari 12, Firefox 39 and Edge 14 — so the woff and ttf copies were pure duplication.

**Result:** `dist/maidr.css` goes from **1,458,693 B (1424.5 KiB)** to **367,893 B (359.3 KiB)** — a 74.8% reduction, 1,065 KiB off every page that loads MAIDR's stylesheet. Gzipped: 264 KB.

### How

`scripts/vite-plugin-woff2-only.js` — a `generateBundle` transform over emitted CSS assets, not a vendored copy of KaTeX's stylesheet, so it stays correct across KaTeX upgrades.

Three things a naive implementation gets wrong, all pinned by tests:

- **Splitting a `src` list needs to be paren-aware.** An inlined `url(data:font/woff2;base64,...)` contains both a comma and a semicolon, so `value.split(',')` shreds it.
- **It needs to be comment-aware too.** A `;` or `:` inside a `/* ... */` would otherwise split a declaration in the wrong place. Minified output carries no comments today, so this was latent rather than live, but a change in minifier settings would surface it silently. `splitTopLevel` now steps over comments wholesale, and classification runs against a comment-stripped copy so a commented-out `format("woff2")` cannot make a legacy source look current. Comments are only ignored for *inspection* — the original text, comments included, is what gets emitted.
- **A face must never end up without a source.** The transform only trims a face that actually offers a woff2 alternative; a face without one is left untouched (and reported via `this.warn`) rather than emptied. It also throws outright rather than emit an empty `src`. A future KaTeX that drops woff2 from some face would grow the bundle back rather than break maths rendering.

`local(...)` alternatives, and any source the transform does not positively recognise as a pre-woff2 font, are preserved. A comment hiding a closing brace, or an unterminated comment, degrades to "no change" rather than emitting half a rule.

Deliberately **not** switching to postcss — pulling in a parser is a much larger change than this PR should carry.

The plugin is registered in the shared `createViteConfig` factory in `scripts/build.js`, unconditionally — 11 of the 12 bundles emit a byte-identical `maidr.css`, and `runParallel`'s merge step fails with `Merge collision` if they diverge. It is mirrored in `vite.config.ts` so a build driven from either config agrees.

## 2. Dead `rehype-mathjax` dependency removed

`rehype-mathjax` was the original renderer for chat LaTeX (`import rehypeMathjax from 'rehype-mathjax/svg'`), replaced by `rehype-katex` in #433. The dependency was never cleaned up — it is imported nowhere in `src/`, `examples/`, `scripts/` or any config file (re-verified by grep across the repo; the only hit was the `package.json` line itself).

KaTeX is also the more accessible of the two: the removed SVG mode emitted `<svg role="img" focusable="false">` with no MathML and no accessible name, where KaTeX emits real hidden MathML alongside the visual output. So KaTeX stays.

Removing it prunes 10 packages from the install tree (`mathjax-full`, `speech-rule-engine`, `mj-context-menu`, `mhchemparser`, `@types/mathjax`, …). `rehype-katex`, `remark-math`, `rehype-sanitize` and `remark-gfm` are untouched.

`package-lock.json` is updated to match. **Correcting an earlier claim in this PR description and in the review: this repo *does* track a lockfile, and the release workflow does not regenerate it** — its `rm -f package-lock.json` only deletes the file on the CI runner before `npm install`, and nothing commits one back. Leaving `package.json` and the lockfile disagreeing would have meant (a) the next `npm install` dragging an unrelated lockfile diff into someone else's commit and (b) `npm audit`/Dependabot continuing to scan `mathjax-full`, `speech-rule-engine` and friends for a dependency the project no longer has. Local `npm audit` goes from 32 findings to 30.

The lockfile diff was kept to exactly the intended change, verified programmatically — **0 packages added, 0 version changes, 0 dependency ranges touched**:

- **10 entries removed:** `rehype-mathjax@5.0.0`, `mathjax-full@3.2.1`, `speech-rule-engine@4.1.3` (+ its pinned `commander@13.1.0`), `mj-context-menu@0.6.1`, `mhchemparser@4.2.1`, `esm@3.2.25`, `@xmldom/xmldom@0.9.9`, `wicked-good-xpath@1.3.0`, `@types/mathjax@0.0.37`.
- **51 packages reclassified `dev: true`.** `rehype-mathjax` depended on `jsdom@^22`, so jsdom and its tree (`tough-cookie`, `whatwg-url`, `ws`, `saxes`, …) were reachable from a *production* dependency; they are now reachable only through the jest devDependencies. That is the point of the change, not churn.
- **Two npm-11.7.0 artifacts reverted**, since both reproduce when regenerating from an unmodified `main` and so are unrelated to this PR: bumping the lock's own `version` 3.74.0 → 3.75.0 (it trails because the release workflow deletes the file), and dropping the `libc` field from 21 optional linux binaries.

`npm ci` reinstalls cleanly from the committed result and leaves the file byte-identical.

## Explicitly out of scope

**Splitting KaTeX out of the default stylesheet, or loading it lazily, is intentionally left to a follow-up PR.** This change only removes redundancy from what already ships — the remaining 359 KB is the single woff2 copy of each face, still eagerly loaded.

## Verification

- `node scripts/build.js core` — succeeds.
- `node scripts/build.js` (all 12 bundles) — succeeds in 235 s, **zero** `Merge collision`, all emissions of `maidr.css` identical at 367.89 kB, zero plugin warnings.
- Programmatic assertions on the rebuilt `dist/maidr.css`: 367,893 B; 20 `@font-face` blocks; exactly **one** `src` source per face, all 20 `format("woff2")` with a `data:font/woff2` URI; **0** `font/ttf`, **0** `font/woff`, 0 other font data URIs, 0 unresolved font urls.
- Browser check (Playwright/Chromium, page linking the rebuilt CSS with a rendered KaTeX fraction): the browser parses 20 `CSSFontFaceRule`s, and for all 20 exact `(family, style, weight)` triples `await document.fonts.load(...)` resolves to one face and `document.fonts.check(...)` is `true`. Zero failures; the only console error is a favicon 404.
- `npm run type-check` — clean.
- `npx jest --coverage=false` — **97 suites / 1222 tests, all passing**, including 13 new cases for the transform. (`test/adapters/recharts/panelDom.test.ts` passes; the `recharts` devDependency is installed.)
- `npm ci` — reinstalls cleanly from the committed lockfile and leaves it byte-identical.
- `npx eslint` on all changed files — clean, also with `CI=true` (which enables the rules antfu's config relaxes in-editor).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,6 @@
         "react-markdown": "^10.1.0",
         "react-redux": "^9.2.0",
         "rehype-katex": "^7.0.1",
-        "rehype-mathjax": "^5.0.0",
         "rehype-sanitize": "^6.0.0",
         "remark-gfm": "^4.0.1",
         "remark-math": "^6.0.0"
@@ -5254,6 +5253,7 @@
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/@tootallnate/once/-/once-2.0.0.tgz",
       "integrity": "sha512-XCuKFP5PS55gnMVu3dty8KPatLqUoy/ZYzDzAGCQ8JNFCkLXzmI7vNHCR+XpbZaMWQK/vQubr7PkYq8g470J/A==",
+      "dev": true,
       "license": "MIT",
       "engines": {
         "node": ">= 10"
@@ -5773,12 +5773,6 @@
       "version": "0.16.8",
       "resolved": "https://registry.npmjs.org/@types/katex/-/katex-0.16.8.tgz",
       "integrity": "sha512-trgaNyfU+Xh2Tc+ABIb44a5AYUpicB3uwirOioeOkNPPbmgRNtcWyDeeFRzjPZENO9Vq8gvVqfhaaXWLlevVwg==",
-      "license": "MIT"
-    },
-    "node_modules/@types/mathjax": {
-      "version": "0.0.37",
-      "resolved": "https://registry.npmjs.org/@types/mathjax/-/mathjax-0.0.37.tgz",
-      "integrity": "sha512-y0WSZBtBNQwcYipTU/BhgeFu1EZNlFvUNCmkMXV9kBQZq7/o5z82dNVyH3yy2Xv5zzeNeQoHSL4Xm06+EQiH+g==",
       "license": "MIT"
     },
     "node_modules/@types/mdast": {
@@ -6574,15 +6568,6 @@
         "@xtuc/long": "4.2.2"
       }
     },
-    "node_modules/@xmldom/xmldom": {
-      "version": "0.9.9",
-      "resolved": "https://registry.npmjs.org/@xmldom/xmldom/-/xmldom-0.9.9.tgz",
-      "integrity": "sha512-qycIHAucxy/LXAYIjmLmtQ8q9GPnMbnjG1KXhWm9o5sCr6pOYDATkMPiTNa6/v8eELyqOQ2FsEqeoFYmgv/gJg==",
-      "license": "MIT",
-      "engines": {
-        "node": ">=14.6"
-      }
-    },
     "node_modules/@xtuc/ieee754": {
       "version": "1.2.0",
       "resolved": "https://registry.npmjs.org/@xtuc/ieee754/-/ieee754-1.2.0.tgz",
@@ -6602,6 +6587,7 @@
       "resolved": "https://registry.npmjs.org/abab/-/abab-2.0.6.tgz",
       "integrity": "sha512-j2afSsaIENvHZN2B8GOpF566vZ5WVk5opAiMTvWgaQT8DkbOqsTfvNAvHoRGU2zzP8cPoqys+xHTRDWW8L+/BA==",
       "deprecated": "Use your platform's native atob() and btoa() methods instead",
+      "dev": true,
       "license": "BSD-3-Clause"
     },
     "node_modules/acorn": {
@@ -6668,6 +6654,7 @@
       "version": "6.0.2",
       "resolved": "https://registry.npmjs.org/agent-base/-/agent-base-6.0.2.tgz",
       "integrity": "sha512-RZNwNclF7+MS/8bDg70amg32dyeZGZxiDuQmZxKLAlQjr3jGyLx+4Kkk58UO7D2QdgFIQCovuSuZESne6RG6XQ==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "debug": "4"
@@ -6905,6 +6892,7 @@
       "version": "0.4.0",
       "resolved": "https://registry.npmjs.org/asynckit/-/asynckit-0.4.0.tgz",
       "integrity": "sha512-Oei9OH4tRh0YqU3GxhX79dM/mwVgvbZJaSNaRk+bshkj0S5cfHcgYakreBjrHwatXKbz+IoIdYLxrKim2MjW0Q==",
+      "dev": true,
       "license": "MIT"
     },
     "node_modules/axios": {
@@ -7309,6 +7297,7 @@
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/call-bind-apply-helpers/-/call-bind-apply-helpers-1.0.2.tgz",
       "integrity": "sha512-Sp1ablJ0ivDkSzjcaJdxEunN5/XvksFJ2sMBFfq6x0ryhQV/2b/KwFe21cMpmHtPOSij8K99/wSfoEuTObmuMQ==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "es-errors": "^1.3.0",
@@ -7716,6 +7705,7 @@
       "version": "1.0.8",
       "resolved": "https://registry.npmjs.org/combined-stream/-/combined-stream-1.0.8.tgz",
       "integrity": "sha512-FQN4MRfuJeHf7cBbBMJFXhKSDq+2kAArBlmRBvcvFE5BB1HZKXtSFASDhdlz9zOYwxh8lDdnvmMOe/+5cdoEdg==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "delayed-stream": "~1.0.0"
@@ -8123,6 +8113,7 @@
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/cssstyle/-/cssstyle-3.0.0.tgz",
       "integrity": "sha512-N4u2ABATi3Qplzf0hWbVCdjenim8F3ojEXpBDF5hBpjzW182MjNGLqfmQ0SkSPeQ+V86ZXgeH8aXj6kayd4jgg==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "rrweb-cssom": "^0.6.0"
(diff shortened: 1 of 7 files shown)


jooyoungseo:
Pushed two follow-ups in response to the review. PR description updated to match.

## 1. `splitTopLevel` is now comment-aware (183638f9)

The review named the one genuine fragility — a `;` or `:` inside a CSS comment could mis-split a declaration. Closed:

- `splitTopLevel` steps over `/* ... */` regions wholesale.
- `classifySource` and the `src` property check run against a comment-stripped copy, so a commented-out `format("woff2")` cannot make a legacy source look current.
- Comments are ignored only for *inspection* — the original text, comments included, is what gets emitted.

Conservative philosophy preserved: a comment hiding a closing brace truncates the `@font-face` match, and an unterminated comment swallows the rest of the value; both degrade to "no change" rather than emitting half a rule. A segment that is nothing but a comment is dropped from a rewritten `src` list, since carrying it through would leave a dangling comma that invalidates the whole descriptor once the browser strips the comment.

Six cases added (13 total for the transform): a comment holding a `;`, one holding a `:`, one between sources in a `src` list, a commented-out format hint, a hidden closing brace, and an unterminated comment. Each asserts the face is still trimmed to woff2 and that the emitted rule has exactly one non-empty, non-dangling `src`.

Not switching to postcss, per the review's own judgement that the current tradeoff is reasonable.

## 2. `package-lock.json` pruned (700855a6)

**Correction to the record:** the review says *"Leaving `package-lock.json` for the release workflow to regenerate is the right call per this repo's conventions."* That is not right, and it originated from my brief, not from the reviewer's analysis. This repo **does** track a lockfile (`git ls-files package-lock.json`), and `release.yml`'s `rm -f package-lock.json` only deletes it on the runner before `npm install` — nothing ever commits one back. Leaving it stale would have meant the next `npm install` dragging an unrelated diff into someone else's commit, and `npm audit`/Dependabot continuing to scan `mathjax-full`, `speech-rule-engine` and friends for a dependency the project no longer has.

The diff is exactly the intended removal, asserted programmatically — **0 packages added, 0 version changes, 0 dependency ranges touched**:

- **10 entries removed:** `rehype-mathjax@5.0.0`, `mathjax-full@3.2.1`, `speech-rule-engine@4.1.3` (+ pinned `commander@13.1.0`), `mj-context-menu@0.6.1`, `mhchemparser@4.2.1`, `esm@3.2.25`, `@xmldom/xmldom@0.9.9`, `wicked-good-xpath@1.3.0`, `@types/mathjax@0.0.37`.
- **51 packages reclassified `dev: true`** — `rehype-mathjax` depended on `jsdom@^22`, so jsdom and its tree (`tough-cookie`, `whatwg-url`, `ws`, `saxes`, …) were reachable from a *production* dependency and now are not. Verified this is caused by the removal, not pre-existing drift, by regenerating the lockfile from an unmodified `main` and confirming zero dev flips appear there.
- **Two npm-11.7.0 artifacts reverted** for the same reason — they reproduce from unmodified `main`, so they are unrelated scope: the lock's own `version` 3.74.0 → 3.75.0 (it trails because the release workflow deletes the file), and the `libc` field dropped from 21 optional linux binaries.

Local `npm audit` goes from 32 findings to 30. `npm ci` reinstalls cleanly from the committed lockfile and leaves it byte-identical.

## Re-verified end to end

`node scripts/build.js core` and a full `node scripts/build.js` (12 bundles, 235 s, **0** `Merge collision`, 0 plugin warnings); `dist/maidr.css` still **367,893 B** with 20 `@font-face` blocks, exactly one woff2 source each and zero `font/ttf` / `font/woff`; `npm run type-check` clean; `npx jest --coverage=false` **97 suites / 1222 tests passing**; `CI=true npx eslint` clean on all changed files.

jooyoungseo:
Acted on the coverage gap from the second review round; leaving the other two observations as they are, with reasoning.

## Added: `generateBundle` is now covered directly (3444fc88)

The gap was real — every existing case drove `stripNonWoff2FontSources`, so the plugin object that `scripts/build.js` actually loads was covered only by the manual verification in the PR body. Seven cases now drive `woff2OnlyFonts().generateBundle` against a fake Rollup bundle with a stubbed `{ warn, info }` context:

| case | asserts |
|---|---|
| string `source` | rewritten in place on `output.source` |
| `Uint8Array` `source` | decode branch produces the same result as the string path |
| non-`.css` asset | untouched |
| chunk entries | skipped, no throw |
| no face offers woff2 | `this.warn` called, source left intact, no `this.info` |
| nothing to rewrite | `output.source` never reassigned, no `this.info` |
| rewritten stylesheet | `this.info` called, reported saving equals the real byte delta |

Plus plugin metadata, since `enforce: 'post'` is load-bearing — the CSS asset doesn't exist in the bundle until `vite:css-post` has assembled it.

Two implementation notes:

- `output.source` is defined with an **intercepting setter**, so "did the hook reassign it?" is *observed* rather than inferred from the resulting text. That's what gives the two no-op cases any teeth.
- The chunk fixtures include one named **`inlined.css`**. A chunk carries no `source`, so reading one throws — which makes the `type !== 'asset'` guard testable independently of the extension guard.

**Amortized to one spawn**, per the note about subprocess overhead: all seven share a single child process collected in `beforeAll`. They run in ~5 ms total versus ~180 ms each for the spawn-per-assertion transform suite. Existing suites left untouched.

**Verified load-bearing by mutation**, rather than assuming green means covered:
- replacing `Buffer.from(output.source).toString('utf8')` with `String(output.source)` → fails **only** `decodes and rewrites a css asset whose source is bytes`
- dropping `output.type !== 'asset'` from the guard → fails **only** `leaves non-css assets and chunks alone`

## Not acting on: hand-maintained `.d.ts` drift

Agreed. A guard that only compared exported *names* would pass while a signature silently diverged, so it would be reassurance rather than protection. Comparing real type signatures means generating declarations from the JS and diffing them — more machinery than a two-function surface justifies. The `.d.ts` carries a "keep both files in sync" note, and `tsc --noEmit` still catches any *use* that stops type-checking.

## Not acting on: `WOFF2_URL_RE` being mime-shaped

Agreed, and I confirmed the trace empirically rather than taking it on faith. For `url(data:application/font-woff2;base64,…)` **with no `format()` hint** — the only shape where the URL check is load-bearing:

```
rewritten: 0 | skipped: 1
unchanged: true
ttf still present: true
```

So it classifies as `'other'`, the face looks like it has no woff2 alternative, and the skip-and-warn path leaves the rule completely intact — every format retained, text still renders, and a `this.warn` names it. Safe and loud, exactly as designed. With a `format("woff2")` hint present (what KaTeX actually emits) it trims normally: `rewritten: 1 | skipped: 0 | ttf dropped: true`. Tightening the regex would buy nothing today, and the cost of being wrong is a bigger bundle rather than broken rendering.

## Verification

`npm run type-check` clean; `CI=true npx eslint` clean on the changed file; `npx jest --coverage=false` → **97 suites / 1229 tests passing** (was 1222; +7). No rebuild — the transform itself is unchanged.

Issue #671, opened 2026-07-30
a11y: evaluate MathJax browser component with a11y extensions for chat maths


jooyoungseo:
`rehype-mathjax` was removed in #668 as an unused dependency. That was correct — it was imported nowhere — but the history behind it points at an accessibility opportunity worth evaluating properly rather than letting it disappear with the dependency.

## History

MathJax was the original renderer. PR 433 (`feat: fix subplot nav announcements & cleanup junk in chat component`) swapped it out:

```diff
- import rehypeMathjax from 'rehype-mathjax/svg';
+ import rehypeKatex from 'rehype-katex';
+ import 'katex/dist/katex.min.css';
```

The commit message gives no reason for the change, and it landed inside an otherwise unrelated cleanup PR. The `rehype-mathjax` entry in `package.json` was simply never removed.

## What the two actually emit

Rendering `\frac{a}{b}` through each:

| | KaTeX (current) | `rehype-mathjax/svg` (removed) |
|---|---|---|
| hidden MathML `<math>` | ✅ | ❌ |
| visual layer `aria-hidden` | ✅ | ❌ |
| `aria-label` | ❌ | ❌ |
| `aria-braillelabel` | ❌ | ❌ |
| output | HTML + MathML | `<svg role="img" focusable="false">` |

The removed renderer emitted an `<svg role="img">` with **no accessible name at all** — a screen reader gets nothing. KaTeX at least emits real MathML. So the swap improved accessibility, even though that does not appear to have been the intent, and keeping KaTeX today is the right call.

## The opportunity

MathJax's accessibility strength lives in the **browser runtime components**, not in build-time SVG pre-rendering. Per the [MathJax 4.0 accessibility docs](https://docs.mathjax.org/en/latest/basic/accessibility.html), `a11y/explorer` and `a11y/speech` generate speech strings and insert them as `aria-label` and `aria-braillelabel`, and provide step-by-step traversal of sub-expressions with Nemeth output to a connected braille display. `rehype-mathjax/svg` pre-renders at build time and discards all of that.

That capability set — speech, Nemeth braille, interactive keyboard exploration — is exactly what MAIDR already does for charts. Applying it to maths in chat responses is a natural fit.

[...]

PR #675, opened 2026-07-30
fix(chat): keep KaTeX's MathML through markdown sanitisation

jooyoungseo:
# Pull Request

## Description

Maths in AI chat responses reached the page with its structure destroyed. Evaluating #671 — should MAIDR move from KaTeX to MathJax's browser components for their a11y extensions — meant first measuring what KaTeX actually delivers today. It delivers nothing: MAIDR's own sanitisation allowlist was stripping it.

Chromium's accessibility tree for `$x = \frac{-b \pm \sqrt{b^2-4ac}}{2a}$`, rendered through the exact plugin chain in `TypingEffect`:

| | before | after |
|---|---|---|
| MathML structural roles | **0** | **21** |
| distinct roles | *(none)* | `MathMLFraction`, `MathMLSquareRoot`, `MathMLSup`, `MathMLRow`, `MathMLIdentifier`, `MathMLOperator`, `MathMLNumber` |

Before the fix the `<math>` element held only flat `StaticText` in *visual* order — `x`, `=`, `2`, `a`, `−`, `b`, `±` … — with no fraction, radical or exponent anywhere in the tree. The markup behind it:

```html
<math>x=−b±b2−4ac2ax = \frac{-b \pm \sqrt{b^2-4ac}}{2a}</math>
```

Three causes, all in the allowlist:

1. **`math` was allowed; nothing inside it was.** `rehype-sanitize` drops a disallowed element but keeps its text, so the expression collapsed into its own leaf characters. The fraction, radical and exponents were gone before the markup rendered.

2. **ARIA attributes were spelled as HTML attribute names.** `rehype-sanitize` matches on hast property names, so `aria-hidden` matched nothing. KaTeX marks its visual layer `aria-hidden="true"` precisely so assistive technology reads the MathML instead of the glyph spans; that attribute was stripped and both layers were exposed:

   ```
   before  "The root is x=−b±b2−4ac2ax = \frac{…}{2a}x=2a−b±b2−4ac."   ← visual layer too
   after   "The root is x=−b±b2−4ac2ax = \frac{…}{2a}."
   ```

3. **KaTeX's visual layer had the same disease** (found while acting on review feedback). `<line>` was not named at all, so `\cancel` and `\not` lost their strike entirely; `preserveAspectRatio` was stripped from `<svg>`, which stretchy delimiters and arrows rely on to crop rather than scale. Rendering defects rather than announcement ones — both sit inside the `aria-hidden` wrapper — but lost for exactly the reason the MathML was.

## Related Issues

Refs #671. The evaluation that issue asks for is [posted there](https://github.com/xability/maidr/issues/671#issuecomment-5135932595) — measurements for all four of its open questions, and the reasoning for staying on KaTeX. This PR is the defect that evaluation surfaced; it does not close the issue.

Two follow-ups were opened from review on this PR:

- **#677** — the same allowlist drops GFM tables, `h1`–`h6`, strikethrough and task-list checkboxes. Larger user impact than the maths bug: a table of chart values renders as `BarValueJan45.2Feb51.8`. Kept separate because `input`, `id` and `data-*` need their own security review.
- **#678** — an ESM-aware Jest project, so the pipeline can be tested end to end rather than only the schema's contents.

## Changes Made

- **`src/util/markdownSanitize.ts`** (new) — the chat allowlist, naming every element and attribute KaTeX emits across both of its layers, with ARIA entries under their hast spellings.

  It lives in `util/` rather than beside its caller because `TypingEffect.tsx` imports `react-markdown`, which is ESM-only and so is not loadable by the Jest suite as currently configured — CommonJS, with `allowJs` off, so nothing under `node_modules` is transformed. A module with no runtime imports can be tested. (That configuration is changeable; see #678.)

- **`src/ui/components/TypingEffect.tsx`** — uses the shared schema; the inline literal is gone. Built once at module scope rather than per render, which matters here because the typing animation re-renders every 10 ms.

- **`test/util/markdownSanitize.test.ts`** (new) — renders a 39-case TeX corpus through the pinned KaTeX in both display modes and fails when the output carries an element or attribute the allowlist does not name, in either layer. So a KaTeX upgrade that starts emitting something new fails here rather than silently losing the markup.

### Security

`href` and `src` are excluded from the MathML attribute list. MathML accepts `href` on any element, the value arrives from an LLM response, and the accessible tree needs neither. `src`/`alt` belong to `mglyph`, which only `\includegraphics` produces and which KaTeX disables under its untrusted defaults — the defaults MAIDR renders with.

`target` is dropped from links. Markdown has no syntax for it and raw HTML is escaped to text, so it admitted an attribute that could not arrive — while standing to allow a `target="_blank"` without `rel="noopener"` the day one could.

The shared attribute list is narrowed to `className` and `ariaHidden`. `role`, `ariaLabel`, `ariaBusy`, `ariaLive` and `ariaAtomic` were present but misspelled, and so inert; correcting the spelling would have made them live for the first time. Nothing in the pipeline emits them — the `role` and `aria-label` on `<pre>` and `<a>` are React props applied after sanitisation — so they are dropped rather than fixed.

`style` stays on `span` and `svg` because KaTeX needs it (`\vec`, `\hat` and `\dot` size their accent SVG with `width:`). It is not a passthrough: `\htmlStyle` and `\htmlClass` are disabled under KaTeX's untrusted defaults and raw HTML is escaped, so a response cannot reach a `url()`.

The schema deliberately declares no key beyond `tagNames` and `attributes`. `hast-util-sanitize` resolves its config as `{...defaultSchema, ...options}` (`lib/index.js:235`) — shallow, one level deep — so the default `protocols` map is what still strips `javascript:` from a link. A partial `protocols` key would replace that map rather than extend it, and a test pins that none is added.

Verified against the compiled module: `javascript:`, `vbscript:` and `data:` URLs in links and images, `\href{javascript:…}` inside maths, `\includegraphics`, raw HTML with an inline handler, and an `aria-live` injection attempt all remain stripped; ordinary `https:` links survive.

Not applicable — the accessibility-tree change is shown in the table above. The equation looked correct throughout, which is why this went unnoticed.

## Checklist

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass — 1291 tests across 101 suites, `markdownSanitize.ts` at 100% coverage. CI is green on `c35b75a`, all 13 checks.

## Additional Notes

**Not verified here:** how NVDA, JAWS and VoiceOver each render the restored MathML. Chromium's accessibility tree is what screen readers consume through the platform APIs, so the table above is meaningfully closer than a DOM assertion — but it is not a screen reader, and confirmation with real assistive technology is worth having before #671 is closed.

An earlier revision of this description claimed the raw TeX annotation was being read aloud. That was overstated: the TeX reaches the DOM text but not the accessibility tree, in either version. The defect is the lost structure, not a spoken annotation.

`<annotation encoding="application/x-tex">` is kept. MathML Core renders only the first child of `<semantics>`, Chromium's AX tree confirms it is excluded, and keeping it preserves copy-as-TeX.

### DIFF
--- a/src/ui/components/TypingEffect.tsx
+++ b/src/ui/components/TypingEffect.tsx
@@ -1,6 +1,7 @@
 import { Box } from '@mui/material';
 import { useViewModelState } from '@state/hook/useViewModel';
 import { containsLatex, ensureKatexStylesheet } from '@util/katex';
+import { createChatSanitizeSchema } from '@util/markdownSanitize';
 import React, { memo, useEffect, useMemo, useState } from 'react';
 import ReactMarkdown from 'react-markdown';
 import rehypeSanitize from 'rehype-sanitize';
@@ -14,6 +15,9 @@ type RehypePlugins = NonNullable<
 
 /** Stable empty list, so a message without maths never re-renders on identity. */
 const NO_MATH_PLUGINS: RehypePlugins = [];
+
+/** Built once: the allowlist is the same for every message on the page. */
+const SANITIZE_SCHEMA = createChatSanitizeSchema();
 
 interface TypingEffectProps {
   text: string;
@@ -142,39 +146,9 @@ export const TypingEffect: React.FC<TypingEffectProps> = memo(({ text, isUser, m
         <ReactMarkdown
           rehypePlugins={[
             // Before rehypeSanitize, so KaTeX's own markup goes through the
-            // allowlist below rather than around it.
+            // allowlist rather than around it.
             ...mathPlugins,
-            [rehypeSanitize, {
-              attributes: {
-                '*': ['className', 'aria-label', 'aria-hidden', 'role', 'aria-busy', 'aria-live', 'aria-atomic'],
-                'a': ['href', 'target'],
-                'img': ['src', 'alt'],
-                'math': ['display'],
-                'span': ['style'],
-                'svg': ['aria-hidden', 'role', 'xmlns', 'width', 'height', 'viewBox'],
-                'path': ['d'],
-              },
-              tagNames: [
-                'p',
-                'br',
-                'b',
-                'i',
-                'em',
-                'strong',
-                'a',
-                'pre',
-                'code',
-                'ul',
-                'ol',
-                'li',
-                'blockquote',
-                'img',
-                'math',
-                'span',
-                'svg',
-                'path',
-              ],
-            }],
+            [rehypeSanitize, SANITIZE_SCHEMA],
           ]}
           remarkPlugins={[remarkGfm, remarkMath]}
           components={{
--- a/src/util/markdownSanitize.ts
+++ b/src/util/markdownSanitize.ts
@@ -4,10 +4,12 @@ import type { Options as SanitizeSchema } from 'rehype-sanitize';
  * The sanitisation allowlist for Markdown rendered in AI chat responses.
  *
  * Lives here rather than beside its one caller in `TypingEffect.tsx` because
- * that module imports `react-markdown`, which is ESM-only and cannot be loaded
- * by the Jest suite. Keeping the schema in a module with no runtime imports is
- * what lets `test/util/markdownSanitize.test.ts` check it against the MathML
- * KaTeX actually emits.
+ * that module imports `react-markdown`, which is ESM-only and so is not
+ * loadable by the Jest suite as it is configured — CommonJS, with `allowJs`
+ * off, so nothing under `node_modules` is transformed. Keeping the schema in a
+ * module with no runtime imports is what lets
+ * `test/util/markdownSanitize.test.ts` check it against the MathML KaTeX
+ * actually emits.
  *
  * `rehype-sanitize` *replaces* its default schema with the one it is given
  * rather than merging into it, so everything the chat is allowed to render has
@@ -167,7 +169,7 @@ export function createChatSanitizeSchema(): SanitizeSchema {
       // load-bearing for stretchy delimiters and arrows, which KaTeX draws at
       // `width="400em"` over a `viewBox` 400000 units wide and expects
       // `xMaxYMin slice` to crop rather than scale.
-      'svg': ['ariaHidden', 'xmlns', 'width', 'height', 'viewBox', 'preserveAspectRatio', 'style'],
+      'svg': ['xmlns', 'width', 'height', 'viewBox', 'preserveAspectRatio', 'style'],
       'path': ['d'],
       // `\cancel` and `\not` draw their strike as an SVG line.
       'line': ['x1', 'x2', 'y1', 'y2', 'strokeWidth'],
--- a/src/util/markdownSanitize.ts
+++ b/src/util/markdownSanitize.ts
@@ -160,12 +160,25 @@ export function createChatSanitizeSchema(): SanitizeSchema {
   return {
     attributes: {
       '*': [...GLOBAL_ATTRIBUTES],
-      'a': ['href', 'target'],
+      // No `target`. Markdown has no syntax for it, raw HTML is escaped to
+      // text, and nothing in the plugin chain adds one, so it was allowing an
+      // attribute that could not arrive — and a `target="_blank"` without
+      // `rel="noopener"` is reverse tabnabbing waiting for the day one could.
+      // Same reasoning as the ARIA entries dropped from GLOBAL_ATTRIBUTES.
+      'a': ['href'],
       'img': ['src', 'alt'],
+      // `style` here and on `svg` below carries KaTeX's computed layout —
+      // `height:1em;vertical-align:-0.25em`, `top:-4em`, `width:0.471em`. The
+      // values are generated, never passed through: `\htmlStyle` and
+      // `\htmlClass` are disabled by the untrusted defaults KaTeX renders with,
+      // and a raw `<span style>` in a response is escaped to text. So a
+      // response cannot reach a `url()` and turn a rendered equation into a
+      // tracking beacon.
       'span': ['style'],
       // KaTeX's visual layer. It sits inside an `aria-hidden` wrapper, so what
       // is lost here is rendering rather than announcement — but it is lost the
-      // same way the MathML was, by not being named. `preserveAspectRatio` is
+      // same way the MathML was, by not being named. `style` is what `\vec`,
+      // `\hat` and `\dot` size their accent with. `preserveAspectRatio` is
       // load-bearing for stretchy delimiters and arrows, which KaTeX draws at
       // `width="400em"` over a `viewBox` 400000 units wide and expects
       // `xMaxYMin slice` to crop rather than scale.
--- a/src/util/markdownSanitize.ts
+++ b/src/util/markdownSanitize.ts
@@ -114,20 +114,22 @@ export const MATHML_ATTRIBUTES: readonly string[] = [
 /**
  * Attributes allowed on every element, spelled as hast property names.
  *
- * `ariaHidden` is the one that carries weight today: KaTeX marks its visual
- * layer with it so assistive technology reads the MathML instead of the glyph
- * spans beside it. The rest are allowed for markup that may carry them; no
- * plugin in the chain emits them at present, and raw HTML in a response never
- * reaches this tree because the pipeline runs without `rehype-raw`.
+ * Both entries are load-bearing. KaTeX puts `class` on every span it builds,
+ * and marks its visual layer `aria-hidden` so assistive technology reads the
+ * MathML instead of the glyph spans beside it.
+ *
+ * The list used to also carry `role`, `ariaLabel`, `ariaBusy`, `ariaLive` and
+ * `ariaAtomic`. Nothing in the pipeline emits any of them — the `role` and
+ * `aria-label` that `TypingEffect` puts on `<pre>` and `<a>` are React props,
+ * applied to the tree after it has been sanitised — and while they were
+ * misspelled they were inert. Correcting the spelling would have made them
+ * live for the first time, so they are dropped instead: a response body cannot
+ * reach an attribute today (the pipeline runs without `rehype-raw`, so raw HTML
+ * is escaped to text), and this way it still could not were that to change.
  */
 const GLOBAL_ATTRIBUTES: readonly string[] = [
   'className',
-  'ariaLabel',
   'ariaHidden',
-  'role',
-  'ariaBusy',
-  'ariaLive',
-  'ariaAtomic',
 ];
 
 /** Every MathML element, mapped to the presentation attributes it may keep. */
@@ -141,6 +143,15 @@ function mathmlAttributes(): Record<string, string[]> {
  * The schema `TypingEffect` hands to `rehype-sanitize`.
  *
  * Built fresh on each call so a caller cannot mutate the shared allowlists.
+ *
+ * Deliberately declares only `tagNames` and `attributes`. `hast-util-sanitize`
+ * resolves its configuration as `{...defaultSchema, ...options}` — a shallow
+ * merge, one level deep — so every key left out here keeps its default. That is
+ * what still strips `javascript:` from a link or an image without `protocols`
+ * appearing anywhere below: the default map covers `href`, `src`, `cite` and
+ * `longDesc`. Adding a partial `protocols` key would replace that map outright
+ * rather than extend it, which is why the test asserts this schema declares no
+ * key beyond the two it means to override.
  * @returns The allowlist of tags and attributes an AI chat response may render.
  */
 export function createChatSanitizeSchema(): SanitizeSchema {
@@ -150,8 +161,16 @@ export function createChatSanitizeSchema(): SanitizeSchema {
       'a': ['href', 'target'],
       'img': ['src', 'alt'],
       'span': ['style'],
-      'svg': ['ariaHidden', 'role', 'xmlns', 'width', 'height', 'viewBox'],
+      // KaTeX's visual layer. It sits inside an `aria-hidden` wrapper, so what
+      // is lost here is rendering rather than announcement — but it is lost the
+      // same way the MathML was, by not being named. `preserveAspectRatio` is
+      // load-bearing for stretchy delimiters and arrows, which KaTeX draws at
+      // `width="400em"` over a `viewBox` 400000 units wide and expects
+      // `xMaxYMin slice` to crop rather than scale.
+      'svg': ['ariaHidden', 'xmlns', 'width', 'height', 'viewBox', 'preserveAspectRatio', 'style'],
       'path': ['d'],
+      // `\cancel` and `\not` draw their strike as an SVG line.
+      'line': ['x1', 'x2', 'y1', 'y2', 'strokeWidth'],
       ...mathmlAttributes(),
     },
     tagNames: [
@@ -172,6 +191,7 @@ export function createChatSanitizeSchema(): SanitizeSchema {
       'span',
       'svg',
       'path',
+      'line',
       ...MATHML_TAG_NAMES,
     ],
   };
--- a/src/util/markdownSanitize.ts
+++ b/src/util/markdownSanitize.ts
@@ -1,0 +1,178 @@
+import type { Options as SanitizeSchema } from 'rehype-sanitize';
+
+/**
+ * The sanitisation allowlist for Markdown rendered in AI chat responses.
+ *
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Thanks — both substantive points were right, and chasing the second one turned up two more gaps. Pushed in c30ccbe.

## Correction to my own description first

I wrote that the raw TeX annotation "was read out" and that a screen reader "read that literally". I checked it properly in Chromium and **that part was overstated** — the TeX reaches the DOM text but not the accessibility tree, in either version. What I should have claimed is what the AX tree actually shows:

| `$x = \frac{-b \pm \sqrt{b^2-4ac}}{2a}$` | before | after |
|---|---|---|
| MathML structural roles | **0** | **21** |
| distinct roles | *(none)* | `MathMLFraction`, `MathMLSquareRoot`, `MathMLSup`, `MathMLRow`, `MathMLIdentifier`, `MathMLOperator`, `MathMLNumber` |
| raw TeX in AX tree | no | no |

Before the fix the `<math>` element held only flat `StaticText` in *visual* order — `x`, `=`, `2`, `a`, `−`, `b`, `±` … — with no fraction, radical or exponent anywhere in the tree. The defect is that the structure was gone, not that the TeX was spoken. Text not covered by `aria-hidden` also confirms the duplicate-announcement half:

```
BEFORE  "The root is x=−b±b2−4ac2ax = \frac{…}{2a}x=2a−b±b2−4ac."   ← visual layer exposed too
AFTER   "The root is x=−b±b2−4ac2ax = \frac{…}{2a}."
```

Chromium's AX tree is still not a screen reader, so the caveat in the PR description stands.

## `annotation-xml` blind spot — fixed

Correct, and it mattered more than "low risk": a pattern that cannot match an element doesn't fail on it, it *drops it from the survey*, so the guarantee was silently narrower than advertised. Widened to `/^(?:semantics|annotation[a-z-]*|m[a-z-]+)$/`.

## `protocols` — verified, and pinned

I checked the mechanism rather than taking it on trust, since the guidance depended on it. `hast-util-sanitize/lib/index.js:235`:

```js
schema: options ? {...defaultSchema, ...options} : defaultSchema,
```

Shallow, one level deep — so you're right on both counts: omitting `protocols` keeps the default map, and declaring a partial one would replace it outright.

Your suggested test isn't available, though: `rehype-sanitize` is ESM-only and can't be loaded by the CommonJS Jest suite — that constraint is the reason the schema is in its own module in the first place. So I pinned the invariant that actually guards the gap you described:

```ts
expect(Object.keys(schema).sort()).toEqual(['attributes', 'tagNames']);
```

That fails the moment someone adds a `protocols` key, which is the failure mode you named. The `javascript:`/`vbscript:`/`data:` cases are verified against the compiled module out of band, and they stay clean.

## What the second point turned up

Codifying it meant surveying KaTeX's *whole* output rather than just the MathML branch — and the visual layer had the same disease:

- **`<line>` was not in `tagNames` at all**, so `\cancel{x}` and `\not=` lost their strike entirely.
- **`preserveAspectRatio` was stripped from `<svg>`**. KaTeX draws stretchy delimiters and arrows at `width="400em"` over a `viewBox` 400000 units wide and relies on `xMaxYMin slice` to crop rather than scale, so `\xrightarrow` and `\left(…\right)` rendered wrong.

Both are inside the `aria-hidden` wrapper, so these are rendering rather than announcement defects — but they were lost for precisely the reason the MathML was. The suite now surveys both layers, so neither can rot again.

## Global ARIA attributes — acted on

You flagged this as pre-existing, but it isn't quite: while those entries were misspelled they were **inert**, and correcting the spelling would have made `aria-live` and `role` live for the first time. That's my change enabling them, not a carried-over state. Since nothing in the pipeline emits any of them — the `role`/`aria-label` on `<pre>` and `<a>` are React props applied after sanitisation — I dropped them. `GLOBAL_ATTRIBUTES` is now `className` and `ariaHidden`.

`style` on `span` I've left alone: genuinely pre-existing, KaTeX needs it for positioning, and removing it would break rendering outright. Worth its own issue as you suggest.

## Not changed

- **Per-element MathML attributes.** Agreed it's breadth over precision; everything in the list is a non-URL presentation attribute, and the URL-bearing ones are excluded by name.
- **Fresh schema per call.** Fair that there's one caller. Leaving it — it's three lines, and the alternative is a shared mutable object reachable from an export.

`lint:fix`, `type-check`, `build` and `test` all pass — 1289 tests across 101 suites, `markdownSanitize.ts` at 100%.

---

jooyoungseo:
Point 2 was right and is fixed in 6c6dafd. On point 1 — you're right that it's the gap worth closing, and I was wrong to call it impossible. Detail below, along with why I'm still not doing it in this PR.

## Correcting myself: it *is* testable

I've been saying `rehype-sanitize` "cannot be loaded by the Jest suite". That's true of the current configuration, but I stated it as a hard limit and it isn't. A second Jest project loads the whole ESM stack:

```js
// jest.esm.config.mjs
export default {
  testEnvironment: 'node',
  extensionsToTreatAsEsm: ['.ts'],
  testMatch: ['**/test/**/*.esm-test.ts'],
  transform: { '^.+\\.ts$': ['ts-jest', { useESM: true, tsconfig: { module: 'ESNext', moduleResolution: 'bundler' } }] },
}
```

```
$ NODE_OPTIONS=--experimental-vm-modules npx jest -c jest.esm.config.mjs
PASS esm test/util/__probe.esm-test.ts
  ✓ loads the ESM markdown stack (22 ms)
```

I also checked the flag is safe for what's already there — the existing 101 suites / 1289 tests all pass under `--experimental-vm-modules`. So there's no blocker, just work.

## Why not in this PR

Wiring it into `npm test` needs a second config, a changed `test` script, and — since the repo has no `cross-env` and `NODE_OPTIONS=… jest` is not valid on Windows — either a new devDependency or a `scripts/` wrapper. That's test infrastructure landing inside an accessibility fix, against the repo's own one-logical-change-per-commit rule, and it widens the review surface of a change that's otherwise contained.

I'd rather hand you the finished feasibility work than smuggle it in. Happy to open it as a follow-up, or to add it here if you'd prefer it in one go — your call.

What I did do is stop the comments claiming it's impossible; they now say the current config can't load it and what it would take.

## On your e2e suggestion specifically

Worth flagging that the Playwright route doesn't work for this. `e2e_tests/` drives charts, and the sanitiser only runs on **assistant** messages in `TypingEffect`, which needs a real LLM response — so a spec would need an API key in CI. The ESM-project route reaches the same code without that.

## Meanwhile, verified out of band

Against the compiled module, so it's the committed schema and not a copy:

```
javascript: link     clean      \href in maths       clean
javascript: image    clean      \includegraphics     clean
data: html link      clean      raw html + handler   clean
vbscript: link       clean      aria-live injection  clean
normal https link survives: true
```

Not a substitute for the automated test — just a statement of what's currently known to hold.

## Points 3 and 4

Both correct observations, both pre-existing, both left alone deliberately:

- **`target` on `a`** — nothing in the pipeline produces it, so it is dead allowance. I removed the dead ARIA entries because *this PR* would have made them live for the first time; `target` has no such transition, so removing it is unrelated cleanup. Your `rel="noopener"` note would be the right call if it ever became live, and belongs with whatever change does that.
- **`title` on images and links** — also absent from the pre-existing schema, so nothing changes here. Agreed it isn't accessibility-critical: `alt` and the link text carry the announcement.

`lint:fix`, `type-check`, `build` and `test` all pass — 1289 tests across 101 suites, `markdownSanitize.ts` at 100%.

---

jooyoungseo:
Both points tested rather than reasoned about, and they landed differently. efb7edd.

## `target` — removed, you were right

I pushed back on this last round on the grounds that the ARIA entries were inert *because of the bug I was fixing* whereas `target` was merely unproduced. That distinction doesn't hold up as a reason to keep it, and your consistency argument does. Removed.

I first measured it wrong, which is worth recording. A naive `/target=/` over the rendered HTML said `[x](https://example.com){target=_blank}` produced one — but that was the regex matching the literal text markdown leaves after the link, not an attribute:

```
<p><a href="https://example.com">x</a>{target=_blank}</p>
```

Parsing the DOM and counting real `<a target>` elements gives **0** for every case — markdown link syntax, the `{target=…}` attribute syntax, raw HTML, GFM autolinks. Provably dead, so removing it changes no behaviour. `schema.attributes.a` is now `['href']`, pinned by a test, and `href` keeps the default `protocols` filtering.

## `style` on `<svg>` — kept, and it isn't an accepted risk

Two things needed checking here.

**It is load-bearing.** Searching the 39-case corpus for `svg[style]`:

```
display=false  \vec{v}\hat{n}\dot{x}  style="width:0.471em"
display=true   \vec{v}\hat{n}\dot{x}  style="width:0.471em"
```

`\vec`, `\hat` and `\dot` size their accent SVG with it. Dropping it breaks accent rendering, so this isn't a gratuitous widening.

**The vector you describe isn't reachable.** Your concern needs an attacker-controlled `url()` in a style value. Nothing can put one there:

| input | result |
|---|---|
| `$\htmlStyle{background:url(https://evil.test/x.gif)}{x}$` | blocked — *"HTML extension is disabled on strict mode"* |
| `$\htmlClass{evil}{x}$` | blocked, same |
| `<span style="background:url(…)">hi</span>` | escaped to text |

Every `style` that survives is KaTeX-computed layout — `height:1em;vertical-align:-0.25em`, `top:-4em`, `border-right-width:1em`, `width:0.471em`. `\htmlStyle` and `\htmlClass` need `trust: true`, and `rehype-katex` runs with KaTeX's defaults.

So rather than note it as accepted risk, I've recorded *why* it's safe in the module, next to the allowance — the guarantee depends on `trust` staying off, and that's the thing a future reader needs to know.

## On your last section

Fair to flag that you couldn't run the suite. For the record, on efb7edd: `lint:fix`, `type-check`, `build` and `test` all pass — **1290 tests across 101 suites**, `markdownSanitize.ts` at 100%. CI is the authority, not my local run.

And agreed that real screen-reader verification is still outstanding. That is the one open item on #671, and it's the reason this PR refs the issue rather than closing it.

---

jooyoungseo:
Note 1 confirmed and worse than described — filed as #677. Note 3 addressed in ab830cb. Note 2: CI is green on `efb7edd`, all 13 checks.

## Note 1 — real, and it goes past tables

You were right to flag it, and measuring it turned up more than the tag families you named. Rendering a corpus through the pipeline with and without sanitisation:

| Markdown | Renders as | Lost |
|---|---|---|
| a 2×2 GFM table | `BarValueJan45.2Feb51.8` | `table`/`thead`/`tbody`/`tr`/`th`/`td` |
| `## Summary` | `Summary` | `h1`–`h6` |
| `~~gone~~` | `gone` | `del` |
| `- [x] done` | `done` | `input` |
| `text[^1]` | partial | `sup`, `section` |
| `---` | *(nothing)* | `hr` |

Headings weren't on your list and are a significant loss on their own — heading navigation is a primary way AT users move through long text. And `del` plus the task-list checkbox both **invert meaning** rather than just flattening layout: a retraction reads as an assertion, and checked is indistinguishable from unchecked.

The table case is the one that matters most for this project specifically. A table of values is the likeliest thing an LLM produces when describing a chart, and `BarValueJan45.2Feb51.8` gives a screen reader user a run of digits with nothing binding them to a label — in a library that exists to make data readable non-visually.

## Why I stopped rather than fixing it here

I went back and forth on this, since the fix is small and the failure mode is already proven. Landing it here would be wrong anyway:

- It's a **separate logical change** — GFM/CommonMark output, not KaTeX — and the repo's convention is to split unrelated fixes rather than bundle them.
- The security decisions want their own review, not the tail end of a maths PR. `input` is a form control that should only ever be admitted as a *disabled checkbox*; `id` on footnote anchors is a DOM-clobbering vector (handled by `hast-util-sanitize`'s `user-content-` prefix, which the "no other schema key" test in this PR is what keeps in force — but that deserves confirming, not assuming); `data-footnote-ref`/`data-footnote-backref` should be named explicitly rather than via a `data-*` wildcard.
- This PR is four review rounds in. Widening it again trades a reviewable change for a sprawling one.

#677 has the measurements, the missing tag and attribute lists, and the suggested approach — the same corpus-diff method used here, plus the warning about hast property names that caused half of this bug.

## Notes 2 and 3

**CI** — green on `efb7edd`: Unit Tests, Type Check, Lint Code, Lint Commit Messages, Build Project, Docs Build, E2E Smoke, CodeQL, both CodeQL analyses, Copilot Instructions In Sync, preview build. Fair to have flagged that you couldn't verify my numbers; the checks are the authority.

**JSDoc vs call site** — good catch, that would read as a contradiction. Noted in the doc comment: the freshness is the function's own contract, while `TypingEffect` hoists one instance because the allowlist never varies and the animation re-renders every 10 ms.

Thanks for four rounds of genuinely useful review — between them they turned up the `svg`/`line` layer, the newly-live ARIA entries, the dead `target`, and now #677, none of which were in the original diagnosis.

---

jooyoungseo:
Note 1 fixed in c35b75a. Note 2 was already filed as #677 just before this review landed. Thanks for independently re-deriving the `rehype-raw` and `trust: false` checks rather than taking the description's word — those two facts carry most of the security argument here, and they're worth more having been checked twice.

## Note 1 — the test was weaker than the contract

Right, and the gap was in the place that mattered. The old assertions were:

```ts
expect(first).not.toBe(second);
expect(first.tagNames).not.toBe(second.tagNames);
```

Every allowlist actually lives one level below `attributes`, so a shared `attributes` map — or a shared array inside it — would have passed both while leaving the lists mutable across callers. Now checked directly, plus a case stating the property as the failure it prevents:

```ts
first.tagNames?.push('script');
first.attributes?.['*']?.push('onclick');
first.attributes?.math?.push('href');

const second = createChatSanitizeSchema();
expect(second.tagNames).not.toContain('script');
expect(second.attributes?.['*']).not.toContain('onclick');
expect(second.attributes?.math).not.toContain('href');
expect(MATHML_TAG_NAMES).not.toContain('script');
expect(MATHML_ATTRIBUTES).not.toContain('href');
```

As you found, the implementation was already correct — only the assertions were behind it.

## Note 2 — filed

You and the previous round both landed on this; it's #677. Measured through the pipeline, it goes past tables and strikethrough: `h1`–`h6` are dropped too (so heading navigation is gone), and `del` plus the task-list checkbox **invert meaning** rather than flattening layout. A 2×2 GFM table renders as `BarValueJan45.2Feb51.8`.

Kept out of this PR deliberately — separate logical change, and `input`/`id`/`data-*` want their own security review rather than arriving at the tail of a maths fix. Reasoning and suggested approach are in the issue.

## Note 3 / process note

Agreed on the ARIA removal being the right call — the deciding detail was that those entries were *dead*, not merely unused, so fixing the spelling would have made `aria-live` injectable for the first time rather than preserving a status quo.

On not being able to run the suite: CI is the authority and it's green. On `c35b75a`: **1291 tests across 101 suites**, `markdownSanitize.ts` at 100%, plus `lint:fix`, `type-check` and `build`.

---

jooyoungseo:
Filed as #678, with the working config so it isn't re-derived a fourth time.

This is the third round to raise end-to-end coverage, and I agree it's the last real gap — I proved the ESM Jest project works two rounds ago and offered to add it here. Putting it in an issue instead, for one concrete reason: wiring it into `npm test` needs either a new `cross-env` devDependency or a `scripts/` wrapper, because `NODE_OPTIONS=… jest` isn't valid on Windows. That's a maintainer's call about the repo's tooling, not something to decide at the tail of a bug fix. #678 has both options, the verification that the flag is safe for the existing 101 suites, and the specific assertions worth writing.

One correction on the alternative you suggested: **the Playwright route doesn't reach this code.** `e2e_tests/` drives charts, and the sanitiser only runs on *assistant* messages in `TypingEffect` — a spec would need a live LLM response and an API key in CI. The manual Chromium measurement in the PR body worked because it rendered the pipeline's output directly rather than going through MAIDR's chat. The ESM-project route reaches the same code with neither obstacle.

`mathmlAttributes()` breadth: deliberate, and now recorded in the module — every entry is a presentation attribute with no injection surface, and the URL-bearing ones (`href`, `src`) are excluded by name and pinned by a test. Round one raised the same point and reached the same conclusion.

Nothing else actionable here. CI is green on `c35b75a` — 1291 tests across 101 suites, `markdownSanitize.ts` at 100%.

For anyone picking this up: the two follow-ups are #677 (tables, headings, strikethrough and task lists are dropped by the same allowlist — larger user impact than the maths bug) and #678 (this).

---

PR #674, opened 2026-07-30
test(e2e): restore the e2e suite to green and run all of chromium on PRs

jooyoungseo:
# Pull Request

## Description

Seven specs were failing in `beforeAll`. Because Playwright skips an entire `describe` block behind a failed hook, that also took out 185 more tests — **about 62% of the e2e suite was not executing**, and the failure count badly understated the loss. barplot, both boxplots, heatmap, multiLayer, multiLineplot and violin had no effective e2e coverage at all.

Baseline on `main`, chromium:

```
106 passed, 7 failed, 185 did not run   (298 total)
```

After this PR, chromium **and** firefox:

```
298 passed (0 failed, 0 skipped)
```

None of it was a product regression. Every failure was stale test code.

## Related Issues

Fixes #626

## Changes Made

Three commits, one root cause each.

### 1. `fix(e2e)` — resolve the maidr schema the way the library does

`extractMaidrData()` read only the `maidr-data` attribute. But `main()` in `src/index.tsx` accepts **three** inputs, in this order:

1. a JSON-shaped `maidr` attribute (`Constant.MAIDR_JSON_SELECTOR` = `[maidr^="{"]`)
2. a `maidr-data` attribute
3. the `window.maidr` global

The examples are split across all three, so a spec that read one of them passed or failed on which mechanism its example happened to use:

| example | mechanism | before |
|---|---|---|
| `dodged-barplot`, `stacked-barplot`, `lineplot`, `histogram`, `multilayer_plot` | `maidr-data="{…}"` | ✅ |
| `barplot`, `boxplot-horizontal`, `boxplot-vertical`, `heatmap` | `var maidr = {…}` global | ❌ |
| `violin`, `multi-lineplot` | `maidr="{…}"` | ❌ |

Since all three are supported product inputs, the test helper was the stale side. The eleven near-identical copies are replaced with one in `e2e_tests/utils/maidr-data.ts` that mirrors the entry point's resolution order (−257/+88 lines).

Separately, `multiLayer-page.ts` navigated to `examples/multiLayer_plot.html` while the file on disk is `multilayer_plot.html`. That resolves on a case-insensitive filesystem (macOS) and 404s on the Linux CI runner — which is why it reproduced on CI and not necessarily on a maintainer's laptop.

### 2. `test(e2e)` — update expectations that drifted while the specs were skipped

Unblocking the hooks surfaced 50 failures inside the previously-skipped blocks. Each was verified against the running product before being changed:

[...]

PR #676, opened 2026-07-30
test(e2e): keep the original error when page objects wrap a failure

jooyoungseo:
# Pull Request

## Description

Every page-object method wraps failures like this:

```ts
} catch (error) {
  throw new MultiLayerPlotError('Failed to navigate to Multi Layer Plot');
}
```

The original error — and its stack — is thrown away. So when `multiLayer-page.ts` pointed at `examples/multiLayer_plot.html` while the file on disk is `multilayer_plot.html`, the CI report said only:

```
MultiLayerPlotError: Failed to navigate to Multi Layer Plot
```

No 404. No path. Nothing to grep for. That is a large part of why the bug in #626 sat undiagnosed for about two months.

After this change the same failure reads:

```
MultiLayerPlotError: Failed to navigate to Multi Layer Plot
    at MultiLayerPlotPage.navigateToMultiLayerPlot (…/multiLayer-page.ts:44:13)
  [cause]: Error: Navigation failed to path "examples/multiLayer_plot.html":
           page.goto: net::ERR_FILE_NOT_FOUND at file:///…/examples/multiLayer_plot.html
      at MultiLayerPlotPage.navigateTo (…/base-page.ts:58:13)
    [cause]: Error: page.goto: net::ERR_FILE_NOT_FOUND at file:///…/examples/multiLayer_plot.html
```

The 404 and the exact bad path are right there. That output is from an actual run on this branch, not a mock-up.

## Related Issues

Follow-up to #674 (review feedback), addressing the diagnostic weakness behind #626. Independent of #674 — neither blocks the other.

## Changes Made

**`e2e_tests/utils/errors.ts`** — the eleven plot error classes, plus `TestError` and `AssertionError`, take a standard `ErrorOptions` parameter and forward it to `super`:

```ts
constructor(message: string, options?: ErrorOptions) {
  super(message, options);
  this.name = 'BarPlotError';
}
```

`ErrorOptions` rather than a bespoke `cause` parameter, so it is the platform contract and `tsconfig`'s `ESNext` target already supports it. `KeypressError` already folded a cause into its message but did not attach it; it now also passes `{ cause }` to `super`, so the stack survives too.

**The twelve page objects** — `{ cause: error }` added at 198 wrapping sites (192 single-line, 6 multi-line ones that interpolated `errorMessage` into the message but still dropped the stack). The two `KeypressError` sites already passed their cause positionally and now benefit from the `super` change.

**`e2e_tests/specs/errorCause.spec.ts`** — new, pins the contract so this cannot silently regress:
- a failed navigation keeps the underlying browser error (asserts `ERR_FILE_NOT_FOUND` reaches the caller through the wrapper)
- all eleven plot error classes propagate a cause through their constructor

## Checklist

- [ ] I have updated the documentation, if applicable. — no user-facing behaviour changed.

## Additional Notes

**Please read this section before merging — there is a real coverage gap in how this PR gets validated.**

Verification on this branch:

| check | result |
|---|---|
| `npm run lint` | clean |
| e2e — chromium | 108 passed, 7 failed, 185 did not run |

Those 7 failures and 185 skips are **`main`'s existing baseline**, not something this PR introduces — they are exactly the failures #674 fixes. The relevant delta is `106 → 108` passed: `main`'s baseline plus the two new tests in `errorCause.spec.ts`. Nothing regressed.

**The gap:** `main`'s `ci.yml` still runs only `liveData.spec.ts` and `dialogAccessibility.spec.ts` on PRs, so **`errorCause.spec.ts` will not execute in this PR's CI**. It passes locally, and it starts running in CI automatically once #674 lands (that PR switches the job to the full chromium project). If you merge this one first, its new test is unverified by CI until #674 follows. Merging #674 first avoids that entirely.

No `src/` product code is touched — this is test infrastructure and CI-visible diagnostics only.

---

### DIFF
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -55,7 +55,7 @@ export class BasePage {
       });
     } catch (error) {
       const errorMessage = error instanceof Error ? error.message : String(error);
-      throw new Error(`Navigation failed to path "${path}": ${errorMessage}`);
+      throw new Error(`Navigation failed to path "${path}": ${errorMessage}`, { cause: error });
     }
   }
 
@@ -304,7 +304,7 @@ export class BasePage {
       await this.clickElement(this.selectors.helpModalClose);
       await this.closeModal(this.selectors.helpModal);
     } catch (error) {
-      throw new AssertionError('Failed to show help menu');
+      throw new AssertionError('Failed to show help menu', { cause: error });
     }
   }
 
@@ -331,6 +331,7 @@ export class BasePage {
       const errorMessage = error instanceof Error ? error.message : String(error);
       throw new AssertionError(
         `Failed to show settings menu: ${errorMessage}`,
+        { cause: error },
       );
     }
   }
@@ -385,6 +386,7 @@ export class BasePage {
       const errorMessage = error instanceof Error ? error.message : String(error);
       throw new AssertionError(
         `Failed to close settings menu with Escape: ${errorMessage}`,
+        { cause: error },
       );
     }
   }
@@ -423,6 +425,7 @@ export class BasePage {
       const errorMessage = error instanceof Error ? error.message : String(error);
       throw new AssertionError(
         `Settings menu did not survive a backdrop click: ${errorMessage}`,
+        { cause: error },
       );
     }
   }
@@ -451,6 +454,7 @@ export class BasePage {
       const errorMessage = error instanceof Error ? error.message : String(error);
       throw new AssertionError(
         `Failed to show chat dialog: ${errorMessage}`,
+        { cause: error },
       );
     }
   }
@@ -644,7 +648,7 @@ export class BasePage {
         timeout: 10000,
       });
     } catch (error) {
-      throw new Error('Plot failed to load correctly');
+      throw new Error('Plot failed to load correctly', { cause: error });
     }
   }
 
@@ -708,7 +712,7 @@ export class BasePage {
       await this.page.keyboard.press(TestConstants.TAB_KEY);
       await this.verifySvgFocused();
     } catch (error) {
-      throw new Error('Failed to activate MAIDR');
+      throw new Error('Failed to activate MAIDR', { cause: error });
     }
   }
 
@@ -725,7 +729,7 @@ export class BasePage {
       await this.page.click(svgSelector);
       await this.verifySvgFocused();
     } catch (error) {
-      throw new Error('Failed to activate MAIDR by clicking');
+      throw new Error('Failed to activate MAIDR by clicking', { cause: error });
     }
   }
 
@@ -740,7 +744,7 @@ export class BasePage {
       const text = await this.getElementText(notificationSelector);
       return text.replace(/\s+/g, ' ').trim();
     } catch (error) {
-      throw new Error('Failed to get instruction text');
+      throw new Error('Failed to get instruction text', { cause: error });
     }
   }
 
@@ -761,7 +765,7 @@ export class BasePage {
       const notificationText = await this.getElementText(notificationSelector);
       return notificationText === modeMessages[mode];
     } catch (error) {
-      throw new Error(`Failed to check ${mode} status`);
+      throw new Error(`Failed to check ${mode} status`, { cause: error });
     }
   }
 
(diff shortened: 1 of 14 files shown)


jooyoungseo:
Took all three actionable points in `4491bb3`.

**1. `ElementNotFoundError` missing the `options` parameter — fixed.** You're right that it's currently unused, but leaving one of the four `TestError` subclasses without the parameter is exactly the kind of half-migrated state that bites whoever wires it up later. It now matches its siblings.

**2. The navigation test not covering the full chain — fixed, and this was the best catch of the review.** You were correct: it called `BasePage.navigateTo` directly, so it pinned the inner link and quietly skipped the `MultiLayerPlotError` wrapper that the PR description's example output is actually about.

It now drives the real `navigateToMultiLayerPlot()` and asserts all three levels:

```ts
expect(error).toBeInstanceOf(MultiLayerPlotError);
expect(error.message).toBe('Failed to navigate to Multi Layer Plot');

const navigationError = error.cause as Error;      // which path was requested
expect(navigationError.message).toContain('Navigation failed to path');

const browserError = navigationError.cause as Error; // why the browser refused
expect(browserError.message).toContain('page.goto');
```

On your suggestion of "a temporarily-bad path" — I went a different way deliberately. This branch is cut from `main`, where the capital-L typo still exists, so `navigateToMultiLayerPlot()` already 404s and the test would have passed for free. It would then have silently stopped testing anything the moment #674 corrects the filename. Instead it blocks the navigation with `page.route(/\.html$/, route => route.abort())` — routing does apply to `file://` URLs, which I checked rather than assumed. I verified the robustness claim by patching the filename locally to what #674 makes it and re-running: still 3 passed.

**3. `KeypressError` / `AssertionError` / `TestError` not in the loop — added**, along with `ElementNotFoundError`, as a separate case since their constructor signatures differ. The `KeypressError` assertion checks both halves of its contract: the cause is attached *and* its message still folds in the original text.

**On the nit** (duplicated inner-error text in both the message and the structured cause) — agreed it's intentional, and I'd keep it. The flattened text is what shows on the reporter's one-line summary; dropping it would mean the top line goes back to being uninformative, which is the thing this PR exists to fix. The structured cause serves tooling, the interpolated text serves the human scanning CI output.

**On the CI gap** — thanks for independently confirming it against `ci.yml:113` rather than taking my word for it. Worth restating for whoever merges: **merging #674 first** means these tests are covered by automation immediately; merging this one first leaves them locally-verified only until #674 follows.

Re-verified after the change: lint clean, type-check clean, chromium `109 passed / 7 failed / 185 did not run` — the 7 and 185 being `main`'s untouched baseline that #674 fixes, and 109 = baseline 106 + the 3 tests here.

---

jooyoungseo:
Observation 1 was a real bug, not just a curiosity — fixed in `5e2abd0`. Thanks for flagging it; I'd have shipped it.

You hedged it as "unlikely to break anything," but it reproduces, and the effect runs directly against the point of this branch. Since `InstallErrorCause` keys off `HasProperty(options, "cause")` rather than the value, `{ cause: undefined }` installs the own property, and Node's inspect prints it:

```
KeypressError: Failed to press key "ArrowRight" during navigation
  [cause]: undefined
```

That is a noise line on every keypress failure with nothing to attach — in the very output this PR exists to make readable. Verified both directions before and after:

```
hasOwnProperty(err, 'cause')  →  { cause } : true      cause ? { cause } : undefined : false
util.inspect contains [cause] →  { cause } : true      cause ? { cause } : undefined : false
```

The fix passes the options object only when there is a cause, and `errorCause.spec.ts` now asserts the absence of the own property. I checked the assertion actually bites by reverting the fix locally and re-running — test 3 fails with `Expected: false, Received: true`, and passes again with it restored. A regression test that has never been seen to fail isn't worth much.

**Observation 2** — agreed that the positional-vs-`ErrorOptions` split is a copy-the-wrong-pattern hazard, so I put the reasoning where a contributor will actually hit it rather than leaving it in the PR description:

```ts
/**
 * Unlike the other errors here, this one takes its cause positionally rather
 * than as `ErrorOptions`, because it also folds the cause's text into its own
 * message. New error classes should follow the `options?: ErrorOptions`
 * pattern the rest of this file uses.
 */
```

**On the coverage check** — you grepped for wrapping sites still missing a cause and correctly classified the remainder as validation throws and direct assertion failures with no original error to attach. That matches what I found, and it's the check I'd most want a second pair of eyes on, since a mechanical 198-site edit is exactly where a silent miss hides.

**On not being able to run the suite** — noted, and reasonable to say so plainly rather than implying a green run. For the record from this side: lint clean, type-check clean, chromium `109 passed / 7 failed / 185 did not run`, where the 7 and 185 are `main`'s untouched baseline that #674 fixes.

---

PR #679, opened 2026-07-30
test(e2e): make the WebKit project pass

jooyoungseo:
# Pull Request

## Description

WebKit had never actually been verified. #674 and #676 both shipped with "webkit — not verified" in their notes, because its Playwright build needs system libraries the sandbox lacked. I installed them (`npx playwright install-deps webkit`) and ran it.

It was badly broken, for a reason that had nothing to do with WebKit rendering:

```
before   22 failed,  49 passed    (three barplot specs only)
after     0 failed,  71 passed
```

Full three-browser suite, `CI=true`, on this branch:

```
903 passed (chromium 301, firefox 301, webkit 301)   0 failed
```

## Related Issues

Follows #674 (restored the e2e suite) and #676 (error causes). Closes the WebKit gap both left open.

### 1. The control modifier was read from the wrong machine

Playwright's WebKit on Linux is a **Mac-flavoured WebKit** — it reports `navigator.platform` as `MacIntel`. Measured across the three projects:

| project | `navigator.platform` | app binds | test sent |
|---|---|---|---|
| chromium | `Linux x86_64` | Control | Control ✅ |
| firefox | `Linux x86_64` | Control | Control ✅ |
| **webkit** | **`MacIntel`** | **Command** | Control ❌ |

`src/util/platform.ts` reads `navigator.platform`, so MAIDR was right to bind Command. The test constants read `process.platform` — the *host* OS — which is `linux` on the runner:

```ts
static readonly META_KEY = process.platform === 'darwin' ? 'Meta' : 'Control';
```

Host and browser happen to agree for Chromium and Firefox, which is why nobody noticed. On WebKit they disagree, every Ctrl-combination shortcut silently no-ops, and help, settings, extreme navigation and autoplay all fail.

Replaced with `modifierKey(page)` in a new `e2e_tests/utils/platform.ts`, which asks the browser the same way the app does. Both constants are deleted rather than left in place, so the next caller can't pick the broken one.

### 2. A cross-browser defect I introduced in #676

`errorCause.spec.ts` blocked its navigation with `page.route(/\.html$/, route => route.abort())`. **Playwright only intercepts `file://` navigations in Chromium** — which is the only browser I verified it on. In Firefox and WebKit the navigation just succeeded and the test failed its own guard.

This was already on `main` and would have failed the scheduled workflow. Closing the page instead makes `page.goto` reject identically everywhere; the asserted chain is unchanged.

### 3. A racy assertion

`isModeActive` snapshotted the live region immediately after the keypress, racing MAIDR's announcement. It now waits for the expected text first, then re-reads and compares as before.

### 4. Retries on CI — a mitigation, stated as one

Several other getters read live regions the same snapshot way. Under parallel load WebKit occasionally reads first, and **a different test failed on each full run**:

| run | WebKit failure |
|---|---|
| 1 | multiLayer → toggle sound mode |
| 2 | multiLineplot → toggle braille mode |
| 3 | multiLayer → display Y-axis Title |
| 4 | boxplotVertical → Y-axis Title, dodgedBarplot → bottom level |

Every one passed 3/3 in isolation. That is load-induced flakiness, not four bugs.

`retries: process.env.CI ? 2 : 0` absorbs it without hiding it — Playwright still marks a retried test **flaky** in the report, and a test that fails every attempt still fails the run. Local runs keep 0 retries so a real failure stays immediate.

**This is not a cure.** The cure is for those getters to wait for the region to update rather than snapshotting it, the way `isModeActive` now does. The config comment says exactly that and points at the example. I did not do it here because it means touching every reader and risks masking genuine "never announced" bugs — which are accessibility bugs, and the ones this suite exists to catch.

## Checklist

- [ ] I have updated the documentation, if applicable. — no user-facing behaviour changed.

## Additional Notes

| check | result |
|---|---|
| `npm run lint` | clean |
| `npm test` (jest) | 1277 passed, 100 suites |
| e2e — chromium | 301 passed |
| e2e — firefox | 301 passed |
| e2e — **webkit** | **301 passed** |
| e2e — all three, `CI=true` | **903 passed, 0 failed** |

No `src/` product code is touched. The product was correct throughout — on a browser reporting itself as a Mac it binds Command, which is right. Only the tests were wrong.

Two things worth flagging beyond this diff:

- **`e2e.yml` and `e2e_tests.yml` are near-duplicates** — same `0 2 */2 * *` cron, both running the full three-browser suite, so every scheduled window does the work twice. Only `e2e_tests.yml` files the report issue. Left alone as unrelated, but worth collapsing.
- **The PR job still runs chromium only.** Now that WebKit and Firefox pass, adding them is defensible, but it roughly triples that job's runtime (~3m40s → ~9m). I'd rather the maintainer make that call than spend it by default.

---

### DIFF
--- a/e2e_tests/config/test-config.ts
+++ b/e2e_tests/config/test-config.ts
@@ -6,6 +6,7 @@
  */
 import type { PlaywrightTestConfig } from '@playwright/test';
 import path from 'node:path';
+import process from 'node:process';
 import { fileURLToPath } from 'node:url';
 
 // The package is ESM ("type": "module"), so CommonJS `__dirname` does not
@@ -40,6 +41,18 @@ const config: PlaywrightTestConfig = {
   expect: {
     timeout: 5000,
   },
+
+  // The page objects read MAIDR's live regions immediately after dispatching a
+  // key. Under parallel load WebKit — the slowest of the three — occasionally
+  // reads before the announcement lands, and a different test flakes each run.
+  // Retries absorb that without hiding it: Playwright still marks a test that
+  // needed one as "flaky" in the report. A test that fails every attempt is a
+  // real failure and still fails the run.
+  //
+  // This is a mitigation, not the cure. The cure is for those getters to wait
+  // for the region to update rather than snapshotting it once — see
+  // `isModeActive` in base-page.ts for the shape that fix takes.
+  retries: process.env.CI ? 2 : 0,
 
   // Test reporters
   reporter: [
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -762,6 +762,16 @@ export class BasePage {
     mode: string,
     modeMessages: Record<string, string>,
   ): Promise<boolean> {
+    // The announcement lands asynchronously after the keypress, so reading the
+    // region once races the update — fast enough to pass in Chromium and
+    // Firefox, and a flake in WebKit under full-suite load. Wait for the
+    // expected text instead, and report absence as false rather than throwing:
+    // callers assert on the boolean, and "the mode never announced" is a
+    // failed expectation, not a broken page object.
+    await expect(this.page.locator(notificationSelector))
+      .toHaveText(modeMessages[mode], { timeout: 5000 })
+      .catch(() => { /* fall through to the re-read below */ });
+
     try {
       const notificationText = await this.getElementText(notificationSelector);
       return notificationText === modeMessages[mode];
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -317,7 +317,7 @@ export class BasePage {
   public async showHelpMenu(): Promise<void> {
     try {
       await this.pressKeyCombination(
-        await this.resolveModifier('key combination'),
+        await this.resolveModifier('show help menu'),
         TestConstants.SLASH_KEY,
         'show help menu',
       );
@@ -341,7 +341,7 @@ export class BasePage {
   public async showSettingsMenu(): Promise<void> {
     try {
       await this.pressKeyCombination(
-        await this.resolveModifier('key combination'),
+        await this.resolveModifier('show settings menu'),
         TestConstants.COMMA_KEY,
         'show settings menu',
         100,
@@ -369,7 +369,7 @@ export class BasePage {
    */
   public async openHelpMenu(): Promise<void> {
     await this.pressKeyCombination(
-      await this.resolveModifier('key combination'),
+      await this.resolveModifier('open help menu'),
       TestConstants.SLASH_KEY,
       'open help menu',
     );
@@ -382,7 +382,7 @@ export class BasePage {
    */
   public async openSettingsMenu(): Promise<void> {
     await this.pressKeyCombination(
-      await this.resolveModifier('key combination'),
+      await this.resolveModifier('open settings menu'),
       TestConstants.COMMA_KEY,
       'open settings menu',
       100,
@@ -396,7 +396,7 @@ export class BasePage {
   public async closeSettingsMenuWithEscape(): Promise<void> {
     try {
       await this.pressKeyCombination(
-        await this.resolveModifier('key combination'),
+        await this.resolveModifier('show settings menu'),
         TestConstants.COMMA_KEY,
         'show settings menu',
         100,
@@ -428,7 +428,7 @@ export class BasePage {
   public async verifySettingsMenuIgnoresBackdropClick(): Promise<void> {
     try {
       await this.pressKeyCombination(
-        await this.resolveModifier('key combination'),
+        await this.resolveModifier('show settings menu'),
         TestConstants.COMMA_KEY,
         'show settings menu',
         100,
@@ -541,7 +541,7 @@ export class BasePage {
     useMetaKey = false,
   ): Promise<void> {
     if (useMetaKey) {
-      await this.pressKeyCombination(await this.resolveModifier('key combination'), key, action);
+      await this.pressKeyCombination(await this.resolveModifier(action), key, action);
     } else {
       await this.pressKey(key, action);
     }
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -176,26 +176,29 @@ export class BasePage {
 
   /**
    * Presses a key combination (e.g., Command + key)
-   * @param modifierKey - The modifier key (e.g., Command, Shift)
+   *
+   * Named `modifier` rather than `modifierKey` so it does not shadow the
+   * imported `modifierKey()` helper, which callers pass the result of.
+   * @param modifier - The already-resolved modifier key (e.g., Meta, Shift)
    * @param key - The key to press
    * @param context - Context description for error reporting
    * @param delay - Optional delay between key presses
    * @throws KeypressError if key combination fails
    */
   protected async pressKeyCombination(
-    modifierKey: string,
+    modifier: string,
     key: string,
     context: string,
     delay = 50,
   ): Promise<void> {
     try {
-      await this.page.keyboard.down(modifierKey);
+      await this.page.keyboard.down(modifier);
       await this.page.waitForTimeout(delay);
       await this.pressKey(key, context);
-      await this.page.keyboard.up(modifierKey);
+      await this.page.keyboard.up(modifier);
     } catch (error) {
       throw new KeypressError(
-        `${modifierKey}+${key}`,
+        `${modifier}+${key}`,
         context,
         error instanceof Error ? error : undefined,
       );
@@ -517,7 +520,8 @@ export class BasePage {
    * Moves to a specific data point using a key combination
    * @param key - The key to press
    * @param action - Description of the movement action
-   * @param useMetaKey - Whether to use the Meta key
+   * @param useMetaKey - Whether to hold the control/command modifier, which
+   * resolves to Meta or Control depending on what the browser reports
    * @throws KeypressError if operation fails
    */
   protected async moveToDataPoint(
@@ -767,6 +771,11 @@ export class BasePage {
    * @param modeMessages - Map of mode values to expected messages
    * @returns Promise resolving to true if mode is active, false otherwise
    * @throws Error if mode status cannot be checked
+   *
+   * Note: a false result costs the full wait timeout, because the wait can only
+   * end early on a match. Every caller today asserts the mode IS active, so
+   * that path is not hit; a future "assert mode X is NOT active" test would pay
+   * ~5s per call and should take a shorter timeout rather than live with it.
    */
   protected async isModeActive(
     notificationSelector: string,
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -2,6 +2,7 @@ import type { Locator, Page } from '@playwright/test';
 import { expect } from '@playwright/test';
 import { TestConstants } from '../utils/constants';
 import { AssertionError, KeypressError } from '../utils/errors';
+import { modifierKey } from '../utils/platform';
 
 /**
  * Base page object that all other page objects extend
@@ -291,7 +292,7 @@ export class BasePage {
   public async showHelpMenu(): Promise<void> {
     try {
       await this.pressKeyCombination(
-        TestConstants.COMMAND_KEY,
+        await modifierKey(this.page),
         TestConstants.SLASH_KEY,
         'show help menu',
       );
@@ -315,7 +316,7 @@ export class BasePage {
   public async showSettingsMenu(): Promise<void> {
     try {
       await this.pressKeyCombination(
-        TestConstants.COMMAND_KEY,
+        await modifierKey(this.page),
         TestConstants.COMMA_KEY,
         'show settings menu',
         100,
@@ -343,7 +344,7 @@ export class BasePage {
    */
   public async openHelpMenu(): Promise<void> {
     await this.pressKeyCombination(
-      TestConstants.COMMAND_KEY,
+      await modifierKey(this.page),
       TestConstants.SLASH_KEY,
       'open help menu',
     );
@@ -356,7 +357,7 @@ export class BasePage {
    */
   public async openSettingsMenu(): Promise<void> {
     await this.pressKeyCombination(
-      TestConstants.COMMAND_KEY,
+      await modifierKey(this.page),
       TestConstants.COMMA_KEY,
       'open settings menu',
       100,
@@ -370,7 +371,7 @@ export class BasePage {
   public async closeSettingsMenuWithEscape(): Promise<void> {
     try {
       await this.pressKeyCombination(
-        TestConstants.COMMAND_KEY,
+        await modifierKey(this.page),
         TestConstants.COMMA_KEY,
         'show settings menu',
         100,
@@ -402,7 +403,7 @@ export class BasePage {
   public async verifySettingsMenuIgnoresBackdropClick(): Promise<void> {
     try {
       await this.pressKeyCombination(
-        TestConstants.COMMAND_KEY,
+        await modifierKey(this.page),
         TestConstants.COMMA_KEY,
         'show settings menu',
         100,
@@ -514,7 +515,7 @@ export class BasePage {
     useMetaKey = false,
   ): Promise<void> {
     if (useMetaKey) {
-      await this.pressKeyCombination(TestConstants.META_KEY, key, action);
+      await this.pressKeyCombination(await modifierKey(this.page), key, action);
     } else {
       await this.pressKey(key, action);
     }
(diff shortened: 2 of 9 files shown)


jooyoungseo:
All three taken in `7daf027`. The `BarPlotPage` question was the right one to ask.

**1. `BarPlotPage.isModeActive` — you're right, and it was missed, not intentional.**

It narrows the base signature (`(mode, expectedMessage)` vs `(notificationSelector, mode, modeMessages)`), which TypeScript accepts as an override, so all four barplot toggle tests resolved to the snapshot version and never reached the base class's new wait. Exactly as you say: leaning on the blanket retries instead of the actual fix, in the one page object that has its own copy. I grepped for other overrides — `BarPlotPage` is the only one.

Fixed by delegating rather than by duplicating the wait, so there stays one implementation:

```ts
return await super.isModeActive(this.selectors.notification, mode, {
  [mode]: expectedMessage,
});
```

The `BarPlotError` wrapping is preserved around it.

**2. Whitespace disagreement — real, and closer to the surface than "probably never triggers".**

`toHaveText` normalises; the follow-up read compared raw `textContent` with `===`. Worth noting `getNotificationText` in that same file *already* did `.replace(/\s+/g, ' ').trim()` — because MAIDR's announcement markup genuinely wraps across lines. So the base comparison was the odd one out, not a hypothetical. Both sides now go through a shared `normalizeText`, which is also the rule `toHaveText` applies.

**3. Swallowed error in the wait — kept, with the reasoning made explicit.**

I did reconsider this one. The alternative is `try { await expect(...); return true } catch { return false }`, which removes the `.catch` but also removes the diagnostic: a genuinely missing notification container would silently return `false` instead of throwing `Failed to check ${mode} status`. Since the read below is authoritative and does surface real errors, discarding the timeout is the intended behaviour rather than an oversight, so I rewrote the comment to say that outright instead of the vaguer "fall through":

```
// ... A timeout here is deliberately not fatal: the read
// below is the authoritative check and answers either way, so rethrowing
// would only turn a "mode never announced" expectation into a page-object
// error. Callers assert on the boolean.
```

Happy to switch to the stricter shape if you'd rather have the terser code than the error.

**4. Redundant round-trips — hoisted** at both autoplay sites (`base-page.ts` and `barplot-page.ts`). One `const modifier = await modifierKey(this.page)` per autoplay instead of one for the keydown and another for the keyup. The remaining call sites are one per logical action, so nothing to save there.

Re-verified after the change, `CI=true`, all three projects: **903 passed, 0 failed** — and with the barplot toggles now waiting rather than retrying, that number is doing slightly more honest work than it was before.

---

jooyoungseo:
All three taken in `a263cfe`. Two of them were my doing, so worth fixing rather than deferring.

**2. Shadowed parameter — renamed.** `pressKeyCombination(modifierKey: string, ...)` predates this branch, but importing `modifierKey()` into the same file is what made it ambiguous, and the call site you quoted is the proof:

```ts
await this.pressKeyCombination(await modifierKey(this.page), key, action);
```

Argument and parameter sharing a name while resolving to different things. Now `modifier`, with a line on the method saying why, so nobody renames it back.

**3. Stale `useMetaKey` doc — corrected.** It now reads "the control/command modifier, which resolves to Meta or Control depending on what the browser reports" — which is the whole point of the branch, so leaving it saying "the Meta key" would have been a small contradiction sitting right next to the fix.

**1. The `false`-costs-5s asymmetry — documented on the method.** You're right that nothing hits it: I re-checked, every caller asserts the mode *is* active. But the cost is real and invisible at the call site, so a note in the JSDoc rather than only in this thread:

```
Note: a false result costs the full wait timeout, because the wait can only
end early on a match. Every caller today asserts the mode IS active, so
that path is not hit; a future "assert mode X is NOT active" test would pay
~5s per call and should take a shorter timeout rather than live with it.
```

I considered making the timeout a parameter now, but with no caller needing it that's speculative API surface — the note tells the next person what to reach for.

Re-verified after the change: lint clean, type-check clean, `CI=true` across all three projects **903 passed, 0 failed**.

Worth restating for whoever merges, since two reviews have now flagged it: this PR's own CI runs **chromium only**, so the WebKit result it's named for is not demonstrated by the green checks here — only by the local three-browser runs quoted above and, after merge, by the scheduled `e2e.yml` / `e2e_tests.yml` workflows. Those are the ones that have been red since 2026-06-01 and should be the real confirmation.

---

jooyoungseo:
Both taken in `88abb47`. The unwrapped-error path was a real gap and partly self-inflicted, so I'd rather respin than defer it.

**Unwrapped `modifierKey()` resolution.** Your reading of the evaluation order is exactly right: `pressKeyCombination` takes an already-resolved `modifier: string`, so `await modifierKey(this.page)` runs in the *caller's* scope before the callee's `try` is entered. `openHelpMenu`, `openSettingsMenu` and `moveToDataPoint` all document `@throws KeypressError` and would have let a page-evaluate failure through raw.

Worth owning that **both `startAutoplay` cases were my doing** — I hoisted the resolution above the `try` in the previous review round to save the duplicate round-trip, and in doing so moved it out from under the guard. Fixing one review note created the next one.

Rather than patch each site, everything now goes through one wrapper:

```ts
protected async resolveModifier(context: string): Promise<string> {
  try {
    return await modifierKey(this.page);
  } catch (error) {
    throw new KeypressError('modifier', context, error instanceof Error ? error : undefined);
  }
}
```

`grep` confirms the only remaining direct `modifierKey(this.page)` call is inside that wrapper. The autoplay resolution also moved back inside its `try` so `BarPlotError`/`Error` still holds there — the hoist is preserved, just relocated under the guard, so it's still one resolution per autoplay.

**`normalizeText` duplication.** Fair, and I'd under-scoped it: I kept the helper module-private to hold the diff down, which left the copy in `getInstructionText` a few hundred lines below it and a third in `BarPlotPage.getNotificationText`. Now `e2e_tests/utils/text.ts`, imported by both files — `grep` for the regex returns one hit.

**One result worth reporting rather than rounding off.** The verification run came back:

```
902 passed, 1 flaky, 0 failed
  [webkit] multiLayer › Speed Controls › should be able to speed up
```

That is the retry mechanism doing precisely what the PR claims for it — a WebKit live-region read lost the race, the retry passed, and Playwright surfaced it as **flaky** instead of swallowing it. The test reads the speed indicator, which is one of the snapshot-reading getters the PR explicitly lists as still exposed. So the disclosed gap is real and now has an observed instance attached to it, and the mitigation behaved as advertised rather than as hoped.

Lint and type-check clean.

---

jooyoungseo:
Fixed in `20f9f42`. You caught me doing the small version of the thing this branch exists to prevent.

`resolveModifier('key combination')` was mine from the previous commit — a bulk replacement that dropped a generic string in at every site, while the descriptive context was sitting one argument below on every single one of those calls, and `boxplotVertical-page.ts` was already passing its real name. Losing the shortcut's identity in an error message is exactly the failure mode #676 was about, just smaller.

All seven now pass the same context their adjacent `pressKeyCombination` gets:

```ts
await this.pressKeyCombination(
  await this.resolveModifier('show help menu'),
  TestConstants.SLASH_KEY,
  'show help menu',
);
```

and `moveToDataPoint` passes its `action` parameter. `grep` confirms no generic contexts remain.

**On the retry sign-off** — agreed that's a maintainer call, not mine, and your framing of the risk is the right one: a blanket CI retry absorbs *future* unrelated flakiness too, not just the races characterised here. Two things that might help the decision:

Across the verification runs on this branch the flaky set is never the same twice, and it is always WebKit and always a snapshot-reading getter:

| run | flaky |
|---|---|
| 1 | — (903 passed clean) |
| 2 | webkit multiLayer → speed up |
| 3 | webkit dodgedBarplot → speed up, webkit multiLineplot → Go To options |

Latest run: **901 passed, 2 flaky, 0 failed.** So the retries are load-bearing, they are visibly load-bearing, and every instance so far falls inside the gap the PR discloses rather than outside it. If you'd rather not take the blanket retry, the alternative is to land the modifier fix and hold the `retries` line — WebKit would then be genuinely green but intermittently red on the scheduled run until those getters are converted to wait. I'd rather you choose that than have me assume it.

Lint and type-check clean.

---

jooyoungseo:
Took the message-shape note in `8c45ac2`; the other two need no change.

**`KeypressError('modifier', …)`** — you're right that it reads wrong, and it's the small version of this PR's own theme, so worth fixing rather than waving through as cosmetic. `KeypressError` renders `Failed to press key "<key>" during <context>`, so the literal `'modifier'` produced a sentence about pressing a key named "modifier":

```
before:  Failed to press key "modifier" during open help menu: page closed
after :  Failed to press key "Meta/Control" during open help menu: page closed
```

Naming the two candidates says what could not be decided.

I deliberately did **not** switch to `TestError`, which was the other obvious fix. Eighteen methods reach this path via `moveToDataPoint` and every one documents `@throws KeypressError`; changing the type to satisfy a message-shape nit would mean sweeping all of them, and the type is already accurate — the keypress genuinely did not happen. One word beat eighteen doc edits.

**`boxplotVertical-page.ts` line length** — confirmed clean. There is no `max-len` rule in the flat config, and `npm run lint` passes with no warnings; I re-ran it after this change.

**The two maintainer-decision items** — agreed, and thanks for saying so explicitly rather than leaving them ambiguous. They stay flagged, not absorbed.

Latest verification, `CI=true`, all three projects: **900 passed, 3 flaky, 0 failed.** Flaky this run: boxplotVertical X-axis title, multiLayer Y-axis title, multiLineplot speed-up — all WebKit, all axis-title or speed-indicator reads, which is squarely the `getAxisTitle`/speed-indicator snapshot class the PR documents as still exposed. Across six runs on this branch the flaky set has been different every time and has never once fallen outside that class, which is the strongest evidence I can offer that the characterisation is right and the boundary is where the description says it is.

---

jooyoungseo:
Checked your `.catch` observation rather than reflexively patching it, and it turns out not to bite — but the verification run that came with it **falsifies something I claimed two comments ago**, so that part matters more.

**On the broad `.catch`: no change needed, and here is why.** A strict-mode violation is not absorbed into a bare `false`. The soft-wait swallows it, but `getElementText` below raises the same violation, and the surrounding catch surfaces it as `Failed to check <mode> status` with the violation as the cause. Measured against a live page rather than argued:

```
--- selector "div" matches 7 elements ---
  WAIT: expect(locator).toHaveText(expected) failed        <- swallowed
  READ: rejected -> strict mode violation: locator('div') resolved to 7 elements
--- selector "svg g" matches 65 elements ---
  READ: rejected -> strict mode violation: locator('svg g') resolved to 65 elements
```

So only a genuine "never announced" reaches the boolean. Recorded in the code (`ee4857f`) so the next reader doesn't have to re-derive it.

**Correction to my previous comment.** I wrote that across runs the flaky set "has never once fallen outside" the snapshot-reading getters the PR leaves exposed. The next run broke that:

```
901 passed, 2 flaky, 0 failed
  [webkit] histogram → should toggle text mode on and off
  [webkit] stackedBarplot → should toggle sound mode on and off
```

Both go through `isModeActive` — the path this PR *fixes* with a wait. My generalisation was drawn from five runs and stated more strongly than five runs support.

Looking at why, the wait is weaker than I described. These tests assert a **sequence of transient announcements**:

```ts
await toggleTextMode();  const terse   = await isTextModeActive(TERSE);
await toggleTextMode();  const off     = await isTextModeActive(OFF);
await toggleTextMode();  const verbose = await isTextModeActive(VERBOSE);
```

Waiting fixes reading *too early*. It cannot fix arriving *too late*: if the region has already advanced past the expected value before the wait starts, no amount of waiting brings it back, and the 5s simply elapses. So for transient-sequence assertions the wait is a strong mitigation, not a cure — same category as the retries, which is not how I characterised it.

The real fix for that shape is the one the repo's own testing rules already prescribe for live regions — observe with a `MutationObserver` and assert against the recorded sequence rather than sampling the current value. That is a larger change across the mode-toggle tests and I would not smuggle it into this PR.

Net effect on the merge decision: unchanged — WebKit goes from 22 hard failures to 0, and the residual is flake absorbed by retries. But the retries are load-bearing for a slightly wider set than I said, and you should have the accurate version before signing off on them.

---

jooyoungseo:
Note 3 taken in `c3bf085`. Note 1 sent me to check something I should have checked before writing my last comment — the answer is reassuring.

**1. Could retries be masking a product bug rather than a read-timing artifact?**

This got sharper after my correction two comments up, where the flaky set turned out to include `isModeActive` paths. "MAIDR sometimes doesn't announce" and "the test sampled at the wrong moment" produce identical symptoms, and only one of them is an accessibility bug. So I stopped sampling and recorded the region with a `MutationObserver` instead — the approach `.claude/rules/testing.md` already prescribes for live regions — through the exact toggle sequence that flakes, on WebKit, five times:

```
ANNOUNCEMENTS: ["Text mode is terse","Text mode is off","Text mode is verbose"]   x5
MODES OBSERVED: terse, off, verbose (3/3)
```

Every announcement fires, every time. The product is not dropping them; the test is arriving after the value has moved on. So the retries are absorbing a **test-sampling artifact, not a silent accessibility regression**.

Caveat, since this is the sort of claim I just had to walk back: that ran in isolation, not under the parallel load where the flakes actually occur. It shows the announcement pipeline is sound and matches the mechanism, but it does not prove announcements never drop under load. The observer-based assertion is what would prove it, and that is the follow-up this thread keeps circling.

**3. `BarPlotPage.isModeActive` incompatible override — renamed.** You're right that it's pre-existing, but it's worth more than a follow-up note: that override is *precisely* how the barplot toggles bypassed the base wait, and it was invisible at both the call sites and the base method. Now `isModeMessageActive`, still delegating to `super.isModeActive`, so a future base-signature change cannot silently miss it.

**2. `KeypressError('Meta/Control', …)`** — agreed on the trade-off, leaving it.

**4. Jest coverage for `normalizeText` / `modifierKey`** — declining, and not only on convention. `e2e_tests/` is outside jest's `roots` (`src`, `test`), and `.claude/rules/testing.md` draws the three suites as an explicit boundary. Adding an e2e utility to the unit suite would be the first crossing of it, which is a call for a maintainer, not a drive-by. `normalizeText` is also exercised on every mode assertion in a 900-test three-browser run.

Latest: **901 passed, 2 flaky, 0 failed** (WebKit speed-indicator reads — the snapshot class).

---

jooyoungseo:
Measured the `noImplicitOverride` suggestion rather than just agreeing with it, and the result is more nuanced than either of us assumed.

**It would not have rejected the bad signature.** Your reasoning was that TS "would have required an explicit `override` keyword and then rejected the incompatible signature outright." The first half is right; the second isn't. Method parameters are bivariant and a shorter parameter list is assignable to a longer one, so the exact shape that caused the bug still compiles cleanly *with* the keyword:

```ts
class Base {
  protected async isModeActive(sel: string, mode: string, msgs: Record<string, string>): Promise<boolean>
}
class Sub extends Base {
  protected override async isModeActive(mode: string, expected: string): Promise<boolean>  // compiles
}
```

`noImplicitOverride: true`, `strict: true`, exit 0.

**The flag is still worth having, for a different reason.** Its value here is *visibility*, not rejection: `override` at the declaration site would have made the shadowing obvious to a reader, and that is precisely what was missing — the bug was invisible at the call sites *and* at the base method. So the suggestion is good, the mechanism is just "forces disclosure" rather than "blocks the mistake."

**What it costs, measured on this tree:**

| area | errors |
|---|---|
| `e2e_tests/page-objects/plots` | 80 |
| `src/model` | 64 |
| `src/state/viewModel` | 12 |
| `test/model` | 3 |
| `src/service` | 3 |
| **total** | **162** |

161 are `TS4114` (missing `override`) and one is `TS4116` (an abstract override in `src/model/abstract.ts:631`). All mechanical — I checked specifically for a non-mechanical error hiding in there, hoping the flag had caught a real latent bug, and it hadn't.

So it's a ~162-line mechanical sweep across `src/`, `test/` and `e2e_tests/`. That is a reasonable follow-up but not something to bolt onto this PR: it would touch product code, and "no `src/` changes" is a property several reviews here have leaned on when assessing risk. Config restored, `npm run type-check` clean.

**On filing follow-up issues** — you're the second reviewer to ask for one (the earlier round wanted one for the remaining snapshot-reading getters). I'm checking with the maintainer on whether they want these as issues rather than opening them unprompted on their repo; there are now three candidates: the duplicate `e2e.yml`/`e2e_tests.yml` cron, the snapshot getters, and `noImplicitOverride` with the numbers above.

---

jooyoungseo:
Follow-ups filed, so nothing here depends on this thread staying findable:

- **#680** — make the live-region getters wait instead of snapshotting. This is the real cure for the WebKit flakiness that `retries` currently absorbs. Carries the six-run flaky table, the `MutationObserver` evidence that MAIDR announces reliably (so this is a test-sampling artifact, not an accessibility bug), and the distinction between getters that need a wait and mode toggles that need a recorded sequence. Its acceptance test is deliberately sharp: **`retries` can be deleted and a full three-browser run is still clean.**
- **#681** — `e2e.yml` and `e2e_tests.yml` run the same suite on the same cron. Includes the side-by-side table and the observation that only one of them files the report issue, so a red run in the other is effectively silent — worth knowing when triaging.
- **#682** — enable `noImplicitOverride`, with the measured 162-error breakdown and the correction that it discloses shadowing rather than preventing it.

That resolves every follow-up raised across the ten review rounds on this PR. This branch itself stays as it is — test infrastructure only, no `src/` changes.

---

PR #684, opened 2026-07-31
test(e2e): synchronise assertions with MAIDR's announcements

### LINKED ISSUE #680: test(e2e): make the live-region getters wait instead of snapshotting
## Background

#679 made the WebKit e2e project pass (22 hard failures → 0) and, as part of that, taught `BasePage.isModeActive` to wait for the expected announcement instead of reading the live region once. The remaining getters were left snapshotting, and #679 added `retries: process.env.CI ? 2 : 0` as an explicitly-labelled mitigation rather than a cure.

This issue is that cure.

## The pattern

The page objects read MAIDR's live regions immediately after dispatching a key:

```ts
const notificationText = await this.getElementText(selector);
return notificationText === expected;   // may run before the announcement lands
```

Chromium and Firefox are fast enough that this almost always wins the race. WebKit under parallel load is not, so a different test flakes on each full run.

Still on the snapshot pattern:

- `getAxisTitle` — axis-title assertions
- `getCurrentDataPointInfo` — navigation assertions
- `getPlaybackSpeed` — speed-control assertions
- `waitForElementContent` callers reading the info region

`isModeActive` in `e2e_tests/page-objects/base-page.ts` is the shape to follow.

## Evidence

Across full three-browser runs on the #679 branch the flaky set was **different every time** and **always WebKit**:

| run | flaky |
|---|---|
| 1 | — (903 passed clean) |
| 2 | multiLayer → speed up |
| 3 | dodgedBarplot → speed up, multiLineplot → Go To options |
| 4 | boxplotVertical X-axis, multiLayer Y-axis, multiLineplot speed up |
| 5 | histogram text mode, stackedBarplot sound mode |
| 6 | heatmap reset speed, stackedBarplot speed up |

**This is not a product bug.** Recording the live region with a `MutationObserver` through the toggle sequence that flakes, on WebKit, five times, showed every announcement firing:

```
ANNOUNCEMENTS: ["Text mode is terse","Text mode is off","Text mode is verbose"]   x5
MODES OBSERVED: terse, off, verbose (3/3)
```

MAIDR announces reliably; the tests sample at the wrong moment.

## Why waiting is only part of the answer

Run 5 above is the important one — those two are `isModeActive` paths, which #679 already taught to wait. Waiting fixes reading **too early**. It cannot fix arriving **too late**: the mode-toggle tests assert a sequence of transient announcements

```ts
await toggleTextMode();  const terse   = await isTextModeActive(TERSE);
await toggleTextMode();  const off     = await isTextModeActive(OFF);
await toggleTextMode();  const verbose = await isTextModeActive(VERBOSE);
```

and if the region has already advanced past the expected value before the wait starts, no amount of waiting brings it back.

So this issue has two halves:

1. **Getters that assert a settled value** (axis title, current data point, playback speed) → convert to wait-then-read, exactly like `isModeActive`.
2. **Tests that assert a sequence of transient announcements** (the mode toggles) → observe the region with a `MutationObserver` and assert against the recorded sequence. `.claude/rules/testing.md` already prescribes this for live regions: *"A live region announces on the DOM mutation, not on the state update … observe the region with a `MutationObserver` — an assertion on the text alone passes even when nothing moved."*

## Done when

- The getters above wait rather than snapshot.
- The mode-toggle tests assert against a recorded announcement sequence.
- `retries` can be dropped from `e2e_tests/config/test-config.ts` and a full three-browser run is clean without it.

That last point is the real acceptance test: today the retries are load-bearing, and they should not be.

Refs #679, #626

jooyoungseo:
# Pull Request

## Description

#679 left the WebKit suite green only because it added `retries: process.env.CI ? 2 : 0`, and said so plainly. #680 wrote down the acceptance test for removing them: **a full three-browser run is clean without retries.**

This PR meets it.

## Related Issues

Closes #680. Follows #679, #676, #674. Spun off #685.

### The race, and why waiting was not enough

The page objects read the live region straight after dispatching a key. #679 taught `isModeActive` to wait for the expected text, which fixes reading **too early** — but not arriving **too late**. The region holds one message at a time, so an announcement already replaced cannot be waited for. That is why mode toggles kept flaking after #679.

The fix is to stop sampling and start observing, and to move the wait to *before* the keypress.

### 1. Record the announcements

`e2e_tests/utils/announcements.ts` observes `document.body` and counts inserted `role="alert"` nodes.

Three things here are deliberate, and all three were wrong in earlier revisions of this PR:

- **The alert node, not `#maidr-text-container`.** `Text.tsx` renders `current` (gated by the `announce` flag) into the alert and `visual` (ungated) into the container. During autoplay the container advances every step while nothing is announced — measured: container changed `true`, announcements recorded `0`.
- **Walk the `MutationRecord`s, don't sample the live node.** A callback batches every record since the last flush, so two announcements landing before one flush would collapse into a single entry.
- **`querySelectorAll`, not `querySelector`.** A record lists only the *root* of an inserted subtree, so two alerts mounted under one new node collapse the same way, one level down.

### 2. Wait for the announcement the action causes

`pressKeyAwaitingAnnouncement` marks the count, presses, then waits for it to advance — mode toggles, speed changes, data-point moves, layer switches. A silent keypress resolves `false` and the caller's assertion still decides.

### 3. Assert against what was announced, not what is on screen

Waiting for *any* new announcement is not sufficient. The layer switches did not await, so their message could land during the **next** action and satisfy its wait early:

```
announcements since mark: ["Layer 2 of 2: single line plot at X values is 0, …",
                           "Sound is off"]
```

That made the multiLayer sound toggle fail on the third acceptance run. `isModeActive` now asks what the action announced. The mark is **consumed** after each check, so a check with no preceding awaited action falls back to the region rather than matching stale history — a silent false positive otherwise.

### 4. Axis titles wait on the display, not on an announcement

The axis title only reaches the alert once the cursor is on a data point, and those specs assert straight after activation:

```
without prior navigation:  2082 ms, announcements []
after navigation:            59 ms, announcements ["X label is Petal Length (cm)"]
```

An announcement wait there finds nothing and spends its whole budget — so those tests passed because a 2 s timeout acted as a sleep, which is the anti-pattern this branch exists to remove. They now wait for the container the assertion reads: **2082 ms → 78 ms**.

### 5. Drop the retries

`retries` is gone, so a flake is a real failure again. `trace` becomes `retain-on-failure`, because `on-first-retry` can no longer fire and a genuine failure would otherwise leave only a screenshot.

### 6. Make the invariant enforceable, not remembered

Five review rounds each found one more page object dispatching a key straight at the keyboard, so the invariant is now lint-enforced rather than left to inspection:

- `keyboard.press` is banned across `page-objects/**`, with one inline disable on the sanctioned call site in `pressKey`. Widening it from `plots/**` to the whole directory immediately caught a bypass no re-read had: **`activateMaidr` pressed Tab directly**, so activation neither wrapped its failure nor cleared the mark.
- `pressKeyCombination` is banned in `plots/**`. It routes through `pressKey`, so nothing goes unwrapped — but it does not *wait*, and a modified data-point move announces. `moveToTop` on the vertical boxplot was the live case.
- The mark is invalidated by any unwrapped keypress, and cleared in `awaitingAnnouncement` itself rather than relying on `pressKey` being reached — `pressKeyCombination` holds the modifier down first, so a throw on `keyboard.down` would never reach the clearing site.

Each of these is pinned by a test in `announcements.spec.ts` and verified to fail without its fix, with one exception stated plainly: the `keyboard.down` case has no test, because nothing swallows a `KeypressError` today and the stale mark has no route to a check. The change removes the dependency on that staying true.

The rules deliberately stop at `page-objects/**`. Three specs do press keys directly (`multiLineplot` 20, `liveData` 13, `dialogAccessibility` 3), but each keypress is followed by a *waiting* assertion on text that persists rather than a sample of the transient alert — polling something that persists cannot arrive too late, which is exactly the property `role="alert"` lacks.

### Waits swallow a timeout and nothing else

Both waits distinguish "nothing happened" from "the run is broken", rather than folding every rejection into a no-op:

```
TOHAVETEXT       name= "Error"        ctor= ExpectError
WAITFORFUNCTION  name= "TimeoutError" ctor= TimeoutError
```

`expect().not.toHaveText()` offers nothing to narrow on, which is why the axis-title wait is written with `page.waitForFunction` — the consistency is in behaviour, not just in comments.

## Checklist

- [ ] I have updated the documentation, if applicable. — no user-facing behaviour changed. `.claude/agents/test-runner.md` corrected, since it described the retry config this PR removes.

## Additional Notes

| check | result |
|---|---|
| `npm run lint` | clean |
| `npm test` (jest) | 1291 passed, 101 suites |
| e2e — webkit | 308 passed (3.8m) |
| e2e — chromium | 308 passed (3.5m) |
| e2e — firefox | 308 passed (4.1m) |

Run one project at a time at `338acba`, clean tree verified before and after, exit 0 on all three. No retries, so every one is a hard-failure run.

This PR's own CI job runs chromium only — firefox and WebKit live in the scheduled workflows — so the three-browser criterion from #680 is met locally and re-run on every push rather than read off this PR's checks.

**Cost.** An earlier revision of this description claimed "~2% slower". That was true but was measuring dead time: 22 axis toggles per project each burning a 2 s timeout. With change 4 that is gone — WebKit went 4.1 min → 3.6 min. The remaining synchronisation resolves as soon as the announcement lands.

**What this does not claim.** Consecutive clean runs are good evidence, not proof. If something flakes again, the recorded announcements are available to the test, so the diagnosis starts from what MAIDR actually announced rather than from one sampled string.

**Scope.** No `src/` product code is touched. The investigation on #679 established that MAIDR announces reliably — five for five on WebKit — so this was always a test-synchronisation defect, not an accessibility one.

### DIFF
--- a/.claude/agents/test-runner.md
+++ b/.claude/agents/test-runner.md
@@ -32,7 +32,7 @@ You are a testing specialist for the MAIDR accessibility library.
 - Test config: `e2e_tests/config/test-config.ts`
 - Utils: `e2e_tests/utils/constants.ts`, `e2e_tests/utils/errors.ts`
 - Tests run against `file://` protocol (no server needed)
-- 30s timeout, retries: 2
+- 30s timeout, no retries (a flake is a real failure; see e2e_tests/utils/announcements.ts)
 
 ### Existing E2E spec files
 - `barplot.spec.ts` — Bar plot navigation and sonification
--- a/e2e_tests/config/test-config.ts
+++ b/e2e_tests/config/test-config.ts
@@ -6,7 +6,6 @@
  */
 import type { PlaywrightTestConfig } from '@playwright/test';
 import path from 'node:path';
-import process from 'node:process';
 import { fileURLToPath } from 'node:url';
 
 // The package is ESM ("type": "module"), so CommonJS `__dirname` does not
@@ -41,18 +40,6 @@ const config: PlaywrightTestConfig = {
   expect: {
     timeout: 5000,
   },
-
-  // The page objects read MAIDR's live regions immediately after dispatching a
-  // key. Under parallel load WebKit — the slowest of the three — occasionally
-  // reads before the announcement lands, and a different test flakes each run.
-  // Retries absorb that without hiding it: Playwright still marks a test that
-  // needed one as "flaky" in the report. A test that fails every attempt is a
-  // real failure and still fails the run.
-  //
-  // This is a mitigation, not the cure. The cure is for those getters to wait
-  // for the region to update rather than snapshotting it once — see
-  // `isModeActive` in base-page.ts for the shape that fix takes.
-  retries: process.env.CI ? 2 : 0,
 
   // Test reporters
   reporter: [
--- a/e2e_tests/config/test-config.ts
+++ b/e2e_tests/config/test-config.ts
@@ -55,8 +55,13 @@ const config: PlaywrightTestConfig = {
     // Browser settings
     viewport: null,
 
-    // Capture traces and screenshots on failure
-    trace: 'on-first-retry',
+    // Capture traces and screenshots on failure.
+    //
+    // `retain-on-failure`, not `on-first-retry`: there are no retries any more,
+    // so a trace keyed to a retry attempt would never be written and a real
+    // failure would leave only a screenshot. A failure is now always the first
+    // and only attempt, which is exactly when the trace is worth having.
+    trace: 'retain-on-failure',
     screenshot: 'only-on-failure',
   },
 
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -1048,9 +1048,10 @@ export class BasePage {
       await this.page.keyboard.down(TestConstants.SHIFT_KEY);
       await this.pressKey(arrowKey, `start ${directionName} autoplay`);
 
+      // Only the two modifiers are still held: `pressKey` above is a full
+      // press, so the arrow key was already released.
       await this.page.keyboard.up(modifier);
       await this.page.keyboard.up(TestConstants.SHIFT_KEY);
-      await this.page.keyboard.up(arrowKey);
 
       if (expectedContent) {
         await this.waitForElementContent(
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -699,8 +699,28 @@ export class BasePage {
    * @throws KeypressError if operation fails
    */
   protected async toggleAxisTitle(axisKey: string, axisName: string): Promise<void> {
+    // Wait on the displayed text, not on an announcement. The axis title only
+    // reaches the alert once the cursor is on a data point, and these specs
+    // assert it straight after activation — so an announcement wait finds
+    // nothing and spends its whole timeout, which passes only because the
+    // delay works as a sleep. The assertion reads this container, so a change
+    // here is the event it actually depends on.
+    //
+    // This also settles the two-keypress sequence: entering label scope warns
+    // when text mode is off, and that announcement lands inside this wait
+    // rather than leaking into the next action's.
+    const container = this.page.locator(
+      `#${TestConstants.MAIDR_NOTIFICATION_CONTAINER}`,
+    );
+    const before = normalizeText((await container.textContent()) ?? '');
+
     await this.pressKey(TestConstants.LABEL_KEY, 'label scope');
-    await this.pressKeyAwaitingAnnouncement(axisKey, axisName);
+    await this.pressKey(axisKey, axisName);
+
+    await expect(container)
+      .not
+      .toHaveText(before, { timeout: TestConstants.ANNOUNCEMENT_TIMEOUT })
+      .catch(() => { /* the caller's assertion decides */ });
   }
 
   /**
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -23,14 +23,13 @@ export class BasePage {
    * region's current text, so a later announcement replacing theirs — or an
    * earlier one arriving late — cannot turn a real pass into a failure.
    *
-   * The contract is positional and not enforced by the types: `isModeActive`
-   * assumes the action it is checking was the last one to set this. Every
-   * caller follows `toggleXMode()` with `isXModeActive()`, which holds it. A
-   * check that is NOT preceded by an awaited action reads the window since
-   * some unrelated earlier action instead, so widen this to an explicit
-   * handle rather than adding such a caller.
+   * Null until an awaited action sets it, and reset to null once a check has
+   * consumed it. That makes the positional contract self-limiting: a check
+   * with no awaited action before it takes the region fallback instead of
+   * matching against an unrelated earlier announcement, which would be a
+   * silent false positive rather than a visible failure.
    */
-  private actionAnnouncementMark = 0;
+  private actionAnnouncementMark: number | null = null;
 
   /**
    * Common selectors used across different pages
@@ -858,13 +857,20 @@ export class BasePage {
     // been replaced — an announcement from an earlier, un-awaited action
     // arriving late is enough — and sampling the current text would then report
     // a mode that did toggle as inactive.
-    const announced = await announcementsSince(this.page, this.actionAnnouncementMark);
-    if (announced.length > 0) {
-      return announced.some(text => normalizeText(text) === normalizeText(expected));
+    const mark = this.actionAnnouncementMark;
+    this.actionAnnouncementMark = null;
+    if (mark !== null) {
+      const announced = await announcementsSince(this.page, mark);
+      if (announced.length > 0) {
+        return announced.some(text => normalizeText(text) === normalizeText(expected));
+      }
     }
 
-    // No announcements recorded, so this was not reached through an awaited
-    // action. Fall back to waiting on the region. A timeout here is
+    // Nothing recorded for this action, so fall back to the region. Note this
+    // path reads `#maidr-text-container`, which is the displayed text and not
+    // the `role="alert"` node a screen reader hears — the two agree for every
+    // `notify()`-driven message, which is all of these, but the fallback does
+    // not carry the same guarantee as the recorded path above. A timeout here is
     // deliberately not fatal: the read below is the authoritative check and
     // answers either way, so rethrowing would only turn a "mode never
     // announced" expectation into a page-object error. Callers assert on the
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -1,7 +1,6 @@
 import type { Locator, Page } from '@playwright/test';
 import { expect } from '@playwright/test';
 import {
-  announcementCount,
   announcementsSince,
   installAnnouncementRecorder,
   waitForAnnouncementAfter,
@@ -234,8 +233,9 @@ export class BasePage {
    * @param act - The key action to run
    */
   private async awaitingAnnouncement(act: () => Promise<void>): Promise<void> {
-    await installAnnouncementRecorder(this.page);
-    this.actionAnnouncementMark = await announcementCount(this.page);
+    // Install and mark in one page call: this path now runs on nearly every
+    // interaction in the suite, so a second round trip per action is not free.
+    this.actionAnnouncementMark = await installAnnouncementRecorder(this.page);
     await act();
     await waitForAnnouncementAfter(this.page, this.actionAnnouncementMark);
   }
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -1,5 +1,11 @@
 import type { Locator, Page } from '@playwright/test';
 import { expect } from '@playwright/test';
+import {
+  announcementCount,
+  announcementsSince,
+  installAnnouncementRecorder,
+  waitForAnnouncementAfter,
+} from '../utils/announcements';
 import { TestConstants } from '../utils/constants';
 import { AssertionError, KeypressError } from '../utils/errors';
 import { modifierKey } from '../utils/platform';
@@ -11,6 +17,14 @@ import { normalizeText } from '../utils/text';
  */
 export class BasePage {
   protected readonly page: Page;
+
+  /**
+   * Announcement count captured immediately before the most recent action that
+   * waited for one. Assertions read the window after this mark rather than the
+   * region's current text, so a later announcement replacing theirs — or an
+   * earlier one arriving late — cannot turn a real pass into a failure.
+   */
+  private actionAnnouncementMark = 0;
 
   /**
    * Common selectors used across different pages
@@ -190,6 +204,27 @@ export class BasePage {
   }
 
   /**
+   * Presses a key and waits for MAIDR to announce the result.
+   *
+   * This is the synchronisation point the assertions depend on. Reading the
+   * live region straight after a keypress races the announcement, and waiting
+   * afterwards cannot recover one that has already been replaced — so the wait
+   * has to start before the key is pressed.
+   *
+   * Silent keypresses are fine: the wait resolves false and the caller's own
+   * assertion still decides.
+   * @param key - The key to press
+   * @param context - Context description for error reporting
+   * @throws KeypressError if the key press itself fails
+   */
+  protected async pressKeyAwaitingAnnouncement(key: string, context: string): Promise<void> {
+    await installAnnouncementRecorder(this.page);
+    this.actionAnnouncementMark = await announcementCount(this.page);
+    await this.pressKey(key, context);
+    await waitForAnnouncementAfter(this.page, this.actionAnnouncementMark);
+  }
+
+  /**
    * Presses a key combination (e.g., Command + key)
    *
    * Named `modifier` rather than `modifierKey` so it does not shadow the
@@ -279,7 +314,7 @@ export class BasePage {
    * @throws KeypressError if toggling fails
    */
   protected async toggleMode(key: string, modeName: string): Promise<void> {
-    await this.pressKey(key, modeName);
+    await this.pressKeyAwaitingAnnouncement(key, modeName);
   }
 
   /**
@@ -496,7 +531,7 @@ export class BasePage {
    * @throws KeypressError if operation fails
    */
   protected async changeSpeed(key: string, action: string): Promise<void> {
-    await this.pressKey(key, action);
+    await this.pressKeyAwaitingAnnouncement(key, action);
   }
 
   /**
@@ -545,9 +580,12 @@ export class BasePage {
     useMetaKey = false,
   ): Promise<void> {
     if (useMetaKey) {
+      await installAnnouncementRecorder(this.page);
+      this.actionAnnouncementMark = await announcementCount(this.page);
       await this.pressKeyCombination(await this.resolveModifier(action), key, action);
+      await waitForAnnouncementAfter(this.page, this.actionAnnouncementMark);
     } else {
-      await this.pressKey(key, action);
+      await this.pressKeyAwaitingAnnouncement(key, action);
     }
   }
 
@@ -647,7 +685,7 @@ export class BasePage {
    */
   protected async toggleAxisTitle(axisKey: string, axisName: string): Promise<void> {
     await this.pressKey(TestConstants.LABEL_KEY, 'label scope');
-    await this.pressKey(axisKey, axisName);
+    await this.pressKeyAwaitingAnnouncement(axisKey, axisName);
   }
 
   /**
@@ -799,13 +837,22 @@ export class BasePage {
   ): Promise<boolean> {
     const expected = modeMessages[mode];
 
-    // The announcement lands asynchronously after the keypress, so reading the
-    // region once races the update — fast enough to pass in Chromium and
-    // Firefox, and a flake in WebKit under full-suite load. Wait for the
-    // expected text first. A timeout here is deliberately not fatal: the read
-    // below is the authoritative check and answers either way, so rethrowing
-    // would only turn a "mode never announced" expectation into a page-object
-    // error. Callers assert on the boolean.
+    // Prefer what the action actually announced. The region holds one message
+    // at a time, so by the time this runs the mode message may already have
+    // been replaced — an announcement from an earlier, un-awaited action
+    // arriving late is enough — and sampling the current text would then report
+    // a mode that did toggle as inactive.
+    const announced = await announcementsSince(this.page, this.actionAnnouncementMark);
+    if (announced.length > 0) {
+      return announced.some(text => normalizeText(text) === normalizeText(expected));
+    }
+
+    // No announcements recorded, so this was not reached through an awaited
+    // action. Fall back to waiting on the region. A timeout here is
+    // deliberately not fatal: the read below is the authoritative check and
+    // answers either way, so rethrowing would only turn a "mode never
+    // announced" expectation into a page-object error. Callers assert on the
+    // boolean.
     // The catch is broad on purpose but not lossy: a structural problem such as
     // a strict-mode violation from a selector matching several elements is
     // raised again by `getElementText` below and surfaces as "Failed to check
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -901,9 +901,13 @@ export class BasePage {
    * @throws Error if mode status cannot be checked
    *
    * Note: on the recorded path a false result is immediate — the window is
-   * already known, so a mismatch costs nothing. Only the fallback below pays
-   * the full wait timeout, and it is reached only when no awaited action
-   * preceded the check.
+   * already known, so a mismatch costs nothing. The fallback below is what
+   * pays the full wait timeout, and it is reached in three cases: no awaited
+   * action preceded the check, the action announced nothing, or this is the
+   * second check after a single action, since the first read consumes the
+   * mark. All three still answer correctly, just from the region rather than
+   * from what was announced — so a check that wants the stronger guarantee
+   * should follow its own awaited action.
    */
   protected async isModeActive(
     notificationSelector: string,
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -251,9 +251,16 @@ export class BasePage {
     // interaction in the suite, so a second round trip per action is not free.
     const mark = await installAnnouncementRecorder(this.page);
 
-    // Publish only once the action has actually run. `act()` presses keys and
-    // `pressKey` clears the mark, so holding it in a local until then is what
-    // lets that clearing be unconditional. It also means a throwing action
+    // Clear here rather than leaving it to `pressKey`. `act()` is sometimes
+    // `pressKeyCombination`, which holds the modifier down before it reaches
+    // `pressKey` — so an action that throws on `keyboard.down` would never
+    // reach the clearing site and would leave an earlier action's mark in
+    // place. Nothing swallows a KeypressError today, so that stale mark has no
+    // route to a later check; this makes the invalidation independent of that
+    // staying true.
+    this.actionAnnouncementMark = null;
+
+    // Publish only once the action has actually run, so a throwing action
     // leaves the mark cleared: a caller that swallowed the failure and carried
     // on cannot be handed a window belonging to an action that never happened.
     await act();
@@ -945,7 +952,7 @@ export class BasePage {
     // <mode> status" with the violation as its cause. Only a genuine
     // "never announced" reaches the boolean. Verified, not assumed.
     await expect(this.page.locator(notificationSelector))
-      .toHaveText(expected, { timeout: 5000 })
+      .toHaveText(expected, { timeout: TestConstants.REGION_FALLBACK_TIMEOUT })
       .catch(() => { /* the read below decides */ });
 
     try {
--- a/e2e_tests/page-objects/base-page.ts
+++ b/e2e_tests/page-objects/base-page.ts
@@ -23,6 +23,13 @@ export class BasePage {
    * waited for one. Assertions read the window after this mark rather than the
    * region's current text, so a later announcement replacing theirs — or an
    * earlier one arriving late — cannot turn a real pass into a failure.
+   *
+   * The contract is positional and not enforced by the types: `isModeActive`
+   * assumes the action it is checking was the last one to set this. Every
+   * caller follows `toggleXMode()` with `isXModeActive()`, which holds it. A
+   * check that is NOT preceded by an awaited action reads the window since
+   * some unrelated earlier action instead, so widen this to an explicit
+   * handle rather than adding such a caller.
    */
   private actionAnnouncementMark = 0;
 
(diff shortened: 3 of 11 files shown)


jooyoungseo:
You were right, and this was the most valuable catch of the whole series. Fixed in `870b5fb`.

## The recorder was reading the wrong node

Confirmed against `src/ui/component/Text.tsx` exactly as you described:

```ts
current = message || (announce && navText)   // -> role="alert" node
visual  = message || navText                 // -> #maidr-text-container
```

Only `current` is gated by `announce`, and only the alert node is what a screen reader hears. I was observing the display container and calling it "what MAIDR announced." Every path this branch synchronises goes through `notify()` where the two strings agree — which is exactly why four clean runs held and hid it.

**And it is observable, not just theoretical.** Autoplay is your example, and it reproduces:

```
container changed during autoplay:      true
announcements recorded during autoplay: 0     (with the fix)
```

The container advances on every autoplay step while nothing is announced. The old recorder would have logged a phantom announcement per step. Anyone later writing a silent-navigation assertion on top of it would have been measuring the display and believing they were measuring the live region — which, in a repo whose first principle is that accessibility is the product, is the worst possible place for that confusion to live.

The fix reads the alert, scoped through the container's parent so MUI's `role="alert"` in the settings dialog cannot be mistaken for it, and keys on **node identity** rather than text: the alert is keyed by a revision counter and re-mounts per announcement, so a new node means a new announcement and a mutation elsewhere means none.

## Pinned rather than left to manual checking

Your last note asked for a regression test. It's `e2e_tests/specs/announcements.spec.ts`, in the e2e suite rather than jest — `e2e_tests/` sits outside jest's roots and the three-suite split is explicit in `.claude/rules/testing.md`, and a jsdom fake could not have caught this bug anyway, because the bug was about which real node MAIDR renders what into. Three tests:

1. one entry per announcement, in order
2. a repeated identical message registers every time
3. **display updates that were never announced are ignored** — asserting both that the container did change and that nothing was recorded

Test 3 is the one that tells the two nodes apart; it fails against the previous implementation.

## Other notes

- **Duplicated install/mark/wait** — folded into one `awaitingAnnouncement(act)` helper taking the key action as a callback. You're right that the second caller is what earns the abstraction under the project's own rule.
- **`actionAnnouncementMark` contract** — documented on the field, including what breaks if a check is *not* preceded by an awaited action, and pointing at the explicit-handle fix if such a caller ever appears.
- **Magic `2000`** — now `TestConstants.ANNOUNCEMENT_TIMEOUT`.

## Re-verification after repointing

The recorder now reads a different node, so the acceptance evidence was re-earned rather than carried over:

| project | result |
|---|---|
| chromium | 304 passed |
| firefox | 304 passed |
| webkit | 304 passed ×3 |
| `announcements.spec.ts`, 3 browsers ×3 | 27 passed |

Still no retries, so every one of those is a hard-failure run.

---

jooyoungseo:
Both actionable notes taken in `fb5e652`. The trace one was a real regression I introduced and walked straight past.

**1. Dead `trace` config — fixed, and worse than it first looks.** Removing `retries` made `trace: 'on-first-retry'` unreachable, so a genuine CI failure would have produced a screenshot and nothing else. On the branch whose entire purpose is making these failures diagnosable, that is the wrong direction, and I made it while congratulating myself about diagnosability.

The line above it already read:

```ts
// Capture traces and screenshots on failure
trace: 'on-first-retry',
```

so the comment was aspirational before and became false the moment I dropped retries. Now `retain-on-failure`, with a note saying why the retry-keyed variant cannot fire any more.

**2. Recorder installed per action — fixed, but not the way you suggested.** You're right about the cost: that path now runs on nearly every interaction. I did not move installation into `activateMaidr()` though, because a navigation replaces `window` and takes the recorder with it — pinning it to one earlier moment makes it go *quietly* blind, which is the failure mode this whole PR exists to eliminate.

Instead the install call now returns the count, so install-and-mark is one round trip rather than two while staying lazy:

```ts
this.actionAnnouncementMark = await installAnnouncementRecorder(this.page);
```

Same number of page calls per action as before your note minus one, no fragility added. That left `announcementCount` with no callers, so it's gone rather than sitting there as dead exported surface.

**3. Positional contract** — thanks for actually walking the call sites to confirm it holds rather than taking the doc comment's word for it. No change, as you say.

**On not being able to run anything:** stating that plainly is the right call, and it is the third review in this series to do so. Re-verified here after these changes, no retries so every run is hard-failure:

| project | result |
|---|---|
| chromium | 304 passed |
| firefox | 304 passed |
| webkit | 304 passed |
| `announcements.spec.ts` × 3 browsers | 9 passed |

CI was green on the previous head including `E2E (chromium)`; this commit re-triggers it.

---

jooyoungseo:
All four taken in `eb32c20`. Three turned out to be fixable rather than merely documentable, which is a better outcome than the "known tradeoff" framing I had given two of them.

**3. MutationObserver batching — fixed, and this changes the recorder's correctness argument.** You're right that sampling the live alert once per callback undercounts when two land in one batch. It now walks the `MutationRecord`s and counts added alert nodes, which is what an announcement physically *is*:

```ts
new MutationObserver((mutations) => {
  for (const mutation of mutations) {
    for (const node of mutation.addedNodes) { /* … */ }
```

That also drops the node-identity bookkeeping, so the recorder is now correct by construction rather than correct because batching is "unlikely today". Notably this is the second time on this branch that a reviewer found the recorder measuring a proxy for announcements rather than announcements — first the wrong node, now the wrong sampling. Worth naming the pattern.

**1. Positional contract — made self-limiting instead of documented.** Your sharper reading was the useful bit: I had written that a check without a preceding action "reads the window since some unrelated earlier action," but you pointed out the consequence is a **false positive** matching history, not a fallback. A silent wrong answer is worse than a slow right one.

`actionAnnouncementMark` is now `null` until an action sets it and `null` again once a check consumes it, so that case takes the region fallback rather than matching stale announcements. The convention is no longer load-bearing on convention.

**2. Fallback reads the container, not the alert — noted in place.** Fair catch that the distinction commit 3 established doesn't hold on the fallback path. I left the behaviour (every `notify()`-driven message makes the two agree, and with the change above this path is now genuinely unreachable in the current suite) but said so at the point of use, since silently losing that distinction is exactly what the previous commit existed to prevent.

**4. `.claude/agents/test-runner.md` — fixed here rather than deferred.** It said `retries: 2`, and this branch is what makes that false, so it belongs in this diff. `npm run sync:copilot` reports the generated side already in sync.

Re-verified after the rewrite — the recorder counts differently now, so the evidence was re-earned rather than carried:

| project | result |
|---|---|
| chromium | 304 passed |
| firefox | 304 passed |
| webkit | 304 passed |
| `announcements.spec.ts` × 3 browsers × 2 | 18 passed |

No retries, so each of those is a hard-failure run.

---

jooyoungseo:
Your `toggleAxisTitle` question was the right thread to pull, and measuring it turned up something worse than the gap you described — in my own change. Fixed in `d6e366a`.

**What I found.** I went to check whether the label-scope press could leak an announcement, and instead discovered the *second* press wasn't doing what I thought either. The axis title only reaches the alert once the cursor is on a data point, and the axis-title specs assert straight after activation without navigating:

```
without prior navigation:  2082 ms, announcements []
after navigation:            59 ms, announcements ["X label is Petal Length (cm)"]
```

So the announcement wait I added found nothing and spent its entire 2 s budget, on all 22 axis toggles per project. **Those tests have been passing because that timeout behaves as a sleep** — the precise anti-pattern this branch exists to remove, reintroduced by the branch itself. I had measured the suite getting ~2% slower and accepted it without asking *where*.

**The fix** waits for the displayed text to change, which is the container the assertion actually reads:

```
toggleXAxisTitle: 2082 ms -> 78 ms
```

Real synchronisation instead of an accidental delay, and 26× faster on that path.

**And it answers your original question.** `warnIfTextOff()` does call `notify()`, but only when text mode is off — confirmed in `textViewModel.ts:202-208`, and reproduced:

```
text mode ON  -> []
text mode OFF -> ["Text mode is off. To enable, press the T key."]
```

The axis specs run with text on, so it is latent rather than live — but with the display wait covering both keypresses, any such warning now lands *inside* that wait rather than leaking into the next action's marked window. The leak is closed rather than documented.

**On your two other observations:** `replayCurrentPoint` and autoplay's initial arrow press are deliberately unawaited — autoplay turns `announce` off outright, so there is nothing to wait for, and the recorder spec's third test pins exactly that. Neither is followed by an announcement-dependent assertion. And yes, the fallback's full-timeout cost is real; it is now unreachable in the current suite since the mark is consumed, so it costs nothing in practice.

Re-verified, no retries, every run hard-failure:

| project | result |
|---|---|
| chromium | 304 passed |
| firefox | 304 passed |
| webkit | 304 passed ×3 |

WebKit is now 3.6 min against 4.1 before this commit — the dead waiting is gone.

I'll correct the "~2% slower" line in the PR description, since with this the suite should be at or below where it started rather than above it.

---

jooyoungseo:
Both taken in `34d2434`, and the second one was bigger than either of us thought.

**2. Raw keypresses bypassing the synchronisation — it's two files, not one.** You flagged `boxplotVertical-page.ts`. Grepping for the pattern rather than trusting the one example:

```
violinPlot-page.ts:187,199,211,223  ArrowRight / ArrowLeft / ArrowUp / ArrowDown
violinPlot-page.ts:235,247          PageDown / PageUp        <- layer switches
boxplotVertical-page.ts:320,332     ArrowDown / End
```

Six more in `violinPlot-page.ts`, and two of those are **PageUp/PageDown layer switches** — the exact announcing action whose late message leaked into the next action's marked window and made the multiLayer sound toggle fail on the third acceptance run of this branch. I fixed that bug in `multiLayer-page.ts` and never opened the violin file to check whether it had the same shape. It did.

You were right that it wasn't obviously live today, and it wasn't the thing failing — but "the same class of race sitting untouched in one file" turned out to be understating it, and I'd rather close it than file it. All eight now go through `pressKeyAwaitingAnnouncement`. The single raw press left is barplot's Tab, which moves focus to activate MAIDR and announces nothing.

**1. Implicit `autoplayDuration` margin — derived instead of documented.** Confirmed at `src/type/settings.ts:139`. Your read is right: the flat 2000 ms only holds because the default is 4000 and autoplay hasn't reached the bound, where `announce` flips back on and a real announcement lands. It now samples at half the actual setting:

```ts
await page.waitForTimeout(DEFAULT_SETTINGS.general.autoplayDuration / 2);
```

so tuning the default moves the sample point with it rather than turning this into a mystery flake.

**3.** Confirmed — both catches are the same deliberate pattern, not an oversight.

Re-verified after the change, no retries so each run is hard-failure:

| project | result |
|---|---|
| chromium | 304 passed |
| firefox | 304 passed |
| webkit | 304 passed |
| `announcements.spec.ts` × 3 browsers × 2 | 18 passed |

Worth noting what this round says about the branch as a whole: five review rounds have now each found one more place where a keypress and its assertion weren't synchronised. That's a good argument that the grep — `this.page.keyboard.press` outside `base-page.ts` — belongs in review habit rather than in any one person's memory.

---

jooyoungseo:
All three taken in `88bb920`. Your sharp-edge note got the stronger of the two options you offered.

**2. The bypass hazard — lint rule, not a comment.** You suggested "an ESLint rule of thumb or a comment." A comment would not have helped here, and the branch's own history is the evidence: **five review rounds each turned up one more call site**, most recently the violin layer switches repeating the exact leak already fixed in `multiLayer-page.ts`. I wrote on an earlier round that this "belongs in review habit rather than in any one person's memory" — habit clearly wasn't sufficient either, including mine.

So it's now enforced:

```ts
files: ['e2e_tests/page-objects/plots/**/*.ts'],
rules: {
  'no-restricted-syntax': ['error', {
    selector: 'CallExpression[callee.property.name=\'press\'][callee.object.property.name=\'keyboard\']',
    message: 'Do not press keys directly in a plot page object. Use this.pressKey() …',
  }],
},
```

`base-page.ts` is deliberately out of scope — that's where the wrapping lives. And I checked the rule actually fires rather than assuming, by reintroducing a raw press:

```
247:13  error  Do not press keys directly in a plot page object…  no-restricted-syntax
✖ 1 problem
```

then confirming it goes clean on restore. A lint rule that never fires is decoration.

That also flushed out the last raw press — barplot's Tab. It announces nothing, so it moved to `pressKey` rather than the awaiting variant, which also gets it `KeypressError` wrapping it didn't have.

**1. Stale `isModeActive` JSDoc — corrected.** You're right, and it was mine: the "every false costs the full timeout" note predates the recorded-window path, which answers immediately on a mismatch. Only the fallback pays it now, and it's reached only when no awaited action preceded the check.

**3. `replayCurrentPoint` — removed.** Confirmed zero callers across `e2e_tests/` and `src/`. Consistent with dropping `announcementCount` and the `META_KEY`/`COMMAND_KEY` constants earlier on this line of work: dead exported surface is what the next person reaches for by mistake.

Re-verified, no retries, every run hard-failure:

| project | result |
|---|---|
| chromium | 304 passed |
| firefox | 304 passed |
| webkit | 304 passed |

---

jooyoungseo:
Both taken in `912c6b8` — including the one you offered as "not a strong ask", which turned out to be testable after all.

**The batching gap — pinned.** You were right that it's the one claim in this branch with no test behind it, and right that two MAIDR notifies landing in a single flush isn't something a spec can force. But the claim is about the *observer's* batching behaviour, and that is directly testable: insert two alert nodes in one synchronous block and assert both are recorded.

I checked it actually pins the thing rather than just passing. Against the previous sample-once-per-callback implementation:

```
✘ records both announcements when two land in one flush
  - Expected  - 1
  + Received  + 0
```

and green against the current one. The test says outright that it drives the DOM rather than MAIDR, and why.

**The comment — trimmed, and a correction on the citation.** The substance of your point is right and I've made the change: "five review rounds on #684" is noise in a year, and the rule's rationale should stand alone. It now reads:

```ts
// A key dispatched straight at the keyboard skips BasePage's
// synchronisation, so its announcement can land during the NEXT action and
// satisfy that action's wait early — the assertion then reads the region
// before its own announcement arrives. Layer switches are the usual case.
```

But the quote attributed to this repo's `CLAUDE.md` — *"Don't reference the current task, fix, or callers … since those belong in the PR description and rot as the codebase evolves"* — **is not in this repo.** I grepped `CLAUDE.md`, `.claude/rules/`, and every markdown file under `.claude/`; it appears nowhere. The nearest thing is `.claude/rules/testing.md:104`, which points the other way in a narrow case: pin a known bug with `it.failing` *and the issue number*.

Flagging that because a fabricated citation to project policy is more durable than the suggestion it was attached to — someone could later enforce it as a rule that was never agreed. The advice was good on its own merits and didn't need the borrowed authority.

**Your other checks** match what I found: the fallback does get a second chance rather than short-circuiting to `false`, `base-page.ts` is the only other file with `keyboard.press` and both uses are intentional, and the concurrent-announcement scenario is real but not reachable while specs run sequentially — worth revisiting only if a spec ever interleaves autoplay with another action.

Re-verified, no retries:

| project | result |
|---|---|
| chromium | 305 passed |
| firefox | 305 passed |
| webkit | 305 passed |

---

jooyoungseo:
Both taken in `036f0f4`. The first one has a better answer than "probably fine".

**1. The `toggleAxisTitle` / `isModeActive` asymmetry — real, but not a hole.** You're right that `toggleAxisTitle` has no re-read after its catch, so nothing plays the role `getElementText` plays in `isModeActive`. It doesn't need one, and the reason is two lines above the catch rather than after it:

```ts
const before = normalizeText((await container.textContent()) ?? '');   // <- no catch
...
await expect(container).not.toHaveText(before, …).catch(() => {});
```

That `textContent()` uses the *same locator* with nothing catching it, and locator calls enforce strict mode — so a selector matching several elements raises there, before the wait is reached. Checked rather than argued:

```
locator.textContent: Error: strict mode violation: locator('div') resolved to 3 elements
```

Only the timeout gets to the catch. That's now recorded in the code, since "the same locator was already exercised without a catch" is not something the next reader should have to reconstruct.

**2. Stale mark if the action throws — closed.** Correct that nothing does this today, since a `KeypressError` aborts the test. But it is precisely the false positive the consumed-mark change existed to prevent, just reached from the other side, so leaving it to "no caller does that yet" felt like the wrong call. The mark is now cleared before the error propagates.

**On the process observation** — thanks for noting the commit messages carry their measurements. That has been load-bearing on this branch rather than decorative: the axis-title fix only exists because I measured `2082 ms` and asked where it went, and three separate claims here turned out to need a number rather than an argument.

Re-verified, no retries:

| project | result |
|---|---|
| chromium | 305 passed |
| firefox | 305 passed |
| webkit | 305 passed |

---

jooyoungseo:
Four review rounds since my last reply. Three of them found real gaps and produced changes; the fourth was a documentation correction. Details below, then the acceptance run.

## The stale mark (`3c77b97`)

Correct, and I had missed it. The mark was consumed on read but nothing invalidated it, so a keypress that went around the awaited path left it intact for the next check. I checked the specs first — no current spec interleaves an axis-title toggle with a mode check, so this was latent rather than live — but it is the same bug class the earlier rounds kept turning up, so I made it structural instead of leaving it to inspection:

- `pressKey` now clears the mark unconditionally.
- `awaitingAnnouncement` publishes its mark **after** the action runs, so the wrapped path is unaffected.

That ordering also removed the `catch` that used to clear the mark on failure: an action that throws never publishes one.

Verified by deleting just the clearing line: 1 failed (the new test), 5 passed.

## The `toggleAxisTitle` catch (`f4f039f`)

Also correct — it was the one wait left in the PR folding every rejection into a no-op. But it could not be narrowed as suggested. I probed both:

```
TOHAVETEXT       name= "Error"        ctor= ExpectError
WAITFORFUNCTION  name= "TimeoutError" ctor= TimeoutError
```

`expect().not.toHaveText()` gives nothing to test against. So I rewrote the wait as `page.waitForFunction`, which does reject with a named `TimeoutError`. The two waits are now consistent in behaviour rather than only in comment. The strict-mode check is unaffected — the `textContent()` read above still exercises the locator uncaught.

## Widening the lint rule (`a9e6d48`) — and what it caught

The point about `base-page.ts` being outside the rule's scope was right, and worth more than the "not a real problem today" framing it was given. I widened it to all of `page-objects/**` with one inline disable on the sanctioned call site in `pressKey`.

Widening it immediately found a real bypass I had not found by re-reading: **`activateMaidr` pressed Tab straight at the keyboard.** Activation neither wrapped its failure as a `KeypressError` nor cleared the mark. That is the fourth file this class of bug has appeared in, and the first one lint found rather than a human.

## `moveToTop` (`c3e3b49`) — worse than "slower fallback"

Confirmed, and it is not only a performance note. `moveToTop` is an *announcing* modified move that was never awaited, which is exactly the shape that broke acceptance run 3 on the multiLayer sound toggle: the announcement can land inside the next action's window and satisfy its wait early. Routed through `moveToDataPoint`, which wraps the combination.

The keypress rule could not catch it — it is not a raw `keyboard.press` — so `pressKeyCombination` is now banned in `plots/**` too. `base-page` keeps it: its own direct callers are the menu and dialog openers, asserted on visibility rather than on an announcement. Both selectors are repeated in the `plots/**` block because flat config *replaces* a rule's options rather than merging them; listing only the new one would have switched the keypress ban off for the exact directory it was written for. Verified both fire.

Its catch also reported `"Failed to move to last box"`, copied from the method above. Corrected.

## The `isModeActive` note (`cdac6ac`)

My comment was wrong, or at least incomplete — it named one of three cases. The fallback is also reached when the action announced nothing, and when a second check follows a single action. Corrected in place. Comment only.

## Declined, with reasons

**The selector is name-based, not type-aware.** Agreed, and there is no fix: `no-restricted-syntax` cannot see types. It is a convention guard, and I would rather have it than not.

**`ANNOUNCEMENT_TIMEOUT` margin under load.** This one is not load-bearing, which is worth stating rather than measuring. If a real announcement took longer than 2 s, `pressKeyAwaitingAnnouncement` resolves `false` and the recorded window is empty — and an empty window falls through to the region wait, which is a separate 5 s budget on the text the assertion actually reads. The timeout being too tight costs time, not correctness.

**The observer rooted at `document.body`.** Raised twice; unchanged. There is no stable node to attach to — MAIDR builds its UI on activation and re-mounts the alert on every announcement. Its cost is measured, not assumed: WebKit went 4.1 → 3.6 min.

## Acceptance run

Three full suites at `c3e3b49`, one project at a time, clean tree verified before and after:

| project | result |
|---|---|
| webkit | 307 passed (3.8m) |
| chromium | 307 passed (3.6m) |
| firefox | 307 passed (4.2m) |

Exit 0 on all three, no retries, so every one is a hard-failure run.

Two earlier runs are not reported here because I edited `base-page.ts` while they were mid-suite. Those results were against a mixed tree and I killed them rather than quote them.

`cdac6ac` is on top of the tested commit and is a comment only — no behaviour delta from what ran above.

---

jooyoungseo:
Three more rounds since the last reply. Two found real gaps, and one of those was a genuine correctness hole rather than a documentation nit.

## `querySelector` recorded only the first alert per added node (`a835598`)

Correct, and it is the same failure the record walk was written to prevent, one level down. A `MutationRecord` lists only the root of an inserted subtree, so two alerts mounted under one new node registered as one entry — exactly the collapse that walking records instead of sampling was meant to avoid.

Now `querySelectorAll`, pinned by a new case that inserts one wrapper containing two alerts. Verified both ways: with `querySelector` that case fails and the other six pass; with `querySelectorAll` all seven pass.

## The mark could survive a throw before `pressKey` was reached (`244e7e9`)

Also correct. The invalidation guarantee rested on `pressKey` being reached, but `pressKeyCombination` holds the modifier down first — so an action throwing on `keyboard.down` would never reach the clearing site and would leave an earlier action's mark in place. `awaitingAnnouncement` now clears it directly.

**No test pins this one.** Nothing swallows a `KeypressError` today, so the stale mark has no route to a later check; the change makes the invalidation stop depending on that staying true. I would rather say that than imply coverage it does not have.

## Dead `keyboard.up(arrowKey)` (`244e7e9`)

Real. `pressKey` is a full press, so the arrow was already released. Removed, with a note on why only the two modifiers remain.

## Changed the name, not the number (`244e7e9`)

The suggestion to fold the mode check's hardcoded `5000` into `ANNOUNCEMENT_TIMEOUT` I did not take, because `ANNOUNCEMENT_TIMEOUT` is `2000` — it would have cut the region fallback from 5 s to 2 s. That path is reached *only* after the recorded window has already come back empty, so it is precisely where a real-but-late message still has to be caught. Shortening it trades a magic number for a flake.

Named it `REGION_FALLBACK_TIMEOUT` instead, value unchanged, with the reason it differs written next to it.

## `some(...)` rather than the last announcement (`ebcfe18`)

Documented rather than narrowed, because `some` is the correct semantic. An awaited action can announce more than once and the mode message is not necessarily last — entering label scope warns when text mode is off *before* announcing the label. Matching only the last entry would report those as inactive.

It is safe because the window holds one action's announcements and no one else's: the mark is taken immediately before the action, and any unwrapped keypress invalidates it. The types say none of that, so the comment now does.

## Left alone, with reasons

- **The observer is never disconnected.** Its lifecycle is already on the function: idempotent, survives for the life of the page, and a navigation replaces `window` and takes it with it. A second note would restate that.
- **`ANNOUNCEMENT_TIMEOUT` under load.** Not load-bearing. If a real announcement took longer than 2 s, the wait resolves `false`, the recorded window is empty, and the check falls through to the region wait — a separate budget on the text the assertion actually reads. Too tight costs time, not correctness.
- **Per-keypress `page.evaluate` round trip.** The wall-clock comparison asked for is already measured: WebKit went 4.1 → 3.6 min, because the axis-title change removed 22 dead 2 s timeouts per project. The round trip is inside that number.
- **The fallback reads the container, not the alert.** Documented at the call site as the weaker guarantee, and reached only when there is no mark to consult.

## Acceptance run

At `244e7e9`, one project at a time, clean tree verified before and after:

| project | result |
|---|---|
| webkit | 308 passed (3.8m) |
| chromium | 308 passed (3.6m) |
| firefox | 308 passed (4.1m) |

Exit 0 on all three, no retries. `ebcfe18` sits on top and is a comment only.

On the note about not being able to run anything in the review sandbox — that is what these local runs are for, and the three-browser-clean-without-retries criterion from #680 is the bar I have been holding this to on every push.

---

jooyoungseo:
Two more rounds. One found something I got wrong, and one asked me to confirm something that turned out not to be true.

## The dead key release was in the wrong copy (`338acba`)

This was a straight miss on my part. The previous round pointed at `BarPlotPage.startAutoplay`'s redundant `keyboard.up(arrowKey)`; I removed it there and never checked whether the base implementation it was copied from had the same line. It did — and `BasePage.startAutoplay` is the copy that `histogram`, `lineplot`, `heatmap`, both boxplot orientations, `multiLineplot`, `multiLayer`, `stackedBarplot` and `dodgedBarplot` all inherit. So the fix landed in the less-used copy first and I called it done.

Fixed. The de-duplication point is right too, and this is a concrete instance of the two copies drifting — filed as #685 rather than widening this branch further.

## "Confirm no spec calls `page.keyboard.press` directly" — they do

I checked rather than confirmed, and the assumption does not hold. Three spec files press keys directly, 36 call sites:

| file | sites |
|---|---|
| `multiLineplot.spec.ts` | 20 |
| `liveData.spec.ts` | 13 |
| `dialogAccessibility.spec.ts` | 3 |

**They are not the race this PR fixes, and I do not think they should be converted here.** The pattern is materially different. Every one of those keypresses is followed by a *waiting* assertion against text that persists, not a sample of a transient region:

```ts
// multiLineplot.spec.ts
await page.keyboard.press('ArrowRight');
await expect(textContainer).toContainText(/x\D*3/i, { timeout: 2000 });

// liveData.spec.ts — waitForAriaText polls with waitForFunction
await page.keyboard.press('ArrowRight');
await waitForAriaText(page, '52');
```

`#maidr-text-container` holds the displayed description, which stays until the next move, and `waitForAriaText` polls `[id^="react-container"]`'s cumulative `textContent`. Polling something that persists cannot arrive too late — which is precisely the property `role="alert"` does *not* have, since it re-mounts per announcement and holds one message at a time. That asymmetry is the whole reason the page objects needed observing rather than waiting.

`liveData.spec.ts` also documents at the top why it drives fixtures the page objects do not model, so the raw keypresses there are a considered exception rather than an oversight.

That is also why the lint rules stop at `page-objects/**`: extending them to specs would flag 36 sites that are not broken, and force a conversion this PR has no business doing.

## Noted, not changed

- **Two `pressKeyAwaitingAnnouncement` calls before one check loses the first mark.** That is the intended reading — the check asks what the *last* awaited action announced. Losing the earlier mark is the design, not a leak.
- **The inline whitespace normalisation could drift from `normalizeText`.** True, and unavoidable: the predicate runs in the page, where `normalizeText` does not exist. The cross-reference comment is the available mitigation.
- **Per-keypress round trip.** Measured, and inside the numbers below: WebKit 4.1 → 3.6 min, because the axis-title change removed 22 dead 2 s timeouts per project.

## Acceptance run

At `338acba`, one project at a time, clean tree verified before and after:

| project | result |
|---|---|
| webkit | 308 passed (3.8m) |
| chromium | 308 passed (3.5m) |
| firefox | 308 passed (4.1m) |

Exit 0 on all three, no retries.

On the repeated ask to confirm the full three-browser run before merge — that is what these are. This PR's CI job runs chromium only; firefox and WebKit live in the scheduled workflows, so the three-browser criterion from #680 is met locally and re-run on every push rather than read off this PR's checks.

---

PR #686, opened 2026-07-31
ci: drop the duplicate scheduled e2e workflow

### LINKED ISSUE #681: ci: e2e.yml and e2e_tests.yml run the same suite on the same schedule
## The overlap

Two scheduled workflows run the full three-browser Playwright suite on an identical cron:

| | `e2e.yml` | `e2e_tests.yml` |
|---|---|---|
| name | Scheduled End-to-End Tests | Scheduled Playwright Tests |
| cron | `0 2 */2 * *` | `0 2 */2 * *` |
| node | `lts/*` | `22` |
| steps | `npm ci` → `e2e:install` → `build` → `npm run e2e` | same |
| timeout | 60 min | 60 min |
| files a report issue | no | **yes** |
| re-raises the failure | yes (direct) | yes (`continue-on-error` + explicit `exit 1`) |

Every scheduled window therefore runs the same ~900-test three-browser suite twice, on the same runner type, at the same minute.

## Why it matters beyond wasted minutes

Only `e2e_tests.yml` files the rolling report issue — that is the workflow that produced #626. `e2e.yml` fails silently by comparison: it has no reporting step, so a red run there is only visible to someone browsing the Actions tab. Two workflows with the same trigger and different observability is a confusing signal when triaging, and it doubles the load exactly when the suite is slowest.

Worth noting they can also disagree: they pin different Node versions (`lts/*` vs `22`), so in principle one can go red while the other stays green, and nothing surfaces that.

## Suggested resolution

Keep `e2e_tests.yml` (it has the reporting step that made #626 visible in the first place) and delete `e2e.yml`, or give them genuinely different jobs — for example one on a different cadence, or split by browser so the wall-clock halves instead of doubling.

Also worth deciding at the same time: the cron uses minute `0`, which is the busiest minute on GitHub's scheduler and a common cause of delayed scheduled runs. An off-minute would get more predictable start times.

Left out of #679 deliberately as unrelated to that fix.

Refs #679, #626

jooyoungseo:
# Pull Request

## Description

`e2e.yml` and `e2e_tests.yml` ran the same three-browser Playwright suite on the same cron (`0 2 */2 * *`), on the same runner type, at the same minute. Every scheduled window paid for the ~900-test suite twice.

## Related Issues

Closes #681. Follows #684, #679, #626. Spun off #687.

### Which one to keep

`e2e_tests.yml`. It is the one that files the rolling report issue — the mechanism that surfaced #626 in the first place. `e2e.yml` had no reporting step, so a red run there was visible only to someone browsing the Actions tab.

### What a plain delete would have cost

`e2e.yml` did one thing the survivor did not: **upload the Playwright report.** That is worth more now than when the issue was filed, because #684 set `trace: 'retain-on-failure'` — so a scheduled failure writes a trace for every failed test, and without an upload step it would leave the reporter's text and nothing to open.

Folded in before deleting, mirroring the pattern already in `ci.yml`'s e2e job (both `playwright-report/` and `test-results/`, `if-no-files-found: ignore`).

**The gate is `always() && steps.e2e.outcome != 'success'`, not `failure()`.** Two separate reasons, both worth stating because `failure()` is the change a future reader is most likely to try:

- The test step runs under `continue-on-error` so the report step always gets its output. That masks the step's *conclusion* to `success`, and only `outcome` keeps the real result — `failure()` would never fire.
- `always()` additionally keeps the step from being skipped when the job is cancelled, which a default `success()`-gated step would be.

Scoped honestly: not skipped is not the same as reached. On the 60-minute timeout the job is cancelled and then force-killed, so the upload runs only if `npm run e2e` exits inside the grace window. Worth having for the case where it does; not a guarantee, and the comment in the file says so rather than implying one.

**One deliberate behaviour change:** `e2e.yml` uploaded on `always()`; this uploads only on non-success, so a green scheduled run no longer produces a downloadable report. That is intentional for two reasons that only hold together after #684 — `retain-on-failure` writes no traces for a passing run, so a green artifact would be an html pass list and nothing more, and the report issue this workflow files already records the reporter's full output for every run including green ones. Recorded in the file so the narrower gate does not read as an oversight.

### The two open questions in the issue

**Node version.** The two workflows pinned different versions (`lts/*` vs `22`), so in principle one could go red while the other stayed green and nothing would surface it. Settled toward `lts/*`, which is what all eight jobs in `ci.yml` use — so the scheduled suite and the chromium job that gates PRs now run the same Node, and a scheduled failure is reproducible from a pull request rather than possibly being a version difference. Picked up `cache: npm` at the same time, which `e2e.yml` had and `e2e_tests.yml` did not.

`lts/*` is Node 24 today and the suite had only ever been run on 22, so the bump is backed by an actual run rather than by the argument that `ci.yml` does it too — see the table below.

**Cron minute.** Moved off minute 0 to `37 2 */2 * *`. Minute 0 is the busiest slot on GitHub's scheduler and runs queued there are routinely delayed; for a suite with a 60-minute timeout, that delay competes with the timeout. Same day cadence, 9:37 PM CDT instead of 9:00.

## Checklist

- [ ] I have updated the documentation, if applicable. — nothing references `e2e.yml`; checked across markdown, YAML, JSON and TypeScript.
- [ ] I have added appropriate unit tests, if applicable. — not applicable to workflow config.

## Additional Notes

Static checks:

| check | result |
|---|---|
| `actionlint` on `e2e_tests.yml` | clean |
| `actionlint` on all of `.github/workflows/` | clean |
| YAML parses, 9 steps in order | confirmed |
| references to `e2e.yml` anywhere in the repo | none |

`actionlint` validates the `steps.e2e.outcome` reference in the new condition, so the gating is checked rather than assumed.

The Node bump, run on **24.18.1** — what `lts/*` resolves to — at `a14f400`, clean tree before and after:

| step | result |
|---|---|
| `npm ci` | exit 0 |
| `npm run build` | exit 0 |
| e2e — chromium | 308 passed (4.3m) |
| e2e — firefox | 308 passed (5.2m) |
| e2e — webkit | 308 passed (4.9m) |

Same 308 per project as on Node 22, no retries. Worth doing because every earlier three-browser run in this thread was on 22.22.2; `ci.yml` already proved chromium on `lts/*`, but firefox and WebKit on Node 24 were proven by nothing.

**What this does not change.** The suite itself, its timeout, its cadence in days, or the report issue's behaviour. The wall-clock of one scheduled run is unchanged — what goes away is the second, quieter copy of it.

**What is not verified here.** No CI on this repository can exercise a `schedule`-triggered workflow, so the trigger itself is proven only by the next scheduled run. What the run above does cover is the part that changed underneath it — the Node version the job will use.

**Not done here.** The issue also floated splitting by browser so wall-clock halves instead of doubling. That is a different change with its own trade-offs (three jobs, three reports to reconcile against one rolling issue), and it is not needed to stop running the suite twice.


jooyoungseo:
The timeout caveat is a fair catch, and it lands on my wording rather than on the condition. Fixed in `79689e8`.

The comment claimed `!= 'success'` means a timeout-cancelled run "still ships whatever it managed to write." That overstates it. `always()` only stops the step being *skipped*; the job is cancelled and then force-killed, so the upload runs only if `npm run e2e` exits inside the grace window. Not skipped is not the same as reached. The comment now says that, and the PR body is corrected too — I would rather it read as "worth having for the case where it does" than as a guarantee someone later relies on.

While there I also spelled out the `failure()` reasoning in the file itself, since that is the change a future reader is most likely to try and the two reasons are easy to conflate: `continue-on-error` masks the step's *conclusion* to `success` so `failure()` can never fire, and separately `always()` covers cancellation. Your description of the mechanism matches what I found.

On the one-logical-change nit — I agree with your read, and it was a deliberate call rather than an oversight. All three trace to the same issue's own open questions: #681 explicitly asks which workflow to keep, flags the differing Node pins as something that can diverge unnoticed, and says the cron minute is "worth deciding at the same time." Splitting them would have produced two commits that only make sense read against the third. The Node change in particular is not incidental tidying — it is the fix for the "one can go red while the other stays green" half of the issue, which survives the de-duplication only if the survivor is the version the PR gate actually runs.

Also confirmed your two independent checks match mine: `actionlint` is clean across the whole workflows directory (so the deletion left nothing dangling), and no reference to `e2e.yml` exists anywhere in the repo.

---

jooyoungseo:
Right on the upload condition, and it deserved to be stated rather than inherited. Recorded in `2820d23`.

`e2e.yml` uploaded on `always()`; this uploads only when the run is not a success. So a green scheduled run no longer produces a downloadable report — a real behaviour change, and one I made by copying `ci.yml`'s convention rather than by deciding it out loud.

It is intentional, for two reasons that only hold together after #684: `retain-on-failure` writes no traces for a passing run, so a green artifact is an html pass list and nothing more, and the report issue this workflow files already carries the reporter's full output for every run including green ones. Uploading it would store something already written down. Now said in the file, where someone would otherwise read the narrower gate as an oversight and "fix" it back.

Worth noting the second half of that only became true recently — before `retain-on-failure` landed, `always()` was a more defensible default than it is now.

Nothing to do on the other two. The artifact name was chosen for exactly the reason you give (browsing across workflows, not correctness), and the cron minute stands.

---

jooyoungseo:
Comment density: fair, and taken — `a14f400`. Twenty lines for a seven-line step, and looking at how it got there, it accreted one paragraph per review round rather than being written that way. Same four points in eleven lines; nothing dropped, it just stopped restating itself. I'd rather keep them inline than in a linked doc for the reason you give — this file's history *is* subtle gating bugs, and a doc link is one more thing to not read.

**Retention days:** the premise does not hold any more, which is worth spelling out since it changed inside this PR. Artifacts accumulate every 2 days only if every run uploads, and after the failure-only gate they upload only when the suite is red. On a green cadence that is zero artifacts, and a red one produces a report someone is about to look at. If the suite were failing often enough for retention to matter, retention would not be the problem. Adding `retention-days` would be a knob for a rate that no longer exists.

**PR description length:** true, and the diff is genuinely small. I would rather over-explain a two-line gate that three review rounds have now each probed a different corner of than leave the next person to re-derive it.

One thing in your verification worth confirming back, since you found it independently: the root `playwright.config.ts` with `trace: 'on'` really is unused by `npm run e2e`, which loads `e2e_tests/config/test-config.ts`. That stale second config was flagged on #684 as well. It is out of scope here, but it is the third time it has come up — I will file it rather than let it keep surfacing in reviews.

---

jooyoungseo:
Followed up on the stale root config and it is worse than stale — filed as #687.

It does not just sit unused. `npm run e2e:ui` and `npm run e2e:debug` do not pass `--config`, so they load it, and it throws before running anything:

```
ReferenceError: __dirname is not defined in ES module scope
    at file:///home/user/maidr/playwright.config.ts:21:22
```

The package is `"type": "module"`. `e2e_tests/config/test-config.ts` hit exactly this and derives its directory from `import.meta.url` with a comment explaining why; the root copy never got the same fix. It also points `testDir` at `test/e2e/specs`, which does not exist — the specs are in `e2e_tests/specs/`. So the documented interactive path in `.claude/rules/testing.md` is one nobody can currently follow.

`npm run e2e` and both workflows pass `--config` explicitly, so the suite and this PR are unaffected.

Left out of here deliberately: #686 is about the duplicated scheduled workflows, and the fix for that one is a `package.json` change plus a deletion, which has nothing to do with this diff.

---

jooyoungseo:
Nothing to change from this round. Two confirmations and one repeat.

**Squash-merge:** it is the repo default, verified rather than assumed — `main` has zero merge commits, and every recent entry is a single squashed commit carrying its PR number, including #684 which was 21 commits on the branch. So the three `docs(ci)` commits collapse to one on merge. Worth saying explicitly though: they exist because each was a correction a review round earned, and I would rather that be visible on the branch than folded in silently as I went.

**Early-step failure:** your reasoning matches mine and matches the intent. If `npm ci` or the browser install fails, the e2e step never runs, `steps.e2e.outcome` is empty, `!= 'success'` is satisfied, and the upload step runs and finds nothing — no-op via `if-no-files-found: ignore`. That is the behaviour I want. The alternative gate (`== 'failure'`) would skip it, which is fine in that specific case but wrong for the timeout case, and I would rather have one gate that is right about the case that matters than two that are each right about half.

**`retention-days`:** raised on the previous round too, and the answer is the same — it would be a knob for a rate that no longer exists. Artifacts accumulate every two days only if every run uploads; after the failure-only gate they upload only when the suite is red, so a green cadence produces none.

`actionlint` is clean on the changed file and on the whole `.github/workflows/` directory, including after the condensing commit — noting since you flagged you could not run it in your environment.

---

jooyoungseo:
Closing a gap in my own verification before this merges.

The Node change is the one part of this PR with a failure mode nothing here exercises, and I had not actually checked it. `lts/*` resolves to **24.18.1** today, while every three-browser run quoted in this thread — the 308×3 on #684 and everything since — ran on **22.22.2**, which is this container's default. `ci.yml` proves chromium on `lts/*`, because that job already uses it. Firefox and WebKit on Node 24 were proven by nothing.

Low risk, since Playwright ships its own browser binaries and Node mostly affects the runner. But low is not verified, and the failure mode is exactly what this thread has been clearing up: the scheduled suite goes red in two days and files a report issue.

So I ran it, reproducing what CI does — `npm ci`, `npm run build`, then each project — on Node 24.18.1 at `a14f400`, clean tree before and after:

| step | result |
|---|---|
| `npm ci` | exit 0 |
| `npm run build` | exit 0 |
| e2e — chromium | 308 passed (4.3m) |
| e2e — firefox | 308 passed (5.2m) |
| e2e — webkit | 308 passed (4.9m) |

Same 308 per project as on Node 22, exit 0 everywhere, no retries.

Had firefox or WebKit been unhappy on 24, the honest fix would have been to keep the `'22'` pin and say why — a version bump nobody has run is not worth the consistency it buys. As it is, the bump is now backed by a run rather than by the argument that `ci.yml` does it too.

---

PR #688, opened 2026-07-31
test(e2e): point the interactive e2e scripts at the real config

### LINKED ISSUE #687: test(e2e): npm run e2e:ui and e2e:debug crash on the stale root playwright.config.ts
## The bug

`npm run e2e:ui` and `npm run e2e:debug` do not start. Both fail before running anything:

```
$ npx playwright test --config=playwright.config.ts --list
ReferenceError: __dirname is not defined in ES module scope
    at file:///home/user/maidr/playwright.config.ts:21:22
    at loadUserConfig (node_modules/playwright/lib/common/configLoader.js:107:46)
```

## Why they reach it

The scripts in `package.json` disagree about the config:

```json
"e2e":       "npx playwright test --config=e2e_tests/config/test-config.ts",
"e2e:ui":    "npx playwright test --ui",
"e2e:debug": "npx playwright test --debug",
```

Only `e2e` passes `--config`. Without it Playwright picks up the root `playwright.config.ts`, which is the stale one.

## Two things wrong with the root config

1. **It cannot load.** It calls `__dirname` twice, and the package is `"type": "module"`. `e2e_tests/config/test-config.ts` hit this and solved it — it derives the directory from `import.meta.url` and carries a comment saying why. The root copy never got the same treatment.

2. **It points at a directory that does not exist.** `testDir: path.join(__dirname, 'test/e2e/specs')`. The specs live in `e2e_tests/specs/`. Even with the ESM problem fixed it would collect nothing.

It also sets `trace: 'on'`, which contradicts the `trace: 'retain-on-failure'` #684 deliberately chose — more evidence nothing reads it.

## Impact

`.claude/rules/testing.md` documents `npm run e2e:ui` as the interactive way to run the suite, so the documented path is one nobody can currently follow. The suite itself is unaffected: `npm run e2e` and both CI workflows pass `--config` explicitly and use the real config.

## Suggested resolution

Delete the root `playwright.config.ts` and add `--config=e2e_tests/config/test-config.ts` to `e2e:ui` and `e2e:debug`, so all three scripts load the one config that works. Worth checking `e2e:install` and any editor/extension integration at the same time, since a Playwright VS Code extension will also look for a root config by default.

## Notes

Surfaced three times in review now — on #684 and twice on #686 — as "the root config looks stale". Filing it so it stops resurfacing. Out of scope for #686, which is about the duplicated scheduled workflows.

Refs #686, #684

jooyoungseo:
# Pull Request

## Description

`npm run e2e:ui` and `npm run e2e:debug` do not start. Neither passes `--config`, so Playwright falls back to the root `playwright.config.ts`, which throws before collecting anything:

```
ReferenceError: __dirname is not defined in ES module scope
    at file:///home/user/maidr/playwright.config.ts:21:22
    at loadUserConfig (node_modules/playwright/lib/common/configLoader.js:107:46)
```

`.claude/rules/testing.md` documents `npm run e2e:ui` as the interactive way to run the suite, so the documented path is one nobody can currently follow.

## Related Issues

Closes #687. Surfaced during review of #686; follows #684.

### Three things wrong with the old root config

1. **It cannot load.** It calls `__dirname` twice and the package is `"type": "module"`. `e2e_tests/config/test-config.ts` hit exactly this and solved it — it derives the directory from `import.meta.url` and carries a comment saying why. The root copy never got the same treatment.

2. **Its `testDir` has never existed.** It points at `test/e2e/specs`. `git log --all -- 'test/e2e'` returns nothing: that path has never been in this repository. The specs have always been in `e2e_tests/specs/`. Even with the ESM problem fixed it would have collected zero tests.

3. **It sets `trace: 'on'`**, contradicting the `trace: 'retain-on-failure'` that #684 chose deliberately — more evidence nothing reads it.

It arrived in #549 and has been broken since. Nothing noticed because `npm run e2e` and both CI workflows pass `--config` explicitly.

### The fix

Give `e2e:ui` and `e2e:debug` the `--config` that `e2e` already has:

```diff
  "e2e":       "npx playwright test --config=e2e_tests/config/test-config.ts",
- "e2e:ui":    "npx playwright test --ui",
- "e2e:debug": "npx playwright test --debug",
+ "e2e:ui":    "npx playwright test --config=e2e_tests/config/test-config.ts --ui",
+ "e2e:debug": "npx playwright test --config=e2e_tests/config/test-config.ts --debug",
```

…and replace the root config with a re-export of the real one:

```ts
export { default } from './e2e_tests/config/test-config';
```

**The re-export is the point.** Repairing the old file would have meant maintaining a second set of settings, which is exactly how it drifted into a `testDir` that never existed. A file holding one line and a comment cannot drift that way. It exists so the paths that look for a config at the repository root resolve to the same one the npm scripts use. Deleting it outright (the first revision of this PR) left those collecting zero tests, silently.

### Guarding the file that caused this

`test/scripts/playwrightRootConfig.test.ts` pins the root config as a re-export: it must re-export the real config, must declare none of the seven top-level settings the deleted copy carried, and must stay one line of code.

Asserted on the file's text, not by importing both configs and comparing them — that stronger check is not available here, and I confirmed it rather than assumed it:

```
● Test suite failed to run
  SyntaxError: Cannot use 'import.meta' outside a module
```

`test-config.ts` reads `import.meta.url` and ts-jest transpiles to CommonJS, where that is a syntax error.

Verified against the real regression: restoring the exact pre-fix `playwright.config.ts` fails all 9 cases; the re-export passes all 9. Had this existed in #549, the drift would have been caught the day it landed.

### The references CI caught

Deleting the file broke `test/scripts/claudeRuleGlobs.test.ts`, which asserts every glob in the generated instruction files matches at least one tracked file:

```
● rule globs › testing.instructions.md: "playwright.config.ts"
  should match at least one tracked file
```

Both test-runner agent docs also named the deleted config on one line and the real one two lines down; collapsed to a single `Config:` line. `.github/agents/test-runner.agent.md` was additionally still claiming **"2 retries"**, which #684 removed — `sync:copilot` mirrors `CLAUDE.md` and `.claude/rules/` but not `.claude/agents/`, so that copy is hand-maintained and had been drifting since then with nothing to catch it.

**Why the local run missed it.** The glob test matches with `git ls-files`, which reads the index, and the deletion was unstaged when I first ran `npm test` — so the local suite saw a tree where the file still existed. `git add -A` before running is what reproduces CI here.

## Checklist

- [x] I have updated the documentation, if applicable. — both test-runner agent docs and `.claude/rules/testing.md`; `.github/instructions/testing.instructions.md` regenerated via `npm run sync:copilot`.
- [x] I have added appropriate unit tests, if applicable. — `test/scripts/playwrightRootConfig.test.ts`, verified failing against the config that caused this issue.

## Additional Notes

The fix is in the scripts and the config resolution, so the proof is running them:

| | before | after |
|---|---|---|
| `npm run e2e:ui` | `ReferenceError: __dirname is not defined` | loads config, reaches browser launch |
| `npm run e2e:debug` | same crash, same line | `Running 924 tests using 1 worker` |
| bare `npx playwright test --list` | same crash | `924 tests in 16 files` |

924 is 308 × 3 browsers — the real suite. `e2e:ui` gets no further than launching its browser here only because this container has no display; the point is that it now gets past config loading, which is where it used to die.

Everything else, re-run with the changes staged:

| check | result |
|---|---|
| `npm run build` | exit 0, all 12 builds complete |
| `npm test` | 1300 passed, 102 suites |
| `sync-copilot --check` | in sync, 11 files |
| guard vs. pre-fix config / vs. re-export | 9 failed / 9 passed |

Test counts moved twice during this branch and both moves are the same two cases: dropping the `playwright.config.ts` glob took its two parametrized checks with it (1291 → 1289), restoring the glob brought them back (→ 1291), and the new guard adds nine more (→ 1300).

**One claim here is inference, not verification.** The re-export is stated to restore discovery for editor integrations such as the Playwright VS Code extension. That follows from the extension's documented behaviour of looking for a config at the workspace root, but I have not run VS Code — the part I did verify is the bare `npx playwright test`, which went from 0 tests to 924.


jooyoungseo:
You were right, and CI failed on exactly it while I was reading your comment. Fixed in `9b829e9`.

```
FAIL test/scripts/claudeRuleGlobs.test.ts
  ● rule globs › testing.instructions.md: "playwright.config.ts"
    should match at least one tracked file
```

There is a test asserting every glob in the generated instruction files matches at least one tracked file — `test/scripts/claudeRuleGlobs.test.ts`, written for exactly this failure mode ("a typo'd path that matches nothing anywhere"). Deleting the file while leaving the glob is that failure mode.

**Why my local run missed it, since that matters more than the fix.** I ran `npm test` before committing and got 1291 passing. The test matches with `git ls-files`, which reads the *index* — and my deletion was still unstaged, because a `git stash`/`pop` in an unrelated before/after experiment had unstaged it. So the local run was against a tree where the file still existed as far as git was concerned. `git add -A` first is what reproduces CI.

### What changed

- **`.claude/rules/testing.md`** — dropped the `"playwright.config.ts"` glob rather than repointing it. `e2e_tests/**` in the same list already covers the real config, so a repointed entry would match nothing new.
- **`.claude/agents/test-runner.md`** — it named the deleted config on one line (`Config:`) and the real one two lines down (`Test config:`). Collapsed to a single `Config:` line, parallel with the Jest block above it, plus a note that all three e2e scripts pass `--config`.
- **`.github/agents/test-runner.agent.md`** — same duplicate, and **also still claiming "2 retries"**, which #684 removed. Worth flagging: `sync:copilot` mirrors `CLAUDE.md` and `.claude/rules/` but **not** `.claude/agents/`, so that copy is hand-maintained and had been drifting since #684 with nothing to catch it. The glob test guards rule globs, not agent prose.

### Verification, staged this time

| check | result |
|---|---|
| `sync-copilot --check` | in sync, 11 files |
| `npm test` | 1289 passed, 101 suites |

1289 rather than 1291 because the dropped glob carried two parametrized cases of its own — not two tests lost.

---

jooyoungseo:
Added the shim — `5b0abc7`. Closes the one residual risk this PR had been carrying.

```ts
export { default } from './e2e_tests/config/test-config';
```

The root file holds no settings, only that line and a comment saying to edit the real config. That is what makes it safe rather than a repeat of the problem: the file it replaces was a full second copy of the settings, which is precisely how it drifted into a `testDir` that never existed and a `trace` mode contradicting the suite's. There is nothing here to drift.

### It falsified two things I had just written

Worth spelling out, because the shim is one line and the ripple was not:

- Both test-runner agent docs said **"there is no root `playwright.config.ts`"**. Now untrue. Corrected to describe the re-export.
- I had dropped the `"playwright.config.ts"` glob from `.claude/rules/testing.md` on the grounds that the file no longer existed. It exists again, so it is restored.

The glob test count moving **56 → 58** is the check that the restored glob is live rather than silently ignored — those are its two parametrized cases coming back.

### Verification, staged

| check | before shim | after |
|---|---|---|
| bare `npx playwright test --list` | 0 tests, module noise | **924 tests in 16 files** |
| `npm run e2e:debug` | 924 tests | 924 tests |
| `claudeRuleGlobs.test.ts` | 56 passed | 58 passed |
| `npm test` | 1289 passed | 1291 passed, 101 suites |
| `sync-copilot --check` | in sync | in sync, 11 files |
| `npm run lint:fix` / `type-check` | clean | clean |

Back to 1291 for the same reason the count dropped to 1289 earlier — the glob's two cases, not tests appearing or disappearing.

---

jooyoungseo:
Took the test suggestion — `f46d9f8` — though not in the form you proposed, because that form does not run here.

**The deep-equal is not available.** I probed it rather than assumed:

```
● Test suite failed to run
  SyntaxError: Cannot use 'import.meta' outside a module
```

`test-config.ts` reads `import.meta.url` (that is the fix for the ESM problem this whole PR is about), and ts-jest transpiles to CommonJS, where `import.meta` is a syntax error. Importing either config into a Jest test fails at load.

**So the assertions are on the file's text**, which is where the regression is actually visible. `test/scripts/playwrightRootConfig.test.ts` pins three things: the root file re-exports the real config, it declares none of the settings the deleted copy carried (`testDir`, `testMatch`, `projects`, `timeout`, `reporter`, `use` — named individually so a failure says which one came back), and it stays one line of code so settings cannot hide in it.

Weaker than a semantic comparison, and the file says so rather than implying otherwise. But it catches the thing that happened: someone replacing the re-export with settings of their own.

**Verified against the real regression, not a mock.** Restoring the exact pre-fix `playwright.config.ts` and re-running:

```
✕ should re-export the real config
✕ should not declare its own testDir
✕ should not declare its own testMatch
✕ should not declare its own projects
✕ should not declare its own timeout
✕ should not declare its own reporter
✕ should not declare its own use
✕ should stay small enough that settings could not hide in it
Tests: 8 failed, 8 total
```

All 8 pass against the re-export. Had this existed in #549, the drift would have been caught the day it landed instead of surviving to #687.

| check | result |
|---|---|
| `npm test` | 1299 passed, 102 suites |

On the commit-structure note: leaving them. Squash-merge is this repo's default — `main` has zero merge commits and one squashed entry per PR — so the history collapses regardless, and each commit here is a correction a review round or CI actually earned. Folding them in silently would hide that the glob was dropped and restored for a reason.

Also confirming your `tsconfig` observation, since it is load-bearing for the shim: no `include` filter, so `playwright.config.ts` is type-checked by `npm run type-check` even as a one-line re-export. A signature mismatch with `test-config.ts` would surface there.

---

jooyoungseo:
Taken — `71972ae`. `expect` was a genuine gap, not just tidiness.

Rather than adding the one key and calling it done, I checked whether the list was now actually complete: extracted the old config's top-level keys and diffed them against what the guard names.

```
old config          guard names
----------          -----------
expect              expect
projects            projects
reporter            reporter
testDir             testDir
testMatch           testMatch
timeout             timeout
use                 use
```

Exact match on all seven. `expect` was the only one missing.

You are right that the risk was small — the one-line-of-code assertion is a backstop that catches anything, including a lone `expect` block. What the named cases buy is the failure message: "should not declare its own expect" says what came back, where the line-count failure just says the file grew.

| check | result |
|---|---|
| guard vs the re-export | 9 passed |
| guard vs the pre-fix config | 9 failed |
| `npm test` | 1300 passed, 102 suites |
| `npm run lint:fix` / `type-check` | clean |

On not being able to run the suite in your sandbox — noted, and the reported numbers are from runs with the changes staged, which matters here: an earlier round of this PR reported a false green precisely because `git ls-files`-based assertions read the index and my deletion was unstaged.

---

PR #690, opened 2026-07-31
build: enable noImplicitOverride

### LINKED ISSUE #682: build: enable noImplicitOverride so shadowed methods are visible
## Why

`tsconfig.json` has `strict: true` but not `noImplicitOverride`. #679 found a bug that flag would have made obvious:

```ts
// e2e_tests/page-objects/base-page.ts
protected async isModeActive(notificationSelector: string, mode: string, modeMessages: Record): Promise

// e2e_tests/page-objects/plots/barplot-page.ts — silently overrode the above
protected async isModeActive(mode: string, expectedMessage: string): Promise
```

The subclass kept a snapshot-reading copy while the base class was taught to wait, so all four barplot mode-toggle tests quietly resolved to the un-fixed version. It was invisible at the call sites *and* at the base method, and was only caught by a reviewer reading the diff.

## What the flag does and does not do

> **Corrected while implementing this (see #690).** The original text below said the value is "disclosure, not prevention". That is half right, and the half it misses is the more valuable one.

`noImplicitOverride` would **not** have rejected #679's signature. Method parameters are bivariant and a shorter parameter list is assignable to a longer one, so the exact shape still compiles with the keyword present:

```ts
class Base {
  protected async isModeActive(sel: string, mode: string, msgs: Record): Promise
}
class Sub extends Base {
  protected override async isModeActive(mode: string, expected: string): Promise  // compiles
}
```

Verified with `strict: true, noImplicitOverride: true` — exit 0.

**But it does prevent a different failure, and that one is not merely cosmetic.** Once the keyword is present, removing or renaming a base member turns every stray override into a compile error. Measured on this repo by renaming `AbstractPlot.dispose`:

```
src/model/abstract.ts(391,19): error TS4113: This member cannot have an 'override'
  modifier because it is not declared in the base class 'AbstractPlot<TraceState>'
src/model/plot.ts(226,19): error TS4113: …
src/model/plot.ts(417,19): error TS4113: …
```

Without the flag those three silently become new methods that nothing calls — the base class stops dispatching to them and no tool says a word. So the value is:

- **disclosure** — the keyword makes shadowing visible to a reader, which is what was missing in #679;
- **prevention** — orphaned overrides become compile errors rather than dead code.

## Cost

Measured on `main` — 162 errors, all mechanical:

| area | errors |
|---|---|
| `e2e_tests/page-objects/plots` | 80 |
| `src/model` | 64 |
| `src/state/viewModel` | 12 |
| `test/model` | 3 |
| `src/service` | 3 |
| **total** | **162** |

161 are `TS4114` (missing `override`); one is `TS4116` (`src/model/abstract.ts:631`, an abstract-method override). I checked specifically for a non-mechanical error in case the flag had caught a real latent bug elsewhere — it had not.

So this is a wide but shallow sweep touching `src/`, `test/` and `e2e_tests/`. It was left out of #679 because that PR's clean property — no product code touched — is something several reviews there leaned on when assessing risk.

## Done when

- `noImplicitOverride: true` is in `tsconfig.json`.
- `npm run type-check` is clean.
- The full suite still passes.

Refs #679. Implemented in #690.

jooyoungseo:
# Pull Request

## Description

`tsconfig.json` has `strict: true` but not `noImplicitOverride`, so a method that overrides a base member looks identical to one that does not.

#679 lost time to exactly that. A subclass kept a snapshot-reading `isModeActive` while the base class was taught to wait, so every barplot mode toggle quietly resolved to the un-fixed copy — invisible at the call sites *and* at the base method, and caught only by a reviewer reading the diff.

## Related Issues

Closes #682. Follows #679, #684.

> **Merge order — resolved.** #689 has merged and this branch is rebased onto it (`35001a4`). The one predicted conflict, the `override` on the `startAutoplay` that #689 deletes, is resolved in `main`'s favour. Every number below was re-measured after the rebase.

## Changes Made

One flag, and the 161 members that now need the keyword:

| area | members |
|---|---|
| `e2e_tests/page-objects` | 79 |
| `src/model` | 64 |
| `src/state/viewModel` | 12 |
| `src/service` | 3 |
| `test/model` | 3 |

**Nothing else changes.** Two checks rather than an assurance:

- **Every one of the 161 changed lines differs from its original by exactly the inserted `override `.** Compared line by line, not by eye — 161 added, 161 removed, 161 matching after removing the keyword.
- **The emitted bundle is unchanged.** Built this branch and `main` and hashed the output: all **34** js/css files match byte for byte. `override` is erased at compile time, so that is the expected result — and now a verified one rather than an assumption.

### The issue undersold the flag

#682 said the value is "disclosure, not prevention", on the grounds that bivariant parameters mean the keyword would not have rejected #679's incompatible signature. That part is right, but it is not the whole story, and I checked rather than assumed:

```
$ # rename AbstractPlot.dispose, then type-check
src/model/abstract.ts(391,19): error TS4113: This member cannot have an 'override'
  modifier because it is not declared in the base class 'AbstractPlot<TraceState>'
src/model/plot.ts(226,19): error TS4113: …
src/model/plot.ts(417,19): error TS4113: …
```

Renaming a base member turns its now-orphaned overrides into compile errors. Without the flag they silently become new methods nothing calls. That is prevention, of a real failure mode. The issue body has been corrected.

### How it was applied

Mechanically, from the compiler's own diagnostics — 161 hand edits would be 161 chances to fumble one.

Worth recording one trap, because the obvious implementation is wrong: the reported column points at the *member name*, which sits after `async`, `readonly`, `get` and `set`. TypeScript wants `override` before those, so inserting at the column produces `public async override activateMaidr()`, which does not parse. The script backs up over them:

```python
while True:
    head = text[:at].rstrip(' ')
    for word in ('async', 'readonly', 'get', 'set'):
        if head.endswith(word) and not head[-len(word) - 1].isalnum():
            at = len(head) - len(word)
            break
    else:
        break
```

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas. — no new code; the change is a keyword.
- [ ] I have updated the documentation, if applicable. — no documented behaviour changed. `.claude/rules/typescript.md` describes conventions, not compiler flags.
- [ ] I have added appropriate unit tests, if applicable. — the compiler is the test; `npm run type-check` fails without the keywords.

## Additional Notes

All measured on the rebased head `34cad8a`, against `main` at `35001a4`:

| check | result |
|---|---|
| `npm run lint:fix` | clean, no reformatting |
| `npm run build` | exit 0 |
| bundle vs `main` | **34/34 files byte-identical** |
| `npm test` | 1300 passed, 102 suites |
| e2e — chromium | 308 passed (3.6m) |
| e2e — firefox | 308 passed (4.3m) |
| e2e — webkit | 308 passed (3.8m) |

Run one project at a time, clean tree verified before and after, exit 0 on all three, no retries.

**This is the first PR in this thread to touch `src/`.** #674, #676, #679, #684, #686, #688 and #689 all leaned on "no product code touched" when reviewers assessed risk, so that property is gone here and worth saying plainly rather than leaving a reviewer to notice. What replaces it is the bundle comparison above: the risk is not that this changes behaviour — it provably does not — but that a mechanical sweep across 41 files is tedious to review. The line-by-line check is there so it does not have to be read that way.

**Why it could not be split.** `noImplicitOverride` is a single `tsconfig` flag. Turning it on requires all 161 keywords at once; there is no "e2e first, `src/` later" version of this change.

### DIFF
--- a/e2e_tests/page-objects/plots/barplot-page.ts
+++ b/e2e_tests/page-objects/plots/barplot-page.ts
@@ -18,7 +18,7 @@ export class BarPlotPage extends BasePage {
   /**
    * Selectors for various UI elements
    */
-  protected readonly selectors: {
+  protected override readonly selectors: {
     notification: string;
     info: string;
     speedIndicator: string;
@@ -75,7 +75,7 @@ export class BarPlotPage extends BasePage {
    * @returns Promise resolving when the condition is met
    * @throws BarPlotError if timeout is reached before the condition is met
    */
-  public async waitForElementContent(
+  public override async waitForElementContent(
     selector: string,
     expectedContent: string,
     options: { timeout?: number; pollInterval?: number } = {},
@@ -116,7 +116,7 @@ export class BarPlotPage extends BasePage {
    * @returns Promise resolving when verification is complete
    * @throws BarPlotError if plot is not loaded correctly
    */
-  public async verifyPlotLoaded(): Promise<void> {
+  public override async verifyPlotLoaded(): Promise<void> {
     try {
       await this.page.waitForLoadState('domcontentloaded');
       await expect(this.page.locator(this.selectors.svg)).toBeVisible({
@@ -132,7 +132,7 @@ export class BarPlotPage extends BasePage {
    * @returns Promise resolving when MAIDR is activated
    * @throws BarPlotError if MAIDR cannot be activated
    */
-  public async activateMaidr(): Promise<void> {
+  public override async activateMaidr(): Promise<void> {
     try {
       await this.verifyPlotLoaded();
       await this.pressKey(TestConstants.TAB_KEY, 'activate maidr');
@@ -147,7 +147,7 @@ export class BarPlotPage extends BasePage {
    * @returns Promise resolving when MAIDR is activated via click
    * @throws BarPlotError if MAIDR cannot be activated by clicking
    */
-  public async activateMaidrOnClick(): Promise<void> {
+  public override async activateMaidrOnClick(): Promise<void> {
     try {
       await this.verifyPlotLoaded();
       await this.page.click(this.selectors.svg);
@@ -177,7 +177,7 @@ export class BarPlotPage extends BasePage {
    * @returns Promise resolving to the instruction text
    * @throws BarPlotError if instruction text cannot be retrieved
    */
-  public async getInstructionText(): Promise<string> {
+  public override async getInstructionText(): Promise<string> {
     return this.getNotificationText();
   }
 
@@ -313,7 +313,7 @@ export class BarPlotPage extends BasePage {
    * @returns Promise resolving to the current speed value
    * @throws BarPlotError if speed cannot be retrieved
    */
-  public async getPlaybackSpeed(): Promise<number> {
+  public override async getPlaybackSpeed(): Promise<number> {
     try {
       const speedText = await this.getElementText(this.selectors.speedIndicator);
       return Number.parseFloat(speedText);
@@ -336,7 +336,7 @@ export class BarPlotPage extends BasePage {
    * @returns Promise resolving to the current data point information
    * @throws BarPlotError if data point information cannot be retrieved
    */
-  public async getCurrentDataPointInfo(): Promise<string> {
+  public override async getCurrentDataPointInfo(): Promise<string> {
     return this.getInfoText();
   }
 
--- a/e2e_tests/page-objects/plots/boxplotHorizontal-page.ts
+++ b/e2e_tests/page-objects/plots/boxplotHorizontal-page.ts
@@ -11,7 +11,7 @@ export class BoxplotHorizontalPage extends BasePage {
   /**
    * Selectors for various UI elements
    */
-  protected readonly selectors = {
+  protected override readonly selectors = {
     notification: `#${TestConstants.MAIDR_NOTIFICATION_CONTAINER} ${TestConstants.PARAGRAPH}`,
     info: `#${TestConstants.MAIDR_INFO_CONTAINER} ${TestConstants.PARAGRAPH}`,
     speedIndicator: `#${TestConstants.MAIDR_SPEED_INDICATOR}${TestConstants.BOXPLOT_HORIZONTAL_ID}`,
@@ -50,7 +50,7 @@ export class BoxplotHorizontalPage extends BasePage {
    * @returns Promise resolving when MAIDR is activated
    * @throws BoxplotHorizontalError if MAIDR cannot be activated
    */
-  public async activateMaidr(): Promise<void> {
+  public override async activateMaidr(): Promise<void> {
     try {
       await super.activateMaidr(this.selectors.svg, TestConstants.BOXPLOT_HORIZONTAL_ID);
     } catch (error) {
@@ -63,7 +63,7 @@ export class BoxplotHorizontalPage extends BasePage {
    * @returns Promise resolving when MAIDR is activated via click
    * @throws BoxplotHorizontalError if MAIDR cannot be activated by clicking
    */
-  public async activateMaidrOnClick(): Promise<void> {
+  public override async activateMaidrOnClick(): Promise<void> {
     try {
       await super.activateMaidrOnClick(this.selectors.svg, TestConstants.BOXPLOT_HORIZONTAL_ID);
     } catch (error) {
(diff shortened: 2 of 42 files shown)


jooyoungseo:
Thorough review — the base-member spot checks are the part I could not do for myself as cheaply, particularly confirming `AbstractPlot.dispose` is the true root of that chain rather than a coincidental same-named method. Two things to settle.

### #689 has not merged

> since #689 already merged into `main` (per `git log`), this branch will need the rebase mentioned in the description

It has not. Checked rather than taken on trust, because acting on it would mean an unnecessary rebase:

```
$ git log origin/main --oneline -1
dfbe863 test(e2e): point the interactive e2e scripts at the real config (#688)

$ git merge-base --is-ancestor 577d2ef origin/main && echo merged || echo "not merged"
not merged
```

`origin/main` is still at #688, and this PR's own base SHA is that same `dfbe863`. #689 is open with all 13 checks green.

So **the merge-order caveat still stands as written**: #689 first, then I rebase this one. Worth correcting rather than letting it pass, because the natural reading — "the caveat is already resolved" — would invite merging this first, which is the ordering that actually creates the conflict.

### The issue body is corrected

Done, in #682. It now separates the two halves rather than collapsing them:

- **disclosure** — the keyword makes shadowing visible, which is what #679 lacked;
- **prevention** — removing or renaming a base member turns stray overrides into `TS4113` errors instead of dead code nothing calls.

The original text asserted only the first and dismissed the second, which was my error when filing it.

### On not re-running the checks

Understood, and no need — the two claims worth doubting are both mechanical and re-derivable from the diff you already read. If it helps, the bundle comparison was: build this branch, hash every emitted js/css, `git stash`, build `main`, hash again, compare. 34 files, identical. The three-browser suite is still running and I will post those numbers here rather than leave the table saying "running".

---

jooyoungseo:
Three-browser suite is done, so the table in the body no longer says "running".

| project | result |
|---|---|
| chromium | 308 passed (4.0m) |
| firefox | 308 passed (4.7m) |
| webkit | 308 passed (4.4m) |

Run one project at a time at `4d0722f`, clean tree verified before and after, exit 0 on all three, no retries. CI on this head is **13/13 green** as well.

That is the complete evidence set for "nothing changed but the keyword":

| claim | how it was checked |
|---|---|
| no line changed except the insertion | 162 added / 162 removed / 162 matching after stripping `override ` |
| no runtime change | built this branch and `main`, hashed output — 34/34 js/css byte-identical |
| nothing broke | jest 1300 / 102 suites, three-browser e2e 308 × 3, CI 13/13 |

Still gated on merge order: **#689 first, then I rebase this.** #689 is open and green; `origin/main` remains at `dfbe863`.

---

jooyoungseo:
Your one actionable point — "confirm that merge-order dependency is still accurate at merge time" — resolved itself while you were writing it. **#689 merged, and this is now rebased.** `4d0722f` → `34cad8a`, sitting on `35001a4`.

The conflict was the single line predicted: my branch re-adding `startAutoplay` with `override`, `main` having deleted it. Took `main`'s side.

### Re-verified after the rebase, not carried over

Every number in the body was measured against the *old* `main`, so quoting it now would be stale. Re-run against `35001a4`:

| check | result |
|---|---|
| diff vs `main` | **161 lines, every one a pure `override ` insertion** |
| bundle vs `main` | **34/34 js/css byte-identical** |
| `npm run lint:fix` | clean, no reformatting |
| `npm test` | 1300 passed, 102 suites |
| e2e — chromium | 308 passed (3.6m) |
| e2e — firefox | 308 passed (4.3m) |
| e2e — webkit | 308 passed (3.8m) |

161 rather than 162 because the 162nd was the keyword on the method #689 deleted. The conflict resolution was checked the same mechanical way rather than by eye: `barplot-page.ts` now differs from `main` by exactly 8 insertions, down from 9.

### A correction to how I was checking this

Worth recording, because I nearly got it wrong in the other direction. I had been verifying "has #689 landed?" with:

```
git merge-base --is-ancestor 577d2ef origin/main
```

**That is unreliable in this repo.** Merges here are squashed, so the branch commit never becomes an ancestor — the command answered "no" even after #689 was demonstrably on `main` as `35001a4`. Had I trusted only that, I would have concluded the rebase was not needed yet. The PR number in the commit subject is the thing to read.

That also means my earlier "#689 has not merged" reply to the previous round was right by luck of a second signal (`origin/main` was still `dfbe863`), not by the check I quoted.

The merge-order note in the body now says the rebase is done.

---

Issue #678, opened 2026-07-30
test: add an ESM-aware Jest project so the markdown pipeline can be tested end to end


jooyoungseo:
Raised three times in review on #675, so worth capturing with the feasibility work already done rather than leaving it to be re-derived.

## The gap

`test/util/markdownSanitize.test.ts` checks what the sanitisation schema *contains* — every element and attribute KaTeX emits is named, ARIA entries use hast property names, no key beyond `tagNames`/`attributes` is declared. What it cannot check is `rehype-sanitize` actually applying that schema, which is the crux of the fix in #675.

So these are currently verified by hand only:

- KaTeX's MathML survives sanitisation with its structure intact
- `aria-hidden` survives on the visual layer
- `javascript:`, `vbscript:` and `data:` URLs are stripped from links and images
- raw HTML in a response is escaped rather than parsed

## Why the suite can't do it today

`rehype-sanitize` and the rest of the unified stack are ESM-only. The Jest suite is CommonJS, and `tsconfig.json` has `allowJs: false`, so nothing under `node_modules` is transformed:

```
SyntaxError: Unexpected token 'export'
    at node_modules/rehype-katex/index.js:5
```

This is why the schema lives in `src/util/markdownSanitize.ts` with no runtime imports — `TypingEffect.tsx` imports `react-markdown`, and importing it into a test would fail the same way.

## It works — verified

A second Jest project loads the whole stack:

```js
// jest.esm.config.mjs
export default {
  testEnvironment: 'node',
  extensionsToTreatAsEsm: ['.ts'],
  testMatch: ['**/test/**/*.esm-test.ts'],
  transform: {
    '^.+\\.ts$': ['ts-jest', {
      useESM: true,
      tsconfig: { module: 'ESNext', moduleResolution: 'bundler' },
    }],
  },
  moduleNameMapper: { '^@util/(.*)$': '<rootDir>/src/util/$1' },
}
```

```
$ NODE_OPTIONS=--experimental-vm-modules npx jest -c jest.esm.config.mjs
PASS esm test/util/__probe.esm-test.ts
  ✓ loads the ESM markdown stack (22 ms)
```

The flag is safe for what already exists — the full suite (101 suites, 1291 tests) passes under `--experimental-vm-modules`.

## The decision this needs

`NODE_OPTIONS=… jest` is not valid on Windows and the repo has no `cross-env`, so wiring it into `npm test` means either:

1. adding `cross-env` as a devDependency, or
[...]

Issue #677, opened 2026-07-30
a11y: chat sanitiser destroys tables, headings, strikethrough and task lists


jooyoungseo:
Found while reviewing #675, which fixes the same defect for maths. The chat markdown allowlist does not name most of what the pipeline can produce, and `rehype-sanitize` drops a disallowed element while keeping its text — so the content survives as prose with its structure gone.

`TypingEffect` enables `remarkGfm`, so an AI response can contain all of this. None of it renders.

## Measured through the real pipeline

A table of chart values — the single most likely thing an LLM emits when describing a plot:

```markdown
| Bar | Value |
| --- | ----: |
| Jan | 45.2 |
| Feb | 51.8 |
```

before sanitisation:

```html
<table><thead><tr><th>Bar</th><th>Value</th></tr></thead>
<tbody><tr><td>Jan</td><td>45.2</td></tr>…
```

after:

```
BarValueJan45.2Feb51.8
```

No table, no headers, no row or cell boundaries. A screen reader user gets a run of digits with nothing binding them to a label, and none of the table navigation AT provides. In a library whose purpose is making data readable non-visually, this is the worst possible thing to lose.

The rest of the same survey:

| Markdown | Renders as | Lost |
|---|---|---|
| `## Summary` | `Summary` | `h1`–`h6` — heading navigation, a primary AT navigation mode |
| `~~gone~~` | `gone` | `del` — **meaning inverted**, a retraction reads as an assertion |
| `- [x] done` / `- [ ] todo` | `done` / `todo` | `input` — checked and unchecked are indistinguishable |
| `text[^1]` | partially | `sup` on the reference, `section` on the footnotes block |
| `---` | *(nothing)* | `hr` |

`del` and the task-list checkbox are the sharpest: both change what the sentence means rather than merely how it looks.

## Cause

Identical to #675. `src/util/markdownSanitize.ts` names what KaTeX emits — that part is now tested against KaTeX's real output — but the GFM and CommonMark side was never surveyed the same way. Missing from `tagNames`:

```
del  h1 h2 h3 h4 h5 h6  hr  input  section  sup
table  thead  tbody  tr  th  td
```

and the attributes those need: `align` on `th`/`td`, `type`/`checked`/`disabled` on `input`, `id` on the footnote anchors.

## Why not fixed in #675

It is a separate logical change, and the security decisions deserve their own review rather than arriving at the tail of a maths fix:

[...]

PR #695, opened 2026-08-01
fix(chat): keep tables, headings and task lists through sanitisation

jooyoungseo:
# Pull Request

## Description

The chat allowlist names what KaTeX emits — that side was surveyed in #675 — but the GFM and CommonMark side never was. `rehype-sanitize` drops a disallowed element **while keeping its text**, so none of this failed visibly. It flattened.

A table of chart values, the single most likely thing an LLM emits when describing a plot:

```markdown
| Bar | Value |
| --- | ----: |
| Jan | 45.2 |
| Feb | 51.8 |
```

reached the user as:

```
BarValueJan45.2Feb51.8
```

No headers, no row or cell boundaries, and none of the table navigation assistive technology provides — a run of digits with nothing binding them to a label. In a library whose purpose is making data readable non-visually, that is the worst thing on the list.

The rest of the same survey:

| Markdown | Rendered as | Lost |
|---|---|---|
| `## Summary` | `Summary` | `h1`–`h6`, and heading navigation with them |
| `~~gone~~` | `gone` | `del` — **meaning inverted**, a retraction reads as an assertion |
| `- [x] done` | `done` | `input` — checked and unchecked indistinguishable |
| `2.` `3.` | renumbered from 1 | `ol[start]` — the reader gets different numbers than were written |
| `text[^1]` | partial | `sup`, `section` |
| `---` | *(nothing)* | `hr` |

## Related Issues

Refs #677. Follows #675, which fixed the same class of defect for maths.

Filed from this PR: #696 (footnote anchors), #697 (`.sr-only` is not styled), #698 (`Link: [object Object]`). Blocked-on: #678.

## Changes Made

Adds the elements and attributes a survey of the **real pipeline** reports — `remark-parse` → `remark-gfm` → `remark-rehype`, walking the resulting hast tree rather than reasoning from the GFM spec.

Two things the issue did not list turned up in that survey:

- **`a[ariaLabel]` and `a[ariaDescribedBy]` are genuinely emitted.** #675 removed both from the global list on the reasoning that *"nothing in the pipeline emits any of them"*. That holds for maths and stops holding the moment footnotes are admitted: the backref's only text is its `aria-label` (`"Back to reference 1"`), so without it the link is unlabelled.
- **`h2[id]` and `section[dataFootnotes]`**, for the footnotes block.

Three security decisions, as the issue asked:

- **`input`** is admitted only as the disabled checkbox remark-gfm emits, with `type` **pinned to its value** rather than merely named. The tuple form matters: it replaces a disallowed value instead of dropping it, and an `input` whose `type` is dropped renders as a **text box**, so the bare name failed open in both directions.

  ```
  [['type','checkbox'], …]   type=hidden  ->  type="checkbox"
  ['type', …]                type=hidden  ->  type="hidden"
  ```

  `form`, `name` and `value` stay out: none is emitted, so allowing them could only ever admit something a response invented.
- **The footnote data attributes are named individually**, not admitted by a `data-*` wildcard. A test asserts no wildcard appears anywhere in the schema.
- **`id` relies on the default `clobberPrefix`**, which still applies because the schema declares only `tagNames` and `attributes` — the existing test pinning that is what keeps it true, and it is now load-bearing for three elements where it guarded none before.

Also fixes one defect a layer up: `TypingEffect`'s `a` override set `aria-label` unconditionally after the spread, so it discarded the one the pipeline had just been permitted to carry. The footnote backref announced as *"Link: ↩"* rather than *"Back to reference 1"* — the exact information admitting `ariaLabel` was meant to preserve. It falls back now, matching the `img` override beside it.

`MessageBubble.tsx` was checked as the issue suggested: it delegates to `TypingEffect`, so this one allowlist covers both.

## Checklist

### Verified end to end, not just against the schema

Checking the allowlist against itself would pass while the rendering stayed broken, so every claim here was run through the real pipeline with `hast-util-sanitize`:

```html
<table><thead><tr><th align="left">Bar</th><th align="right">Value</th></tr></thead>
<tbody><tr><td align="left">Jan</td><td align="right">45.2</td></tr></tbody></table>

<li class="task-list-item"><input type="checkbox" checked disabled> done</li>

ol  {"start":2}
a   {"href":"https://e.com","title":"the title"}
```

And the footnote references, resolved against the ids actually present after sanitisation:

```
aria-describedby  user-content-footnote-label   RESOLVES
href              user-content-fn-1             DANGLING
href              user-content-fnref-1          DANGLING
```

### Known limitation: footnote anchors dangle

The issue said the `id` decision was *"worth confirming rather than assuming"*. Confirmed, and it is a real trap: `mdast-util-to-hast` prefixes the ids **derived from user text** with `user-content-` and leaves its own literal `footnote-label` bare, then `hast-util-sanitize` prefixes everything in `clobber` again. So `href="#user-content-fn-1"` points at a target that became `user-content-user-content-fn-1`.

`aria-describedby` survives because it is in `clobber` alongside `id`, so both sides of that pair move together. `href` is not, so it cannot follow its target — that, rather than the double-prefixing itself, is the bug.

Not fixed here, and **#696 now records why the obvious fix does not work**: a per-message `clobberPrefix` leaves `footnote-label` colliding, because the library hardcodes that id and documents that it always does. Full de-duplication needs a post-processing pass, which is a different change from this one.

Leaving `id` out entirely would be worse, not better: `aria-describedby` would then point at nothing, and the anchors would be dead either way.

### On the tests

Each new case was verified to fail against the defect it names — removing `table`, `del`, `input`, `ol[start]` or `br` from the allowlist, adding `name`/`value` to the checkbox, or loosening the `type` pin each turns the relevant case red.

The corpus in the test file is a **transcript** of the pipeline survey rather than a live one: the whole remark/rehype chain is ESM and this Jest suite is CommonJS, so it cannot run here. Review proved the sharp edge of that — `ol[start]`, `a[title]`, `img[title]` and `br` were all missed on the first pass, every one of them because the corpus behind the transcript was too narrow to emit them. That limitation is now recorded next to the transcript, and #678 has a ranked list of what its live survey should pin first.

| check | result |
|---|---|
| `npm test` | 1326 passed, 103 suites |
| `npm run build` | exit 0 |

---

### DIFF
--- a/src/ui/components/TypingEffect.tsx
+++ b/src/ui/components/TypingEffect.tsx
@@ -155,8 +155,14 @@ export const TypingEffect: React.FC<TypingEffectProps> = memo(({ text, isUser, m
             pre: ({ node, ...props }) => (
               <pre {...props} role="text" aria-label="Code block" />
             ),
+            // Fall back rather than overwrite, the way the `img` override
+            // below already does. A prop after a spread wins unconditionally,
+            // so this used to discard any `aria-label` the pipeline had set —
+            // and the one case that has one is the footnote backref, whose
+            // visible text is a bare return arrow. It announced as "Link: ↩"
+            // in place of "Back to reference 1".
             a: ({ node, ...props }) => (
-              <a {...props} aria-label={`Link: ${props.children}`} />
+              <a {...props} aria-label={props['aria-label'] ?? `Link: ${props.children}`} />
             ),
             img: ({ node, ...props }) => (
               <img {...props} alt={props.alt || 'Image in message'} />
--- a/src/util/markdownSanitize.ts
+++ b/src/util/markdownSanitize.ts
@@ -179,13 +179,18 @@ export function createChatSanitizeSchema(): SanitizeSchema {
       // every dataset key a future plugin invents.
       'a': [
         'href',
+        // `[text](url "title")`. Text only — no protocol, nothing to filter.
+        'title',
         'id',
         'ariaLabel',
         'ariaDescribedBy',
         'dataFootnoteRef',
         'dataFootnoteBackref',
       ],
-      'img': ['src', 'alt'],
+      'img': ['src', 'alt', 'title'],
+      // `start` only appears when a numbered list does not begin at 1, so
+      // without it a list written "2." "3." silently renumbers from 1.
+      'ol': ['start'],
       // The footnotes block and the anchors its links resolve against.
       'section': ['dataFootnotes'],
       'h2': ['id'],
--- a/src/util/markdownSanitize.ts
+++ b/src/util/markdownSanitize.ts
@@ -202,7 +202,13 @@ export function createChatSanitizeSchema(): SanitizeSchema {
       // and `value` are deliberately absent: none is emitted, and together
       // they are what would turn a rendered response into a submittable
       // control.
-      'input': ['type', 'checked', 'disabled'],
+      //
+      // `type` is pinned to the one value rather than merely allowed. The
+      // tuple form replaces a disallowed value instead of dropping it, which
+      // is the safer failure here: a dropped `type` leaves an `input` that
+      // defaults to a text box, where a replaced one stays the inert checkbox
+      // this entry exists to admit.
+      'input': [['type', 'checkbox'], 'checked', 'disabled'],
       // Column alignment from `:--` / `--:`. Carries no meaning for a screen
       // reader but is the difference between a table that reads as a table
       // visually and one that does not.
--- a/src/util/markdownSanitize.ts
+++ b/src/util/markdownSanitize.ts
@@ -121,13 +121,19 @@ export const MATHML_ATTRIBUTES: readonly string[] = [
  * MathML instead of the glyph spans beside it.
  *
  * The list used to also carry `role`, `ariaLabel`, `ariaBusy`, `ariaLive` and
- * `ariaAtomic`. Nothing in the pipeline emits any of them — the `role` and
- * `aria-label` that `TypingEffect` puts on `<pre>` and `<a>` are React props,
- * applied to the tree after it has been sanitised — and while they were
+ * `ariaAtomic`, on the reasoning that nothing in the pipeline emitted any of
+ * them — the `role` and `aria-label` `TypingEffect` puts on `<pre>` and `<a>`
+ * are React props, applied after sanitisation — and that while they were
  * misspelled they were inert. Correcting the spelling would have made them
- * live for the first time, so they are dropped instead: a response body cannot
- * reach an attribute today (the pipeline runs without `rehype-raw`, so raw HTML
- * is escaped to text), and this way it still could not were that to change.
+ * live for the first time, so they were dropped instead.
+ *
+ * That reasoning held only while footnotes were being dropped too. remark-gfm
+ * emits `ariaLabel` and `ariaDescribedBy` on the reference and backref links,
+ * so both are admitted now — on `a` alone rather than here. Which is the point
+ * this comment is really making: the narrow grant is what keeps an attribute
+ * from being available on every element that never needed it, and a response
+ * body still cannot reach one, since the pipeline runs without `rehype-raw`
+ * and raw HTML is escaped to text.
  */
 const GLOBAL_ATTRIBUTES: readonly string[] = [
   'className',
--- a/src/util/markdownSanitize.ts
+++ b/src/util/markdownSanitize.ts
@@ -169,8 +169,40 @@ export function createChatSanitizeSchema(): SanitizeSchema {
       // attribute that could not arrive — and a `target="_blank"` without
       // `rel="noopener"` is reverse tabnabbing waiting for the day one could.
       // Same reasoning as the ARIA entries dropped from GLOBAL_ATTRIBUTES.
-      'a': ['href'],
+      // `id` and the two ARIA entries are the footnote wiring remark-gfm
+      // emits: the reference anchor is described by the footnotes heading, and
+      // the backref carries "Back to reference 1" as its only text. Without
+      // them the backref is an unlabelled link.
+      //
+      // The `data-footnote-*` pair is named rather than admitted by a `data-*`
+      // wildcard, so this allows the two attributes footnotes need and not
+      // every dataset key a future plugin invents.
+      'a': [
+        'href',
+        'id',
+        'ariaLabel',
+        'ariaDescribedBy',
+        'dataFootnoteRef',
+        'dataFootnoteBackref',
+      ],
       'img': ['src', 'alt'],
+      // The footnotes block and the anchors its links resolve against.
+      'section': ['dataFootnotes'],
+      'h2': ['id'],
+      'li': ['id'],
+      // GFM's task list. A form control in a chat transcript is admitted only
+      // as the disabled checkbox remark-gfm produces — `checked` is the whole
+      // point, since checked and unchecked are otherwise indistinguishable,
+      // and `disabled` is what keeps it from being operable. `form`, `name`
+      // and `value` are deliberately absent: none is emitted, and together
+      // they are what would turn a rendered response into a submittable
+      // control.
+      'input': ['type', 'checked', 'disabled'],
+      // Column alignment from `:--` / `--:`. Carries no meaning for a screen
+      // reader but is the difference between a table that reads as a table
+      // visually and one that does not.
+      'th': ['align'],
+      'td': ['align'],
       // `style` here and on `svg` below carries KaTeX's computed layout —
       // `height:1em;vertical-align:-0.25em`, `top:-4em`, `width:0.471em`. The
       // values are generated, never passed through: `\htmlStyle` and
@@ -208,6 +240,34 @@ export function createChatSanitizeSchema(): SanitizeSchema {
       'blockquote',
       'img',
       'span',
+      // Heading levels. `rehype-sanitize` keeps the text of an element it
+      // drops, so a missing heading did not vanish — it flattened into the
+      // paragraph flow, taking heading navigation with it.
+      'h1',
+      'h2',
+      'h3',
+      'h4',
+      'h5',
+      'h6',
+      // `del` and `input` change what a sentence means rather than how it
+      // looks: a retraction read as an assertion, and a done task
+      // indistinguishable from an outstanding one.
+      'del',
+      'input',
+      'hr',
+      // Footnotes: `sup` on the reference, `section` on the block.
+      'sup',
+      'section',
+      // The table family. An LLM describing a chart reaches for a table first,
+      // and losing it leaves the numbers as a run of digits with nothing
+      // binding them to a label — no headers, no row or cell boundaries, and
+      // none of the table navigation assistive technology provides.
+      'table',
+      'thead',
+      'tbody',
+      'tr',
+      'th',
+      'td',
       'svg',
       'path',
       'line',
--- a/test/util/markdownSanitize.test.ts
+++ b/test/util/markdownSanitize.test.ts
@@ -239,6 +239,11 @@ describe('mathML allowlist', () => {
  *
  * #678 wires up the ESM Jest project that closes that gap; this list is what
  * the survey should reproduce when it does.
+ *
+ * A transcript is also only as good as the corpus behind it. `ol[start]`,
+ * `a[title]` and `img[title]` are absent from a corpus whose lists all begin
+ * at 1 and whose links carry no title, and were missed on the first pass for
+ * exactly that reason.
  */
 const GFM_ELEMENTS: readonly string[] = [
   'a',
@@ -279,13 +284,17 @@ const GFM_PROPERTIES: readonly (readonly [string, string])[] = [
   ['a', 'dataFootnoteRef'],
   ['a', 'href'],
   ['a', 'id'],
+  ['a', 'title'],
   ['h2', 'id'],
   ['img', 'alt'],
   ['img', 'src'],
+  ['img', 'title'],
   ['input', 'checked'],
   ['input', 'disabled'],
   ['input', 'type'],
   ['li', 'id'],
+  // Emitted only when a numbered list does not begin at 1.
+  ['ol', 'start'],
   ['section', 'dataFootnotes'],
   ['td', 'align'],
   ['th', 'align'],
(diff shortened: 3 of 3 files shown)


jooyoungseo:
`ol[start]` was a real miss — `799162a` — and chasing it turned up two more. The ARIA-role question I ran rather than reasoned about, as you asked.

### `ol[start]`: right, and the corpus is why I missed it

```
2. two          ->  ol[start]=2
3. three

1. one          ->  (no start attribute)
2. two
```

`start` appears **only** when the list does not begin at 1, and every list in my corpus began at 1. So the transcript could not contain it.

The failure mode is the quiet kind this PR is about: nothing disappears, the list just renumbers from 1 and the reader is given different numbers from the ones that were written.

### Widening the corpus found two more of the same shape

Your catch said the corpus was thin, so I went looking rather than fixing the one case:

| missing | comes from | why the corpus missed it |
|---|---|---|
| `ol[start]` | `2.` `3.` | every list started at 1 |
| `a[title]` | `[text](url "title")` | no link carried a title |
| `img[title]` | `[alt](src "title")` | no image carried one |

All three now allowed, and confirmed through the real pipeline rather than against the schema:

```
ol  {"start":2}
a   {"href":"https://e.com","title":"the title"}
img {"src":"https://e.com/i.png","alt":"a","title":"cap"}
```

`title` is text with no protocol, so nothing is given up by admitting it.

### The ARIA-role question: no roles on 13.2.1

You couldn't run it, so:

```
mdast-util-to-hast: 13.2.1
remark-gfm:         4.0.1

footnote branch emits:
  a[ariaDescribedBy] a[ariaLabel] a[className]
  a[dataFootnoteBackref] a[dataFootnoteRef] a[href] a[id]
  h2[className] h2[id] section[className] section[dataFootnotes]
```

No `role` on any footnote element — not on the reference, the backref, or the `section`. Checked with two footnotes as well as one, in case the role only appeared on a later reference. So there is nothing being silently stripped there on the pinned version.

Worth having asked, though: it is the same failure mode, and "an unlisted attribute is dropped, not the whole element" is exactly why it would have been invisible.

### The two I am not changing

**Exact-array assertions are brittle** — agreed, and this round proved it: adding `title` broke `should allow a link only what the pipeline puts on one` and I had to update it. I am keeping it anyway, because that break is the point. It is the only thing that would make someone notice an attribute appearing on `a` that nobody decided to allow.

**The submittable-control test is subsumed** by the exact-equality test above it — true today. I am keeping both because they fail on different things: the exact one breaks whenever the list legitimately changes and gets updated, and the negative one is what stops `name` or `value` being added *during* that update. Subsumption holds for the current value, not across the edit that changes it.

### The transcript's limit, recorded next to it

You were right that this shouldn't get lost, so it is now in the file rather than only in this thread — a transcript is only as good as the corpus behind it, with these three named as the cases that proved it.

| check | result |
|---|---|
| `npm test` | 1325 passed, 103 suites |

---

jooyoungseo:
`br` was a real hole — `dfb08f1`. Confirming the other two as deliberate.

### `br`, and the check I should have run the first time

You were right, and rather than add the one element I diffed the whole set — the previous round had already shown that fixing a single reported miss leaves the corpus just as thin:

```
survey:  a blockquote br code del em h1 h2 h3 h4 h5 h6 hr img input li ol
         p pre section strong sup table tbody td th thead tr ul

in survey, missing from GFM_ELEMENTS:  ['br']
in GFM_ELEMENTS, not in survey:        []
```

So `br` was the only one, and nothing in the list is dead weight. That is the check that should have run before I wrote the transcript, not after two reviewers found holes in it by reading.

Worth being precise about the cost: `br` was **already in `tagNames`**, so nothing was broken and no user lost a line break. What was broken is the transcript's claim to be complete — and it is load-bearing now, where it wasn't before. Removing `br` from the allowlist fails `should name every element the markdown pipeline emits`; before this commit it passed.

### `li: ['id']` being unscoped — deliberate, and you have the reason exactly right

`hast-util-sanitize`'s `attributes` map is per-tag, with no per-ancestor form, so scoping `id` to footnote definitions is not expressible. The same is true of every entry here — `a[dataFootnoteRef]` is equally unscoped.

The blast radius argument holds: the pipeline runs without `rehype-raw`, so raw HTML in a response is escaped to text and nothing but remark-gfm can put an `id` on an `li` in the first place. And an `id` that does arrive is clobber-prefixed, which is the protection that matters.

### Order-sensitive `toEqual` — keeping it, for the reason it annoyed me

Agreed it will cause churn. It already did this round: adding `title` broke that assertion and I had to update it by hand.

I'm keeping it because that is the only mechanism here that makes an unreviewed addition to `a` visible. `arrayContaining` would have let `title` — or anything else — appear without anyone deciding it should. The churn is the notification.

### Verification

| check | result |
|---|---|
| `npm test` | 1325 passed, 103 suites |

Thanks for checking that #677, #678 and #696 actually exist and say what the PR claims — that is the sort of thing that rots quietly, and it is a more useful check than it looks.

---

jooyoungseo:
Both findings are real. One is fixed here — `70a1763` — because it falsified a claim in this PR's own description. The other two are filed.

### The `aria-label` was admitted and then thrown away

This is the important one, and you are right that it made the PR's justification for `ariaLabel` false end to end.

```tsx
a: ({ node, ...props }) => (
  <a {...props} aria-label={`Link: ${props.children}`} />
),
```

A prop after a spread wins unconditionally, so the sanitiser preserved `aria-label="Back to reference 1"` and this discarded it one layer later. The backref's visible text is a bare return arrow, so it announced as *"Link: ↩"* — the exact information the attribute was admitted to preserve.

Fixed by falling back rather than overwriting, which is what the `img` override two lines below already does. Your suggested shape, unchanged.

**A second defect in the same line, which I am not fixing here.** `${props.children}` is a template interpolation over React children, so it is only a string for a link whose text is unformatted:

```
[plain](url)      ->  "Link: plain"
[**bold**](url)   ->  "Link: [object Object]"
[a `b` c](url)    ->  "Link: a,[object Object],c"
```

Any link containing emphasis, code or an image announces as `[object Object]`. That is a live defect on ordinary links rather than a footnote edge case, but fixing it means extracting text from a React tree, which is a different change from the one-line fallback. Filing it separately rather than growing this PR.

### `sr-only` — checked, and it is worse than the footnotes heading

You were right to flag this as a runtime question the diff cannot answer. It is not styled anywhere:

```
grep -c "sr-only" dist/maidr.css dist/maidr-math.css
  dist/maidr.css:0
  dist/maidr-math.css:0

tailwind / bootstrap in dependencies:  []
```

The project's actual mechanism is the inline `visuallyHidden` object in `Text.tsx`; `Constant.SR_ONLY_CLASS = 'maidr-sr-only'` is a third spelling that nothing styles or uses.

So the bigger instance is the one you noted in passing: **`TypingEffect`'s own live region** carries `className="sr-only"` and holds the full message text once typing finishes. On a page defining no `.sr-only` — which MAIDR cannot assume, since it embeds in arbitrary pages — every completed assistant message renders **twice**. Filed as #697.

One correction to the framing, though: for the footnotes heading this is **not a regression from this PR**. Before it, `h2` was disallowed, and `rehype-sanitize` drops the element while keeping its text — so "Footnotes" was already visible, just as loose text. This PR makes it a real heading, which is strictly better; it does not change whether it shows.

### Cross-message id collisions

Correct, and it constrains the follow-up rather than this PR. Added to #696, with the part that matters: it makes the two candidate fixes there **insufficient on their own**.

| option | fixes dangling | fixes collision |
|---|---|---|
| `clobberPrefix: ''` | yes | no — both messages still emit `user-content-fn-1` |
| rewrite hrefs | yes | no, unless the rewrite also namespaces per message |

`remark-rehype` takes `clobberPrefix`, and `TypingEffect` already has `messageId` in scope, so a per-message prefix through `remarkRehypeOptions` is the shape that gives uniqueness and the anti-clobber prefix at once. Noted there for whoever picks it up.

Also took your suggestion that #678's live survey should include a multi-message case — a single-document transcript cannot see this class of bug at all.

### The two notes already handled

`br` was added in `dfb08f1` after the previous round, along with a full diff of survey-against-transcript confirming it was the only gap. The order-sensitive `toEqual` is deliberate, for the reason it already earned this cycle: adding `title` broke it and forced a look, which is the only thing that makes an unreviewed addition to `a` visible.

### Verification

| check | result |
|---|---|
| `npm test` | 1325 passed, 103 suites |
| `npm run build` | exit 0 |

The `TypingEffect` fix has no test: that module imports `react-markdown`, which is ESM-only and unloadable by this CommonJS suite — the same constraint that put the schema in its own module in the first place, and the same one #678 removes.

---

jooyoungseo:
Point 2 taken — `7155b14` — and it was better than "optional hardening" turned out to be. Point 1 I can confirm with the run you were missing.

### The `type` pin fails *open* without it

You framed this as making an implicit guarantee explicit. It is more than that: the two forms differ in what happens when a value does arrive.

```
[['type','checkbox'], …]   type=hidden  ->  type="checkbox"
['type', …]                type=hidden  ->  type="hidden"
```

The tuple **replaces** a disallowed value rather than dropping it — and dropping would have been the worse of the two anyway, because an `input` with no `type` renders as a **text box**. So the bare name was failing open in both directions, and the pin is the only form that keeps it a checkbox.

Confirmed the pipeline still round-trips:

```
input {"type":"checkbox","checked":true,"disabled":true}
input {"type":"checkbox","checked":false,"disabled":true}
```

Loosening the pin fails both checkbox cases.

### Point 1: your deduction is right, and here is the run

You reasoned that `footnote-label` must not be pre-prefixed while `fn-1`/`fnref-1` are, which is why one picks up `clobberPrefix` once and the others twice. That is exactly it, and I had the data from the survey:

**Before sanitisation** — note which ids already carry the prefix:

```
h2 id="footnote-label"              <- NOT pre-prefixed
a  id="user-content-fnref-1"        <- pre-prefixed
li id="user-content-fn-1"           <- pre-prefixed
```

**After:**

```
h2 id="user-content-footnote-label"            <- prefixed once, matches ariaDescribedBy
a  id="user-content-user-content-fnref-1"      <- twice
li id="user-content-user-content-fn-1"         <- twice
```

`mdast-util-to-hast` namespaces the label-derived ids because those come from user text, and leaves its own fixed `footnote-label` bare. `aria-describedby` survives because it is in `clobber` alongside `id`, so both sides of that pair move together; `href` is not, so it stays pointing at the pre-sanitisation name.

Recorded in #696 rather than only here, since that is where someone will need it.

You are also right that it should be re-confirmed against a live pipeline once #678 lands rather than resting on a transcript — that is the same argument I made about the corpus, and it applies to this just as much.

### Point 3: agreed, and worth stating as the invariant it now is

`h2`, `li` and `a` can carry an `id` where nothing could before, and the only thing standing behind that is the untouched default `clobberPrefix`. So `should override no schema key other than tagNames and attributes` went from pinning a tidiness property to being the guard for three elements.

Not changing anything — the test already exists and already fails if someone adds a key — but the escalation is real and I would rather it be said in the thread than discovered later.

### On your caveat

Noted, and it is the right instinct: CI on this head is green other than E2E and the review job still running, and every claim above is from a run in this workspace rather than reasoning. The `type=hidden` behaviour in particular I would not have predicted correctly without executing it — I expected it to drop the attribute.

| check | result |
|---|---|
| `npm test` | 1326 passed, 103 suites |

---

jooyoungseo:
Point 2 taken — `390e070`. The other two I have collected on #678 rather than left in this thread.

### The stale comment was mine to have caught

You are right, and I had already noticed it: my own notes from the round that admitted `ariaLabel` say #675's claim "stops holding once footnotes are in scope". I wrote that down and then did not go back and fix the comment two sections above the code contradicting it.

Corrected. The claim that was falsified is gone; the argument it was making — grant an attribute as narrowly as the markup needs, which is why `ariaLabel` went on `a` rather than back into the global list — is what the comment now says. Comment only, no schema change.

### 1 and 3: agreed, and moved somewhere they will be found

Both are "pin it when #678 lands", and a note in a merged PR thread is not a place anyone will look. Written up on **#678** as a ranked checklist, with these two at the top because they are the only items that are regressions-waiting-to-happen in shipped code rather than gaps in observation:

1. **the `aria-label` fallback** — the bug this PR exists to prevent, and the one thing here with no regression guard. Including your point that it depends on `hast-util-to-jsx-runtime` leaving `aria-*` hyphenated, which is worth asserting rather than assuming since the fallback reads `props['aria-label']`.
2. **`['type','checkbox']` replacing rather than dropping** — the distinction the safety argument rests on, currently verified by a one-off run and nothing in CI.

The rest of the list is the live survey, the multi-message case from #696, and the clobber mechanism.

### On the caveat

Worth saying that your static read got every technical claim right, including the `hast-util-to-jsx-runtime` hyphenation, which is the sort of thing that is easy to get backwards. The one thing this round that a static read could not have settled is the `type=hidden` behaviour from the previous round — I predicted the attribute would be dropped, and it is replaced.

| check | result |
|---|---|
| `npx jest test/util/markdownSanitize.test.ts` | 23 passed |

CI on the previous head was green across everything except `E2E (chromium)` and the review job still running; this commit touches a comment only.

---

jooyoungseo:
I tried to fix the collision here rather than only track it, and the attempt is why I am not shipping it — it does not work.

### The obvious fix leaves one collision behind

You suggested a `clobberPrefix` derived from `messageId`, which is also what I had written into #696. `react-markdown` exposes `remarkRehypeOptions` and `TypingEffect` already has `messageId`, so it is about one line. Ran it against two independently rendered messages first:

```
--- default prefix ---
m1 a#user-content-user-content-fnref-1   m2 …fnref-1              <- collide
m1 li#user-content-user-content-fn-1     m2 …fn-1                 <- collide
m1 h2#user-content-footnote-label        m2 …footnote-label       <- collide

--- per-message prefix ---
m1 a#user-content-msg-m1-fnref-1         m2 …msg-m2-fnref-1       <- fixed
m1 li#user-content-msg-m1-fn-1           m2 …msg-m2-fn-1          <- fixed
m1 h2#user-content-footnote-label        m2 …footnote-label       <- STILL COLLIDES
```

`clobberPrefix` reaches ids derived from the footnote label, because that is user text. The heading's id is not derived from anything — the library hardcodes it and says so:

```
lib/state.js:132   > **Note**: `id: 'footnote-label'` is always added, because footnote
lib/footer.js:236                     id: 'footnote-label'
```

No option scopes it. `footnoteLabelProperties` sets the heading's other properties; the `id` is appended unconditionally afterwards.

So shipping this would trade "three colliding ids" for "one colliding id", while pre-empting #696's design with half of it and leaving the anchors dangling regardless. That is a worse position than either doing it properly there or not starting.

Written up on #696, where it now rules out one of the four options and gives a second reason to prefer the post-processing one — the plugin that re-points `href` at the clobbered `id` is the natural place to namespace all three ids.

### Where I think we differ, slightly

You called it "a fresh accessibility regression this same change makes newly reachable", and I would put it as newly *reachable* but not yet *reaching*: today the only reference to the surviving duplicate is `aria-describedby`, and every footnotes heading says "Footnotes", so resolving against the wrong message announces identically. The `fn`/`fnref` collisions matter only once anchors resolve — which is #696.

That is not an argument for ignoring it, and duplicate ids in one document are worth removing on their own. It is an argument that the right place is the change that makes anchors work, not this one.

### The tuple point

Agreed, and already on **#678** as item 2 in a ranked list of what the ESM project should pin first — with the same framing you used, that the comment's reasoning leans on which behaviour it is. Worth adding one correction there, though: it is not quite true that both outcomes are safe. A dropped `type` leaves an `input` that renders as a **text box**, which is why the tuple form is the one that fails safe:

```
[['type','checkbox'], …]   type=hidden  ->  type="checkbox"
['type', …]                type=hidden  ->  type="hidden"
```

---

Issue #696, opened 2026-08-01
a11y: footnote links in chat point at ids that sanitisation has renamed


jooyoungseo:
## Problem

A footnote reference in an AI chat response renders, is announced, and cannot be followed. Both of its links point at ids that no longer exist by the time the DOM is built.

Measured after sanitisation, resolving each reference against the ids actually present:

```
aria-describedby  user-content-footnote-label   RESOLVES
href              user-content-fn-1             DANGLING
href              user-content-fnref-1          DANGLING
```

## Cause

Two layers prefix, and only one of them prefixes both sides.

`remark-rehype` already namespaces footnote ids with `user-content-`, and emits matching hrefs:

```
a  id="user-content-fnref-1"  href="#user-content-fn-1"
li id="user-content-fn-1"
```

`hast-util-sanitize` then applies its own `clobberPrefix`, which covers `clobber: ['ariaDescribedBy', 'ariaLabelledBy', 'id', 'name']` — but not `href`:

```
a  id="user-content-user-content-fnref-1"  href="#user-content-fn-1"
li id="user-content-user-content-fn-1"
```

`aria-describedby` survives because both it and `id` are in `clobber`, so the pair moves together. `href` is not, so it is left pointing at the pre-sanitisation name.

## Why it was left out of #695

#695 admits `id` deliberately — dropping it would be worse, since `aria-describedby` would then dangle too and the anchors would be dead either way. But the fix for the anchors is a security trade rather than an oversight:

Setting `clobberPrefix: ''` on the chat schema makes the ids survive as `user-content-fn-1`, matching the hrefs remark-rehype wrote. It works, and it moves the anti-clobbering guarantee out of the sanitiser and onto a remark-rehype default — so the day that default changes, chat content could define `id="search"` and shadow `document.forms.search` and friends with nothing catching it.

It also adds a third key to a schema that deliberately declares only `tagNames` and `attributes`, which is what keeps every other `hast-util-sanitize` default in force (`protocols`, and so the `javascript:` filtering on `href`). That invariant has a test pinning it.

## Options

1. **`clobberPrefix: ''` on the schema.** Smallest diff. Delegates anti-clobbering to remark-rehype and needs the existing "declares no other key" test relaxed with a stated reason.
2. **Rewrite the hrefs.** A small rehype plugin after sanitisation that re-points `href="#x"` at the clobbered id. Keeps both defaults intact; adds a plugin to maintain.
[...]

PR #692, opened 2026-08-01
ci: close the rolling test-failure issue when the suite goes green

jooyoungseo:
# Pull Request

## Description

A passing scheduled run files its own `test-report` issue and leaves the `test-failure` one untouched. So a failure report outlives the fix, and no number of green runs ever closes it.

**#683 is the live example.** It was filed at `79bde6d` — after #674 and #676, but before #679 and #684 — and reports **80 failed / 823 passed**. Every parsed failure is WebKit:

| spec | failures |
|---|---|
| `multiLayer.spec.ts` | 10 |
| `multiLineplot.spec.ts` | 7 |
| `stackedBarplot.spec.ts` | 7 |
| `lineplot.spec.ts` | 6 |
| `violinPlot.spec.ts` | 2 |

Those are precisely what #679 (cross-browser modifier key) and #684 (announcement synchronisation) fixed. WebKit has since run 308/308 locally on every branch in that series. The suite is green; the issue saying otherwise simply had no way to close.

## Related Issues

Refs #683 — which I am closing separately, since this fix cannot retroactively close it. Follows #626, #679, #684, #686.

## Changes Made

On a green run, comment on every open `test-failure` issue and close it as `completed`.

Three deliberate details:

- **Every open one, not just the newest.** The title carries a timestamp, so a period of failures before this existed could have left more than one behind.
- **Pull requests filtered out.** `issues.listForRepo` returns PRs as well, and a PR that happened to carry the label is not a report.
- **Failure behaviour untouched.** A red run still updates the rolling issue in place and closes nothing.

The comment says a future failure files a fresh issue rather than reopening this one, so the history stays per-incident rather than one issue accumulating unrelated runs.

## Checklist

- [x] I have read the Contributor Guidelines.
- [x] I have performed a self-review of my own code and ensured it follows the project's coding standards.
- [x] I have tested the changes locally, and all tests related to this pull request pass. — see the harness below; a scheduled workflow cannot be exercised from a PR.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have updated the documentation, if applicable. — no documented behaviour changed.
- [ ] I have added appropriate unit tests, if applicable. — not applicable to workflow config; the harness is the evidence.

## Additional Notes

Syntax checks pass, but syntax passing is not the logic being right — so I extracted the `github-script` body and ran it against stubbed octokit calls:

```
SUCCESS + open #683      : listForRepo(test-report), create(test-report),
                           listForRepo(test-failure), comment(#683),
[...]

PR #699, opened 2026-08-01
fix(chat): hide the live region that was rendering every message twice

### LINKED ISSUE #697: a11y: the chat live region is not visually hidden — MAIDR ships no .sr-only rule
## Problem

`TypingEffect.tsx` renders its screen-reader live region as

```tsx
<div className="sr-only" aria-live={settings.general.ariaMode} aria-atomic="true">
  {isTyping ? '' : text}
</div>
```

Nothing in this project defines `.sr-only`:

```
grep -c "sr-only" dist/maidr.css dist/maidr-math.css
  dist/maidr.css:0
  dist/maidr-math.css:0

tailwind / bootstrap in dependencies:  []
```

So unless the host page happens to define the class, that div is an ordinary block element. Once typing finishes it holds the **full message text** — which means every completed assistant message renders twice on screen, the typed copy and the live-region copy.

MAIDR is embedded in arbitrary pages, so "the host might define it" is not something the library can rely on. Bootstrap 4 shipped `.sr-only`, Bootstrap 5 renamed it to `.visually-hidden`, and Tailwind provides it only if the page uses Tailwind. A page with none of those gets the duplicate.

## The project already has the right mechanism, unused here

`src/ui/component/Text.tsx` hides its live region with an inline style object:

```ts
const visuallyHidden: React.CSSProperties = {
  position: 'absolute', width: '1px', height: '1px',
  padding: 0, margin: '-1px', overflow: 'hidden',
  clip: 'rect(0, 0, 0, 0)', whiteSpace: 'nowrap', border: 0,
};
```

That is the standard clip pattern and it does not depend on a stylesheet reaching the element. `TypingEffect` predates or simply missed it.

There is also `Constant.SR_ONLY_CLASS = 'maidr-sr-only'` in `src/util/constant.ts`, which nothing styles and nothing appears to use — worth resolving in the same pass rather than leaving a third spelling of the same idea.

## Second instance, from #695

#695 admits the GFM footnote markup, which includes a heading `mdast-util-to-hast` hardcodes as

```html
<h2 class="sr-only" id="footnote-label">Footnotes</h2>
```

Same class, same outcome: the word "Footnotes" is visible in the chat bubble.

Note this is **not a regression from #695** — before it, `h2` was disallowed and `rehype-sanitize` dropped the element while keeping its text, so "Footnotes" was already showing, just as loose text rather than a heading. #695 makes it a real heading, which is an improvement; it does not make it any more or less visible.

## Suggested fix

Use the `visuallyHidden` style object for the `TypingEffect` live region, and either ship a `.sr-only` rule in the library's own CSS or map the class to that style — the footnotes heading arrives with `class="sr-only"` from the pipeline and cannot be given an inline style from React.

Shipping the rule is probably the one that covers both, since any future markup carrying `sr-only` is then handled too.

## Acceptance

- The chat live region is announced but not visible, on a page that defines no `.sr-only` of its own
- A response containing footnotes does not show a visible "Footnotes" heading
- `Constant.SR_ONLY_CLASS` is either used or removed

jooyoungseo:
# Pull Request

## Description

`TypingEffect`'s screen-reader live region was marked hidden with a class nothing defines:

```tsx
<div className="sr-only" aria-live={settings.general.ariaMode} aria-atomic="true">
  {isTyping ? '' : text}
</div>
```

```
grep -c "sr-only" dist/maidr.css dist/maidr-math.css
  dist/maidr.css:0
  dist/maidr-math.css:0

tailwind / bootstrap in dependencies:  []
```

So it was an ordinary block element holding the **full message text**, and every finished assistant message rendered a second time, in full, below the typed one.

This is not an oversight that happened to be harmless — MAIDR embeds in pages it does not control, so "the host might define `.sr-only`" is not something the library can lean on. Bootstrap 4 shipped it, Bootstrap 5 renamed it to `.visually-hidden`, Tailwind provides it only if the page uses Tailwind. Any page with none of those gets the duplicate.

## Related Issues

Closes #697. Found while reviewing #695.

## Changes Made

The project already had the right mechanism, used in exactly one place — `Text.tsx`'s clip-pattern style object. Lifted to `src/ui/visuallyHidden.ts` now that there is a second caller, and used for the chat live region.

**A global `.sr-only` rule was the other option and is worse.** `sr-only` is a common name, so a rule shipped by an embedded library either loses to the host page's or silently wins over it for the rest of the page — restyling markup that has nothing to do with MAIDR. An inline style cannot reach past the elements MAIDR renders.

It also would not have been possible without changing the build: `dist/maidr.css` is a deliberate **placeholder**, emitted so integrations that link it by name keep working after KaTeX's stylesheet was split out. `scripts/vite-plugin-math-stylesheet.js` documents that MAIDR has no static CSS left at all.

Two smaller things in the same area:

- **The footnotes heading** arrives from the markdown pipeline as `<h2 class="sr-only">` — `mdast-util-to-hast` hardcodes both, and expects a stylesheet to honour the class. React cannot put an inline style on markup the pipeline generates, so the class is matched in the existing `h2` component override. This is **not a regression from #695**: before it, `h2` was disallowed and `rehype-sanitize` dropped the element while keeping its text, so "Footnotes" was already visible, just as loose prose.
- **`Constant.SR_ONLY_CLASS`** is removed. It named a third spelling of the same idea — `'maidr-sr-only'` — that nothing styled and nothing used, which is worse than absent because it reads like the mechanism.

## Checklist

### Verified in a real browser

`TypingEffect` cannot be rendered here — the chat needs an LLM backend, and it imports ESM-only `react-markdown` besides (#678). But the style is shared with `Text.tsx` and needs neither. Chromium, with the exact style object and three `role="alert"` elements:

```
#hidden   box=1x1          (visuallyHidden)
#visible  box=1264x18      (ordinary block)
#none     box=null         (display:none)
```

```
ARIA snapshot of <main>:
  - alert: Announced but unseen
```

Three went in, **one came out**. The `visuallyHidden` element is invisible *and* announced; `display: none` and `visibility: hidden` are absent from the accessibility tree entirely — which is why the unit test pins both as unset rather than trusting a comment about what they do.

### On the guard

The two ways this regresses silently are guarded separately.

**The class returning** is a `no-restricted-syntax` rule in `eslint.config.ts`, matched on the AST. It began as a regex over `className=` in a test and was wrong three times — a word boundary beside a hyphen, truncation at the first `}` inside a template, a brace inside a string literal — each failing by *silently passing*, which is the wrong failure mode for the only protection a fix has. The selector gets a real parse for free, runs on every file rather than only when one test does, and needs no allowlist: scoping to `JSXAttribute[name.name='className']` leaves the `h2` override alone, since that reads the class in a `style` attribute rather than assigning one.

That rule is itself pinned by a test that lints a fixture through **the repo's own eslint binary and config**, not `RuleTester` — a reconstruction can agree with itself while disagreeing with what CI enforces. Nine cases, both directions:

```
className="sr-only"                        FLAGGED
className={'sr-only'}                      FLAGGED
className={`sr-only`}                      FLAGGED
className={cond ? 'sr-only' : undefined}   FLAGGED
className={`${cond} sr-only`}              FLAGGED
className={cond ? 'a{b' : 'sr-only'}       FLAGGED
className="not-sr-only"                    ignored
className={`${cond}-sr-only`}              ignored
className={someVariable}                   ignored
```

**The style being "simplified"** to `display: none` or `visibility: hidden` is a unit assertion. Both are shorter and look equivalent, and the browser check above shows what they actually do.

### What is still not verified

That the duplicate text is gone and the region announces in a real screen reader. Chromium's accessibility tree is not a screen reader any more than jsdom is, and the component cannot be rendered by this suite at all. That needs #678 and a human with assistive technology.

| check | result |
|---|---|
| `npm test` | 1328 passed, 104 suites |
| `npm run build` | exit 0 |

---

### DIFF
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -86,6 +86,32 @@ const config: ReturnType<typeof antfu> = antfu({
       message: 'Do not call pressKeyCombination() from a plot page object: it presses the key but does not wait for the announcement. Use this.moveToDataPoint(key, action, true) for a modified move.',
     }],
   },
+}, {
+  // `sr-only` hides nothing here. MAIDR ships no stylesheet — `dist/maidr.css`
+  // is a placeholder and the UI is styled at runtime by emotion — so an element
+  // given that class is an ordinary visible one. `TypingEffect`'s live region
+  // carried it and rendered every finished chat message a second time, in full.
+  //
+  // Matched on the AST rather than by scanning source text. This started as a
+  // regex over `className=` in a test and took four rounds to stop being wrong
+  // — word boundaries around a hyphen, truncation at the first `}` inside a
+  // template, a brace inside a string literal — and every one of those failed
+  // by silently passing. The selector gets a real parse for free and runs on
+  // every file rather than only when that test does.
+  //
+  // The `sr-only` in `TypingEffect`'s `h2` override is untouched: it sits in a
+  // `style` attribute, reading the class the markdown pipeline set rather than
+  // assigning one.
+  files: ['src/**/*.ts', 'src/**/*.tsx'],
+  rules: {
+    'no-restricted-syntax': ['error', {
+      selector: 'JSXAttribute[name.name=\'className\'] Literal[value=/(^|\\s)sr-only(\\s|$)/]',
+      message: 'Nothing defines .sr-only in MAIDR, so this element is visible. Use the visuallyHidden style object from @ui/visuallyHidden.',
+    }, {
+      selector: 'JSXAttribute[name.name=\'className\'] TemplateElement[value.raw=/(^|\\s)sr-only(\\s|$)/]',
+      message: 'Nothing defines .sr-only in MAIDR, so this element is visible. Use the visuallyHidden style object from @ui/visuallyHidden.',
+    }],
+  },
 });
 
 export default config;
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -87,21 +87,14 @@ const config: ReturnType<typeof antfu> = antfu({
     }],
   },
 }, {
-  // `sr-only` hides nothing here. MAIDR ships no stylesheet — `dist/maidr.css`
-  // is a placeholder and the UI is styled at runtime by emotion — so an element
-  // given that class is an ordinary visible one. `TypingEffect`'s live region
-  // carried it and rendered every finished chat message a second time, in full.
+  // `sr-only` hides nothing here — MAIDR ships no stylesheet to define it, so
+  // the class leaves an ordinary visible element. `src/ui/visuallyHidden.ts`
+  // holds the reasoning and the replacement.
   //
-  // Matched on the AST rather than by scanning source text. This started as a
-  // regex over `className=` in a test and took four rounds to stop being wrong
-  // — word boundaries around a hyphen, truncation at the first `}` inside a
-  // template, a brace inside a string literal — and every one of those failed
-  // by silently passing. The selector gets a real parse for free and runs on
-  // every file rather than only when that test does.
-  //
-  // The `sr-only` in `TypingEffect`'s `h2` override is untouched: it sits in a
-  // `style` attribute, reading the class the markdown pipeline set rather than
-  // assigning one.
+  // Matched on the AST because the source-text version of this guard was wrong
+  // three times over, each in a way that passed silently. Scoping to
+  // `className` also leaves `TypingEffect`'s `h2` override alone, since that
+  // reads the class in a `style` attribute rather than assigning one.
   files: ['src/**/*.ts', 'src/**/*.tsx'],
   rules: {
     'no-restricted-syntax': ['error', {
--- a/src/ui/component/Text.tsx
+++ b/src/ui/component/Text.tsx
@@ -1,22 +1,7 @@
 import { useViewModelState } from '@state/hook/useViewModel';
+import { visuallyHidden } from '@ui/visuallyHidden';
 import { Constant } from '@util/constant';
 import React from 'react';
-
-/**
- * Visually-hidden style that keeps an element in the accessibility tree
- * but invisible on screen (standard "sr-only" / "clip" pattern).
- */
-const visuallyHidden: React.CSSProperties = {
-  position: 'absolute',
-  width: '1px',
-  height: '1px',
-  padding: 0,
-  margin: '-1px',
-  overflow: 'hidden',
-  clip: 'rect(0, 0, 0, 0)',
-  whiteSpace: 'nowrap',
-  border: 0,
-};
 
 const Text: React.FC = () => {
   const { enabled, announce, value, revision, message } = useViewModelState('text');
--- a/src/ui/components/TypingEffect.tsx
+++ b/src/ui/components/TypingEffect.tsx
@@ -170,9 +170,8 @@ export const TypingEffect: React.FC<TypingEffectProps> = memo(({ text, isUser, m
             ),
             // The footnotes heading arrives as `<h2 class="sr-only">`, which
             // mdast-util-to-hast hardcodes and expects a stylesheet to honour.
-            // MAIDR has none, so the class alone leaves a literal "Footnotes"
-            // heading in the bubble. Applied here rather than as a global rule
-            // because a global `.sr-only` would collide with the host page's.
+            // Nothing can style pipeline-generated markup inline, so the class
+            // is matched here instead — see `visuallyHidden` for why not a rule.
             h2: ({ node, className, ...props }) => {
               const hidden = (className ?? '').split(/\s+/).includes('sr-only');
               return <h2 {...props} className={className} style={hidden ? visuallyHidden : undefined} />;
--- a/src/ui/components/TypingEffect.tsx
+++ b/src/ui/components/TypingEffect.tsx
@@ -1,5 +1,6 @@
 import { Box } from '@mui/material';
 import { useViewModelState } from '@state/hook/useViewModel';
+import { visuallyHidden } from '@ui/visuallyHidden';
 import { containsLatex, ensureKatexStylesheet } from '@util/katex';
 import { createChatSanitizeSchema } from '@util/markdownSanitize';
 import React, { memo, useEffect, useMemo, useState } from 'react';
@@ -167,6 +168,18 @@ export const TypingEffect: React.FC<TypingEffectProps> = memo(({ text, isUser, m
             img: ({ node, ...props }) => (
               <img {...props} alt={props.alt || 'Image in message'} />
             ),
+            // The footnotes heading arrives as `<h2 class="sr-only">`, which
+            // mdast-util-to-hast hardcodes and expects a stylesheet to honour.
+            // MAIDR has none, so the class alone leaves a literal "Footnotes"
+            // heading in the bubble. Applied here rather than as a global rule
+            // because a global `.sr-only` would collide with the host page's.
+            h2: ({ node, className, ...props }) => (
+              <h2
+                {...props}
+                className={className}
+                style={className?.split(/\s+/).includes('sr-only') ? visuallyHidden : undefined}
+              />
+            ),
           }}
         >
           {displayedText}
@@ -174,7 +187,7 @@ export const TypingEffect: React.FC<TypingEffectProps> = memo(({ text, isUser, m
       </div>
       {/* Visually hidden live region for screen readers */}
       <div
-        className="sr-only"
+        style={visuallyHidden}
         aria-live={settings.general.ariaMode}
         aria-atomic="true"
       >
--- a/src/ui/components/TypingEffect.tsx
+++ b/src/ui/components/TypingEffect.tsx
@@ -173,13 +173,10 @@ export const TypingEffect: React.FC<TypingEffectProps> = memo(({ text, isUser, m
             // MAIDR has none, so the class alone leaves a literal "Footnotes"
             // heading in the bubble. Applied here rather than as a global rule
             // because a global `.sr-only` would collide with the host page's.
-            h2: ({ node, className, ...props }) => (
-              <h2
-                {...props}
-                className={className}
-                style={className?.split(/\s+/).includes('sr-only') ? visuallyHidden : undefined}
-              />
-            ),
+            h2: ({ node, className, ...props }) => {
+              const hidden = (className ?? '').split(/\s+/).includes('sr-only');
+              return <h2 {...props} className={className} style={hidden ? visuallyHidden : undefined} />;
+            },
           }}
         >
           {displayedText}
--- a/src/ui/visuallyHidden.ts
+++ b/src/ui/visuallyHidden.ts
@@ -1,0 +1,28 @@
+import type React from 'react';
+
+/**
+ * Style that keeps an element in the accessibility tree but off the screen.
+ *
+ * The standard clip pattern. Used rather than a `.sr-only` class because MAIDR
+ * has no stylesheet to put one in: `dist/maidr.css` is a placeholder emitted so
+ * integrations that link it by name keep working, and the UI is styled at
+ * runtime by emotion (see `scripts/vite-plugin-math-stylesheet.js`).
+ *
+ * Shipping a global `.sr-only` rule instead would be worse than merely
+ * inconvenient. MAIDR embeds in pages it does not control, and `sr-only` is a
+ * common name — Bootstrap 4 and Tailwind both define it — so a global rule
+ * would either be overridden by the host's or silently override the host's for
+ * the rest of the page. An inline style on the elements MAIDR owns cannot
+ * reach anything outside them.
+ */
+export const visuallyHidden: React.CSSProperties = {
+  position: 'absolute',
+  width: '1px',
+  height: '1px',
+  padding: 0,
+  margin: '-1px',
+  overflow: 'hidden',
+  clip: 'rect(0, 0, 0, 0)',
+  whiteSpace: 'nowrap',
+  border: 0,
+};
--- a/src/ui/visuallyHidden.ts
+++ b/src/ui/visuallyHidden.ts
@@ -22,7 +22,12 @@ export const visuallyHidden: React.CSSProperties = {
   padding: 0,
   margin: '-1px',
   overflow: 'hidden',
+  // `clip` is deprecated and `clipPath` is its replacement, but both are here
+  // on purpose: this is the one idiom every browser still honours `clip` for,
+  // and dropping it would silently widen the set of engines where the element
+  // becomes visible. Belt and braces is the standard form of this pattern.
   clip: 'rect(0, 0, 0, 0)',
+  clipPath: 'inset(50%)',
   whiteSpace: 'nowrap',
   border: 0,
 };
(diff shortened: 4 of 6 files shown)


jooyoungseo:
One taken, one declined — `6045b91`. The declined one is worth showing the working for, because checking it found the same bug in code I had written.

### The regex nit would introduce a false positive

You suggested replacing `className?.split(/\s+/).includes('sr-only')` with `/\bsr-only\b/.test(className ?? '')` to avoid an array per render. Those are not equivalent — `-` is a non-word character, so `\b` sits right beside it:

| class | `\bsr-only\b` | split + includes |
|---|---|---|
| `sr-only` | true | true |
| `foo sr-only` | true | true |
| `not-sr-only` | **true** | false |
| `sr-only-thing` | **true** | false |

So the regex hides any element whose class merely *contains* the substring between hyphens. An array allocation per `h2` in a chat bubble is not worth that.

**But you were right about it in a place I had not looked.** My own guard in the test used exactly the pattern you proposed:

```js
/className=["'][^"']*\bsr-only\b/
```

A file with `className="not-sr-only"` would have failed the suite over a class with nothing to do with hiding anything. It compares whole tokens now.

### `clip` deprecation — taken, both spellings

Correct, and the right moment to fix it: this is being promoted from a private constant to a shared utility, so it should not carry the old form forward unexamined.

Both are set rather than swapping. `clip` is deprecated, but it is the one every engine still honours for this exact idiom, so dropping it widens the set of browsers where the element becomes visible — which is the failure this whole PR is about. The doubled form is the standard one, and the test pins both.

### Test-coverage gap — partly closed, honestly

You noted the guard cannot catch a template-literal reintroduction. Fixed, along with the braced form. Verified all four:

```
className="sr-only"        CAUGHT
className={'sr-only'}      CAUGHT
className={`sr-only`}      CAUGHT
className="not-sr-only"    correctly ignored
```

The other half of your point stands and I am not claiming otherwise: **nothing here catches the `h2` override being deleted.** A static guard can see a class being assigned; it cannot see a component override that stopped existing. That needs a render, which needs #678 — where it is now on the checklist with the rest.

### The two I am acknowledging without changing

**The `h2` override applies to every `h2`, not just the footnotes one.** True. Only the footnotes heading can carry `sr-only` today, because `mdast-util-to-hast` hardcodes that class — and hardcodes the `id` beside it, which is the same coupling that makes #696 harder than it looks. Worth the flag; not worth a narrower selector that would break the moment `footnoteLabelProperties` is used.

**`Constant.SR_ONLY_CLASS` removal** — thanks for independently confirming no references remain. That is the check I would most want a second pair of eyes on, since a stale reference would not fail the build.

| check | result |
|---|---|
| `npm test` | 1328 passed, 104 suites |
| `npm run build` | exit 0 |

---

jooyoungseo:
Both taken — `c491a2f`. The second turned out to close a real gap rather than only document one.

### The optional-chaining density

You are right on both counts: it is correct — the short-circuit covers the trailing `.includes()`, not just the first link — and it reads like a bug.

Took a different route than the suggested comment, though. If a line needs a note saying "this is not the bug it looks like", the line is the thing to change:

```tsx
const hidden = (className ?? '').split(/\s+/).includes('sr-only');
```

No short-circuit to reason about, so there is nothing left to explain.

### The ternary gap — closed rather than accepted

You framed this as a fine tradeoff for a guard rail, and I would have agreed if it were expensive. It was not: the guard now takes the whole braced expression and looks for the class in any string literal inside it.

```
className="sr-only"                        CAUGHT
className={'sr-only'}                      CAUGHT
className={`sr-only`}                      CAUGHT
className={isFoo ? 'sr-only' : undefined}  CAUGHT   <- your case
className="not-sr-only"                    ignored
className={someVariable}                   ignored
```

That last line is load-bearing and not incidental: `className={someVariable}` contains no string literal, which is exactly what keeps the guard from reporting the `h2` override — the one place in the codebase that legitimately *reads* the class rather than assigning it. Widening the regex without that property would have made the guard fail on the fix it exists to protect.

### Still not exhaustive, and I would not claim otherwise

A regex over source text cannot see a class assembled at runtime (`clsx`, a template with a variable in it, a constant defined elsewhere). What it does cover is every form that could plausibly be written by hand at the point where this regressed. The case it structurally cannot cover is the one already recorded on #678 as item 6: the `h2` override being **deleted** — a static guard sees a string appear, never a component disappear.

| check | result |
|---|---|
| `npm test` | 1328 passed, 104 suites |
| `npm run build` | exit 0 |

---

jooyoungseo:
Fixed in `23af3f1`, and checking it showed the gap is wider than reported.

### The truncation misses the class wherever it sits

You described the class being missed when the interpolation comes *before* it. It is missed either way:

```
className={`${cond} sr-only`}   ->  captured  {`${cond}
className={`sr-only ${cond}`}   ->  captured  {`sr-only ${cond}
```

The second one keeps the text, but what survives truncation is an *unterminated* string, so `assignsSrOnly` finds no complete literal and reports clean. Truncating at the first `}` breaks the guard for any interpolated template, not for one arrangement of one.

Counts braces now, and the scan handles nesting properly rather than assuming one level.

### Your example is not actually an offender, which is worth separating out

`` className={`${cond}-sr-only`} `` builds `foo-sr-only` — a different class, not the `sr-only` token. So that exact shape should be, and still is, **ignored**. The mechanism you identified was right; the instance was not an offending class. Both are in the verification table so the distinction is recorded rather than blurred:

```
className="sr-only"                        CAUGHT
className={'sr-only'}                      CAUGHT
className={`sr-only`}                      CAUGHT
className={isFoo ? 'sr-only' : undefined}  CAUGHT
className={`${cond} sr-only`}              CAUGHT   <- was missed
className={`sr-only ${cond}`}              CAUGHT   <- was missed
className="not-sr-only"                    ignored
className={`${cond}-sr-only`}              ignored  <- your example
className={someVariable}                   ignored
```

The three ignores carry as much weight as the catches. The last in particular: `className={someVariable}` holds no string literal, which is what keeps the `h2` override — the one place in the codebase that legitimately *reads* this class rather than assigning it — from being reported by the guard written to protect it.

### On why this got three rounds of attention

You put it well: this test is the only regression guard for the bug being fixed, because the component cannot be rendered here at all (#678). A guard that silently passes is worse than no guard, since it invites exactly the confidence it does not earn — so it is worth more scrutiny than a test whose subject is directly testable, not less.

Still not exhaustive, and the boundary is now explicit in the file: a class assembled at runtime (`clsx`, a constant from elsewhere) is invisible to source text, and the `h2` override being *deleted* is structurally invisible to any static check. That last one is on #678 as item 6.

| check | result |
|---|---|
| `npm test` | 1328 passed, 104 suites |
| `npm run build` | exit 0 |

---

jooyoungseo:
Both taken — `5d7e6b4`. You called the brace one "arguably not worth a follow-up"; I think it was, and the reason is the same one that made the previous round worth it.

### The brace counter

Correct as reported. The counter treated every brace as structure, so one inside a string unbalanced it:

```
className={cond ? 'a{b' : 'sr-only'}   ->  never returns to depth 0
```

What makes it worth fixing rather than noting is the **shape of the failure**, which is identical to the truncation from last round: neither gives a wrong answer. Both drop the expression from the scan and leave the guard reporting clean. A guard whose failure mode is "silently passes" invites exactly the confidence it has not earned — and this one is the only regression protection the fix has, since the component cannot be rendered here at all.

The count skips string contents now. Re-checked the neighbouring cases at the same time, since a change to the scanner can break what it already handled:

```
className={cond ? 'a{b' : 'sr-only'}       CAUGHT   <- was dropped
className={`${cond} sr-only`}              CAUGHT
className="sr-only"                        CAUGHT
className={isFoo ? 'sr-only' : undefined}  CAUGHT
className={cond ? 'a{b' : 'c'}             ignored
className={someVariable}                   ignored
```

You are right that a brace in a class name is not realistic. The fix is cheap and the alternative was a documented hole in the one thing standing between this bug and its return.

### The comment duplication

Taken. The `h2` block restated the case against a global rule that `visuallyHidden.ts` already makes at length, so it now points there instead of arguing it a second time. What it keeps is the part that is only true at that call site: the heading comes from the pipeline, and nothing can style pipeline-generated markup inline.

### On not being able to run the suite

Noted, and worth saying: your static reads have been accurate every round — the `className` / `GLOBAL_ATTRIBUTES` interaction, the Bootstrap 5 comparison, and both scanner defects were all correct from reading alone. CI is green on the previous head and this commit touches a test helper and a comment.

| check | result |
|---|---|
| `npm test` | 1328 passed, 104 suites |
| `npm run build` | exit 0 |

---

jooyoungseo:
Took the ESLint suggestion you explicitly weren't asking for — `e53f965`. You were right, and the history of this PR is the argument for it.

### The scanner deleted, the rule in its place

Your framing — "a hand-rolled scanner over source text is inherently open-ended" — is the accurate diagnosis of what four rounds of review were actually doing. Word boundaries around a hyphen, truncation at the first `}` in a template, a brace inside a string literal. Three separate defects, all in a thing I kept calling a regex while it grew into a small parser, and **every one failed by silently passing**.

That is the failure mode I argued twice in this thread was unacceptable for the only protection this fix has. Continuing to patch it while saying that was the inconsistency you spotted.

`no-restricted-syntax` gets a real parse for free — and this repo already uses it this way, for the page-object key presses. It also runs on every file during lint rather than only when one test executes, and reports at the offending line instead of naming a file.

Verified against every shape the scanner had to be taught, plus the three that must stay ignored:

```
className="sr-only"                        FLAGGED
className={'sr-only'}                      FLAGGED
className={`sr-only`}                      FLAGGED
className={isFoo ? 'sr-only' : undefined}  FLAGGED
className={`${cond} sr-only`}              FLAGGED
className={cond ? 'a{b' : 'sr-only'}       FLAGGED
className="not-sr-only"                    ignored
className={`${cond}-sr-only`}              ignored
className={someVariable}                   ignored
```

**Net 112 lines deleted.** The remaining test keeps only the assertion an AST cannot make: that the style leaves the element in the accessibility tree.

One property worth calling out, because it is what makes the selector clean rather than lucky: scoping to `JSXAttribute[name.name='className']` leaves the `h2` override alone automatically. Its `sr-only` sits in a `style` attribute, *reading* the class the pipeline set rather than assigning one — so no allowlist or exception is needed, where a text scan needed care to avoid reporting the fix by its own guard.

Your `{...{ className: 'sr-only' }}` spread case is still not covered — that is an `ObjectExpression`, not a `JSXAttribute`. I left it, because adding a second selector for a form nobody writes would be reaching for exhaustiveness again, which is the habit that produced the scanner. If it ever appears, the selector is one line.

### The two smaller notes

**`src/ui/visuallyHidden.ts` placement** — I share your read. It is not a component, and both `component/` and `components/` exist as a pre-existing split, so putting it inside either would be picking a side of an inconsistency this PR did not create.

**The `h2` keeping its `className`** — leaving it deliberately. It is inert here, but it is what the pipeline emitted, and dropping it would make the rendered DOM diverge from the source markup for no gain. If a host page happens to define `.sr-only`, keeping the class means their rule applies too and the inline style still wins.

| check | result |
|---|---|
| `npm test` | 1327 passed, 104 suites |
| `npm run build` | exit 0 |

---

jooyoungseo:
Both taken — `f283531`. Point 2 was the right catch: I had replaced a guard that failed silently with a rule that could regress silently.

### The rule now has a test, against the config rather than a copy

You put the problem precisely: *"that verification does not persist."* I checked nine forms by hand, wrote the results in a commit message, and left nothing running. Given the regex needed several attempts, that is the one part of this that earned a test.

It lints a fixture through the repo's **own eslint binary and config**, not `RuleTester`. A reconstruction can agree with itself while disagreeing with the config actually in force — which is the failure this exists to exclude, and the same reason #692's workflow test reads its script out of the YAML rather than duplicating it. Piped through stdin so the fixture never exists under `src/`, where it would be linted and type-checked on its own account.

It asserts **both directions in one comparison**, because either half alone is passable by a broken selector: one that matches everything satisfies a check of the flagged cases, one that matches nothing satisfies a check of the ignored ones.

Verified it fails against two real regressions rather than assuming:

```
(^|\s)sr-only(\s|$)  ->  \bsr-only\b     fails on not-sr-only
TemplateElement selector removed         fails on both template forms
```

**One thing the first version got wrong**, worth recording since it is the same class of bug: `execFileSync` throws on a non-zero exit, and eslint always exits non-zero here because the fixture exists to produce errors. The first run failed outright rather than silently, which was luck — a `catch` that swallowed it would have read "no messages" as "nothing reported" and passed. It now reads stdout from the thrown error and re-throws when that is empty.

### Comment duplication

Fair, and this is the second round on it. The rule's comment now points at `visuallyHidden.ts` for the reasoning instead of restating it, keeping only what is true at the rule site: why the selector is scoped to `className`, which is what leaves the `h2` override alone.

### The one I am leaving

The `h2` override recomputing `split(/\s+/)` per render — agreed, not worth a `useMemo`. Footnote headings are at most one per message, and memoising a string split behind a dependency array costs more to read than it saves.

| check | result |
|---|---|
| `npm test` | 1328 passed, 104 suites |
| `npm run build` | exit 0 |

---

jooyoungseo:
No code change this round — but your point about "looks fixed in code while staying silently broken" prompted a check I had twice said I could not do, and it was only *mostly* true that I could not.

### Verified in a real browser

I cannot render `TypingEffect` — the chat needs an LLM backend, so the live region never appears in the preview. But the *mechanism* is shared with `Text.tsx` and does not need one. Chromium, with the exact style object:

```
#hidden   box=1x1          (visuallyHidden)
#visible  box=1264x18      (ordinary block)
#none     box=null         (display:none)
```

And the accessibility tree for the same markup:

```
- main:
  - alert: Announced but unseen
```

Three `role="alert"` elements went in. **One came out.** The `visuallyHidden` one is present and named; `display:none` and `visibility:hidden` are absent from the tree entirely.

So both halves are now evidenced rather than asserted:

- the element is **not visible** (1×1, clipped) yet **is announced**, which is the fix;
- the "simplification" the unit test guards against would genuinely **mute** the region, which until now was a claim in a comment.

That second one is the more useful result. The test pins `display` and `visibility` as absent, and I had justified it by describing what those properties do. It is better to have watched them do it.

What this still does not prove is the end-to-end behaviour in `TypingEffect` — that the duplicate message text is gone and the region announces in a real screen reader. jsdom is not a screen reader and neither is Chromium's accessibility tree. That remains for #678 and a human with assistive technology.

### The two nits

**Subprocess cost** — measured at ~1.4s in a suite that runs ~40s. I agree the tradeoff is worth it, for the reason you gave: an in-process `RuleTester` reconstruction is exactly what drift would hide.

**`className={className}` forwarding `sr-only`** — intentional, and answered [in the previous round](https://github.com/xability/maidr/pull/699#issuecomment-5152843092). Briefly: it is inert, it is what the pipeline emitted, and dropping it would make the rendered DOM diverge from the source markup for no gain. If a host page does define `.sr-only`, keeping the class means their rule applies too and the inline style still wins.

### Thanks for tracing the schema

Confirming `className` is in `GLOBAL_ATTRIBUTES` and `id` is allowed on `h2` is the check that matters most here — without it the `hidden` branch is dead code that looks alive, and nothing in the suite would say so.

---

PR #700, opened 2026-08-01
fix(chat): let a link's own text be its accessible name

### LINKED ISSUE #698: a11y: a chat link with formatted text announces as "Link: [object Object]"
## Problem

`TypingEffect.tsx` builds each link's accessible name by interpolating its React children into a template literal:

```tsx
a: ({ node, ...props }) => (
  <a {...props} aria-label={props['aria-label'] ?? `Link: ${props.children}`} />
),
```

`props.children` is a string only when the link text is unformatted. Any inline markup inside the link makes it a React element, or an array of them, and template interpolation stringifies those through `Object.prototype.toString`:

| markdown | announced as |
|---|---|
| `[plain](url)` | `Link: plain` |
| `[**bold**](url)` | `Link: [object Object]` |
| `[a \`b\` c](url)` | `Link: a,[object Object],c` |
| `[[alt](img.png)](url)` | `Link: [object Object]` |

An `aria-label` overrides the element's text content entirely, so the visible text is not a fallback — a screen reader user gets `[object Object]` and nothing else. Emphasis inside link text is ordinary in LLM output, so this is not a rare shape.

## Context

Found while reviewing #695, which fixed the adjacent defect in the same line: the override used to set `aria-label` unconditionally after the spread, discarding the one the pipeline sets on footnote back-references. That part is fixed; this part is not, because it needs more than a fallback.

## Worth questioning the whole override

Before extracting text from a React tree, it is worth asking whether the label should exist at all.

A screen reader already announces an `<a href>` as a link — that is its role. Prefixing the accessible name with `"Link: "` makes it announce roughly as "Link: colon, download the data, link", which is the kind of redundancy WCAG's own guidance on `aria-label` warns against. And an `aria-label` derived from the children is, at best, a worse copy of the text content the browser would have used anyway.

So the options are roughly:

1. **Drop the override.** The link's text content becomes its accessible name, which is correct for every case above, removes the redundancy, and keeps the pipeline's `aria-label` on the backref because nothing overwrites it. Smallest change, and it deletes code rather than adding it.
2. **Extract text properly.** Walk `props.children` for string leaves and join them. Keeps the `"Link: "` prefix; adds a helper and the maintenance of it.

Option 1 looks right, but it is a change to how every chat link announces, so it deserves a deliberate decision rather than being folded into an unrelated fix.

The `pre` override on the line above has the same shape (`role="text" aria-label="Code block"`) and is worth reviewing in the same pass — `role="text"` is non-standard and is not in the ARIA specification.

## Acceptance

- A link whose text contains emphasis, code, or an image announces its actual text
- The footnote back-reference still announces "Back to reference N"
- Whatever is decided for the `"Link: "` prefix is stated in a comment, since removing it is a deliberate behaviour change and not an oversight

Test coverage for any of this needs #678 — `TypingEffect` imports `react-markdown`, which is ESM-only and cannot be loaded by the current CommonJS Jest suite.

jooyoungseo:
# Pull Request

## Description

`TypingEffect`'s link override built the accessible name by interpolating React children:

```tsx
a: ({ node, ...props }) => (
  <a {...props} aria-label={props['aria-label'] ?? `Link: ${props.children}`} />
),
```

`children` is a React node. It is a string only when the link text has no markup in it — anything else goes through `Object.prototype.toString`:

| markdown | announced as |
|---|---|
| `[plain](url)` | `Link: plain` |
| `[**bold**](url)` | `Link: [object Object]` |
| `` [a `b` c](url) `` | `Link: a,[object Object],c` |
| `[[alt](img.png)](url)` | `Link: [object Object]` |

`aria-label` **replaces** the accessible name rather than supplementing it, so the visible text is not a fallback: a screen reader user gets `[object Object]` and nothing else. Emphasis inside link text is ordinary in LLM output, so this is not a rare shape.

## Related Issues

Closes #698. Found while reviewing #695, and left out of #699 because it needed a decision rather than a fallback.

## Changes Made

**Removed the override** rather than extracting text from the React tree — the issue listed both options and this one needed deciding, so I checked it in Chromium's accessibility tree rather than reasoning about it:

```
aria-label="Link: [object Object]"    ->  link "Link: [object Object]"
no aria-label                         ->  link "Download the data"
aria-label="Link: Download the data"  ->  link "Link: Download the data"
```

The middle line is the name the element already had, computed from its own content — correct for every case in the table above, including the image-only link, where it resolves to the image's `alt`.

The third line shows what the override bought when `children` happened to be a string: a `"Link: "` prefix duplicating the role a screen reader announces anyway. So the override was, at best, a worse copy of the name it replaced.

Removing it also deletes a fallback expression that had to be right. The footnote backref keeps the `aria-label` remark-gfm gives it because nothing overwrites it now, rather than because the `??` branch is correct — `ariaLabel` is already allowlisted on `a` in `createChatSanitizeSchema`, so it flows through natively.

**A `no-restricted-syntax` rule** so the same construction cannot return. Deriving `aria-label` from `children` is the general form of this bug, not a MAIDR-specific one.

## Checklist

### This is a behaviour change, deliberately

Every chat link stops announcing with a `"Link: "` prefix. That is the point rather than a side effect — an `<a href>` is already announced as a link by its role, so the prefix made a link read roughly as "link, Link: download the data". Worth a maintainer's eye even though I am confident it is an improvement.

### On the rule, and a redundancy it caught in itself

The rule is **one** selector, not two. `props.children` is a `MemberExpression` whose property is an `Identifier` named `children`, so matching the identifier covers the member access too.

I only noticed because of the check that follows every rule in this repo now: I broke the `MemberExpression` selector expecting the test to fail, and it passed. The second selector was doing nothing. That is the same discipline that turned up three defects in the `sr-only` guard on #699, and it is cheap.

```
aria-label={`Link: ${props.children}`}   FLAGGED
aria-label={`Link: ${children}`}         FLAGGED
aria-label={String(props.children)}      FLAGGED
aria-label={props['aria-label']}         ignored
aria-label="Back to reference 1"         ignored
aria-label={label}                       ignored
```

The three ignored cases are names someone decided on, rather than a stringified React node. Breaking the remaining selector fails the test.

The fixture helper moved to `test/ui/restrictedSyntax.ts` now that a second rule uses it. It still lints through the repo's **own** eslint binary and config rather than `RuleTester`, so it cannot drift from what CI enforces.

### Not fixed here

The `pre` override sets `role="text"`, which is not in the ARIA specification, alongside `aria-label="Code block"`. My issue flagged it as worth reviewing in the same pass, so I looked: I could **not** demonstrate a problem. Playwright's ARIA snapshot is normalised and reads the same in Chromium and WebKit, which is where `role="text"` would matter. Rather than assert a bug I could not reproduce, I left it for someone who can check against a real screen reader.

### What is not verified

That a real screen reader announces the link text as expected. Chromium's accessibility tree is not a screen reader, and `TypingEffect` cannot be rendered by this suite at all — it imports ESM-only `react-markdown` (#678).

| check | result |
|---|---|
| `npm test` | 1329 passed, 105 suites |
| `npm run build` | exit 0 |

---

---

### DIFF
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -119,11 +119,14 @@ const config: ReturnType<typeof antfu> = antfu({
       // selector was redundant, and only showed up as redundant because
       // breaking it left the test still passing.
       //
-      // Matches the name, not the value, so a renamed binding
-      // (`const { children: kids } = props`) passes. That is the limit of a
-      // syntactic selector rather than something to widen: guessing at names
-      // would cost false positives, and the shape this exists to stop is the
-      // one that was actually written.
+      // Matches the name, not the value, so it errs in both directions: a
+      // renamed binding (`const { children: kids } = props`) passes, and a
+      // `children` belonging to something other than React would be reported.
+      // Both are the limit of a syntactic selector rather than something to
+      // widen — closing either needs scope or type information the rule does
+      // not have — and the shape this exists to stop is the one that was
+      // actually written. A false report is at least loud, and silenced at
+      // the line with a reason.
       selector: 'JSXAttribute[name.name=\'aria-label\'] Identifier[name=\'children\']',
       message: 'Do not build aria-label from children: it stringifies to "[object Object]" unless the child is a plain string, and aria-label replaces the accessible name the element already had.',
     }],
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -103,6 +103,23 @@ const config: ReturnType<typeof antfu> = antfu({
     }, {
       selector: 'JSXAttribute[name.name=\'className\'] TemplateElement[value.raw=/(^|\\s)sr-only(\\s|$)/]',
       message: 'Nothing defines .sr-only in MAIDR, so this element is visible. Use the visuallyHidden style object from @ui/visuallyHidden.',
+    }, {
+      // `children` is a React node, not a string. Interpolating it yields
+      // "[object Object]" for anything but a bare string — an element for
+      // `[**bold**](url)`, an array for mixed content — and `aria-label`
+      // replaces the accessible name rather than supplementing it, so the
+      // visible text is not a fallback. A chat link announced as
+      // "Link: [object Object]" is where this came from.
+      //
+      // An element's own text is already its accessible name. Deriving one
+      // from children can at best reproduce that and at worst destroy it.
+      // One selector, not two. `props.children` is a MemberExpression whose
+      // property is an Identifier named `children`, so matching the identifier
+      // covers the member access as well — a separate MemberExpression
+      // selector was redundant, and only showed up as redundant because
+      // breaking it left the test still passing.
+      selector: 'JSXAttribute[name.name=\'aria-label\'] Identifier[name=\'children\']',
+      message: 'Do not build aria-label from children: it stringifies to "[object Object]" unless the child is a plain string, and aria-label replaces the accessible name the element already had.',
     }],
   },
 });
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -118,6 +118,12 @@ const config: ReturnType<typeof antfu> = antfu({
       // covers the member access as well — a separate MemberExpression
       // selector was redundant, and only showed up as redundant because
       // breaking it left the test still passing.
+      //
+      // Matches the name, not the value, so a renamed binding
+      // (`const { children: kids } = props`) passes. That is the limit of a
+      // syntactic selector rather than something to widen: guessing at names
+      // would cost false positives, and the shape this exists to stop is the
+      // one that was actually written.
       selector: 'JSXAttribute[name.name=\'aria-label\'] Identifier[name=\'children\']',
       message: 'Do not build aria-label from children: it stringifies to "[object Object]" unless the child is a plain string, and aria-label replaces the accessible name the element already had.',
     }],
--- a/src/ui/components/TypingEffect.tsx
+++ b/src/ui/components/TypingEffect.tsx
@@ -156,15 +156,15 @@ export const TypingEffect: React.FC<TypingEffectProps> = memo(({ text, isUser, m
             pre: ({ node, ...props }) => (
               <pre {...props} role="text" aria-label="Code block" />
             ),
-            // Fall back rather than overwrite, the way the `img` override
-            // below already does. A prop after a spread wins unconditionally,
-            // so this used to discard any `aria-label` the pipeline had set —
-            // and the one case that has one is the footnote backref, whose
-            // visible text is a bare return arrow. It announced as "Link: ↩"
-            // in place of "Back to reference 1".
-            a: ({ node, ...props }) => (
-              <a {...props} aria-label={props['aria-label'] ?? `Link: ${props.children}`} />
-            ),
+            // No `a` override. A link's accessible name comes from its own
+            // text, which is right in every case and needs no help: the one
+            // this used to build — `Link: ${children}` — was a worse copy of
+            // that when children was a string, and `[object Object]` when it
+            // was anything else. `aria-label` replaces the name rather than
+            // supplementing it, so there was no fallback to the visible text.
+            //
+            // Dropping it also lets the footnote backref keep the label
+            // remark-gfm gives it, without a fallback expression to get wrong.
             img: ({ node, ...props }) => (
               <img {...props} alt={props.alt || 'Image in message'} />
             ),
--- a/src/util/markdownSanitize.ts
+++ b/src/util/markdownSanitize.ts
@@ -122,10 +122,10 @@ export const MATHML_ATTRIBUTES: readonly string[] = [
  *
  * The list used to also carry `role`, `ariaLabel`, `ariaBusy`, `ariaLive` and
  * `ariaAtomic`, on the reasoning that nothing in the pipeline emitted any of
- * them — the `role` and `aria-label` `TypingEffect` puts on `<pre>` and `<a>`
- * are React props, applied after sanitisation — and that while they were
- * misspelled they were inert. Correcting the spelling would have made them
- * live for the first time, so they were dropped instead.
+ * them — the `role` and `aria-label` `TypingEffect` puts on `<pre>` are React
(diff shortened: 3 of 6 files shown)


jooyoungseo:
Recorded point 1 at the rule site — `2bdab0a`. The other three are acknowledgments; nothing to change.

### The renamed-binding hole is real, and now written down

Confirmed rather than reasoned about, since "would slip past" is the kind of claim worth testing:

```
FLAGGED  <a aria-label={`Link: ${props.children}`} />
MISSED   const { children: kids } = props; <a aria-label={`Link: ${kids}`} />
```

Exactly as you said. It matches the identifier's *name*, not the value behind it.

Left as a limit rather than widened, and I want to be clear about why, because I have twice argued in these reviews that a guard which silently passes is worse than none:

The `sr-only` scanner's holes were reachable by **ordinary code** — a template with an interpolation, a hyphenated class. This one requires renaming `children` while writing the exact bug the rule names. Catching it needs scope analysis or type information, which `no-restricted-syntax` does not have, and approximating with name patterns buys false positives on every unrelated `kids`/`content`/`nodes`.

So the distinction I would draw is between a guard that is *incomplete* and one that is *wrong about what it covers*. This is the first. It is now stated in the config rather than only in this thread, which is where the next person will be standing.

### 2 — subprocess cost

Agreed, and the number is small enough to be worth stating: two spawns, ~1.4s each, in a suite that runs ~40s. The threshold to watch for is not the count of rules but the count of *files* testing them, since the helper spawns once per call. If a third guard lands, batching its cases into an existing fixture costs nothing and saves a spawn.

### 3 and 4

Both are limits I would rather have on the record than smoothed over, so thank you for reading them as disclosure rather than hedging.

Worth adding one thing to 4, though. You are right that a regression in a *different* shape would go uncaught — but the specific hole is narrower than "any regression". `aria-label` on a chat link now comes from exactly two places: the sanitiser allowlist, which is pinned by `markdownSanitize.test.ts`, and nothing else, because the override is gone. The uncovered case is a *new* override being added, which is a deliberate act rather than a drift.

### On verifying the schema claim

Checking that `ariaLabel` really is allowlisted at `markdownSanitize.ts:191` is the check I would most want a second pair of eyes on — the footnote backref keeping its label depends entirely on it, and if it were absent the removal would silently make that link unlabelled instead of correctly labelled. That is a failure that looks identical to success from the diff.

| check | result |
|---|---|
| `npm test` | 1329 passed, 105 suites |
| `npm run build` | exit 0 |

---

jooyoungseo:
Both taken — `d0068cf`. I put the stale comment in this PR rather than a follow-up, because this change is what made it false.

### The `markdownSanitize.ts` comment

Good catch, and the reason it matters is that the sentence exists to *justify* a decision:

> the `role` and `aria-label` `TypingEffect` puts on `<pre>` and `<a>` are React props, applied after sanitisation

That was the argument for dropping `ariaLabel` from the global list. Removing the link override made half of it false, so the justification now cited evidence that had stopped being true — which is worse than a merely stale comment, because the next person reading it has no way to tell.

You called it adjacent and worth a follow-up. I disagree only on the venue: it is a different file, but it was accurate before this diff and false after it, so it is this PR's to fix.

### The rule errs in *both* directions

You are right, and my comment only described one of them. Checked the other rather than taking it:

```
aria-label={`Link: ${props.children}`}        FLAGGED
const { children: kids } = props … ${kids}    MISSED           (false negative)
aria-label={String(astNode.children.length)}  FLAGGED           (false positive)
```

The second is the one worth noting for a different reason than the first: a `children` that is not React's would be reported. Both directions are now in the config.

Neither is worth widening for — closing either needs scope or type information the rule does not have. But the two are not equally bad, and that is the part I would rather have written down than left implicit: **a false report is loud**, and gets silenced at the line with a reason. A false negative is the shape that has caused every real problem in this area — three times on #699 alone.

### The four you verified

`ariaLabel`/`ariaDescribedBy` allowlisted, no `target` anywhere in the schema, the selector being additive to the existing rule block rather than clobbering the `sr-only` ones, and no other `aria-label`-from-children in `src/`. That third one I had not explicitly checked — `no-restricted-syntax` takes an array and I appended to it, but "appended to the same block" and "did not replace the earlier entries" are different claims, and only the tests happening to still pass told me. Worth having had someone read it.

| check | result |
|---|---|
| `npm test` | 1329 passed, 105 suites |
| `npm run build` | exit 0 |

---

jooyoungseo:
No code change this round — one item recorded on #678, and one that is genuinely a maintainer's call rather than mine.

### The behaviour change needs a maintainer, and I agree

You are right to want this confirmed rather than inferred. Stating it plainly so it is easy to say no to:

**Every chat link stops announcing with a `"Link: "` prefix**, including the ones that were previously correct — a plain-text link went from *"Link: download the data, link"* to *"download the data, link"*.

My argument is that the prefix duplicated the role a screen reader announces anyway, and that no `aria-label` at all is better than one derived from content, because it cannot be wrong. But that is a judgement about announcement style, not a bug fix, and it is reversible: the narrower change is to keep the prefix and extract text properly from the React tree. If a maintainer prefers that, say so and I will do it — it is more code, but it is not hard.

What is **not** optional either way is that `[object Object]` has to go.

### Your test suggestion, recorded with the cases

Added to **#678** as item 7 rather than left here, since that is where the work will start. Kept your framing — assert the rendered accessible name, not the markup — and named the specific cases:

| markdown | expected name |
|---|---|
| `[**bold**](url)` | `bold` |
| `` [a `b` c](url) `` | `a b c` |
| `[[alt](img.png)](url)` | `alt` |
| footnote backref | `Back to reference 1` |

The fourth is the one I would not want dropped from the list. It is the only `aria-label` that should reach a chat link at all, it comes from the pipeline rather than from React, and a test covering only the first three would pass while it silently disappeared. That is the failure this PR's removal is closest to causing, so it is the one worth pinning first.

### On the blind spots

Both documented, both accepted, and the asymmetry between them is in the config now: a false report is loud and can be silenced at the line with a reason, while a false negative is the shape that has caused every real problem in this area.

| check | result |
|---|---|
| `npm test` | 1329 passed, 105 suites |
| `npm run build` | exit 0 |

---

PR #701, opened 2026-08-01
test: add an ESM Jest project so the pipeline can be tested end to end

### LINKED ISSUE #678: test: add an ESM-aware Jest project so the markdown pipeline can be tested end to end

jooyoungseo:
# Pull Request

## Description

`markdownSanitize.test.ts` checks what the schema *contains* — every element KaTeX emits is named, attributes use hast property names, no key beyond `tagNames`/`attributes` is declared. It cannot check `rehype-sanitize` **applying** it, because the whole unified stack is ESM-only and the suite compiles to CommonJS:

```
SyntaxError: Unexpected token 'export'
    at node_modules/rehype-sanitize/index.js
```

So the properties the schema exists for were verified by hand, once, and then never again.

That gap is not theoretical. An allowlist can name `table` while a table still arrives as `BarValueJan45.2`, and it can name an attribute under a spelling that matches nothing. Both happened, in #695 and #675.

## Related Issues

Closes #678. Unblocks the deferred checks recorded there from #695, #699 and #700.

## Changes Made

**A second Jest project, not a second config.** `projects` in `jest.config.ts` means one `npm test` covers both, so `ci.yml` needs no change. The naming keeps them apart with no ignore rule: `*.test.ts` requires a name ending `.test.ts`, and `*.esm-test.ts` ends `-test.ts`.

**`scripts/test.js` supplies `--experimental-vm-modules`.** Jest can only import ESM with it, and it is a process-level flag no config can set. The issue offered `cross-env` or a `scripts/` runner and judged the runner the better fit — it adds no dependency and matches `scripts/build.js`. It appends to `NODE_OPTIONS` rather than replacing it (CI runners set memory limits there), forwards argv so `npm test -- --watch` behaves as before, and reports a signal as failure rather than as success.

`NODE_OPTIONS=… jest` in the npm script would have been the one-liner, but that syntax is invalid in the shell npm uses on Windows, so it would work for some contributors and not others.

**Thirteen cases** through the real chain — `remarkParse` → `remarkGfm` → `remarkMath` → `remarkRehype` → `rehypeKatex` → `rehypeSanitize`: MathML structure surviving, `aria-hidden` surviving, tables with alignment, task lists, list numbering, the footnote label, `javascript:` stripped from links and images, TeX unable to smuggle in `href` or `mglyph`, and raw HTML.

Assertions are on the **hast tree**, not serialised HTML. `rehype-stringify` not being installed is the smaller reason: `react-markdown` renders the tree to React elements and never produces an HTML string, so the tree is the last thing on the real path that exists.

## Checklist

- [x] I have updated the documentation, if applicable. — `.claude/rules/testing.md`, `CLAUDE.md`, and the generated Copilot instructions.

### Two things this turned up immediately, which is the point

**1. The schema described raw HTML wrongly, in two places.** Both said it "is escaped to text". It is not:

```
``&lt;img src=x onerror=alert(1)&gt;``   ->  tags=[]      text=""
<script>alert(1)</script>      ->  tags=[]      text=""
a <b>bold</b> c                ->  tags=["p"]   text="a bold c"
```

A block is **dropped outright** — no element and no text. An inline tag is removed while its words are kept. Neither is escaping, which would have left `onerror` readable in the transcript.

The security conclusion is unchanged and slightly stronger, but both comments cite the mechanism as the reason a decision is safe, so a reader checking the reasoning would have been checking against something false. Corrected.

**2. My own first checkbox case was decorative.** It rendered `- [x] done` and asserted the type was `checkbox` — which passes whether the schema pins the *value* (`['type', 'checkbox']`) or merely names the attribute (`'type'`), because remark-gfm never emits anything else.

The difference only appears for a value the pipeline cannot produce, so it now sanitises a hand-built tree carrying `type="hidden"`. Verified by loosening the pin: the case fails now and did not before.

That one is worth flagging because it is the failure mode this whole project exists to remove — a test that reads like it checks something and does not. Finding one in the first thirteen cases is a reasonable argument for the project.

### Every case verified to fail against the defect it names

| reverted | case that fails |
|---|---|
| `mfrac` removed from the allowlist | fraction keeps its structure |
| `['type','checkbox']` → `'type'` | input type rewritten rather than dropped |
| a partial `protocols` key added | `javascript:` stripped from a link |

The third is the one that shows the value most plainly: the schema-shape suite forbids the *key*, but only a real run shows the consequence — that `javascript:` starts surviving.

### What this unblocks

#678 carries a ranked checklist of what could not be tested before. Items 2, 3 and 5 are covered here. Items 1, 4, 6 and 7 need `TypingEffect` itself rendered under jsdom — the project makes that possible for the first time, since `react-markdown` is ESM, but it is a separate change and not folded in here.

| check | result |
|---|---|
| `npm test` | 1342 passed, 106 suites |
| `npm run build` | exit 0 |

---

---


jooyoungseo:
All four taken — `ce126a2`. Two of them were things I had already argued for elsewhere and then failed to apply here.

### The phantom dependencies

You are right, and it is my own objection coming back around. I raised exactly this against a transitive `js-yaml` on #692, and used it on #695 to decline `rehype-stringify`. Then I imported four undeclared packages here without noticing.

```
unified@11.0.5        PHANTOM
remark-parse@11.0.0   PHANTOM
remark-rehype@11.1.2  PHANTOM
@types/hast@3.0.4     PHANTOM
hast-util-sanitize    PHANTOM
```

Declared at the versions already resolved, so nothing new is downloaded — this pins what was being relied on by luck.

**The fifth is gone rather than declared.** The checkbox case called `hast-util-sanitize` directly; it now runs the hand-built tree through `rehype-sanitize`, which wraps the same code, is already a dependency, and is the one the pipeline itself uses. Fewer dependencies *and* closer to the real path.

**One detail on the lockfile**, since it is the sort of thing that is easier to check than to undo. `npm install --package-lock-only` produced a 65-line diff: it bumped the lockfile's `version` field, and stripped `libc` metadata from optional platform packages as an npm-version artefact. The version drift is **pre-existing on main** — `package.json` is at 3.75.0 and the lockfile at 3.74.0, because semantic-release bumps one and not the other — so "fixing" it here would have been me taking over a field the release process owns.

Reverted and added the four lines by hand instead. `npm ci --dry-run` exits 0 against the pair, which is what CI actually runs.

### The duplicated processor

Correct, and the drift you describe is the real risk — a plugin added to one chain and not the other would make the test quietly exercise a pipeline that does not exist. One processor now, parsed and run by itself.

`.process()` would be the idiomatic single call, but it requires a compiler and there is deliberately none: the tree is the output being asserted on, for the reason in the file header.

### The third "escaped to text"

Found it, and the reason I missed it is worth recording. My grep was `grep -rn "escaped to text"`, and the third occurrence is **wrapped across two lines**:

```
// No `target`. Markdown has no syntax for it, raw HTML is escaped to
// text, and nothing in the plugin chain adds one, ...
```

A phrase search over wrapped prose finds only the ones that happen to fit on one line. Searching for the words separately — `escap\w*` and `raw HTML` — turns up no others, so three is the count.

The irony is not lost: a PR whose point is that comments had drifted from behaviour, with a comment left drifted because of how I searched for them.

### `spawn('jest')` on PATH

Taken. It resolves through `createRequire` now and runs Jest's entry point under `process.execPath`, so `node scripts/test.js` works directly instead of failing with `ENOENT` — verified both ways. That also keeps the flag on the process that needs it rather than depending on a shim.

Separately: the `gFM` typo from the inline comment is fixed. Second time I have made it, both times by copying the `mathML` block above.

| check | result |
|---|---|
| `npm test` | 1342 passed, 106 suites |
| `npm ci --dry-run` | exit 0 |
| `npm run build` | exit 0 |

---

jooyoungseo:
Checked the `tableCellAlignToStyle` claim. The option is real, but it is not in `mdast-util-to-hast` and it does not run before sanitisation — so the conclusion does not follow. Worth writing out, because the mechanism is a good argument *for* the `align` entry rather than against it.

**The option is not in `mdast-util-to-hast` at all.**

```
$ grep -rn "tableCellAlignToStyle" node_modules/mdast-util-to-hast/
$ echo $?
1
```

Nothing, at any version — `13.2.1` is what is installed and what the lockfile pins, so this is the pinned version speaking. What that package does is unconditional:

```js
// node_modules/mdast-util-to-hast/lib/handlers/table-row.js:39
if (alignValue) {
  properties.align = alignValue
}
```

**Where it does live is one layer later.** `hast-util-to-jsx-runtime@2.3.6`, which is what `react-markdown` calls to turn the hast tree into React elements:

```js
// node_modules/hast-util-to-jsx-runtime/lib/index.js:441
if (state.tableCellAlignToStyle && key === 'align' && …) {
  alignValue = value
} else {
  props[key] = value
}
```

and `react-markdown@10.1.0` does not pass the option, so it defaults on (`options.tableCellAlignToStyle !== false`, line 102).

**Which means the ordering is the opposite of the concern.** `rehype-sanitize` is a rehype plugin — it runs on the hast tree. The `align` → `style` rewrite happens *after* that, inside the JSX conversion, on markup React builds itself. The sanitiser never sees a `style` attribute on a cell, so it cannot strip one.

Rendered the real path end to end to confirm — the actual chain from `TypingEffect`, through `renderToStaticMarkup`, against `createChatSanitizeSchema()` unmodified and then with two edits:

```
as-is        : <th style="text-align:left">Bar</th>  <th style="text-align:right">Value</th>
align removed: <th>Bar</th>                          <th>Value</th>
style instead: <th>Bar</th>                          <th>Value</th>
```

Three things fall out of that:

1. Alignment does survive, and the `align` entry on `th`/`td` is what makes it survive — remove it and the styling is gone, which is the failure you were describing, just reached the other way round.
2. Adding `style` to `th`/`td` — the fix the claim implies — does nothing, because there is no `style` attribute at sanitisation time to allow.
3. The assertion in the test (`properties.align === 'left'`) is on the tree, at the only point where the value exists under that name. It is green because the pipeline puts it there, not because CI resolved an older version.

On the wider question — *is this green for the right reason* — that is the right instinct and it is why the checkbox case got rewritten in the first place: it passed against both the pinned tuple and the bare attribute name. This one does not have that problem, since removing the entry fails it. `npm test` on `ce126a2`: **1342 passed, 106 suites, 2 projects**, and all 13 checks are green.

Two follow-ups I am not folding in here:

- The `th`/`td` comment says alignment "carries no meaning for a screen reader". Accurate about `align` and about `text-align`, so the comment stands, but it is now describing an attribute that does not reach the DOM under that name. Not worth a commit on its own.
- Nothing pins `tableCellAlignToStyle` staying on. If `react-markdown` ever passed `false`, `align` would reach the DOM as a deprecated presentational attribute rather than as a style — still allowed by the schema, still rendered by browsers, so it degrades rather than breaks. Not a hazard, but it is the kind of transitive default the ESM project now makes testable, and it belongs on #678's list rather than in this diff.

---

PR #702, opened 2026-08-02
test: render TypingEffect and assert the names a screen reader reads

jooyoungseo:
# Pull Request

## Description

Three chat accessibility bugs were fixed this week — a link announcing as `[object Object]` (#700), a visible "Footnotes" heading (#699), and a footnote backref announcing as "Link: ↩" (#695) — and **all three shipped without a regression guard**. Each is a rendered-output bug, and nothing in the suite could render the component: `TypingEffect` imports `react-markdown`, which is ESM-only, and the CommonJS project cannot load it.

#701 removed that blocker. This is the first thing built on it.

## Related Issues

Covers items **1, 4, 6 and 7** from the checklist on #678, plus the working half of item 5. Adds two `it.failing` pins for #696. Filed #703 for the upstream Jest bug described below.

## Changes Made

**`test/ui/typingEffect.esm-test.tsx`** — seven cases, asserted through `getByRole(role, { name })` rather than attribute reads. The accessible **name** is the contract, and in two of the three bugs above the attribute was present and wrong, so an attribute assertion would have passed.

| case | the bug it pins |
|---|---|
| `[**bold**](url)` → `bold` | #700 — `Link: ${children}` stringified to `[object Object]` |
| `` [a `b` c](url) `` → `a b c` | #700 — same, across three children |
| `[[alt](i.png)](url)` → `alt` | #700 — same, image child |
| footnote backref → `Back to reference 1` | #695 — the override clobbered remark-gfm's own label |
| footnotes heading hidden by clip | #699 — visible heading in every footnoted response |
| heading still in the a11y tree | why the clip pattern and not `display: none` |
| `aria-describedby` resolves | the half of the footnote wiring that works |

**Verified against the defects they name**, not just observed green:

| reverted | cases that fail |
|---|---|
| the `a` override restored | all four link-name cases |
| `visuallyHidden` dropped from the `h2` override | the heading case |
| the clock advance neutered to zero | eight of the nine |

**Two `it.failing` pins for #696**, both confirmed to fail as plain `it` first:

```
● should point the reference at the footnote it names (#696)
  Received: null
● should give two messages distinct footnote ids (#696)
  Expected: 6   Received: 3
```

The first is the double-prefix: `mdast-util-to-hast` writes `#user-content-fn-1`, `hast-util-sanitize` prefixes every `id` again to `user-content-user-content-fn-1`, and `href` is not in `clobber`, so the link announces normally and goes nowhere. The second is the multi-message case a single-document test cannot see at all.

**`scripts/test.js` runs one Jest process per project** — see below. This is the substantive part of the PR, and where eleven rounds of review went.

**`src/util/version.ts` and `jest.config.ts`** now import JSON as a default. Named exports from a JSON module are a bundler extension; a real ES module for JSON exposes only `default`, so `import { version } from '../../package.json'` throws the moment anything reaches it as ESM — which `TypingEffect` does via `@util/katex` → `@util/diagnostics`. `jest.config.ts` had the same thing with `compilerOptions` and hit it once a test in the `esm` project imported the config.

## Checklist

- [x] I have updated the documentation, if applicable. — `.claude/rules/testing.md` and the generated Copilot instructions.

### The Jest bug this turned up

Adding the component test made the suite fail in ways that did not reproduce when either project ran alone: **43 suites red with a warm cache, 3 with `--no-cache`, a different set each run.**

`jest-resolve` memoises "is this file ESM?" keyed on the path alone, ignoring the asking project's `extensionsToTreatAsEsm`:

```js
// node_modules/jest-resolve/build/shouldLoadAsEsm.js:44
let cachedLookup = cachedFileLookups.get(path);
if (cachedLookup === undefined) {
  cachedLookup = shouldLoadAsEsm(path, extensionsToTreatAsEsm);
  cachedFileLookups.set(path, cachedLookup);   // <- key is path only
}
```

The memo is process-wide, so the first project to ask about `src/type/grammar.ts` answers for both. Reduced to two files:

```
$ jest --runInBand "typingEffect.esm-test|version.test"
PASS esm  test/ui/typingEffect.esm-test.tsx
FAIL unit test/util/version.test.ts
     Must use import to load ES Module: /home/user/maidr/src/util/version.ts
```

Order-dependent, which is where the nondeterminism came from: a warm cache reorders by recorded duration, and the component test is slow enough to be scheduled first.

**It was latent before this PR rather than absent.** `main` is green three times over and green under `--runInBand` — but only because the `esm` project imported one leaf module, so the overlap was a single file and the scheduler never happened to put it first. Rendering a component pulls in the store, the view models and the grammar, so the two projects now share most of `src/`.

Nothing in a Jest config scopes a process-wide memo, so `scripts/test.js` runs **one process per project**. #703 tracks reporting it upstream; if it is ever fixed, most of this machinery can go.

### What the split then had to get right

Each of these was a real defect found in review, not a hypothetical:

- **Every project runs even after one fails.** `npm test` is CI's only test step, so stopping early hid every failure in the remaining projects — including the accessibility tests this PR adds.
- **`--selectProjects` is intercepted.** It is variadic, so `--selectProjects unit esm` reached a single process and undid the split silently. It is now taken out of the argument list and added back one project at a time.
- **A path filter still works.** One process per project turns Jest's aggregate "at least one test found" into a per-project question, so `npm test -- test/util/version.test.ts` failed on the project the filter was never aimed at. The runner asks `jest --listTests --json` what the filter matches, scoped to the projects that will run, and tolerates a project matching nothing only when something matched somewhere — so a typo and a broken `testMatch` both stay loud.
- **A `--selectProjects` that names nothing is Jest's to reject**, and one that names something still splits. Handing the latter over untouched ran both projects in one process and reintroduced the memoisation clash.
- **Coverage is written to `coverage/<project>`.** Two runs would otherwise write the same directory and the second would replace the first silently. There is no combined report any more; each is scored against the whole of `src/`, so `coverage/esm` reads as most of the tree at 0%.
- **Watch mode cannot be sequential**, since a watcher does not exit, so it watches one project and says so rather than quietly narrowing.

The decisions live in `scripts/testArgs.js` — side-effect-free, so `test/scripts/testArgs.esm-test.ts` can exercise them directly — and `test/scripts/testRunnerProjects.esm-test.ts` compares the runner's project list against the config's `displayName`s by importing both. A project missing from that list would not fail loudly; its suites would simply never run while `npm test` reported success.

### Verification

| check | result |
|---|---|
| `npm test` | 109 suites, 1371 tests |
| `npm run build` | exit 0 |
| `npx eslint .` | exit 0 |

Runner exit codes, all checked by hand:

| invocation | exit | |
|---|---|---|
| `--selectProjects` | 1 | Jest's own error |
| `--selectProjects=` | 1 | same |
| `--selectProjects --selectProjects unit esm` | 0 | and split, not handed over |
| `--selectProjects=esm test/util/version.test.ts` | 1 | scoped filter matching nothing |
| `test/util/version.test.ts` | 0 | matches one project only |
| `test/util/versoin.test.ts` | 1 | typo |
| `--coverage=false` | 0 | an argument, not a filter |
| `esm` `testMatch` broken | 1 | bare and with a flag |

Run the suite and the build separately, not concurrently — running both at once is what produced the first confusing red I chased.

jsdom is not a screen reader. These pin the wiring; they do not replace checking with real assistive technology.

---


jooyoungseo:
All three runner points were right, and the first one was the one I should have caught myself. Fixed in `fb430a5`.

**Stopping at the first failing project.** `npm test` is CI's only test step, so this was a straight loss of feedback — and the tests this PR exists to add are the ones in the project that gets skipped. Every project runs now; the runner still exits with the first non-zero code. Verified with a deliberately failing unit test:

```
[test] project: unit
Test Suites: 1 failed, 106 passed, 107 total
[test] project: esm
Test Suites: 2 passed, 2 total
$ echo $?
1
```

Before the fix the `esm` block was simply absent.

**`--selectProjects unit esm` re-opening the clash.** Also right, and worth more than a comment — the whole point of the change is an invariant, and one a caller can silently break is not one. The flag is now taken out of the argument list and added back a project at a time, so naming several runs several processes:

```
$ node scripts/test.js --selectProjects unit esm
[test] project: unit
[test] project: esm

$ node scripts/test.js --selectProjects=esm
[test] project: esm
```

Both spellings handled. Watch mode honours an explicit single selection and only prints the narrowing notice when it had to choose.

**Coverage fragmentation.** Correct that it is not just a relocation — `collectCoverageFrom` is repo-wide, so each report scores all of `src/` against what that one project loads and `coverage/esm` reads as most of the tree at 0%. Nothing is gated on it, but "coverage dropped" is the wrong conclusion to leave lying around. `testing.md` now says to read the two separately and that there is no combined number.

**On the `renderMessages()` no-argument case** from the Copilot review — the mechanism checks out, `advanceTimersByTime(-Infinity)` throws `TypeError: Negative ticks are not supported`. I did not take the suggested guard, though: defaulting the value would be worse than the throw, because a call with no messages would render an empty document and quietly satisfy every "should not" assertion in the file. That is the exact failure mode this PR exists to remove. Requiring the first message as its own parameter makes the empty call a type error instead:

```
test/ui/typingEffect.esm-test.tsx(213,1): error TS2555: Expected at least 1 arguments, but got 0.
```

No runtime guard, no silent-pass path.

**On "not verified — no `node_modules` in this environment":** fair flag, and the right instinct given the bug is order-dependent. `npm test` twice on `fb430a5`: **108 suites, 1353 tests**, identical both runs; `build`, `tsc` and `eslint` exit 0. CI on the previous head had Unit Tests, Type Check, Build, Lint, CodeQL and Copilot-instructions-in-sync all green, which is the per-project runner working on a machine that is not mine.

---

jooyoungseo:
Reproduced, and it was worse than "worth a quick local check" — it was a plain break of the thing the runner's own doc comment promised. Fixed in `19931b2`.

```
$ node scripts/test.js --coverage=false test/util/version.test.ts
[test] project: unit
PASS unit test/util/version.test.ts   ✓ 2 passed
[test] project: esm
No tests found, exiting with code 1
Pattern: test/util/version.test.ts - 0 matches
$ echo $?
1
```

Exactly your diagnosis: one process per project turned Jest's aggregate "at least one test found" into a per-project question. And you were right that `test/util` masked it — that directory happens to hold files from both projects, which is why the example in the comment kept working while the single-file case, the one people actually reach for, did not.

**I did not take the suggested fix as written.** `--passWithNoTests` on every per-project run except the unfiltered path repairs the reported case but lets two others through:

- a filter matching nothing **anywhere** is a typo, and should stay loud;
- a project matching nothing on an unfiltered run is a broken `testMatch` — which means its suites silently never ran while `npm test` reported success, the exact failure `testRunnerProjects.test.ts` exists to prevent.

Any rule keyed on "were there arguments" has the same hole, because `--coverage=false` is an argument and not a filter. I wrote that version first and it did let the broken-`testMatch` case pass — worth saying, since it is the reason I went further.

So the runner asks rather than guesses: `--listTests` twice, once with the arguments and once without, and tolerance only when the first is a **non-empty proper subset** of the second. `--listTests` resolves paths and loads no module, so it is safe to ask across all projects in one process — which is precisely the aggregate view the split had lost. Two cheap resolutions, and only on a run that passed arguments; an unfiltered run skips them.

Verified, exit codes:

| invocation | exit | |
|---|---|---|
| `test/util/version.test.ts` | 0 | was 1 — the reported case |
| `test/ui/typingEffect.esm-test.tsx` | 0 | the mirror case |
| `test/util` | 0 | still works |
| `test/util/versoin.test.ts` | 1 | typo, Jest prints its own diagnostic |
| `--coverage=false` | 0 | an argument, not a filter |
| *(bare)* | 0 | |
| `esm` `testMatch` broken, bare | 1 | |
| `esm` `testMatch` broken, with a flag | 1 | the case the simpler rule missed |
| a genuinely failing filtered test | 1 | |

A name filter needs none of this — `-t` narrows tests rather than files, and Jest reports the rest as skipped rather than as none found.

**On the two minor notes:**

*Startup overhead* — agreed, and it is the local loop that pays, not CI. Measured: a filtered single-file run is ~7s wall, of which the two `--listTests` resolutions are a bit over a second. Real, and still the right trade against a correctness bug with no config-level alternative.

*The regex scrape in `testRunnerProjects.test.ts`* — correct that a computed `PROJECTS` would make it throw rather than fail meaningfully. It throws with a written message rather than a regex-match error (`scripts/test.js no longer declares a PROJECTS array`), which is the best I could do without a build step, but the maintenance surface is real and worth knowing about.

**On "not verified — no `node_modules`":** the right flag to raise, and this round is why. `npm test` on `19931b2`: **108 suites, 1353 tests**; `build`, `tsc` and `eslint` exit 0. CI on the previous head had everything green except E2E and this review, both of which were still running.

---

jooyoungseo:
All four addressed in `33b6987`. The coverage point was the right one to lead with — writing the tests it asked for immediately found a bug in the fix for the point above it.

**The missing tests.** Fair, and the repo's own rule is explicit about it. The decisions now live in `scripts/testArgs.js` — `takeSelection`, `hasPathFilter`, `isNarrowed` — with the runner left as the part that spawns things, and `test/scripts/testArgs.esm-test.ts` covers all three, including the comma form you asked about.

They run in the `esm` project: the module is ESM (the repo is `"type": "module"`) and the CommonJS project cannot require it — the same split the runner exists to manage, arrived at from the other side. I tried the `mathStylesheet.test.ts` subprocess arrangement first; it works, but costs a process per call and gives up the type checking `scripts/testArgs.d.ts` provides.

**The `--selectProjects` + filter scope gap — confirmed, and it was not as narrow as either of us thought.** Both `--listTests` calls are now scoped the same way the real runs are. But the fix did not work at first, and why is worth recording: with `--selectProjects`, Jest announces the project on **stdout**, alongside the list.

```
$ jest --listTests test/util/version.test.ts --selectProjects esm 2>/dev/null | cat -A
Running one project: esm$
$
```

The correct answer here is *no matching files*. My parser counted that banner as one, so an empty result looked like a match and granted exactly the tolerance the fix was meant to withhold — the scoped case still exited 0 after the "fix", and only looking at the bytes explained it. Absolute paths only now.

**The probe firing on any argument.** Right, and `npm run test:coverage` was spawning two `--listTests` resolutions to conclude that `--coverage` is not a path filter. It looks for a positional first now. The approximation is one-directional, which is why it is safe: a space-separated flag value like `--maxWorkers 2` looks like a path and costs a wasted probe, while a path can never look like a flag. That case is pinned in the tests with the reasoning attached.

**The comma form.** Documented in `takeSelection`'s contract and pinned: `--selectProjects=unit,esm` yields the single name `unit,esm`, which is left to fail as the unknown project it is rather than being quietly reinterpreted — Jest does not accept the form either.

Exit codes, all verified:

| invocation | exit | |
|---|---|---|
| `--selectProjects=esm test/util/version.test.ts` | 1 | the case you flagged |
| `--selectProjects=esm test/ui/typingEffect.esm-test.tsx` | 0 | |
| `--selectProjects=unit test/util/version.test.ts` | 0 | |
| `test/util/version.test.ts` | 0 | |
| `test/ui/typingEffect.esm-test.tsx` | 0 | |
| `test/util` | 0 | |
| `test/util/versoin.test.ts` | 1 | typo |
| `--coverage=false` | 0 | |
| `esm` `testMatch` broken, bare and with a flag | 1 | |

**On commit hygiene:** you are right about `fb430a5`, and it was three changes. I have kept it rather than rewriting pushed history mid-review, since the force-push would invalidate the review threads above it — but the point is taken and this round is one logical change.

`npm test` on `33b6987`: **109 suites, 1365 tests**; `build`, `tsc` and `eslint` exit 0.

---

jooyoungseo:
One of these was right and one wasn't. `62cf752`.

**The `-t` comment was wrong, and it's the kind of wrong worth fixing.** Confirmed:

```
["-t","some name"]  -> hasPathFilter = true
["--coverage"]      -> hasPathFilter = false
["test/util"]       -> hasPathFilter = true
```

A name filter does reach the check — its value doesn't begin with a dash, so nothing can tell it from a path. It then finds nothing narrowed and the run stays strict, which is the right outcome, but "does not need it" described a route past the check that no argument takes. Same shape as the three "raw HTML is escaped to text" comments corrected on #701: conclusion right, mechanism misdescribed, and the mechanism is what a reader checks the conclusion against. Reworded, and pinned as a case so the claim is executable rather than prose.

**The multi-line reformatting concern doesn't hold.** `[^\]]` is a negated character class, not a dot, so it matches newlines and the pattern already reads across lines. Verified by reformatting the declaration and re-running the guard:

```js
const PROJECTS = [
  'unit',
  'esm',
];
```
```
✓ should name every project the jest config declares
```

What would actually stop it matching is a *computed* list rather than a literal, and the `throw` is there so that says so instead of surfacing as a comparison against an empty array. I've written that down in the helper, since the inference you drew is the natural one to draw from the regex alone.

**On `hasPathFilter` reaching `-t`:** agreed it isn't worth narrowing further. Distinguishing `-t "name"` from a path means parsing which flags take values, which is the yargs quirk this whole area keeps tripping over — two extra `--listTests` resolutions on a `-t` run is the cheaper end of that trade.

`npm test` on `62cf752`: **109 suites, 1366 tests**; `tsc` and `eslint` exit 0. `build` was green on `33b6987` and this round touches only comments and one test case.

---

jooyoungseo:
Not intentional — it fell out of the parse loop. Fixed in `9e20018`.

You asked the right question, because the consequence is worse than the flag being dropped: the runner read "named no project" as "named every project" and ran both. Jest refuses outright:

```
$ jest --selectProjects
The --selectProjects option requires the name of at least one project to be specified.
Example usage: jest --selectProjects my-first-project my-second-project
$ echo $?
1
```

**The obvious fix does not work, which is worth recording.** Putting the bare flag back into the forwarded arguments looks right and isn't — Jest absorbs it once the per-project flag is added alongside:

```
$ jest --listTests --selectProjects --selectProjects unit
/home/user/maidr/test/command/pointerGuidance.test.ts
… (runs normally)
$ echo $?
0
```

So `takeSelection` reports the case as `dangling` and the runner hands the arguments over untouched. That is the only arrangement where the caller sees Jest's own message rather than a paraphrase or a silent full run:

```
$ npm test -- --selectProjects
The --selectProjects option requires the name of at least one project to be specified.
$ echo $?
1
```

Pinned in `testArgs.esm-test.ts`, with the failed approach written down so nobody re-simplifies it back.

**On `test/scripts/` not fitting "`test/` mirrors `src/`":** correct, and now called out in `testing.md` — it covers `scripts/` and the workflow bodies rather than a layer of the application.

**On `renderMessages`'s timer math:** thanks for checking it against `TypingEffect.tsx:125` rather than taking it on trust. Both sides walk the raw `text.length`, so they cannot drift.

Runner exit codes re-checked, including the new case:

| invocation | exit |
|---|---|
| `--selectProjects` *(dangling)* | 1 |
| `--selectProjects=esm test/util/version.test.ts` | 1 |
| `--selectProjects=esm test/ui/typingEffect.esm-test.tsx` | 0 |
| `--selectProjects unit esm` | 0 |
| `test/util/version.test.ts` | 0 |
| `test/util/versoin.test.ts` *(typo)* | 1 |
| `--coverage=false` | 0 |

`npm test` on `9e20018`: **109 suites, 1367 tests**; `build`, `tsc` and `eslint` exit 0.

---

jooyoungseo:
Good catch, and confirmed rather than reasoned about. `1a15ef6`.

Same id, same text, no clock advance either time:

```
first render   ->  queryByText(TEXT) === null      (nothing on screen)
second render  ->  2 matches                        (the whole message, timers never advanced)
```

Exactly your reading: `completedAnimations` is module-level, keyed on id + body, nothing resets it, so the second case rendering a given fixture short-circuits at `TypingEffect.tsx:112` and reads a settled component.

Ids are now unique across the file, and the difference is measurable rather than notional. Neutering the clock advance to zero:

| | cases failing |
|---|---|
| before | some — the rest coasted on the first render's animation |
| after | **8 of 9** |

The one that still passes is an `it.failing` pin, which passes precisely because its body throws on an empty document.

You are right that it changed no result today — every assertion is on the end state — but the order-dependence was real, and the case you describe is the one that would have been bitten: a test asserting the live region stays empty while `isTyping` is true would silently take the wrong branch depending on where it landed in the file. The reasoning is now in the helper's docstring so the next person does not reintroduce it by tidying the ids back to `m0`.

**On the two non-blocking notes:**

*Sequential startup cost* — fair that I gave no numbers. Measured locally on this branch: `unit` ~20s, `esm` ~3s, so the second process is a few seconds of Node/ts-jest startup on a ~25s run. CI's Unit Tests job has stayed in the 60–90s band across every push on this PR, which is where it was before. Not enough to be worth a follow-up issue, and I would rather not claim a precise regression figure from a shared CI runner.

*The `spawn()` check* — thanks for looking. Argument array, no `shell: true`, so forwarded `process.argv` cannot be shell-injected.

`npm test` on `1a15ef6`: **109 suites, 1367 tests**; `tsc` and `eslint` exit 0. `build` was green on `9e20018` and this round touches one test file.

---

jooyoungseo:
Both real. `83ae94c`.

**`--selectProjects=` was worse than "project not found".** Your trace of `takeSelection` is exactly right — `dangling` was only set in the variadic branch. What the runner actually produced:

```
$ npm test -- --selectProjects=
In /home/user/maidr
  0 files checked across 0 projects. Run with `--verbose` for more details.
Pattern:  - 0 matches
```

True, and no help at all. Jest's own answer to that argument turns out to be the *same* error as the bare form:

```
$ jest --selectProjects=
The --selectProjects option requires the name of at least one project to be specified.
```

So the empty value takes the route the bare flag already had, and now says the same thing. Pinned.

**The throw path — you were right that it was the one place a regression could be silent, and testing it corrected my own comment.**

The readers now take source text rather than reading the file, so both branches are reachable. Then the case I wrote to prove the throw *failed*:

```
● should refuse a list it cannot read rather than reporting an empty one
  Expected substring: "no longer declares a PROJECTS array"
  Received function did not throw
```

Your example — `[...OTHER, 'x']` — keeps its brackets, so the pattern matches and returns only the literals. It is still caught, but by the comparison against the config's `displayName`s being short by whatever the spread contributed, not by the throw. "A computed list stops matching" was the tidy claim and the wrong one. What genuinely throws is a list built by a call, which has no brackets at all.

Both are pinned now, with the mechanism named correctly in each:

```
✓ should read an array written one name per line
✓ should refuse a list it cannot read rather than reporting an empty one   (projectNames())
✓ should read a spread array as the names it spells out                    ([...BASE, 'esm'] -> ['esm'])
```

**On the performance note:** measured rather than asserted this time — `unit` ~20s, `esm` ~3s locally, so the second process costs a few seconds of startup on a ~25s run, and a filtered run adds two `--listTests` resolutions worth about a second. CI's Unit Tests job has stayed in the 60–90s band across all nine pushes on this PR. I would rather quote that than a before/after delta from a shared runner, where the noise is larger than the effect.

`npm test` on `83ae94c`: **109 suites, 1372 tests**; `build`, `tsc` and `eslint` exit 0.

---

jooyoungseo:
The first point was the best note this PR has had — "testing the test infrastructure's test infrastructure" is exactly what it was, and you were right that the reason I gave for it was wrong. `d7903f3`.

The runner can't be imported. `scripts/testArgs.js` already can, and exists for that reason. So `PROJECTS` and `WATCH_DEFAULT` moved there and the guard imports both sides.

What that deleted: two regex readers, three cases pinning the regex's own behaviour, and the `throw` that only existed because a regex can stop matching — the very cases from the round before this one.

```
 test/scripts/testRunnerProjects.test.ts | 123 ------------------------------
 6 files changed, 74 insertions(+), 142 deletions(-)
```

A renamed `displayName` or a retyped constant now fails to compile rather than failing to match.

**It needed one more thing, and it's the same bug as `version.ts`.** `jest.config.ts` named-imported `compilerOptions` from `tsconfig.json`:

```
SyntaxError: The requested module './tsconfig.json' does not provide an export named 'compilerOptions'
```

Same bundler-convenience-vs-ESM issue, one config file away, and only reachable once something in the `esm` project imports the config. Default import now; Jest still loads the config as CommonJS, where `esModuleInterop` makes the two equivalent.

**On the `--selectProjects` + positional swallow — checked against Jest rather than reasoned about.** Jest does the same thing:

```
$ jest --listTests --selectProjects unit test/util/version.test.ts
/home/user/maidr/test/util/deepMerge.test.ts
/home/user/maidr/test/command/toggleMonitor.test.ts
…
```

It lists the whole `unit` project — the path was taken as a second project name and the filter had no effect. (Slightly different from your prediction: Jest doesn't reject the unknown name, it ignores it.) Stopping at the first positional would make the runner do something Jest does not, which is the worse surprise, so it's pinned as deliberate with the `=` form noted as the way out:

```
--selectProjects unit test/util/version.test.ts   -> selected: ['unit', 'test/util/version.test.ts']
--selectProjects=unit test/util/version.test.ts   -> selected: ['unit'], rest: ['test/util/version.test.ts']
```

**On `--coverageDirectory`:** you're right and the comment was stale from the single-process design. Separate processes have nothing to merge, so a shared custom directory gets the overwrite back. Reworded — it is an escape hatch for redirecting output, not a combined report, and there is no longer any such thing.

**On performance:** measured rather than left open — `unit` ~20s, `esm` ~3s locally, and a filtered run adds two `--listTests` resolutions worth about a second. CI's Unit Tests job has stayed in the 60–90s band across all ten pushes here. I'd rather quote that than a before/after delta from a shared runner where the noise exceeds the effect.

`npm test` on `d7903f3`: **109 suites, 1369 tests**; `build`, `tsc` and `eslint` exit 0.

---

jooyoungseo:
You flagged `--selectProjects --selectProjects unit` as "almost certainly fine". It is not — it was the one input that could opt back into the bug this PR exists to fix. `bfa7e54`.

`takeSelection` reports the *flag*, not the outcome, so that input comes back `dangling: true` **and** naming a project. The runner's dangling branch handed the arguments over untouched on the strength of `dangling` alone, and Jest does not error on it — with two names it runs both projects in one process:

```
$ node scripts/test.js --selectProjects --selectProjects unit esm
Running 2 projects:
    Must use import to load ES Module: /home/user/maidr/src/type/grammar.ts
    Must use import to load ES Module: /home/user/maidr/src/type/grammar.ts
    …
```

The escape hatch for "Jest will reject this" had become a way to opt back into the memoization clash. The branch now requires nothing to have been selected, which is the only case Jest actually rejects; anything that named a project goes through the split as usual:

```
$ node scripts/test.js --selectProjects --selectProjects unit esm
[test] project: unit    105 passed
[test] project: esm       4 passed

$ node scripts/test.js --selectProjects
The --selectProjects option requires the name of at least one project to be specified.
$ echo $?
1
```

Both pinned, with the mechanism in the test.

**On sequential vs concurrent — you were right that the comments explained why separate processes and not why one after another, so that's now written down.** But the estimate doesn't hold. Measured here:

| | wall clock |
|---|---|
| `unit` alone (105 suites) | 21.5s |
| `esm` alone (3 suites) | 4.5s |
| both, sequential | 25.5s |

The projects are nothing like equal, so the ceiling is about four seconds rather than half. And Jest already spreads its own workers across all four cores, so a concurrent second run would mostly contend with the first — the realised saving would be less than that ceiling. Against which `stdio: 'inherit'` means two runs interleave into one terminal, which is a poor trade for a CI log someone reads to find a failure. Deliberate, and now said so in the code.

**On the end-to-end gap in `scripts/test.js`'s top-level branching:** agreed, and your reasoning for leaving it matches mine — the helpers are where the bugs have been, and a subprocess-per-case suite would be heavy. Worth noting that this round is the counter-example: the bug was in the *branching*, not the helper, and `takeSelection`'s output was correct throughout. I have verified this round's cases by hand as before rather than adding that suite, but if the branching grows another arm it should get one.

`npm test` on `bfa7e54`: **109 suites, 1370 tests**; `build`, `tsc` and `eslint` exit 0.

---

jooyoungseo:
No bugs found this round, and the `--json` suggestion was a good one — taken in `6001529`.

**`--listTests --json`.** You are right that the `isAbsolute` filter was a guess about Jest's output surviving as production code. It fixed a real bug — counting `Running one project: esm` as a match granted a scoped run the tolerance it was meant to be denied — but the next line Jest decides to print there would be the same bug again. `--json` prints the array and nothing else, both with `--selectProjects` and when the filter matches nothing:

```
$ jest --listTests --json --selectProjects esm
["/home/user/maidr/test/ui/typingEffect.esm-test.tsx", …]

$ jest --listTests --json test/util/version.test.ts --selectProjects esm
[]
```

So the heuristic and its comment are gone, and a response that is not an array is now treated as Jest having answered a different question rather than being picked over. All ten runner exit codes re-checked and unchanged.

**Reporting the `jest-resolve` bug upstream** — agreed, and filed as #703 rather than folded in here. It records the repro, notes that `takeSelection`, `listTests`, `hasPathFilter` and `isNarrowed` all exist only to make the per-project split invisible, and says to revisit them if it is ever fixed upstream. Keeping the workaround until then regardless: it is correct and tested, and reverting depends on a fixed Jest being released and adopted.

**On the `.claude/rules/testing.md` discrepancy** — checked, and it is your environment rather than the PR. `npm run sync:copilot` reports `Already in sync (11 files)` with nothing to write, CI's "Copilot Instructions In Sync" job is green on every push here, and the branch merges cleanly:

```
$ git merge-tree $(git merge-base HEAD origin/main) HEAD origin/main | grep -c '^<<<<<<<'
0
```

`npm test` on `6001529`: **109 suites, 1370 tests**; `build`, `tsc` and `eslint` exit 0.

---

jooyoungseo:
The `.d.ts` note was the one with something behind it, and it is worse than you put it. `f073653`.

You said a `.d.ts` that is wrong but self-consistent would not be caught. Checked it, and a `.d.ts` that is *obviously* wrong is not caught either:

```diff
-  dangling: boolean;
+  dangling: string;
```
```
$ npm run type-check
(clean)
```

`tsc` only ever sees the declarations, never the JavaScript, so nothing compares them — and the tests execute the real module, so they were unaffected too. No guard at all, in either direction.

The new case annotates each export with the shape it is supposed to have, which makes that a compile error, and asserts the runtime types alongside — annotations on their own would be satisfied by declarations wrong in the same way the expectation is. Both drifts now fail:

```
dangling: boolean -> string     TS2322: Type '{ …; dangling: string; }' is not assignable to type 'Selection'
PROJECTS: string[] -> number[]  TS2322: Type 'number[]' is not assignable to type 'string[]'
```

Worth noting this applies equally to `vite-plugin-woff2-only.d.ts` and `vite-plugin-math-stylesheet.d.ts`, which have the same arrangement and the same absence of a guard. I have not touched them here — different files, different PR — but the pattern being established in the repo does not currently protect itself.

**On the two things you could not run:**

*`--listTests --json` exit code with a filter matching nothing* — good question, and it is the path taken rather than the fallback:

```
$ jest --listTests --json test/util/versoin.test.ts
[]
$ echo $?
0
```

So `isNarrowed` sees an empty list rather than `null`, and the run stays strict for the right reason. Your read that the `null` case fails safe is right too — it just isn't what happens.

*`typingEffect.esm-test.tsx` and jsdom's accessible-name computation* — run locally on every push here; it passes, and the counterfactuals are the stronger evidence: restoring the `a` override fails all four link-name cases, and neutering the clock advance fails eight of the nine. jsdom is still not a screen reader, which the file says in its own header.

`npm test` on `f073653`: **109 suites, 1371 tests**; `build`, `tsc` and `eslint` exit 0. CI was fully green on `6001529` (13/13, including E2E).

---

jooyoungseo:
No changes this round — the one checkable question has an answer, and the other two are observations I agree with rather than things to act on.

**Whether a `--someFlag <path-like-value>` in CI usage could fire the probe on every run.** It cannot, and the reason is simpler than the heuristic: the probe only runs when `rest` is non-empty *and* contains a positional, and there is exactly one invocation in the repo.

```
$ grep -rn "npm test\|npm run test" .github/workflows/*.yml package.json
.github/workflows/ci.yml:177:        run: npm test
```

Bare, no arguments, so `hasPathFilter([])` is false and CI never spawns a `--listTests`. The `test:watch` and `test:coverage` scripts pass `--watch` and `--coverage`, both flags. The heuristic's over-inclusiveness only costs anything when a human types a path or a spaced flag value, which is the case it was written to be generous about.

**Comment density** — agreed on the trade, and agreed it is a maintenance cost rather than a violation. Worth saying that the drift risk you name is the one this PR kept hitting: three of the eleven review rounds here were comments that stated a true conclusion with a false mechanism (`-t` "does not need" the check; "computed lists throw"; `--coverageDirectory` giving "one report"). Each was found by someone reading the comment against the code. So the density is load-bearing and also the thing most likely to rot, and I do not have a better answer than the one already in play — pin the claim as a test case where it is executable, which is what the `-t` and spread-array cases now are.

**Sequential execution as `unit` grows** — noted, and it is the right place to expect the cost. The lever if it matters later is running the projects concurrently with buffered output rather than `stdio: 'inherit'`; today that buys about four seconds and costs interleaved CI logs, but the arithmetic moves as the projects diverge in size.

**CI on `f073653`:** now 12 of 13 green, including the four you flagged as in progress — `Unit Tests`, `Build Project`, `Docs Build` and `claude-review` all landed green. Only `E2E (chromium)` is still running, and it has passed on every previous head here.

Eleven review rounds in, this one found nothing, which is the first time. The three before it each found something real (the dangling-flag hole reintroducing the memoization clash, the `--listTests` parsing heuristic, the unguarded `.d.ts`), so I do not think the loop was spinning — but it does look like it has converged.

---

PR #704, opened 2026-08-02
fix(chat): make footnote anchors in a chat response resolve

### LINKED ISSUE #696: a11y: footnote links in chat point at ids that sanitisation has renamed

jooyoungseo:
# Pull Request

## Description

A footnote reference in an AI chat response renders, is announced, and goes nowhere. Both of its links name ids that no longer exist by the time the DOM is built.

## Related Issues

Closes #696. Turns the two `it.failing` pins added in #702 into assertions of the fixed behaviour. Filed #705 for a pre-existing gap this makes load-bearing.

## Changes Made

**`src/util/footnoteScope.ts`** — a rehype plugin that runs **after** `rehype-sanitize`, namespaces every `id` in the message, and re-points everything that names them.

### Why the anchors dangle

`mdast-util-to-hast` writes matching pairs. `hast-util-sanitize` then applies its `clobberPrefix` to everything in `clobber` — `ariaDescribedBy`, `ariaLabelledBy`, `id`, `name`. `href` is not in that list:

```
a  href="#user-content-fn-1"             <- untouched
li id="user-content-user-content-fn-1"   <- prefixed again
```

`aria-describedby` survives *precisely because* it is in `clobber`, so it moves in step with the `id` it names. The double-prefixing is not the bug — `aria-describedby` is double-prefixed on neither side and works. The bug is that `href` is not clobbered at all, so it cannot follow its target.

### The second fault, which the first one hides

Footnotes are numbered per document and each message is its own document, so every message with a footnote emits `fn-1`. A transcript is one DOM, which makes them duplicates — and once anchors resolve, `getElementById` returns the first match, so a reference in the second message would jump to the **first** message's note.

That is why this is one change rather than two: fixing the anchors without fixing the ids would turn a dead link into a link to the wrong place, which is worse.

### Why post-processing rather than configuring a prefix

`remark-rehype` takes `clobberPrefix`, and it looked like the smaller fix. It reaches only ids derived from the footnote label:

```
--- per-message prefix ---
m1 a#user-content-msg-m1-fnref-1     m2 a#user-content-msg-m2-fnref-1     <- fixed
m1 li#user-content-msg-m1-fn-1       m2 li#user-content-msg-m2-fn-1       <- fixed
m1 h2#user-content-footnote-label    m2 h2#user-content-footnote-label    <- STILL COLLIDES
```

`mdast-util-to-hast` hardcodes the heading as `id: 'footnote-label'` (`lib/footer.js:236`) and documents that it is "always added". No option scopes it. Rewriting the tree is the only place both faults can be fixed, so that is what this does.

## Checklist

### Found by looking rather than reasoning

**`ariaDescribedBy` is an array, not a space-separated string.** hast models it as `['user-content-footnote-label']`. My first version read it as a string, skipped it, and broke the one reference that had been working — caught by the test that has passed since #702, which is the argument for having written it.

**The `href` names the pre-sanitisation id**, so renaming ids consistently is not enough on its own. The rewrite has to match a fragment against the prefixed form first — the sanitiser's damage undone rather than followed.

**The prefix is read from the library.** `defaultSchema.clobberPrefix` comes from `rehype-sanitize` rather than being written out here, so the two cannot drift apart. That also means no new dependency: `rehype-sanitize` is already one, and it re-exports the schema.

**`name` is a fragment target too, but only on an anchor.** It is in `clobber` for the same reason `id` is — a browser resolves `#foo` against an anchor's `name` — and the schema admits it nowhere today, exactly like `ariaLabelledBy`. Both are handled anyway, because this scopes ids in general and admitting either later must not quietly produce a dangling reference. Scoping `name` *everywhere* would be wrong: on a form control it is what the field submits under, and renaming it would change the submitted data to repair a link that does not exist.

### The security trade, stated

The anti-clobbering guarantee moves from the sanitiser's `clobberPrefix` to this plugin's scope, which keeps `user-content-` for exactly that reason. Losing it would be **silent** — ids would still be unique per message and every other test here would still pass — so it has a test of its own.

The chat schema is untouched, so it still declares only `tagNames` and `attributes` and every other `hast-util-sanitize` default stays in force, including the `protocols` map that strips `javascript:`. The plugin only ever rewrites a fragment `href` to another fragment, so it cannot introduce a protocol.

**The scope is escaped injectively rather than stripped**, `-` included. A message id is `msg-<time>`, `system-<time>` or `resp-<time>-<provider>`, and the providers are `OPENAI`, `ANTHROPIC_CLAUDE`, `GOOGLE_GEMINI` and `OLLAMA` — three of which carry an underscore, so the escape path runs in production. (An earlier version of this description claimed a message id carries a *dot*, citing `gpt-5.5`. That was wrong: `addPendingResponse` uses the `Llm` provider key, not `LlmVersion`. Review caught it.)

Escaping `-` is a second, separate reason. The final id is `user-content-<token>-<id>`, so what must be unique is the whole concatenation, not the token: with `-` intact, message `m` with id `x-y` and message `m-x` with id `y` both spell `user-content-m-x-y`. Keeping `-` out of every token makes the first one the boundary and the decomposition unique — an unenforced invariant turned into a proven one.

### Every case verified against the defect it names

| reverted | cases that fail |
|---|---|
| the plugin removed from `TypingEffect` | 4 of the 11 component cases |
| the scope stripped instead of escaped | the injectivity case |
| `user-content-` dropped from the scope | the anti-clobbering case |
| `ariaDescribedBy`'s array form unhandled | the `aria-describedby` case |
| `name` not scoped at all | the anchor case |
| `name` scoped on **every** element | the form-control case |
| `-` left unescaped in the token | the separator case |
| the repair assuming the href carries `user-content-` | the `clobberPrefix`-independence case |

The fourth is not hypothetical — it is the bug I actually wrote, caught by an existing test. The sixth and eighth are the straightforward readings of two review suggestions, which is why both directions are pinned rather than one.

### Acceptance criteria from #696

| criterion | |
|---|---|
| a reference and its backref both resolve after sanitisation | ✅ and each reaches *its own* note, which is stronger than the pins could assert |
| `javascript:` filtering on `href` still applies | ✅ schema untouched; pinned by `markdownPipeline.esm-test.ts` |
| whatever provides anti-clobbering is named in a comment and pinned by a test | ✅ `footnoteScope.esm-test.ts` |
| two messages produce no duplicate DOM ids | ✅ |
| a backref in the second message returns to its own reference | ✅ |

### What this does not fix

The cross-message guarantee rests on `messageId` being unique, and those are `Date.now()`-based, so two messages created in the same millisecond share a scope. Pre-existing, in another layer, and filed as **#705** rather than folded in — but worth naming here, because this PR is what makes message-id uniqueness load-bearing for the first time.

### Verification

| check | result |
|---|---|
| `npm test` | 110 suites, 1386 tests |
| `npm run build` | exit 0 |
| `npx eslint .` | exit 0 |

jsdom is not a screen reader. These pin the wiring; they do not replace checking with real assistive technology.

---

### DIFF
--- a/src/ui/components/TypingEffect.tsx
+++ b/src/ui/components/TypingEffect.tsx
@@ -1,6 +1,7 @@
 import { Box } from '@mui/material';
 import { useViewModelState } from '@state/hook/useViewModel';
 import { visuallyHidden } from '@ui/visuallyHidden';
+import { rehypeScopeIds } from '@util/footnoteScope';
 import { containsLatex, ensureKatexStylesheet } from '@util/katex';
 import { createChatSanitizeSchema } from '@util/markdownSanitize';
 import React, { memo, useEffect, useMemo, useState } from 'react';
@@ -150,6 +151,13 @@ export const TypingEffect: React.FC<TypingEffectProps> = memo(({ text, isUser, m
             // allowlist rather than around it.
             ...mathPlugins,
             [rehypeSanitize, SANITIZE_SCHEMA],
+            // After it, and it has to be: the sanitiser renames `id` and the
+            // ARIA references but not `href`, so a footnote anchor is left
+            // naming where its target used to be. Scoping both sides here puts
+            // them back in agreement and makes the ids unique to this message,
+            // which they are not otherwise — footnotes are numbered per
+            // document and a transcript is one DOM. See `footnoteScope`.
+            [rehypeScopeIds, { messageId }],
           ]}
           remarkPlugins={[remarkGfm, remarkMath]}
           components={{
--- a/src/util/footnoteScope.ts
+++ b/src/util/footnoteScope.ts
@@ -57,21 +57,62 @@ const CLOBBERED = defaultSchema.clobberPrefix ?? '';
 const ID_REFERENCES = ['ariaDescribedBy', 'ariaLabelledBy'] as const;
 
 /**
+ * Whether this element's `name` is a fragment target rather than a field name.
+ *
+ * `name` is in the sanitiser's `clobber` list for the same reason `id` is —
+ * a browser resolves `#foo` against an anchor's `name` as well as an `id` — so
+ * it has to be scoped alongside `id` or an `href` aimed at one would dangle
+ * exactly as this fixes for `id`.
+ *
+ * Only on an anchor, though. Everywhere else `name` is a form control's field
+ * name, and renaming that would change what a form submits rather than where a
+ * link goes. The chat schema admits neither today — `input` is pinned to
+ * `type`, `checked` and `disabled` precisely to keep `name` out — so this is
+ * about the plugin staying correct if one is ever allowed in, not about markup
+ * that reaches it now.
+ * @param element - The element to test.
+ * @returns True if `name` on this element names a fragment.
+ */
+function namesFragment(element: Element): boolean {
+  return element.tagName === 'a';
+}
+
+/**
  * Turns a message id into something safe to put in an `id`, injectively.
  *
- * A message id is `msg-1234` or `resp-1234-<model>`, and a model name can
- * carry a dot or a slash. Every character outside `[A-Za-z0-9-]` becomes
- * `_<hex>_`, which cannot be confused with a literal because `_` is itself
- * outside the set and so is escaped too. Two different message ids therefore
- * cannot encode to the same token, which is what the uniqueness rests on.
+ * A message id is `msg-1234` or `resp-1234-<model>`, and a model name really
+ * does carry a dot — `gpt-5.5`, `gemini-3.5-flash`. Every character outside
+ * `[A-Za-z0-9-]` becomes `_<hex>_`, underscore included, so no two message ids
+ * can encode to the same token. Stripping instead would map `a.b` and `a-b`
+ * together and quietly reintroduce the collision this exists to prevent.
  * @param messageId - The message's id.
  * @returns An id-safe token.
  */
 function token(messageId: string): string {
   return messageId.replace(
-    /[^\w-]|_/g,
+    /[^a-z0-9-]/gi,
     character => `_${character.charCodeAt(0).toString(16)}_`,
   );
+}
+
+/**
+ * The properties on this element that define a fragment target.
+ * @param element - The element to inspect.
+ * @returns `['id']`, or `['id', 'name']` on an anchor.
+ */
+function targetAttributes(element: Element): readonly string[] {
+  return namesFragment(element) ? ['id', 'name'] : ['id'];
+}
+
+/**
+ * The fragment targets this element defines.
+ * @param element - The element to inspect.
+ * @returns Every non-empty target name on it.
+ */
+function targets(element: Element): string[] {
+  return targetAttributes(element)
+    .map(attribute => element.properties?.[attribute])
+    .filter((value): value is string => typeof value === 'string' && value !== '');
 }
 
 /** Every element in the tree, in document order. */
@@ -130,10 +171,11 @@ export function rehypeScopeIds(options: { messageId: string }): (tree: Root) => 
     const all = elements(tree);
     const present = new Set<string>();
 
+    // Collected in a pass of its own, before anything is written, so the
+    // rewriting below cannot see a half-renamed tree and is order-independent.
     for (const element of all) {
-      const id = element.properties?.id;
-      if (typeof id === 'string' && id !== '') {
-        present.add(id);
+      for (const target of targets(element)) {
+        present.add(target);
       }
     }
 
@@ -154,8 +196,11 @@ export function rehypeScopeIds(options: { messageId: string }): (tree: Root) => 
         continue;
       }
 
-      if (typeof properties.id === 'string' && properties.id !== '') {
-        properties.id = scope + properties.id;
+      for (const attribute of targetAttributes(element)) {
+        const value = properties[attribute];
+        if (typeof value === 'string' && value !== '') {
+          properties[attribute] = scope + value;
+        }
       }
 
       for (const attribute of ID_REFERENCES) {
--- a/src/util/footnoteScope.ts
+++ b/src/util/footnoteScope.ts
@@ -78,19 +78,28 @@ function namesFragment(element: Element): boolean {
 }
 
 /**
- * Turns a message id into something safe to put in an `id`, injectively.
+ * Turns a message id into an id-safe token that no other message id can share.
  *
- * A message id is `msg-1234` or `resp-1234-<model>`, and a model name really
- * does carry a dot — `gpt-5.5`, `gemini-3.5-flash`. Every character outside
- * `[A-Za-z0-9-]` becomes `_<hex>_`, underscore included, so no two message ids
- * can encode to the same token. Stripping instead would map `a.b` and `a-b`
- * together and quietly reintroduce the collision this exists to prevent.
+ * Every character outside `[A-Za-z0-9]` becomes `_<hex>_`. That is injective on
+ * its own — `_` is itself outside the set, so an underscore in the output only
+ * ever begins an escape — and it is exercised in practice: a response id is
+ * `resp-<time>-<provider>` and the providers are `OPENAI`, `ANTHROPIC_CLAUDE`,
+ * `GOOGLE_GEMINI` and `OLLAMA`, three of which carry an underscore.
+ *
+ * `-` is escaped along with everything else, which matters for a reason
+ * injectivity alone does not cover. The final id is
+ * `user-content-<token>-<id>`, so what has to be unique is the whole
+ * concatenation rather than the token: with `-` left intact, token `m` with id
+ * `x-y` and token `m-x` with id `y` both spell `user-content-m-x-y`. Escaping
+ * it keeps `-` out of every token, which makes the first one the boundary and
+ * the decomposition unique — the property the cross-message guarantee actually
+ * rests on.
  * @param messageId - The message's id.
- * @returns An id-safe token.
+ * @returns An id-safe token containing no separator.
  */
 function token(messageId: string): string {
   return messageId.replace(
-    /[^a-z0-9-]/gi,
+    /[^a-z0-9]/gi,
     character => `_${character.charCodeAt(0).toString(16)}_`,
   );
 }
@@ -115,12 +124,22 @@ function targets(element: Element): string[] {
     .filter((value): value is string => typeof value === 'string' && value !== '');
 }
 
-/** Every element in the tree, in document order. */
-function elements(node: Nodes): Element[] {
-  const found: Element[] = node.type === 'element' ? [node] : [];
+/**
+ * Every element in the tree, in document order.
+ *
+ * Accumulates into one array rather than spreading a new one at each level:
+ * this runs on every render, and the typing animation renders every 10 ms.
+ * @param node - Where to start.
+ * @param found - The array being filled.
+ * @returns `found`.
+ */
+function elements(node: Nodes, found: Element[] = []): Element[] {
+  if (node.type === 'element') {
+    found.push(node);
+  }
   if ('children' in node) {
     for (const child of node.children) {
-      found.push(...elements(child));
+      elements(child, found);
     }
   }
   return found;
--- a/src/util/footnoteScope.ts
+++ b/src/util/footnoteScope.ts
@@ -1,0 +1,184 @@
+import type { Element, Nodes, Properties, Root } from 'hast';
+import { defaultSchema } from 'rehype-sanitize';
+
+/** One entry of a hast element's properties, in every shape hast allows. */
+type PropertyValue = Properties[string];
+
+/**
+ * Repairs and namespaces the ids in a rendered chat message, so footnote
+ * anchors resolve and two messages cannot claim the same id.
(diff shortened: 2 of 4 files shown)


jooyoungseo:
All three check out. `dbb7a03` for two of them, #705 for the third.

**`ariaLabelledBy` is unreachable — confirmed.** The schema names `ariaHidden` and `className` globally, and `ariaLabel` and `ariaDescribedBy` on `a`. Nothing else, anywhere.

I did not delete it, and the reason is the same one that makes it worth mentioning: this scopes ids in general, not footnote ids in particular, and if `ariaLabelledBy` is ever admitted to the schema, a plugin that silently skipped it would leave a reference pointing at an id that no longer exists — the exact failure this PR removes. So the answer is to cover it rather than to trust a comment.

**The same is true of a branch you did not mention.** `reference()` handles a space-separated string as well as a list, and `mdast-util-to-hast` only ever emits the list, so that branch had no test either. Both are now driven from a hand-built tree, which is the only way to reach them:

```
✓ should move a list-valued reference with the id it names
✓ should move a string-valued reference with the id it names
✓ should leave a reference to an id this message does not define
```

**On the surrogate pairs** — checked rather than reasoned about. Model names are typed literals (`src/type/llm.ts`) and none carries an astral character. The encoding is injective across them regardless:

```
"a.b"   -> a_2e_b          "a😀b"  -> a_d83d__de00_b
"a-b"   -> a-b             "a😁b"  -> a_d83d__de01_b
"a_b"   -> a_5f_b
```

Worth adding that the dot case is **not** hypothetical: `gpt-5.5` and `gemini-3.5-flash` both contain one, so `resp-<time>-gpt-5.5` takes the escaping path in production. That is why stripping would have been wrong rather than merely untidy.

**The `Date.now()` collision is real and is now #705.** You are right that it is pre-existing and right that this PR is what makes it matter — before this, anchors did not resolve at all, so a shared scope had no visible consequence. The issue records what actually rests on the uniqueness: the id scoping here (a collision brings back the cross-message jump, silently, because everything still renders and resolves — just to the wrong message's note), React keys, and `completedAnimations`, which is keyed on id *and* body and so is harmless. Filed rather than folded in: different layer, and this PR is already a two-fault fix.

**On the duplicated `elements()`** — deliberate. The module's copy is not exported, and I would rather the test walk the tree itself than have the thing under test supply the means of inspecting it. There is no shared tree helper in the repo; `markdownPipeline.esm-test.ts` and `typingEffect.esm-test.tsx` each have their own for the same reason.

**On the snapshot-before-mutation point** — good thing to name. `present` is built in a full pass before anything is written, so the rewrite loop cannot see a half-renamed tree. Collapsing the two into one loop would break it, which is why they are two.

`npm test` on `dbb7a03`: **110 suites, 1382 tests**; `tsc` and `eslint` exit 0. `build` was green on `efd2b4f` and this round touches one test file.

---

jooyoungseo:
You turned my own argument back on `name` and it holds — that is the right way to have caught it. `8483d58`.

`name` is in `clobber` for the same reason `id` is (a browser resolves `#foo` against an anchor's `name` too), and it is as absent from the schema today as `ariaLabelledBy`. Admitted later, an `href` aimed at one would dangle in exactly the way this fixes for `id`, with nothing to catch it.

**Scoped on an anchor only, and following your reasoning precisely is what surfaced why.** Everywhere else `name` is a form control's field name, not a fragment target — renaming it would change what a form submits in order to repair a link that does not exist. The schema keeps `name` off `input` deliberately (it is why the checkbox entry pins `type`, `checked` and `disabled` and nothing else), so this is about the plugin staying correct if one is ever allowed in.

Both directions are pinned, because the straightforward reading of your suggestion — scope `name` alongside `id` — is wrong in a way one test would not show:

| reverted to | case that fails |
|---|---|
| `name` not scoped at all *(the gap you found)* | `should scope it, and point a link at where it went` |
| `name` scoped on **every** element | `should leave a form control's name alone` |

**The regex nit is right and taken.** `\w` includes underscore, which is exactly why `[^\w-]` needed the `|_` alternative to force-escape it; `/[^a-z0-9-]/gi` matches the same set without the alternation. The injectivity case covers the change:

```
"a.b" -> a_2e_b     "a-b" -> a-b     "a_b" -> a_5f_b     "a😀b" -> a_d83d__de00_b
```

**On the snapshot-before-mutation observation** — you and the previous round both landed on it, so I have written it into the code rather than leaving it to be re-derived: the collection pass now says in a comment that it exists to run before anything is written, so the rewrite cannot see a half-renamed tree.

**On not being able to run the suite** — noted, and it is why I ran the counterfactuals above rather than reporting green. `npm test` on `8483d58`: **110 suites, 1384 tests**; `build`, `tsc` and `eslint` exit 0.

---

jooyoungseo:
You were right about the field, and checking it turned the second point into something better than an acknowledgement. `8cc30ba`.

**The dotted-model-name claim was false.** `addPendingResponse` builds `resp-<time>-${action.payload.model}` from a payload typed `Llm`, which is the **provider** key:

```ts
// src/type/llm.ts
export type Llm = 'OPENAI' | 'ANTHROPIC_CLAUDE' | 'GOOGLE_GEMINI' | 'OLLAMA';
```

`gpt-5.5` and `gemini-3.5-flash` live in `LlmVersion`, which never reaches a message id. So no message id contains a dot, and I asserted in a docstring, a commit message and a PR reply that one does.

The escaping is still exercised, by the underscore in three of the four providers — which is the true version of the same point, and is now what the comment says:

```
resp-1770000000000-ANTHROPIC_CLAUDE  ->  resp_2d_1770000000000_2d_ANTHROPIC_5f_CLAUDE
```

Same shape as the three "raw HTML is escaped to text" comments corrected on #701: conclusion right, mechanism named wrong, and the mechanism is what a reader checks the conclusion against. Thank you for reading it against the code rather than past it.

**On the boundary property — you asked only that the gap be named, but it is cheaper to close than to document.** You are right that `token()` proves no two *message ids* collide while what the plugin needs is that no two `user-content-<token>-<id>` collide, and that with `-` passing through unescaped the concatenation is not self-delimiting.

Escaping `-` as well keeps it out of every token, which makes the first `-` after the fixed prefix the boundary and the decomposition unique:

| | `-` intact *(before)* | `-` escaped *(now)* |
|---|---|---|
| message `m`, id `x-y` | `user-content-m-x-y` | `user-content-m-x-y` |
| message `m-x`, id `y` | `user-content-m-x-y` ❌ | `user-content-m_2d_x-y` ✅ |

One character of regex turns an unenforced invariant into a proven one. Pinned by `should keep the scope and the id it prefixes tellable apart`, verified to fail against the previous scheme.

The cost is uglier ids — `user-content-msg_2d_1-…` rather than `user-content-msg-1-…`. Worth it: these are internal, and the alternative is a guarantee that holds only while message ids stay shaped as they are today, which #705 already shows is not a safe assumption to build on.

**The allocation nit is taken** — `elements()` accumulates into one array instead of spreading a new one per level. You are right that it matters more than it looks: it runs on every render, and the animation renders every 10 ms.

`npm test` on `8cc30ba`: **110 suites, 1385 tests**; `build`, `tsc` and `eslint` exit 0.

---

jooyoungseo:
Checked both. The first is not the coupling it looks like, and the second would buy nothing — but the first was worth a test rather than a reply. `3e0c21a`.

**The `href` repair does not depend on `remark-rehype`'s prefix.** It looks for *the sanitiser's* prefix plus whatever the href names, and the sanitiser prefixes whatever `remark-rehype` produced — so mdast's choice cancels out on both sides. Driven through the real pipeline with three different values:

```
clobberPrefix "user-content-"        dangling []
clobberPrefix "totally-different-"   dangling []
clobberPrefix ""                     dangling []
```

You asked for a comment noting the assumption and where it is covered. A test is the better answer, since the property is worth *holding* rather than describing, and the one you named — "should leave every fragment link naming an id that exists" — would not actually have caught a bump: it runs with the default, where both prefixes agree, so a prefix-dependent repair passes it. The new case runs all three.

It has teeth. The realistic version of the mistake — assuming the href carries `user-content-` — fails it:

```diff
- const target = present.has(CLOBBERED + named) ? CLOBBERED + named : named;
+ const target = named.startsWith('user-content-') ? CLOBBERED + named : named;
```
```
✕ should repair the link whatever remark-rehype namespaced it with
Tests: 1 failed, 12 passed
```

Every other case stays green under that change, which is exactly why it needed its own.

**Memoising the `rehypePlugins` array would buy nothing.** `react-markdown` rebuilds the pipeline unconditionally:

```js
// node_modules/react-markdown/lib/index.js:176
export function Markdown(options) {
  const processor = createProcessor(options)   // no cache, no identity check
```

`createProcessor` runs `unified().use(...)` on every call, so a stable array reference changes nothing — the parse you note as dominating the cost happens either way. I had a `useMemo` here in an earlier draft and removed it for this reason; leaving it out.

`npm test` on `3e0c21a`: **110 suites, 1386 tests**; `tsc` and `eslint` exit 0. `build` was green on `8cc30ba`, and this round touches one test file.

---

jooyoungseo:
One thing to do here, and it was the right thing to ask for. `3517ee4`.

The `-` reasoning was already in `token()`'s docstring — what was missing is the **test's name**, which is what makes the claim checkable rather than merely asserted. That is the convention `markdownSanitize.ts` already follows two doc comments over ("`test/util/markdownSanitize.test.ts` pins the rule"), so it should have been there:

> `should keep the scope and the id it prefixes tellable apart`, in `test/util/footnoteScope.esm-test.ts`, is that pair as a test.

Verified the name actually resolves rather than trusting my own copy of it — a pointer to a test that has been renamed is worse than none.

**On the MVVC note** — agreed with your own conclusion. `rehypeScopeIds` is rendering logic with no business state, called the same way `markdownSanitize` and `katex` already are in that file. Worth having checked rather than assumed, given how strict `rules/architecture.md` is about the View layer.

**On the per-tick cost** — noted, and it is the reason `elements()` accumulates instead of spreading. Worth adding one measurement rather than leaving it as a thing to watch: the plugin walks a tree of a few dozen elements and does one `Set` build plus one pass, against a full markdown parse and sanitise on the same tick. If the animation ever profiles badly, `createProcessor` running unconditionally on every render (`react-markdown/lib/index.js:176`) is the larger target, and that is not something this PR can address.

**Where this stands:** four review rounds, and this is the second with no defect found — the last three found the `name` gap, the false dotted-model-name claim with the separator invariant behind it, and nothing. That reads like convergence rather than a quiet patch.

`npm test` on `3517ee4`: **110 suites, 1386 tests**; `tsc` and `eslint` exit 0. `build` was green on `8cc30ba`; the two rounds since touch one test file and one comment. CI was fully green including E2E on `8cc30ba`.

---

PR #706, opened 2026-08-02
fix(chat): give every message an id no other message can have

### LINKED ISSUE #705: fix(chat): message ids are millisecond timestamps and can collide
Chat message ids are generated from `Date.now()` alone, so two messages created in the same millisecond get the same id. Raised in review on #704, which is where it started to matter.

```ts
// src/state/viewModel/chatViewModel.ts
id: `msg-${Date.now()}`                              // addUserMessage
id: `system-${Date.now()}`                           // addSystemMessage
id: `resp-${Date.now()}-${action.payload.model}`     // addPendingResponse
```

A collision needs the same prefix and the same millisecond — and, for a response, the same model. So the exposure is narrow, and a human typing cannot reach it. What can is anything that adds two messages programmatically in one tick.

## What rests on message ids being unique

**Id scoping in the DOM** (#704, `src/util/footnoteScope.ts`). The scope for a message's element ids is derived from its message id, which is what stops two messages' footnotes from colliding in one transcript. Two messages sharing an id share a scope, and the cross-message collision that #704 removes comes back — silently, because everything still renders and the anchors still resolve, just to the wrong message's note.

**The animation memo** (`TypingEffect.tsx:34`). `completedAnimations` is keyed `${messageId}\0${text}`, so a collision only matters when the bodies also match — at which point replaying or not replaying the animation is indistinguishable. Harmless.

**React keys**, wherever the message list is rendered. Two children with the same key is a React warning and can misplace state on reorder.

## Why not fixed in #704

It is pre-existing, it is in a different layer, and #704 is already a two-fault fix. Worth its own change rather than a third thing bolted on.

## Suggested fix

A monotonic counter alongside the timestamp is enough and keeps ids readable and sortable:

```ts
let sequence = 0;
const nextId = (kind: string): string => `${kind}-${Date.now()}-${sequence++}`;
```

`crypto.randomUUID()` would also do it, but loses the ordering the timestamp gives and reads worse in devtools.

Either way the property to pin is the one the name suggests: **no two messages in a conversation share an id**, asserted by adding several in one tick rather than by inspecting the format.

## Not a regression

Nothing here is made worse by #704 — before it, footnote anchors did not resolve at all, so the collision had no visible consequence. It is being written down because #704 makes message-id uniqueness load-bearing for the first time, and a guarantee that rests on two calls landing in different milliseconds is worth stating rather than assuming.

jooyoungseo:
# Pull Request

## Description

Message ids were `Date.now()` alone, so two created in the same millisecond were the same string. A human typing cannot reach that; anything adding two messages programmatically can.

## Related Issues

Closes #705. Raised in review on #704, which is what makes it matter. Review here spun off #707.

## Changes Made

**A monotonic sequence, appended.** The timestamp stays for readability and ordering in devtools; the sequence is what carries the uniqueness, being monotonic for the life of the page.

```diff
- id: `msg-${Date.now()}`
+ id: nextId('msg')            // msg-1770000000000-1
```

**The id is minted in `prepare`, not in the reducer.** `rules/viewmodel.md` asks reducers to "assign the payload, nothing more", and the id is the one thing here that cannot come from the caller — so it is generated where Redux Toolkit puts non-determinism, leaving all three reducers pure. The action creators' signatures are unchanged, so no dispatch site moves.

**Suggestion ids get the same treatment.** They were `suggestion-<time>-1..5`: unique within one batch, colliding across two in the same millisecond. Same shape, same fix.

## Why it matters now

#704 derives each message's DOM id scope from its message id. Two messages sharing an id share a scope, so their footnote ids collide again, and a reference resolves to whichever note comes first in document order — the other message's.

Everything still renders. The failure is silent, which is the same shape as the bug #704 removed, arriving through a different door.

## Checklist

### Three things checked rather than assumed

**The `kind` prefix is load-bearing.** `MessageBubble.tsx:86` tells a system message apart with `message.id.startsWith('system-')`. That is the only thing in `src/` that reads the id's format, and it still holds — with a test, because the fix was one careless refactor away from breaking it.

**A collision never misrouted a response.** `updateResponse` and `updateError` look up by `model` **and** `timestamp`, not by id:

```ts
const message = state.messages.find(m =>
  m.model === action.payload.model
  && m.timestamp === action.payload.timestamp,
);
```

Worth knowing, because it means the real blast radius was the DOM ids from #704 and React keys, rather than the conversation itself.

**A third consumer, surfaced in review.** `TypingEffect.tsx:112` keys its `completedAnimations` set on `` `${messageId}\0${text}` ``. Two same-millisecond messages with identical text shared that key, and the second would have skipped its typing animation because the first had already marked it complete. The message still renders in full on the skip path, so nothing is lost visually or to a screen reader — cosmetic, but a real behavioural difference rather than a no-op.

### The test needed the action creators exported

Dispatching by action type — how the other view model tests reach their reducers — **skips `prepare`**, which is now the only step that generates an id. So an action built by hand carries no id at all.

The first version of this test did exactly that and "passed" on `undefined` being equal to itself, until five ids collapsing to one made me look. `chatActions` is exported so the test can go through the creators, which is the path production takes.

### The clock is frozen

Deliberately. With a real clock the old behaviour passes whenever two dispatches straddle a millisecond boundary — the flakiness being removed, used as the test's own oracle. Freezing it makes every id in the case come from one `Date.now()`, where the old code produces one string five times.

### Verified against all three defects

| reverted | case that fails |
|---|---|
| the sequence dropped, back to `Date.now()` alone | `are distinct for messages added in the same millisecond` — 5 ids collapse to 3 |
| the old per-batch suggestion scheme restored | `are distinct across two batches generated in the same millisecond` — 6 collapse to 3 |
| the id minted in the reducer instead of `prepare` | `are minted in the action, leaving the reducer pure` |

The suggestion case is the one worth spelling out. Asserting that a *single* batch's ids are distinct would not have tested this: they were `suggestion-<time>-1..5`, numbered within the batch, so one batch was already unique before the fix. What changed is across batches, because the `1..5` restarted on every call. In the counterfactual run above, the single-batch assertions in that same case still passed — direct evidence that the obvious test would have been green on the bug.

### Commits

| | |
|---|---|
| `a6ed63c` | the fix |
| `c266ee6` | timers restored in `afterEach` rather than at the end of the test |
| `74476f7` | `footnoteScope` doc corrections — the id shape went stale, and the same sentence had a provider count wrong since #704 |
| `ad4a37d` | the suggestion-id case above |

### Verification

| check | result |
|---|---|
| `npm run build` | exit 0 |
| `npx eslint .` | exit 0 |

---

---

### DIFF
--- a/src/state/viewModel/chatViewModel.ts
+++ b/src/state/viewModel/chatViewModel.ts
@@ -22,39 +22,89 @@ const initialState: ChatState = {
   suggestions: [],
 };
 
+/**
+ * Distinguishes ids minted within the same millisecond.
+ *
+ * These used to be `Date.now()` alone, which is unique only if no two are
+ * created in the same tick — a bound nothing enforces, and one that anything
+ * adding two messages programmatically crosses. What made it matter is that
+ * #704 derives each message's DOM id scope from its message id, so two
+ * messages sharing one share a scope: their footnote ids collide again, and a
+ * reference resolves to the first match in document order, which is the other
+ * message's note. Everything still renders, so the failure is silent.
+ */
+let sequence = 0;
+
+/**
+ * A unique id for a message or a suggestion.
+ *
+ * The timestamp is kept for readability and ordering in devtools; the sequence
+ * is what actually carries the uniqueness, being monotonic for the life of the
+ * page. `kind` leads because `MessageBubble` distinguishes system messages by
+ * it (`message.id.startsWith('system-')`).
+ * @param kind - The id's leading segment.
+ * @returns An id no other call can produce.
+ */
+function nextId(kind: string): string {
+  sequence += 1;
+  return `${kind}-${Date.now()}-${sequence}`;
+}
+
 const chatSlice = createSlice({
   name: 'chat',
   initialState,
   reducers: {
-    addUserMessage: (state, action: PayloadAction<{ text: string; timestamp: string }>) => {
-      state.messages.push({
-        id: `msg-${Date.now()}`,
-        text: action.payload.text,
-        isUser: true,
-        timestamp: action.payload.timestamp,
-        status: 'SUCCESS',
-      });
+    // The three that add a message mint its id in `prepare` rather than in the
+    // reducer. `rules/viewmodel.md` asks reducers to assign the payload and
+    // nothing more, and an id is the one thing here that cannot come from the
+    // caller — so it is generated where Redux Toolkit puts non-determinism,
+    // leaving each reducer pure. The action creators' signatures are unchanged.
+    addUserMessage: {
+      reducer: (state, action: PayloadAction<{ id: string; text: string; timestamp: string }>) => {
+        state.messages.push({
+          id: action.payload.id,
+          text: action.payload.text,
+          isUser: true,
+          timestamp: action.payload.timestamp,
+          status: 'SUCCESS',
+        });
+      },
+      prepare: (message: { text: string; timestamp: string }) => ({
+        payload: { ...message, id: nextId('msg') },
+      }),
     },
-    addSystemMessage: (state, action: PayloadAction<{ text: string; timestamp: string; modelSelections?: { modelKey: Llm; name: string; version: string }[]; isWelcomeMessage?: boolean }>) => {
-      state.messages.push({
-        id: `system-${Date.now()}`,
-        text: action.payload.text,
-        isUser: false,
-        timestamp: action.payload.timestamp,
-        status: 'SUCCESS',
-        modelSelections: action.payload.modelSelections,
-        isWelcomeMessage: action.payload.isWelcomeMessage,
-      });
+    addSystemMessage: {
+      reducer: (state, action: PayloadAction<{ id: string; text: string; timestamp: string; modelSelections?: { modelKey: Llm; name: string; version: string }[]; isWelcomeMessage?: boolean }>) => {
+        state.messages.push({
+          id: action.payload.id,
+          text: action.payload.text,
+          isUser: false,
+          timestamp: action.payload.timestamp,
+          status: 'SUCCESS',
+          modelSelections: action.payload.modelSelections,
+          isWelcomeMessage: action.payload.isWelcomeMessage,
+        });
+      },
+      prepare: (message: { text: string; timestamp: string; modelSelections?: { modelKey: Llm; name: string; version: string }[]; isWelcomeMessage?: boolean }) => ({
+        payload: { ...message, id: nextId('system') },
+      }),
     },
-    addPendingResponse: (state, action: PayloadAction<{ model: Llm; timestamp: string }>) => {
-      state.messages.push({
-        id: `resp-${Date.now()}-${action.payload.model}`,
-        text: 'Processing request...',
-        isUser: false,
-        model: action.payload.model,
-        timestamp: action.payload.timestamp,
-        status: 'PENDING',
-      });
+    addPendingResponse: {
+      reducer: (state, action: PayloadAction<{ id: string; model: Llm; timestamp: string }>) => {
+        state.messages.push({
+          id: action.payload.id,
+          text: 'Processing request...',
+          isUser: false,
+          model: action.payload.model,
+          timestamp: action.payload.timestamp,
+          status: 'PENDING',
+        });
+      },
+      // The model stays in the id, where it has always been, because it is what
+      // makes a pending response identifiable in devtools.
+      prepare: (response: { model: Llm; timestamp: string }) => ({
+        payload: { ...response, id: `${nextId('resp')}-${response.model}` },
+      }),
     },
     updateResponse: (state, action: PayloadAction<{ model: Llm; data: string; timestamp: string }>) => {
       const message = state.messages.find(m =>
@@ -95,6 +145,16 @@ const chatSlice = createSlice({
   },
 });
 const { addUserMessage, addSystemMessage, addPendingResponse, updateResponse, updateError, updateSuggestions, updateWelcomeMessage, reset } = chatSlice.actions;
+
+/**
+ * The slice's action creators.
+ *
+ * Exported because the three that add a message mint the id in `prepare`, so
+ * dispatching those by action type — which is how the other view model tests
+ * reach their reducers — skips the only step that generates one. A test of the
+ * ids has to go through the creators.
+ */
+export const chatActions = chatSlice.actions;
 
 /**
  * View model for managing chat interface state and AI model interactions.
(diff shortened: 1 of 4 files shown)


jooyoungseo:
Two review rounds addressed in `c266ee6` and `74476f7`.

## The timer restoration (Copilot, `chatViewModel.test.ts:53`)

Taken — moved to `afterEach`. The reasoning is right in general, and `rules/testing.md` asks for stateful cleanup there for exactly this reason.

Worth being precise about the blast radius, though, since "can contaminate subsequent tests" overstates what would have happened here. Jest gives each test *file* its own fake timers, so a leak could not reach another suite; and within this file, neither of the other two cases reads the clock — they assert prefixes and reducer purity. So the leak was real but currently inert.

That is still worth fixing rather than arguing, because of *how* it would have surfaced if either of those cases ever grew a time dependency: as a failure in a case that has nothing to do with the one that actually broke.

## The review of the diff overall

Both checkable claims verified rather than taken:

- **`rotorNavigationViewModel.ts` as precedent for exporting action creators** — confirmed, `rotorNavigationViewModel.ts:33`. It exports named creators (`export const { setValue, reset } = ...`) where this exports the object; same precedent, slightly different shape.
- **Nothing else parses the id format** — confirmed by grep across `src/`. `MessageBubble.tsx:86` is the only one, and it is pinned by a test.

On the module-scoped `sequence` being page-global rather than per-`ChatViewModel`: correct, and the note is the right one to leave. It is monotonic for the life of the module, so more instances only strengthen uniqueness — but that is a property of the counter's scope rather than of anything guarding it, which is worth having written down.

## One thing the review missed, which the format-shift note led to

Flagging the `resp-<time>-<model>` → `resp-<time>-<seq>-<model>` shift "for the record" turned out to have a concrete casualty. `footnoteScope.ts` documents the response id shape as part of explaining why it *escapes* unsafe characters rather than stripping them — so the previous commit made that docstring stale.

Correcting it surfaced a second error in the same sentence, this one predating the PR and arriving with #704:

> a response id is `resp-<time>-<provider>` and the providers are `OPENAI`, `ANTHROPIC_CLAUDE`, `GOOGLE_GEMINI` and `OLLAMA`, **three of which carry an underscore**

`Llm` is `OPENAI | ANTHROPIC_CLAUDE | GOOGLE_GEMINI | OLLAMA` (`src/type/llm.ts:7-10`) — **two** of the four carry an underscore, not three. The test comment was worse: it named `OPENAI` as one of the three, and `OPENAI` has none.

Both corrected in `74476f7`. Comments only — the escaping behaviour and the reason it must escape rather than strip are unchanged. The underscore case is still real, just half as common as claimed.

## Verification

`node scripts/test.js` — **111 suites, 1389 tests**, all passing. `npm run build`, `npm run type-check`, `npm run lint:fix` exit 0.

Counterfactual re-run after the timer change, to confirm the case still fails for the reason it is written to catch: with the sequence removed from `nextId`, the five ids collapse to three distinct and the uniqueness assertion fails (`Expected: 5, Received: 3`).

---

jooyoungseo:
The suggestion-id gap is real and is closed in `ad4a37d`. Confirmed the premise first: grep across `test/` finds no reference to `generateSuggestions` or `updateSuggestions` — the only "suggestions" hits are `llm.test.ts`, which is about model suggestions and unrelated.

## The proposed test would have passed on the bug

Worth separating, because the ask was right but the recipe would have produced a test that could never fail:

> freeze the clock, generate suggestions for an advanced-expertise state, assert the 5 ids are distinct

The ids were `suggestion-<time>-1..5` — **numbered within the batch**. One batch was already unique before this PR and is unique after it, so a single-batch distinctness assertion is green on both sides of the fix.

What actually changed is *across* batches: the `1..5` restarted on every call, so two batches generated in the same millisecond were the same five strings. That is the property the new case asserts — generate twice under a frozen clock, assert the union is the sum.

The counterfactual, with the old scheme restored:

```
● chat suggestion ids › are distinct across two batches generated in the same millisecond
  Expected: 6
  Received: 3
```

And the useful part: in that same run the single-batch assertions (`first.length > 0`, `second` same length as `first`) **passed**. That is direct evidence rather than an argument — the one-batch test would have been green on the bug.

## Two notes on how it is written

- It goes through `ChatViewModel`, not the reducer. `generateSuggestions` is a private method on the class, so `updateSuggestions()` is the public surface that reaches it — more setup than the reducer-level cases above, which is why it sits in its own `describe`.
- The services are unpopulated casts. The constructor only dispatches (`loadInitialMessage`), and this path only reads settings and messages, so neither service is touched. If that ever stops being true the call throws immediately rather than reading a plausible stub value.
- No expertise-level manipulation: the default `basic` gives three suggestions, and three per batch shows the collision as well as five does. Setting `advanced` would be testing the expertise branch, which is a different thing from the id fix.

## On the other observations

- **`reset()` not resetting `sequence`** — correct, and correct that it should not. Worth being exact about the mechanism though: `reset()` returns `initialState`, and `sequence` is module-level rather than part of the slice state, so it is not that `reset` declines to touch it — it is not reachable from a reducer at all. The consequence you name is right either way.
- **`chatActions` duplicating the internal destructure** — yes, deliberate, and flagged in the export's own doc comment for exactly the "looks like an oversight at a glance" reason.
- **`token()` being agnostic to the new segment** — right, and that is why `74476f7` was a comment-only change: the escaping behaviour did not need to move, only the sentence describing the id shape.

## Verification

`node scripts/test.js` — **111 suites, 1390 tests**, all passing. `npm run build`, `npm run type-check`, `npm run lint:fix` exit 0, run sequentially.

---

jooyoungseo:
Both points from this round land. CI is green on `ad4a37d` — all 13 checks.

## The `completedAnimations` observation is right, and I had it backwards in my own writeup

Verified at `TypingEffect.tsx:112`:

```ts
const animationKey = `${messageId}\0${text}`;
```

So yes — a third consumer of id uniqueness, and one I did not list. Under the old scheme two same-millisecond messages with identical text shared an animation key, and the second would have skipped its typing animation entirely because the first had already marked the key complete.

Worth being exact about the direction, since "harmless" undersells it slightly: the effect is not a replayed animation but a *missing* one. The message still renders in full — `displayedText` is set to the whole text on the skip path — so nothing is lost visually or to a screen reader. It is cosmetic, but it is a real behavioural difference rather than a no-op.

The reason it stayed invisible is the same as everywhere else in this bug: identical text in the same millisecond is not something a human typing produces.

## The duplicated `modelSelections` type → #707

Real, and this PR did make it worse — `addSystemMessage` went from one copy to two when the reducer split from its `prepare` callback.

Filed rather than fixed here, because the count is five, not two:

| Site | Predates this PR? |
| --- | --- |
| `src/type/llm.ts:96` (`Message.modelSelections`) | yes |
| `chatViewModel.ts:77` (reducer generic) | yes |
| `chatViewModel.ts:88` (`prepare` parameter) | **no — added here** |
| `chatViewModel.ts:132` (`updateWelcomeMessage`) | yes |
| `chatViewModel.ts:223` (`getEnabledModelsData`) | yes |

Extracting the type would touch four sites this PR does not otherwise go near, and it is a `refactor` rather than part of the `fix` — so it goes in its own change, per one-logical-change-per-commit. #707 has the detail.

One thing that surfaced while writing it up and is worth flagging there rather than here: `ModelSelection` is already a React component (`src/ui/components/ModelSelection.tsx`), so the obvious name for the extracted type is taken. Different module, so it would not collide, but it is a naming decision rather than a mechanical extraction.

## Verification

`node scripts/test.js` — **111 suites, 1390 tests**, all passing. `npm run build`, `npm run type-check`, `npm run lint:fix` exit 0, run sequentially. CI green on `ad4a37d`.

---

PR #708, opened 2026-08-02
refactor(chat): name the model-selection shape instead of repeating it

### LINKED ISSUE #707: Name the model-selection shape instead of writing it inline five times
`{ modelKey: Llm; name: string; version: string }[]` is written out five times across two files, with nothing tying the copies together:

- `src/type/llm.ts:96` — inline in `Message.modelSelections`
- `src/state/viewModel/chatViewModel.ts:77` — `addSystemMessage`'s reducer `PayloadAction` generic
- `src/state/viewModel/chatViewModel.ts:88` — `addSystemMessage`'s `prepare` parameter
- `src/state/viewModel/chatViewModel.ts:132` — `updateWelcomeMessage`'s `PayloadAction` generic
- `src/state/viewModel/chatViewModel.ts:223` — `getEnabledModelsData`'s return type (plus its `@returns` doc on line 221, a sixth copy in prose)

`rules/typescript.md` puts shared domain types in `src/type/`, so the fix is an exported interface there — `Message` already owns the canonical shape — with the other four referencing it.

### Why it is filed rather than fixed in #706

Raised as a non-blocking nit in review on #706, correctly: that PR took the `addSystemMessage` case from one copy to two by splitting the reducer from its `prepare` callback. But three of the five predate it, so extracting the type would touch code the PR otherwise does not, and it is a `refactor` rather than part of the `fix`. Filing it keeps #706 to one logical change.

### Naming

`ModelSelection` is the obvious name and is already taken by a React component (`src/ui/components/ModelSelection.tsx`), which `MessageBubble` imports. Different module, so no collision — but a type and a component sharing a name in the same codebase is worth a moment's thought rather than doing by reflex. `ModelSelectionInfo`, or naming it for what it holds rather than where it is shown, would both avoid the overlap.

### Scope

Types only, no behaviour. `npm run type-check` is the whole verification.

jooyoungseo:
# Pull Request

## Description

`{ modelKey: Llm; name: string; version: string }[]` was written out **six** times across three files, with nothing tying the copies together.

## Related Issues

Closes #707. Raised as a non-blocking nit in review on #706.

## Changes Made

`SelectedModel` is exported from `src/type/llm.ts`, next to `Message` — which owns the canonical shape — per `rules/typescript.md` putting shared domain types in `src/type/`.

| site | what it was |
|---|---|
| `type/llm.ts` | inline in `Message.modelSelections` |
| `chatViewModel.ts` | `addSystemMessage`'s reducer `PayloadAction` generic |
| `chatViewModel.ts` | `addSystemMessage`'s `prepare` parameter |
| `chatViewModel.ts` | `updateWelcomeMessage`'s `PayloadAction` generic |
| `chatViewModel.ts` | `getEnabledModelsData`'s return type |
| `ModelSelection.tsx` | `ModelSelectionProps.enabledModels` |

A seventh copy sat in prose, in `getEnabledModelsData`'s `@returns`.

#707 listed five of these. The sixth — `ModelSelectionProps` — turned up while doing the work, and is the one that decided the name.

## On the name

`ModelSelection` is the obvious choice and the wrong one. It is already a React component, and **that component's own file is one of the sites that has to import the type**, so the collision would land inside the file least able to work around it. `MessageBubble` is in the same position from the other side: it imports the `ModelSelection` component *and* reads `message.modelSelections`.

`SelectedModel` matches the field it types and collides with nothing.

## `version` is `LlmVersion`, and that is naming rather than narrowing

Raised in review, by both reviewers, as tightening the field. It tightens nothing, and the docstring now says so, because the alternative is that someone later reads it as validation.

```ts
// src/type/llm.ts — the string-with-autocomplete trick
export type OllamaVersion = string & Record<never, never>;
export type LlmVersion = GptVersion | ClaudeVersion | GeminiVersion | OllamaVersion;

// so a union containing `string` IS `string`:
const nonsense: LlmVersion = 'definitely-not-a-real-model-xyz';  // compiles
declare const anyString: string;
const fromString: LlmVersion = anyString;                        // compiles, no cast
```

This is deliberate, and `OllamaVersion`'s own comment says why: model dropdowns are populated from live provider APIs, so the curated literals are suggestions, and correctness is enforced at runtime by `getValidVersion`.

The change is still worth making — `getValidVersion` is declared to return `LlmVersion` and is where the value comes from, so the field naming that type says what it holds rather than merely how wide it is.

## Checklist

- [ ] I have added appropriate unit tests, if applicable. — types only; `tsc` is the test, see below.

### What was left alone

`Settings.tsx:183` also has a `modelKey: Llm` field, but it is `LlmModelSettingRowProps` — a settings row's props, a different shape that happens to share one member. Not folded in.

### Verification

Types only, no behaviour, so `tsc` carries the whole check. On its own that only proves the code still compiles — it does not show the sites are actually *bound* to the new type rather than each independently compatible with it. So: renaming `name` to `label` on the extracted interface.

```
src/state/viewModel/chatViewModel.ts(243,29): error TS2322: Type '{ modelKey: Llm; name: string; version: LlmVersion; }[]'
  is not assignable to type 'SelectedModel[]'.
src/ui/components/ModelSelection.tsx(125,20): error TS2339: Property 'name' does not exist on type 'SelectedModel'.
src/ui/components/ModelSelection.tsx(129,36): error TS2339: Property 'name' does not exist on type 'SelectedModel'.
```

Three failures spanning the producer (`getEnabledModelsData`) and the consumer (`ModelSelection` reading `.name`) — which is the property that matters. The four `PayloadAction` generics do not appear, and should not: they pass the type through without constructing or reading a value, so there is nothing there for a field rename to break.

Note the contrast with the `version` swap above, which produced zero errors — that was not evidence of anything, since swapping a type for what amounts to its alias cannot fail. `name` is the field where a rename actually bites.

| check | result |
|---|---|
| `npm run build` | exit 0 |
| `npx eslint .` | exit 0 |

CI: 13/13 green on `8743676`.

---

---

### DIFF
--- a/src/state/viewModel/chatViewModel.ts
+++ b/src/state/viewModel/chatViewModel.ts
@@ -2,7 +2,7 @@ import type { PayloadAction } from '@reduxjs/toolkit';
 import type { AudioService } from '@service/audio';
 import type { ChatService } from '@service/chat';
 import type { Suggestion } from '@type/chat';
-import type { Llm, Message } from '@type/llm';
+import type { Llm, Message, SelectedModel } from '@type/llm';
 import type { AppStore, RootState } from '../store';
 import { createSlice } from '@reduxjs/toolkit';
 import { MODEL_VERSIONS } from '@service/modelVersions';
@@ -74,7 +74,7 @@ const chatSlice = createSlice({
       }),
     },
     addSystemMessage: {
-      reducer: (state, action: PayloadAction<{ id: string; text: string; timestamp: string; modelSelections?: { modelKey: Llm; name: string; version: string }[]; isWelcomeMessage?: boolean }>) => {
+      reducer: (state, action: PayloadAction<{ id: string; text: string; timestamp: string; modelSelections?: SelectedModel[]; isWelcomeMessage?: boolean }>) => {
         state.messages.push({
           id: action.payload.id,
           text: action.payload.text,
@@ -85,7 +85,7 @@ const chatSlice = createSlice({
           isWelcomeMessage: action.payload.isWelcomeMessage,
         });
       },
-      prepare: (message: { text: string; timestamp: string; modelSelections?: { modelKey: Llm; name: string; version: string }[]; isWelcomeMessage?: boolean }) => ({
+      prepare: (message: { text: string; timestamp: string; modelSelections?: SelectedModel[]; isWelcomeMessage?: boolean }) => ({
         payload: { ...message, id: nextId('system') },
       }),
     },
@@ -129,7 +129,7 @@ const chatSlice = createSlice({
     updateSuggestions: (state, action: PayloadAction<Suggestion[]>) => {
       state.suggestions = action.payload;
     },
-    updateWelcomeMessage: (state, action: PayloadAction<{ text: string; modelSelections?: { modelKey: Llm; name: string; version: string }[] }>) => {
+    updateWelcomeMessage: (state, action: PayloadAction<{ text: string; modelSelections?: SelectedModel[] }>) => {
       // Find the welcome message (first system message with isWelcomeMessage flag)
       const welcomeMessageIndex = state.messages.findIndex(msg => msg.isWelcomeMessage);
       if (welcomeMessageIndex !== -1) {
@@ -218,9 +218,9 @@ export class ChatViewModel extends AbstractViewModel<ChatState> {
 
   /**
    * Retrieves data about enabled AI models including display names and versions.
-   * @returns {{ enabledModels: string[]; modelSelections: { modelKey: Llm; name: string; version: string }[] }} Enabled models data.
+   * @returns {{ enabledModels: string[]; modelSelections: SelectedModel[] }} Enabled models data.
    */
-  private getEnabledModelsData(): { enabledModels: string[]; modelSelections: { modelKey: Llm; name: string; version: string }[] } {
+  private getEnabledModelsData(): { enabledModels: string[]; modelSelections: SelectedModel[] } {
     const llmModels = this.snapshot.settings.llm.models;
 
     const enabledModels = Object.entries(llmModels)
--- a/src/type/llm.ts
+++ b/src/type/llm.ts
@@ -89,11 +89,17 @@ export interface LlmResponse {
  *
  * Held on the message rather than read from settings at render time, so a
  * message keeps saying which model answered it after the setting changes.
+ *
+ * `version` is `LlmVersion` to agree with `getValidVersion`, which produces it.
+ * That is naming, not narrowing: `OllamaVersion` is `string & Record<never,
+ * never>`, so the union accepts any string — deliberately, as the comment on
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Both reviewers raised the same point independently, and it is worth making — `8743676`. But it does **not** do what either description says, and that is worth pinning down before it becomes folklore.

## The change is right; "tightening" is not

> tightening it to `LlmVersion` could be a good follow-up

> Using `LlmVersion` here keeps the type consistent and **preserves intent for callers/consumers**

`LlmVersion` constrains nothing. `OllamaVersion` is the string-with-autocomplete trick:

```ts
export type OllamaVersion = string & Record<never, never>;
export type LlmVersion = GptVersion | ClaudeVersion | GeminiVersion | OllamaVersion;
```

A union containing `string` **is** `string`. The curated literals survive only as autocomplete suggestions — and this is deliberate, spelled out on `OllamaVersion` itself:

> Deliberate tradeoff: because this type accepts any string, the `LlmVersion` union below also widens to accept any string for every provider. That is intentional — model dropdowns are populated from live provider APIs, so valid versions are not limited to the curated literals; correctness is enforced at runtime (`getValidVersion`, provider responses), not at compile time.

## Checked, not inferred

```ts
const nonsense: LlmVersion = 'definitely-not-a-real-model-xyz';  // compiles
declare const anyString: string;
const fromString: LlmVersion = anyString;                        // compiles, no cast
```

Both clean. `string` → `LlmVersion` needs no assertion, which makes them the same type structurally.

That also reframes my own evidence: I reported "zero type errors" when I first tried the swap and treated it as a green light. It was not evidence of anything — swapping a type for what amounts to its alias cannot produce errors. Had `LlmVersion` genuinely narrowed, zero errors would have been the surprising result worth investigating, since `getEnabledModelsData` builds these values from settings.

## So why make the change at all

Because it is honest naming, which is a real if smaller benefit: `getValidVersion` is declared to return `LlmVersion` and is where the value comes from, so the field naming the same type says what it holds instead of just how wide it is. The docstring now records the distinction, so the next reader does not mistake it for validation:

> `version` is `LlmVersion` to agree with `getValidVersion`, which produces it. That is naming, not narrowing … Nothing here validates a version; `getValidVersion` does, at runtime.

If someone does want a version type that actually rejects a bad value, that is a different and much larger change — it would mean giving up the live-provider-list behaviour the current design chose on purpose. Happy to file it if wanted, but it should not ride along here.

## The rest of the review

- **Shape matches the six copies, no drift** — agreed, and that is what the field-rename check was for.
- **`Llm`/`LlmVersion` still used in both files** — correct. Worth noting I did break this briefly: I dropped `Llm` from `ModelSelection.tsx`'s import while editing and put it back before committing.
- **`Settings.tsx:183` left alone** — yes, `LlmModelSettingRowProps` shares one member and is otherwise a different shape.

## Verification

`node scripts/test.js` — **111 suites, 1390 tests**, all passing. `npm run build`, `npm run type-check`, `npm run lint:fix` exit 0, run sequentially.

---

PR #709, opened 2026-08-02
ci: lint the workflow files with actionlint

### LINKED ISSUE #693: ci: run actionlint on .github/workflows in CI
## Problem

Nothing in CI validates the workflow files. `actionlint` catches real breakage — a bad `if:` expression, an unknown context, a malformed matrix — but today it only runs if someone remembers to run it locally before pushing.

That is a gap with teeth, because a broken scheduled workflow does not announce itself. `e2e_tests.yml` fires on `schedule`, so a syntax or expression error lands silently and surfaces two days later as "the suite did not run" — or not at all, if nobody is watching for a run that never happened.

## Context

Raised during review of #692, which added ~60 lines of logic to `e2e_tests.yml`'s `github-script` step. Both the reviewer and I agreed it was worth doing and that it was out of scope there, since it changes `ci.yml` and affects every PR's pipeline rather than that one workflow.

## Suggested shape

A job in `ci.yml`, alongside the existing lint jobs:

```yaml
- name: Lint workflows
  run: |
    bash <(curl -s https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
    ./actionlint -color
```

Pinning the version rather than tracking `main` is probably worth it, so a new actionlint release cannot turn an unrelated PR red.

Worth checking whether it should be `continue-on-error` at first — actionlint may flag pre-existing issues across the other workflows, and it would be better to see that list before making it blocking.

## Acceptance

- `actionlint` runs on every PR touching `.github/workflows/`
- A deliberately malformed workflow fails the job
- Any pre-existing findings in other workflow files are either fixed or explicitly waived

jooyoungseo:
# Pull Request

## Description

Nothing validated `.github/workflows`. That matters most where a mistake is quietest: `e2e_tests.yml` runs on `schedule`, so a bad `if:` or an unknown context fails nothing — the run simply never happens, and an absence is not something anyone is watching for.

## Related Issues

Closes #693. Raised during review of #692.

## Changes Made

Two commits, in this order because the second depends on the first.

| commit | |
|---|---|
| `add7442` | quote `$GITHUB_OUTPUT` in `release.yml` — the one pre-existing finding |
| `e1a98fa` | add the `Lint Workflows` job to `ci.yml` |

## The question the issue left open

> Worth checking whether it should be `continue-on-error` at first — actionlint may flag pre-existing issues across the other workflows, and it would be better to see that list before making it blocking.

Checked by running it. There was **exactly one** finding across all twelve workflows:

```
.github/workflows/release.yml:60:9: shellcheck reported issue in this script:
  SC2086:info:3:30: Double quote to prevent globbing and word splitting [shellcheck]
```

`echo "new_release=true" >> $GITHUB_OUTPUT`, unquoted. The other three writes to a GitHub file command across the workflows already quote it, so this was the outlier rather than a style preference being imposed. Harmless in practice — `GITHUB_OUTPUT` is a runner-controlled path with no spaces or globs — and fixed in `add7442`.

So there is nothing to phase in, and the job goes in **blocking** rather than `continue-on-error`.

## The shellcheck assertion is not ceremony

The job asserts `shellcheck --version` before linting. That step exists because of how this change nearly went wrong.

The **first** local run of actionlint reported zero problems across every workflow. That looked like a clean bill of health and would have justified going straight to a blocking job with no `release.yml` fix at all. It was wrong: actionlint hands `run:` blocks to shellcheck and, when shellcheck is absent, checks less **without saying so**. Neither `shellcheck` nor `pyflakes` was installed locally. Installing shellcheck and re-running is what surfaced the SC2086 above.

`ubuntu-latest` ships shellcheck, so the job gets the full check today. Asserting it makes the day that stops being true a failure rather than a silently weaker check — which is the exact failure mode that just happened locally.

### A version skew the assertion exposed

The first CI run printed `ShellCheck ... version: 0.9.0`. Local verification used **0.10.0**, so this was never the like-for-like comparison it might read as.

The consequence is small but real: a rule added between 0.9.0 and 0.10.0 would fail locally and pass CI, and one removed would do the reverse. Not pinning shellcheck to close it — that means installing a second binary to make a linter agree with itself, for drift that produces a *confusing* result rather than a wrong one, and the assertion step prints the version into every run's log so any future disagreement is one glance from being explained.

Both versions agree on this tree: SC2086 under 0.10.0 locally before the fix, clean under 0.9.0 in CI after it.

## Pinning

The issue sketched the upstream one-liner:

```yaml
bash <(curl -s https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
```

Used a pinned release tarball with a checksum instead — `1.7.7`, SHA256 `023070a2…`. A new actionlint release cannot turn an unrelated PR red, and the binary is verified rather than trusted.

Verified three ways: against the release's own published `actionlint_1.7.7_checksums.txt`, against an independently re-downloaded artifact, and by the job itself on GitHub's runner (`actionlint.tar.gz: OK`).

## Scope

Runs on every PR, not only those touching `.github/workflows/`. Job-level `paths` filters do not exist, and a workflow-level one would gate every other job in `ci.yml` too. The job takes seconds, so always running it is the cheaper option — and it satisfies the acceptance criterion as a superset.

## Checklist

- [ ] I have added appropriate unit tests, if applicable. — see below.

### On tests

`test/scripts/` covers workflow bodies, but `e2eReportWorkflow.test.ts` is there to exercise a `github-script` **JS body** — executable logic extracted from YAML. This job has no JS, only shell, and nothing in `test/` reads `ci.yml`. A test asserting "`ci.yml` contains an actionlint job" would only re-read the file it is testing, so the verification below is the real thing.

### Acceptance criteria, demonstrated

> A deliberately malformed workflow fails the job

| injected fault | result |
|---|---|
| unknown step context — `steps.no_such_step.outputs.thing` | ✅ exit 1, `property "no_such_step" is not defined` |
| misspelled runner label — `ubuntu-lastest` | ✅ exit 1, `label "ubuntu-lastest" is unknown` |
| `runs:` instead of `run:` | ✅ exit 1, `unexpected key "runs" for "step" section` |
| bogus property on `github.event` | ❌ **passes** — and correctly so |

The last one is worth recording rather than quietly dropping. It was my first attempt at a failing case and it did not fail: webhook payloads vary by event, so actionlint types `github.event` as open and cannot flag an arbitrary property on it. Not a gap in the check — a limit of what is knowable there.

> `actionlint` runs on every PR touching `.github/workflows/`

Superset — runs on every PR. This PR is itself the first live case, and the job passed in 4 s.

> Any pre-existing findings in other workflow files are either fixed or explicitly waived

One finding, fixed in `add7442`. Nothing waived.

### Verification

| check | result |
|---|---|
| `actionlint` locally, shellcheck on PATH | exit 0 on the tree as it stands |
| `actionlint` in CI (`Lint Workflows`) | ✅ 4 s |

`actionlint 1.7.7`; shellcheck `0.10.0` locally, `0.9.0` on the runner — see the skew note above.

---

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,50 @@ jobs:
       - name: Run linter
         run: npm run lint
 
+  # Nothing validated the workflow files, which matters most where a mistake is
+  # quietest: e2e_tests.yml runs on `schedule`, so a bad `if:` or an unknown
+  # context does not fail anything — the run just never happens, and nobody is
+  # watching for an absence. See issue #693.
+  #
+  # Runs on every PR rather than only those touching .github/workflows/. Job-level
+  # path filters do not exist, and a workflow-level one would gate every other job
+  # here too. The job takes seconds, so the cheaper thing is to always run it.
+  actionlint:
+    name: Lint Workflows
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      # actionlint hands `run:` blocks to shellcheck and, when it is missing,
+      # simply checks less without saying so. That is not hypothetical: the
+      # first local run of this change reported zero problems, and the one real
+      # finding it had to fix only appeared once shellcheck was installed.
+      # ubuntu-latest ships it — asserting that here makes the day it stops a
+      # failure rather than a silently weaker check.
+      - name: Verify shellcheck is available
+        run: shellcheck --version
+
+      # Pinned by version and checksum rather than piping the upstream install
+      # script into bash, so a new actionlint release cannot turn an unrelated
+      # PR red and the download is verified rather than trusted.
+      - name: Install actionlint
+        env:
+          ACTIONLINT_VERSION: 1.7.7
+          ACTIONLINT_SHA256: 023070a287cd8cccd71515fedc843f1985bf96c436b7effaecce67290e7e0757
+        run: |
+          curl -sSLo actionlint.tar.gz \
+            "https://github.com/rhysd/actionlint/releases/download/v${ACTIONLINT_VERSION}/actionlint_${ACTIONLINT_VERSION}_linux_amd64.tar.gz"
+          echo "${ACTIONLINT_SHA256}  actionlint.tar.gz" | sha256sum --check
+          tar xzf actionlint.tar.gz actionlint
+
+      - name: Lint workflows
+        run: ./actionlint -color
+
   # Playwright on every PR: chromium only, core bundle only. The whole
   # chromium suite is green, so it all runs here — a subset let the specs rot
   # for two months while PRs stayed green (see issue #626). The other two
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -60,7 +60,7 @@ jobs:
         run: |
           npx semantic-release
           if git describe --tags --exact-match HEAD 2>/dev/null; then
-            echo "new_release=true" >> $GITHUB_OUTPUT
+            echo "new_release=true" >> "$GITHUB_OUTPUT"
           fi
 
       - name: Publish to npm with trusted publishing


jooyoungseo:
**"Lint Workflows" passed on its first live run** — 4 seconds of actual work. That settles three of the four observations with evidence rather than argument.

## Point 3 — the checksum, independently verified twice

You were right not to take my word for it. Two confirmations, neither of which is me repeating myself:

**On GitHub's runner**, a fresh download on a different machine:

```
env:
  ACTIONLINT_VERSION: 1.7.7
  ACTIONLINT_SHA256: 023070a287cd8cccd71515fedc843f1985bf96c436b7effaecce67290e7e0757
actionlint.tar.gz: OK
```

**Locally**, re-fetching the published checksums file and re-downloading the artifact from scratch:

```
$ curl -sSL .../actionlint_1.7.7_checksums.txt | grep linux_amd64
023070a287cd8cccd71515fedc843f1985bf96c436b7effaecce67290e7e0757  actionlint_1.7.7_linux_amd64.tar.gz

$ curl -sSLo fresh.tar.gz .../actionlint_1.7.7_linux_amd64.tar.gz && sha256sum fresh.tar.gz
023070a287cd8cccd71515fedc843f1985bf96c436b7effaecce67290e7e0757  fresh.tar.gz
```

Published checksum, freshly downloaded artifact, the pinned constant, and the runner's own verification all agree.

## The shellcheck assertion earned its keep immediately — and surfaced something I had wrong

```
Run shellcheck --version
ShellCheck - shell script analysis tool
version: 0.9.0
```

Present on `ubuntu-latest`, so the job gets the full check. But note the version: **the runner has 0.9.0 and I used 0.10.0 locally.**

I did not know that when I opened this, and the PR description implies a like-for-like comparison it does not have. The consequence is small but real: local and CI can in principle disagree, and a finding introduced between 0.9.0 and 0.10.0 would pass CI while failing locally, or a rule removed in 0.10.0 would do the reverse.

Not worth pinning shellcheck too — that would mean installing a second binary to make a linter agree with itself, for a class of drift that produces a *confusing* result rather than a wrong one, and the assertion step already prints the version into the log so any future disagreement is one glance away from being explained. Flagging it because the description overstated the equivalence, not because it needs fixing here.

Both versions agree on this tree in any case: `release.yml`'s SC2086 under 0.10.0 locally, and clean under 0.9.0 in CI after the fix.

## Point 1 — `linux_amd64`

Agreed on all counts, including that it fails loudly rather than silently, which is the posture the rest of this argues for. Worth adding one thing: if `ubuntu-latest` ever did move to arm64, the `sha256sum --check` would not be what catches it — the URL 404s first and `curl -sSLo` writes an empty file, so `-e` fails the step at the download. Same outcome, one step earlier than your description suggests.

## Point 2 — caching

Agreed, non-issue. Concretely: the whole job is 4 s and the download is ~2 MB, so `actions/cache` round-tripping its own restore/save would likely cost more than it saves.

## Point 4 — the tautological test

Agreed, and this run is the argument for it. The verification that mattered is the job running green against the real repo, which no self-referential unit test could have told us.

---

Remaining CI is still running (Unit Tests, E2E, Build, Docs). Will report when it settles.

---

Issue #672, opened 2026-07-30
chore: list maidr on cdnjs (blocker cleared by the GPL-3.0 declaration)


jooyoungseo:
The licence blocker is gone as of #661 and the v3.75.0 release, so MAIDR can now be submitted to cdnjs.

## Why bother, given jsDelivr already works

Sandboxed embedding contexts allow a fixed set of CDN hosts, and the two big ones differ. The claude.ai Artifact sandbox, for example, permits both:

```
script-src … https://cdnjs.cloudflare.com https://cdn.jsdelivr.net/npm/ …
```

but guidance given to models inside that product points at cdnjs specifically. Being on both removes a class of "the CDN I was told to use doesn't carry it" failure. It costs one JSON file.

## Requirements, checked against current state

| Requirement | Status |
|---|---|
| npm downloads ≥ 800/month | **2,301/month** ✅ |
| Auto-updatable (npm tagged releases) | semantic-release in place ✅ |
| File types (js/css allowed) | `maidr.js`, `maidr.css` ✅ |
| Appropriate licence permissions | `GPL-3.0-or-later` as of #661, published in v3.75.0 ✅ |
| Name free on cdnjs | no `packages/m/maidr*.json` ✅ |

(The 200-GitHub-stars figure in cdnjs's guidelines applies to git-sourced libraries; npm-sourced ones are judged on downloads. Both remain at maintainer discretion.)

## Submission

Fork [`cdnjs/packages`](https://github.com/cdnjs/packages) — **not** `cdnjs/cdnjs`, which is robot-only — and add `packages/m/maidr.json`:

```json
{
  "name": "maidr",
  "description": "Multimodal Access and Interactive Data Representation — non-visual access to data visualizations via sonification, text descriptions, and braille.",
  "keywords": ["accessibility", "a11y", "data-visualization", "sonification", "braille", "screen-reader"],
  "filename": "maidr.min.js",
  "homepage": "https://maidr.ai",
  "repository": { "type": "git", "url": "https://github.com/xability/maidr.git" },
  "license": "GPL-3.0-or-later",
  "autoupdate": {
    "source": "npm",
    "target": "maidr",
    "fileMap": [
      { "basePath": "dist", "files": ["maidr.js", "maidr.css"] }
    ]
  }
}
```

[...]

PR #713, opened 2026-08-03
chore(cdnjs): add the package manifest for listing maidr on cdnjs

jooyoungseo:
# Pull Request

## Description

Prepares the cdnjs listing that the GPL-3.0-or-later declaration unblocked, and
does the parts of it that live in this repository: the submission payload, the
metadata it declares, and a test that keeps both true about the build.

The submission itself has to be a pull request against
[`cdnjs/packages`](https://github.com/cdnjs/packages) — cdnjs serves packages
from its own repository, and nothing here can put a file there. That step is
still a maintainer action; `cdnjs/README.md` in this PR is the instruction for
it, and the JSON is copy-paste ready.

## Related Issues

Addresses #672. Follows #661 (the `GPL-3.0-or-later` declaration) and #673 (the
KaTeX split that created `maidr-math.css`).

## Changes Made

**`cdnjs/maidr.json`** — the payload, as drafted in #672 with one correction:
the file map mirrors `maidr-math.css` as well as `maidr.js` and `maidr.css`.
`src/util/katex.ts` links the maths stylesheet at runtime, resolving it against
the URL `maidr.js` was loaded from — so on a page served from cdnjs it resolves
to a cdnjs URL, and an unmirrored file makes that a 404. The failure is quiet
and narrow: LaTeX in AI chat responses renders unstyled, everything else works,
and no page that never opens the chat notices.

Three files listed by name, no glob — cdnjs
[asks for that](https://github.com/cdnjs/packages/issues/186), and it matters
more here than most: `dist/*.js` would pull every adapter bundle plus every
sourcemap, and `dist/maidr.js.map` alone is 8.7 MB.

**`cdnjs/README.md`** — why the file is in this repo, how to submit it, what is
mirrored and why only that, and what to do once cdnjs approves (the cdnjs URLs
are deliberately *not* added to `README.md`/`docs/` yet — until the listing
lands they 404). The claims about cdnjs's own process — branch convention,
commit format, and that it generates `maidr.min.js` itself — are quoted from
[their CONTRIBUTING.md](https://github.com/cdnjs/packages/blob/master/CONTRIBUTING.md)
and date-stamped, since that process has changed before and this submission may
not happen for a while.

**`package.json`** — adds `description`, `homepage` and `keywords`. npm shows
none of the three for this package today, so it has no description, no link to
maidr.ai, and nothing for keyword search to match on. They are also what the
manifest declares, so it now has a single source of truth to be checked against.

**`README.md`** — the usage snippet linked `dist/maidr_style.css` and the prose
named `styles.css`. Neither has ever been emitted by this repo's build, which
writes `dist/maidr.css` — the name py-maidr and r-maidr link, and the one this
manifest mirrors. The first snippet a new user copies was asking a CDN for a
file that is not there.

**`.claude/rules/testing.md`** (and its synced `.github/instructions/`
counterpart) — the convention named `test/scripts/` as *the* exception to
`test/` mirroring `src/`, which this PR's `test/cdnjs/` made inaccurate. It now
states the general rule: a test directory mirrors either a layer of `src/` or a
directory at the repository root, under the same name. Both instances are
listed, with the bar for adding a third.

N/A — no user-visible interface change.

## Checklist

## Additional Notes

`test/cdnjs/manifest.test.ts` is the point of the exercise. Nothing in this
repository reads `cdnjs/maidr.json`, so without a test a rename in `dist` or a
dropped field would break the mirror silently, months later, somewhere no CI run
of ours can see. It pins the manifest's metadata against `package.json` and
every mirrored filename against what `scripts/build.js` actually emits — reading
both out of a node subprocess, the same way `test/scripts/buildOutputFilenames`
and `test/scripts/mathStylesheet` do, since those modules are plain ESM JS and
`allowJs` is false, and resolving `fileName(format, entryName)` the way Vite
invokes it rather than assuming the callbacks ignore their arguments.

Mutation-checked throughout: dropping `maidr-math.css`, adding a `dist/*.js`
glob, adding a `.map`, adding an adapter bundle, adding or removing a manifest
field, blanking a value, emptying the nested `files` array, and reordering
`keywords` each fail it.

What it deliberately does not assert is anything about cdnjs's schema. An
earlier revision enumerated cdnjs's required fields, carried over from #672;
that list could not be confirmed against their schema, and since their checker
is what actually decides, an unverified list is worse than none. The test states
what this repository can state — the fields its own listing is built from.

Deliberately out of scope, flagged in #672: the ~107 MB unpacked npm tarball
(mostly sourcemaps) is a separate question, and the narrow file map is this PR's
only interaction with it.

Verified locally: `npm run lint`, `npm run type-check`, `npm run build`, and
`npm test` — 108 unit suites (1353 tests) plus 5 ESM suites (57 tests), all
passing.

### DIFF
--- a/.claude/rules/testing.md
+++ b/.claude/rules/testing.md
@@ -30,9 +30,21 @@ forget, because a project the runner omits simply never runs.
 
 `test/` mirrors `src/` — `test/model/`, `test/service/`, `test/state/`,
 `test/command/`, `test/ui/`, `test/util/`, `test/adapters/`. Put a new unit
-test in the directory matching the layer it covers. `test/scripts/` is the
-exception, and covers `scripts/` rather than a layer of the application: the
-build, the test runner, and the workflow bodies under `.github/`.
+test in the directory matching the layer it covers.
+
+Not everything worth testing is a layer of the application, so a directory here
+may instead mirror one at the repository root, under the same name:
+
+- `test/scripts/` covers `scripts/` — the build, the test runner, and the
+  workflow bodies under `.github/`.
+- `test/cdnjs/` covers `cdnjs/`, the payload for the cdnjs listing, which
+  nothing in this repository reads and which therefore has nowhere else to be
+  checked.
+
+Both follow the same rule as the layer directories: the test lives in the
+directory named after what it covers. Add another only when the thing under
+test genuinely is not application code — a config file that has to stay in step
+with the build is; a new module of `src/` is not.
 
 ```bash
 npm test              # jest, with coverage, one process per project
--- a/.github/instructions/testing.instructions.md
+++ b/.github/instructions/testing.instructions.md
@@ -29,9 +29,21 @@ forget, because a project the runner omits simply never runs.
 
 `test/` mirrors `src/` — `test/model/`, `test/service/`, `test/state/`,
 `test/command/`, `test/ui/`, `test/util/`, `test/adapters/`. Put a new unit
-test in the directory matching the layer it covers. `test/scripts/` is the
-exception, and covers `scripts/` rather than a layer of the application: the
-build, the test runner, and the workflow bodies under `.github/`.
+test in the directory matching the layer it covers.
+
+Not everything worth testing is a layer of the application, so a directory here
+may instead mirror one at the repository root, under the same name:
+
+- `test/scripts/` covers `scripts/` — the build, the test runner, and the
+  workflow bodies under `.github/`.
+- `test/cdnjs/` covers `cdnjs/`, the payload for the cdnjs listing, which
+  nothing in this repository reads and which therefore has nowhere else to be
+  checked.
+
+Both follow the same rule as the layer directories: the test lives in the
+directory named after what it covers. Add another only when the thing under
+test genuinely is not application code — a config file that has to stay in step
+with the build is; a new module of `src/` is not.
 
 ```bash
 npm test              # jest, with coverage, one process per project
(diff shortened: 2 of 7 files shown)


jooyoungseo:
Both "worth a second look" points turned out to be checkable, so I checked them against [cdnjs's CONTRIBUTING.md on `master`](https://github.com/cdnjs/packages/blob/master/CONTRIBUTING.md) rather than leaving them as documented assumptions. Pushed in `117a3bc`.

**`filename: "maidr.min.js"` holds.** Their guide commits to it directly: "for JavaScript and CSS files, we'll automatically generate minified versions of them and make them available at `filename.min.js` or `filename.min.css`." Generation is on by default and only suppressed by the optional `optimization` property, which this manifest does not set. KaTeX's own entry names `katex.min.js` on the same basis. `cdnjs/README.md` now quotes that rather than asserting it, and keeps `maidr.js` as the fallback.

**The submission steps hold too.** The commit convention is `Add <library> w/ npm/git auto-update`, so `Add maidr w/ npm auto-update` is right, and branches want a descriptive name — the doc now suggests `add-maidr`.

**One thing I could not verify, so I removed it.** The doc previously enumerated `name`, `description`, `keywords`, `repository` and `autoupdate` as cdnjs's required fields, carried over from #672. Their schema is not at any path I could reach, and CONTRIBUTING.md does not enumerate them — it points at existing packages instead. Since the checker is what actually decides, an unchecked list is worse than none, so that sentence is gone and the doc now just says CI validates against their schema.

Both verified statements are date-stamped in the file, because cdnjs's process has changed before and this submission may not happen for a while.

On the em-dash note in `package.json`'s `description`: leaving it. npm's registry, website and CLI are all UTF-8 throughout, and the character already appears in the repo's own prose; I'd rather not degrade the description for a rendering problem I can't reproduce.

---

Issue #703, opened 2026-08-02
test: report the jest-resolve cross-project ESM memoisation bug upstream


jooyoungseo:
`scripts/test.js` runs one Jest process per project, and `scripts/testArgs.js` exists to make that work, because of a bug in `jest-resolve`. If that bug were fixed upstream, most of both files could go away.

Raised in review on #702, which is where the workaround landed.

## The bug

`jest-resolve` memoises "is this file ESM?" in a module-scoped `Map` keyed on the **path alone**, ignoring the `extensionsToTreatAsEsm` of the project doing the asking:

```js
// node_modules/jest-resolve/build/shouldLoadAsEsm.js:44
function cachedShouldLoadAsEsm(path, extensionsToTreatAsEsm) {
  let cachedLookup = cachedFileLookups.get(path);
  if (cachedLookup === undefined) {
    cachedLookup = shouldLoadAsEsm(path, extensionsToTreatAsEsm);
    cachedFileLookups.set(path, cachedLookup);   // <- key is path only
  }
  return cachedLookup;
}
```

The map is process-wide, so in a `projects` run the first project to ask about a shared `src/` module answers for both. Whichever runs second gets:

```
$ node scripts/test.js --runInBand --coverage=false "typingEffect.esm-test|version.test"
PASS esm  test/ui/typingEffect.esm-test.tsx
FAIL unit test/util/version.test.ts
     Must use import to load ES Module: /home/user/maidr/src/util/version.ts
```

Which project loses depends on Jest's scheduling, so the symptom is a suite that passes alone and fails in company — 43 suites red with a warm cache, 3 with `--no-cache`, a different set each run.

Jest 29.7.0.

## Why it matters here

The whole "one process per project" design exists to avoid it, and everything that follows from that design exists to make the split invisible:

- `runProjects` re-adds `--selectProjects` one project at a time
- `takeSelection` intercepts the flag so a multi-project selection cannot collapse back into one process
- `listTests` asks Jest what a filter matches so a per-project run does not fail on a project the filter was never aimed at
- `hasPathFilter` / `isNarrowed` decide when a project matching nothing is legitimate

Five of the review rounds on #702 were spent on edge cases in that machinery. None of it would be needed if a project's `extensionsToTreatAsEsm` were part of the cache key.

## What to do

1. Check whether this is already reported on `jestjs/jest` — the memo predates Jest 29 and the `projects` + mixed-module-system combination is not exotic, so it may well be known.
[...]

PR #714, opened 2026-08-03
refactor(ci): move the e2e report body into a module

### LINKED ISSUE #694: refactor(ci): move the e2e report github-script body into a module
## Problem

The report step in `.github/workflows/e2e_tests.yml` now carries ~150 lines of JavaScript inside a YAML block scalar. Two consequences:

**It is the only code in the repo that `eslint` and `tsc` never see** — and it runs with `issues: write`.

**Testing it requires extracting it from the YAML.** `test/scripts/e2eReportWorkflow.test.ts` reads the block out textually so that what is tested is what runs. That works, but the extraction has its own failure modes, and every one of them found so far produced a **passing** suite that was not testing the script:

| defect | symptom |
|---|---|
| a second `script:` block added above it | extracts the wrong block, 14/14 pass |
| block indented differently | slices into the code, or silently keeps leading whitespace |
| `\|-` / `\|2-` block headers | reports "no script block", as if the step were deleted |
| `Error` not shared with the `vm` sandbox | `instanceof Error` is false, so the production branch of `reason()` is never exercised |

Each is now guarded and pinned by a test, but four of #692's fourteen commits exist only to close that class. The class itself is avoidable.

## Suggested shape

Move the body to something like `scripts/ci/e2eReport.mjs` and have the step call it:

```yaml
- name: Create or update test report issue
  if: always()
  uses: actions/github-script@v7
  env:
    TEST_STATUS: ${{ steps.e2e.outcome }}
  with:
    script: |
      const report = require('./scripts/ci/e2eReport.mjs');
      await report({ github, context, core });
```

Then `test/scripts/e2eReportWorkflow.test.ts` imports the module directly, and `reportScript()`, `SCRIPT_HEADER`, the multi-block guard, the indentation measurement, and the shared-`Error` sandbox all get deleted with it. The 17 behavioural cases stay as they are — they exercise behaviour, not extraction.

The script also picks up `eslint` and `tsc` coverage for free, which is the larger win.

## Context

Raised independently by two reviewers on #692. Deliberately not done there: the extraction had just been stabilised and pinned, and swapping it out mid-review would have meant the merged code was not the reviewed code.

## Acceptance

- The workflow step delegates to a module under `scripts/ci/`
- `eslint` and `tsc` cover that module
- The test imports it rather than extracting it, and the extraction helpers are gone
- The existing 17 cases still pass unchanged
- A scheduled run (or `workflow_dispatch`) is observed to still file and retire issues correctly — this is the part no PR check can prove

jooyoungseo:
# Pull Request

## Description

The report step carried ~180 lines of JavaScript inside a YAML block scalar, running with `issues: write`. Testing it meant reading the block back out of the workflow and running it in a `vm` sandbox — and every defect found in that extraction produced a **passing** suite that was testing the wrong block, a mis-dedented one, or nothing at all.

## Related Issues

Closes #694. Raised independently by two reviewers on #692. Review here spun off #715.

## Changes Made

The body moves to `scripts/ci/e2eReport.cjs`; the step requires it. Deleted from the test: `SCRIPT_HEADER`, `reportScript()`, the multi-block guard, the indentation measurement, and the shared-`Error` sandbox. The 17 behavioural cases are unchanged.

| commit | |
|---|---|
| `a6650a6` | the extraction |
| `ea29138` | type the github-script bindings; drop a redundant `noEmit` |
| `0e36ade` | restore `ROOT`; document why `Bindings` is `unknown` |

### `.cjs`, not `.mjs`

`package.json` is `"type": "module"`, so a `.js` here would be ESM — and the sketch's `require('./scripts/ci/e2eReport.mjs')` cannot load an ES module reliably (`require(esm)` is a recent, version-dependent Node affordance, not a guarantee). `.cjs` is unambiguous.

It also keeps the test in Jest's **CommonJS** project. Making the module ESM would move `e2eReportWorkflow.test.ts` to the `esm` project — two projects disagreeing about one path, which is precisely the arrangement #703 is about.

### `tsc` coverage is not free

> The script also picks up `eslint` and `tsc` coverage for free, which is the larger win.

Half right, and the half that is wrong is the one called larger.

**eslint: yes**, and it earned its keep on the first run — it immediately flagged `node/prefer-global/process` in the moved code.

**tsc: no.** The root `tsconfig.json` sets `allowJs: false`, so tsc never loads a `.js`/`.cjs` file at all. Confirmed rather than assumed:

```
$ npx tsc --noEmit --listFiles | grep e2eReport
/home/user/maidr/test/scripts/e2eReportWorkflow.test.ts     ← the test, not the module
$ npx tsc --noEmit --listFiles | grep -c e2eReport.cjs
0
```

So this adds `tsconfig.ci.json` — `allowJs` + `checkJs` scoped to `scripts/ci/**` — and wires it into `type-check`. Turning those on repo-wide would pull in every other script and is a different change.

That found two rounds of real gaps: implicit `any` on the `isOwn` callbacks (`TS7006`), and then, once review pointed out that `@param {any}` on the bindings left the checker with nothing to do, the octokit/context/core surfaces. Both closed with JSDoc `@typedef`s — no `any` anywhere, per `rules/typescript.md`.

The types are load-bearing rather than decorative, which is worth showing rather than asserting:

```
# core.warning → core.warn
error TS2339: Property 'warn' does not exist on type 'Core'.        (×2)
# context.sha → context.shaa
error TS2551: Property 'shaa' does not exist on type 'Context'.     (×2)
```

Under `any`, both compiled clean.

## The seam this refactor opens, and closes

Importing the module directly is the point — but it means the cases pass **whether or not the workflow still calls it**. The old extraction was inherently coupled to the workflow; deleting the step broke the suite.

Verified the gap rather than reasoning about it — replacing the step's body with `core.info(...)`:

```
Tests: 1341 passed, 1341 total     ← all green against a gutted step
```

So two cases now assert the workflow still names the module. First attempt was too weak:

```ts
expect(workflow).toContain('scripts/ci/e2eReport.cjs');
```

That passed on the gutted step, because **the explanatory comment I had added above the step mentions the path**. Only 1 of 2 guards failed. Tightened to the whole require expression and the call:

```
Tests: 2 failed, 1341 passed, 1343 total     ← both guards now catch it
```

## Checklist

### The acceptance criterion this PR cannot meet

> A scheduled run (or `workflow_dispatch`) is observed to still file and retire issues correctly — this is the part no PR check can prove

Still true, and still open. The step runs only on `schedule` or `workflow_dispatch`, and neither fires from a pull request. The workflow does expose `workflow_dispatch`, so it can be triggered manually against this branch before or after merge — **worth doing, since a mistake here surfaces two days later as a report that never arrived.**

### On the exact-string guards

They are coupled to that line's formatting, which three reviews have now flagged. Deliberate, and currently inert: nothing in this repo formats that file.

```
$ npx eslint .github/workflows/e2e_tests.yml --format json
"message": "File ignored because of a matching ignore pattern."
```

It becomes live only if a YAML formatter is ever added, and the fix then is whitespace-tolerant matching rather than dropping the guards.

### On `Bindings` being `unknown`

Nothing type-checks that call against the shapes the module declares: the root config's `allowJs: false` means the project ts-jest checks against never loads the `.cjs`, and `tsconfig.ci.json` covers `scripts/ci/**` rather than `test/`. Restating those shapes in the test would buy a compile-time check at the cost of a second copy free to drift from the JSDoc it mirrors — silently, in the way this PR exists to stop. The gap is named in a comment instead.

### Found in review, deliberately not fixed here → #715

`results()` embeds raw reporter output inside a Markdown code fence without escaping, so output containing a triple-backtick line would break out of it. Pre-existing, and this PR is a line-for-line port — folding a behavioural fix in would make it unreviewable *as a port*.

### Verification

| check | result |
|---|---|
| `npm test` | 112 suites, 1400 tests (unit 107/1343 + esm 5/57) |
| `npm run type-check` | exit 0 — root project **and** `tsconfig.ci.json` |
| `npx eslint .` | exit 0, 0 warnings |
| `npm run build` | exit 0 |
| `actionlint` (shellcheck 0.10.0 on PATH) | exit 0 |
| CI | 14/14 green on `0e36ade` |

Counterfactuals run:

| reverted | result |
|---|---|
| the module's retire path short-circuited | 9 cases fail |
| the workflow's call to the module gutted | 2 cases fail (0 before the guards were added) |
| `core.warning` → `core.warn`; `context.sha` → `context.shaa` | 4 type errors (0 under `any`) |

---

---


jooyoungseo:
Both reviews addressed in `ea29138`.

## The `@param {any}` bindings — taken, and it was the right catch

`checkJs` was doing almost nothing on the three values this module actually calls into. `rules/typescript.md` asks for no `any` for the same reason.

Narrowed by hand rather than importing `@octokit/types` — a handful of call signatures is enough for the checker, and a file CI loads directly is a poor place to add a dependency. `listForRepo` stays `unknown`, since it is only ever handed to `paginate` as a route and never called here.

Worth showing the types do work rather than asserting it, since "now it's typed" is exactly the claim that can be hollow:

```
# core.warning → core.warn
scripts/ci/e2eReport.cjs(252,12): error TS2339: Property 'warn' does not exist on type 'Core'.
scripts/ci/e2eReport.cjs(280,12): error TS2339: Property 'warn' does not exist on type 'Core'.

# context.sha → context.shaa
scripts/ci/e2eReport.cjs(155,24): error TS2551: Property 'shaa' does not exist on type 'Context'. Did you mean 'sha'?
scripts/ci/e2eReport.cjs(271,30): error TS2551: Property 'shaa' does not exist on type 'Context'. Did you mean 'sha'?
```

Under `any`, both compiled clean.

The typedefs also surfaced 6 new `jsdoc/require-property-description` warnings on their own `@property` lines — the repo was at zero, so those are filled in rather than left.

## `noEmit` redundant in `tsconfig.ci.json` — correct

Removed. Checked that inheritance genuinely covers it rather than trusting `extends`: with the key gone, `tsc -p tsconfig.ci.json` still emits nothing.

## The exact-string YAML assertions

Flagged as a known fragility rather than something to fix, and I agree with leaving them — but the specific risk named is not live in this repo:

> a Prettier/eslint reformat of that one script line … will fail these tests

eslint does not format that file at all:

```
$ npx eslint .github/workflows/e2e_tests.yml --format json
"message": "File ignored because of a matching ignore pattern."
```

So no formatter here rewrites the block scalar, and the coupling costs nothing today. It would start to matter if a YAML formatter were ever added — at which point the assertions should become whitespace-tolerant rather than exact. Recording that here so the reasoning is on file rather than rediscovered.

## The inline `@param` on the `.find`/`.filter` callbacks

> worth confirming still type-checks clean once `tsc -p tsconfig.ci.json` actually runs in CI, since I couldn't independently execute it here

Confirmed locally, and CI runs it too — `type-check` is now `tsc --noEmit && tsc -p tsconfig.ci.json`, so the `Type Check` job covers it. Both exit 0.

## Verification

| check | result |
|---|---|
| `npm test` | 112 suites, 1400 tests |
| `npm run type-check` | exit 0 — root **and** `tsconfig.ci.json` |
| `npx eslint .` | exit 0, **0 warnings** |
| `npm run build` | exit 0 |
| `actionlint` (shellcheck 0.10.0 on PATH) | exit 0 |

Still open and unprovable from a PR, as the issue says: an actual `schedule` or `workflow_dispatch` run filing and retiring issues. `workflow_dispatch` is enabled, so it can be triggered manually.

---

jooyoungseo:
Both actionable notes taken in `0e36ade`.

**The duplicated `resolve(__dirname, '../..')`** — real, and mine: the seam guards introduced a second call after I had deleted the `ROOT` constant along with the extraction helpers. Restored; both call sites use it.

**`Bindings` typed `unknown` with no explanation** — the fair version of the criticism is that this file explains every other non-obvious choice and skipped this one. Now stated, and your diagnosis of the cause is right: the root `tsconfig.json` sets `allowJs: false`, so the project ts-jest checks against never loads the `.cjs`, and `tsconfig.ci.json` covers `scripts/ci/**` rather than `test/`.

Worth adding the part the comment now carries, since "unavoidable" undersells it slightly — it is avoidable, just not worth it. I could restate the module's shapes in the test and get a compile-time check on the stubs. That buys a second copy of those shapes, free to drift from the JSDoc it is meant to mirror, and drift there would be silent in exactly the way this PR exists to stop. Naming the gap is the better trade.

**The exact-string coupling** — third mention across two reviews, so worth pinning rather than re-acknowledging. The specific mechanism named (a reformat rewriting that line) cannot happen here today:

```
$ npx eslint .github/workflows/e2e_tests.yml --format json
"message": "File ignored because of a matching ignore pattern."
```

Nothing in this repo formats that file. The coupling is real but currently inert; it becomes live only if a YAML formatter is ever added, and the fix then is whitespace-tolerant matching rather than dropping the guards. Recorded here so the next person meets the reasoning rather than rediscovering it.

## Verification

| check | result |
|---|---|
| `npm test` | 112 suites, 1400 tests |
| `npm run type-check` | exit 0 — root **and** `tsconfig.ci.json` |
| `npx eslint .` | exit 0, 0 warnings |

Unchanged and still open, as the issue says: an actual `schedule`/`workflow_dispatch` run observed filing and retiring issues. No PR check can prove it; `workflow_dispatch` is enabled so it can be triggered by hand.

---

jooyoungseo:
The code-fence observation is a real find and new — filed as #715.

Checked how live it is rather than taking it as theoretical: `grep -rn '```' e2e_tests/` finds nothing today, so it is latent. But the capture is `npm run e2e 2>&1 | tee`, so it carries everything the suite and the app print, not just Playwright's own formatting — and this repo has a lot of Markdown in flight (`TypingEffect`, `markdownSanitize`, `markdownPipeline`). A fenced fixture reaching an e2e failure message is a plausible route rather than a contrived one.

Agreed it does not belong here, and for a sharper reason than "pre-existing": this PR is a line-for-line port, and folding a behavioural fix into it would make the port unreviewable *as a port* — the thing three reviewers have now checked it as. #715 has the CommonMark variable-length-fence fix and notes the ordering constraint (choose the fence after truncation, since the tail is what gets embedded).

The other two are already on file — the exact-string coupling with the evidence that nothing formats that YAML ([here](https://github.com/xability/maidr/pull/714#issuecomment-5169653031)), and the manual `workflow_dispatch` run, which remains the one acceptance criterion no PR check can satisfy.

---

PR #716, opened 2026-08-03
perf(build): stop publishing sourcemaps to npm

jooyoungseo:
## Problem

`maidr@3.75.1` is **102.5 MB unpacked / 23.9 MB to download**, and **77% of it is sourcemaps** — 32 `.map` files, the five largest 8.3–8.7 MB each:

```
vegalite.mjs.map  8.7 MB
amcharts.mjs.map  8.5 MB
chartjs.mjs.map   8.5 MB
vegalite.js.map   8.4 MB
maidr.js.map      8.3 MB
```

Inside each one, **~70% is `sourcesContent`** — the full original source of React, MUI, KaTeX and every other dependency, embedded again in each of the twelve bundles. `maidr.js.map` alone carries 767 sources.

For comparison, measured against the published packages:

| library | `.map` files | total files | unpacked |
|---|---|---|---|
| d3 | 0 | 6 | 0.8 MB |
| katex | 0 | 213 | 3.8 MB |
| victory | 0 | 11 | — |
| **plotly.js** | **0** | 1,115 | 93.6 MB |
| recharts | 1 | 839 | 7.1 MB |
| chart.js | 8 | 117 | — |
| **maidr** | **32** | **80** | **102.5 MB** |

plotly.js is a far larger library and publishes no sourcemaps at all; maidr is bigger than it, and three quarters of that is debug metadata.

## Change

- `package.json` — `"files": ["dist", "!dist/**/*.map"]`
- `scripts/build.js` and `vite.config.ts` — `sourcemap: true` → `sourcemap: 'hidden'`

`'hidden'` still writes every `.map` beside its bundle, so debugging this repo locally is unchanged; it only omits the `sourceMappingURL` comment. That part is required, not cosmetic: dropping the files while leaving the comment would point **every** CDN consumer at a 404 — on jsDelivr, which serves the maps today, as well as on cdnjs, whose manifest deliberately does not mirror an 8.7 MB map.

```
unpacked   102.5 MB -> 23.8 MB   (-77%)
download    23.9 MB ->  6.2 MB   (-74%)
files             80 -> 48
```

## About the eslint exception

`jsonc/sort-array-values` wants the `files` array sorted ascending, which puts `!dist/**/*.map` before `dist`. **npm's `files` is order-sensitive** — a negation only excludes what an earlier pattern included — so sorted, `dist` re-includes every map. Verified with `npm pack --dry-run`: reordered, it goes straight back to **80 files with all 32 maps**, package unchanged at 102.5 MB.

So the rule is turned off for `package.json`, with the reasoning in `eslint.config.ts` because package.json takes no comments. Upstream scopes this rule to `pathPattern: '^files$'`, so nothing else is affected — `keywords` and every other array keep their sorting.

I also tried `.npmignore`, which does not work here: `files` takes precedence and the maps came back.

## Verification

- `npm pack --dry-run` — **0** `.map` files, 48 entries, 23.8 MB unpacked / 6.2 MB packed.
- Full `node scripts/build.js` — all 12 bundles, **32** maps still emitted into `dist`, and `grep -l sourceMappingURL dist/*.js dist/*.mjs` returns **0** files.
- `npm run type-check` — clean.
- `CI=true npx eslint` on all four changed files — clean.
- `npx jest` — **1352 tests passing, 0 test failures**, identical to `main`. Seven suites fail to *run* in my tree (three ESM suites on a parse error, four on "Converting circular structure to JSON"); I re-ran the full suite on unmodified `main` and got the exact same 7 failed / 106 passed / 1352 passing, so they are pre-existing and unrelated to this change.

### DIFF
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -134,9 +134,11 @@ const config: ReturnType<typeof antfu> = antfu({
 }, {
   // npm's `files` is order-sensitive: a negation only excludes what an earlier
   // pattern already included. Sorted ascending, `!dist/**/*.map` lands before
-  // `dist`, which then re-includes everything it just excluded — `npm pack
-  // --dry-run` goes from 36 files back to 80, and the 32 sourcemaps that are
-  // 77% of the published package return.
+  // `dist`, which then re-includes everything it just excluded, and every
+  // sourcemap is published again — verified with `npm pack --dry-run`, which
+  // is also the way to check this if the array is ever touched.
+  // `test/scripts/publishedPackage.test.ts` pins the ordering so the fix
+  // cannot be undone by an autofix.
   //
   // package.json takes no comments, so the warning has to live here. Upstream
   // scopes this rule to `pathPattern: '^files$'`, so turning it off costs
--- a/eslint.config.ts
+++ b/eslint.config.ts
@@ -131,6 +131,20 @@ const config: ReturnType<typeof antfu> = antfu({
       message: 'Do not build aria-label from children: it stringifies to "[object Object]" unless the child is a plain string, and aria-label replaces the accessible name the element already had.',
     }],
   },
+}, {
+  // npm's `files` is order-sensitive: a negation only excludes what an earlier
+  // pattern already included. Sorted ascending, `!dist/**/*.map` lands before
+  // `dist`, which then re-includes everything it just excluded — `npm pack
+  // --dry-run` goes from 36 files back to 80, and the 32 sourcemaps that are
+  // 77% of the published package return.
+  //
+  // package.json takes no comments, so the warning has to live here. Upstream
+  // scopes this rule to `pathPattern: '^files$'`, so turning it off costs
+  // nothing else — `keywords` and every other array stay unaffected.
+  files: ['package.json'],
+  rules: {
+    'jsonc/sort-array-values': 'off',
+  },
 });
 
 export default config;
--- a/package.json
+++ b/package.json
@@ -79,7 +79,8 @@
   },
   "main": "dist/maidr.js",
   "files": [
-    "dist"
+    "dist",
+    "!dist/**/*.map"
   ],
   "scripts": {
     "build": "node scripts/build.js",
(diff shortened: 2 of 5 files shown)


jooyoungseo:
Thanks — three points raised across the two reviews. One is already fixed, one I've now verified properly, one I agree with.

### The stray files are already gone

That review looked at `8cc72b55`. A `git add -A` had swept in three `.claude/worktrees/agent-*` gitlinks, two `.playwright-mcp/` snapshots and a screenshot; I caught it, dropped them with `git rm --cached`, amended and force-pushed to `a4873417`. The branch diff is five files:

```
eslint.config.ts
package.json
scripts/build.js
test/scripts/publishedPackage.test.ts
vite.config.ts
```

The suggestion to add `.playwright-mcp/` to `.gitignore` so it stops recurring is a good one and is now #717, along with `.claude/worktrees/` — worth ignoring for the same reason, and the more dangerous of the two since each worktree carries its own `.git` and stages as an embedded repository rather than being skipped.

### `pathPattern: '^files$'` — verified, since both reviews flagged it as unverifiable

Fair challenge; "trust the description" isn't good enough for a claim that this override might silently stop `keywords` being sort-checked. Run on **`main`**, without the override, so the rule is fully active:

**Control** — scramble `files`:

```
$ CI=true npx eslint package.json
  82:5  error  Expected array values to be in ascending order. 'zzz' should be after 'aaa'  jsonc/sort-array-values
✖ 1 problem (1 error, 0 warnings)
```

**Experiment** — scramble `keywords` the same way:

```
$ CI=true npx eslint package.json
(no output, exit 0)
```

The control proves the rule is live and eslint is really running; the experiment shows `keywords` was never in its scope. So `keywords` does not lose enforcement here — it never had any from this rule. `eslint.calculateConfigForFile('package.json')` reports the options directly:

```json
[2, { "order": { "type": "asc" }, "pathPattern": "^files$" }]
```

### unpkg

Agreed, and worth stating explicitly: unpkg serves the npm tarball too, so once the maps are not in the package there is nothing for it to serve and no dangling comment pointing at one. The three CDN paths differ in what they mirror, not in what the tarball contains, so the reasoning holds for all of them.

PR #723, opened 2026-08-04
feat(step): add a step plot trace type for piecewise-constant data

jooyoungseo:
# Pull Request

## Description

Adds `TraceType.STEP`. A step chart *holds* its value across an interval and then jumps, rather than interpolating between samples the way a line chart implies.

The motivating case is a **hypnogram** — sleep stage against time. Two things distinguish it from a line chart, and both are accessibility problems today:

1. The y axis is an **ordinal category** (Awake / REM / N1 / N2 / N3), not a magnitude. Announcing "3" is useless; it has to say "REM".
2. The interesting events are the **transitions**. Arrowing point-by-point through a whole night of scoring epochs is unusable; a user needs to jump from stage change to stage change.

`StepTrace extends LineTrace`, so everything numeric is inherited unchanged. Notably the audio is *deliberately* left alone: `LineTrace` already emits one discrete tone per point, which is the correct sonification for piecewise-constant data — a run of equal levels sounds as a repeated pitch, and a transition is an audible jump. `SmoothTrace`'s continuous glissando would interpolate between levels, which is exactly what a step chart does not do.

## Related Issues

Companion to the step-plot work in xability/py-maidr#299 and the matching r-maidr change. This is the prerequisite: `TraceFactory.create` throws `Invalid trace type: step` until this lands, so the bindings should merge after it.

## Changes Made

**Grammar** — `TraceType.STEP`, `StepPoint extends LinePoint { label?: string }`, and a layer-level `stepDirection?: 'hv' | 'vh' | 'mid'`. `y` stays strictly numeric because it drives sonification, braille and the min/max range; `label` is the name of the level that `y` encodes. `stepDirection` is omitted rather than defaulted when the producing library does not report one, so the description never names a convention the data did not author.

**`StepTrace`** overrides four things and inherits the rest:
- `state` → `plotType: 'step'`, so every spoken surface says "step plot".
- `text` → substitutes the level name for the number.
- `description` → reports the *runs*: transition count, longest run, level names, direction.
- `getRotorFilterUnits()` / `moveToRotorFilter()` → a **"Transitions" rotor unit**. This is the headline navigation feature and it cost nothing outside `step.ts`: `AbstractTrace` already exposes trace-specific rotor units (`Candlestick` uses them for bull/bear filtering), so the rotor service needed no change at all.

**A highlight-placement fix.** `LineTrace.reconcilePathCoordinates` is extracted from `mapViaPathParsing` as a protected seam — a pure extraction, no behaviour change — so `StepTrace` can override it. This matters: a step path is rendered with the corner vertices the steps introduce, `2N-1` for `hv`/`vh` and `2N` for `mid`. The inherited handling trims the surplus from the *end*, but a step path's surplus is *interleaved*, so every highlight would have landed on the first half of the chart. The override removes it by stride instead. I verified the even-index rule against both matplotlib's `STEP_LOOKUP_MAP` and ggplot2's `stairstep()`, and the r-maidr work confirmed it live: a 10-point series renders 19 / 19 / 20 vertices at `hv` / `vh` / `mid`.

**AnyChart.** `step-line` and `step-area` were already being downgraded to line traces (documented as such). They now map to `TraceType.STEP`. `AnyChartTraceType` is a deliberately narrowed union that forces `buildLayer`'s switch to be exhaustive, so this could not be partial. The area-fill warning now keys on the source type rather than the mapped one, so `step-area` still warns that its fill is lost.

**Registration.** Adding a `TraceType` member produces exactly **one** compile error in this repo (`CHART_TYPE_LABEL`); every other site fails silently or at runtime, so the rest were walked by hand: the factory case, the braille encoder `Map` (a missing entry announces "Braille is on" and then shows nothing), `NESTED_DATA_TYPES` in `liveData.ts`, and the `'line'` branches in `highContrast.ts`.

## Checklist

## Additional Notes

**Verification actually run:** `npm run lint:fix`, `npm run type-check` (both `tsconfig.json` and `tsconfig.ci.json`, so the e2e sources are covered), `npx jest --selectProjects unit` — **1396 passed, 112 suites** — and `npm run build` (all 12 targets). The Playwright spec runs **18/18 in chromium**; browsers were not present in the sandbox, so they were installed first and this is a real run rather than a skipped one.

**One bug worth calling out, because it shipped broken through the unit suite.** `reconcilePathCoordinates` runs during `super(layer)` — `LineTrace`'s constructor calls `mapToSvgElements`, which reaches the override — so `StepTrace`'s own field was not assigned yet. It threw, MAIDR never mounted, and 15 of the 18 e2e assertions died on an uncaught `TypeError` before reaching any step behaviour. The model tests could not see it: they pass no selectors, so `mapViaPathParsing` returns before the override is reached. Fixed in 49423bb, with a jsdom test that constructs a trace against a real staircase and asserts the highlights land on the samples rather than the step corners — it fails on both counts if the field reference is put back.

**Known limitation, stated in `docs/BRAILLE.md`.** Braille reuses `LineBrailleEncoder`, which buckets each row's range into four levels. A hypnogram with five or more stages therefore collapses adjacent stages onto one glyph. Reuse is the established pattern (`SMOOTH` and `VIOLIN_KDE` both map to it) and a finer encoding would mean extending the 8-dot glyph vocabulary in `getBrailleChar`, which is a separate change. Braille conveys the shape of the night; the exact stage name comes from the text announcement.

**Deliberately out of scope.** Other adapters can detect step charts too — Plotly's `line.shape`, Chart.js's `dataset.stepped`, Vega-Lite's `mark.interpolate`, Recharts' `type="stepAfter"`, amCharts' `StepLineSeries`, Highcharts' `plotOptions.series.step`. Each is new detection logic with its own selector/highlight path, not a one-line mapping, and I could not exercise those libraries here. AnyChart is included only because it was already producing wrong output and its narrowed union forced the exhaustive case.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DZocjnGTiRvLDCbcvqYqa3)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ and encourages a multi-modal exploration on visualization.
 
 To use maidr, follow these steps:
 
-1. **Import your plot**: maidr is designed to work seamlessly with scalable vector graphics (SVG) objects for visual highlighting. However, maidr is inherently visual-agnostic, and it also supports other raster image formats such as PNG and JPG without the visual highlight feature. Regardless of the image format, maidr provides support for all non-visual modalities, including Braille, text, and sonification (BTS). Additionally, it offers interactive and artificial intelligence (AI) plot descriptions powered by OpenAI GPT, Anthropic Claude, Google Gemini, or local models running on your own machine via [Ollama](https://ollama.com) (no API key required, suitable for sensitive data). The supported plot types include bar plot, boxplot, heatmap, scatter plot, line plot, histogram, segmented bar plots (e.g., stacked bar plot, side-by-side dodged plot, and normalized stacked bar plot).
+1. **Import your plot**: maidr is designed to work seamlessly with scalable vector graphics (SVG) objects for visual highlighting. However, maidr is inherently visual-agnostic, and it also supports other raster image formats such as PNG and JPG without the visual highlight feature. Regardless of the image format, maidr provides support for all non-visual modalities, including Braille, text, and sonification (BTS). Additionally, it offers interactive and artificial intelligence (AI) plot descriptions powered by OpenAI GPT, Anthropic Claude, Google Gemini, or local models running on your own machine via [Ollama](https://ollama.com) (no API key required, suitable for sensitive data). The supported plot types include bar plot, boxplot, heatmap, scatter plot, line plot, step plot, histogram, segmented bar plots (e.g., stacked bar plot, side-by-side dodged plot, and normalized stacked bar plot).
 
 2. **Create an HTML file**: Include the main script file `maidr.js`. No stylesheet link is needed — maidr styles its own interface at runtime, and fetches the stylesheet for mathematical notation on demand when it is actually required. Add the SVG of your plot to the main HTML body, and add an ID attribute of your choice to the SVG. Note that this can be automated with R. Your HTML file should now have the following structure:
 
@@ -57,7 +57,7 @@ To use maidr, follow these steps:
 
 ## Data Schema
 
-The maidr JSON schema defines how plot data is structured for each supported plot type, including bar plots, boxplots, heatmaps, scatter plots, line plots, histograms, and segmented bar plots.
+The maidr JSON schema defines how plot data is structured for each supported plot type, including bar plots, boxplots, heatmaps, scatter plots, line plots, step plots, histograms, and segmented bar plots.
 For the full schema structure, object properties, and data formats, see the [Data Schema documentation](docs/SCHEMA.md).
 
 ## React Integration
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -178,6 +178,42 @@ In multiline braille displays, all data series are represented simultaneously. H
 
 In single-line braille displays, the user can navigate vertically with the up and down arrow keys to move between lines, and the braille representation updates to show the values for the current line.
 
+## Step plot
+
+A step plot's braille is the line plot encoding above, unchanged: each data
+point becomes one Braille character whose dot height is its numeric `y` relative
+to the row's own range. The staircase corners the chart draws are not encoded —
+braille represents the samples, not the rendered geometry — so a run of equal
+levels reads as a run of identical characters and a transition reads as a change
+in dot height.
+
+### Known limitation: ordinal levels collapse
+
+The line encoder quarters each row's range into **four** buckets:
+
+- ⣀ represents values from 0% to 25%
+- ⠤ represents values from 25% to 50%
+- ⠒ represents values from 50% to 75%
+- ⠉ represents values from 75% to 100%
+
+A step plot whose Y axis is an ordinal scale with **five or more levels**
+therefore has more levels than the encoding has heights, and adjacent levels
+land on the same character. A hypnogram coded N3=1, N2=2, N1=3, REM=4, Awake=5
+quarters into bands of one unit, so N3 and N2 both render as ⣀ and cannot be
+told apart by touch alone.
+
+The exact stage name is still available: it comes from the text announcement,
+which reads the point's `label` ("Sleep stage is N2") rather than its numeric
+code. Use braille for the shape of the night — how long each run is, where the
+transitions fall — and the text output to name the level you are standing on.
+
+### Multiline Displays
+
+Step plots follow the line plot rules. In multiline braille displays, each
+series occupies its own line, so several step series can be compared at once. In
+single-line displays, the up and down arrow keys move between series and the
+braille representation updates to the current one.
+
 ## Multiline Braille Display Support
 
 By leveraging the two-dimensional nature of multiline braille displays, MAIDR can represent multiple lines of a plot simultaneously, allowing users to perceive the distribution of values across all lines at once. This is particularly beneficial for plots with multiple groups or categories, such as grouped boxplots or line plots with multiple lines, and it also applies to scatter plot grid navigation.
--- a/docs/LIVE_DATA.md
+++ b/docs/LIVE_DATA.md
@@ -70,7 +70,7 @@ window.maidrLive.appendData(
 
 The shape of `point` matches the layer's data format (see the [Data Schema](SCHEMA.html)): `{ x, y }` for bar/line/scatter points, a full OHLC object for candlestick, and so on.
 
-**Supported layer types:** any layer whose `data` is an array — bar, line (and multiline), scatter, histogram, candlestick, box, smooth, and segmented bar charts. Heatmaps (object-shaped data) do not support appending; use `setData` instead. Violin KDE layers accept appends structurally, but KDE points are pre-computed density samples — appending raw observations does not recompute the distribution, so prefer `setData` with freshly computed densities for violins.
+**Supported layer types:** any layer whose `data` is an array — bar, line (and multiline), step, scatter, histogram, candlestick, box, smooth, and segmented bar charts. Heatmaps (object-shaped data) do not support appending; use `setData` instead. Violin KDE layers accept appends structurally, but KDE points are pre-computed density samples — appending raw observations does not recompute the distribution, so prefer `setData` with freshly computed densities for violins.
 
 ### Streaming candlesticks (stock ticker)
 
--- a/docs/SCHEMA.md
+++ b/docs/SCHEMA.md
@@ -345,7 +345,7 @@ The data property is defined as a list of objects where each object is a record 
                       "x": 1.0,
                       "y": 2,
                       "label": "N2"
-          },
+          }
         ]
         //add multiple arrays for multiple step series
       ]
--- a/docs/SCHEMA.md
+++ b/docs/SCHEMA.md
@@ -121,7 +121,7 @@ Or multiple plots:
 
 Use the following to define the object properties:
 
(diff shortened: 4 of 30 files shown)


jooyoungseo:
Both multi-series findings were real — thanks, those are exactly the kind that hide behind a single-series motivating case. Fixed in 0e79226 along with the doc corrections from the other review.

**1. The "with N groups" cue.** Confirmed: `Context.getInstruction` gated the suffix on `plotType === 'multiline'`, and a step trace calls itself `'step'` whatever its series count, so it never fired.

I gated on the group count itself rather than teaching the check a second plot-type name:

```ts
const groupCountText
  = state.groupCount && state.groupCount > 1
    ? ` with ${state.groupCount} groups`
    : '';
```

That says how many groups there are regardless of what the trace calls itself, which generalises to the next trace type that reports groups under its own name. Line charts are untouched: `LineTrace` only sets `groupCount` when there is more than one series, so the old `'multiline' && groupCount` and the new `groupCount > 1` agree on every input it produces. The `'multiline'` + `groupCount === 1` downgrade above it is left alone.

I also fixed the **pre-activation** mirror in `maidr-component.tsx`, which you didn't flag but has the same shape — it builds its instruction from raw JSON rather than from the model, so without it the announcement would have changed when the user clicked.

**2. `describe.ts:680`.** Confirmed and fixed — `TraceType.STEP` now takes the `announceMultiLinePosition` branch alongside `LINE`. A step trace navigates identically (series on one axis, samples on the other), so the generic 2-D announcement was the wrong one. You're right that this was the same walk-every-`TraceType.LINE`-site work as the braille map and `highContrast.ts`, and I missed it.

Three tests added under `multi-series step trace`: the group count reaches the state, the transitions unit is offered when only one of several series ever changes level, and a transition jump stays inside the series the cursor is on.

**3. `mid` with non-uniform spacing.** Your algebra is right — the interior average is `(x[i-1] + 2x[i] + x[i+1]) / 4`, which is `x[i]` only under even spacing.

I documented the real guarantee rather than chasing exactness, and added a test that pins it: the highlight always lands **inside the sample's own horizontal run**, never on a neighbour's, so it still reads as the right sample. The exact recovery is `x[i+1] = 2·m[i] - x[i]` iterated from `x[0]`, which is closed-form but numerically unstable — the error doubles and alternates sign each step, so over a night of epochs it would be far worse than a bounded sub-run offset. The two half-width end runs are read off directly and are exact.

`npm run lint:fix`, `npm run type-check` and `npx jest --selectProjects unit` all clean — **1406 passed, 112 suites**.

On merge order, since it came up on the sibling PRs: this one has to land and release first. The Python (xability/py-maidr#299) and R (xability/r-maidr#49) bindings both emit `type: "step"`, and `TraceFactory.create`'s `default` throws on an unknown type, so they are hard failures against any bundle without this.

---

jooyoungseo:
Confirmed and fixed in 9e1a5c1 — this was a real regression I introduced, and your trace through it was exactly right.

I verified each link before changing anything: `stampLineAttributes` does stamp step series (`LINE_LIKE_SERIES_TYPES` contains `step-line` and `step-area`), `resolveSelector`'s `if`-chain had no `STEP` branch and fell through to `return undefined`, and `LineTrace.mapToSvgElements` bails to `null` on a missing selector. So the DOM was stamped and ready with nothing pointing at it.

Worth being precise about the severity: **this worked before my change.** `step-line` used to map to `TraceType.LINE` and got the LINE selector. Re-pointing it at `TraceType.STEP` took it out of that branch. So an AnyChart step chart went from having highlighting to having none — a regression, not a pre-existing gap, and one that says nothing while it happens.

Fixed as you suggested, plus the structural half:

```ts
switch (traceType) {
  case TraceType.BAR: …
  case TraceType.LINE:
  case TraceType.STEP:
    return `${scope}[${LINE_ATTR}^="${stamp}${seriesIndex}-"]`;
  …
  case TraceType.BOX:
  case TraceType.CANDLESTICK:
    return undefined;   // both build their own selectors in their layer builder
}
```

Narrowing the parameter from `TraceType` to `AnyChartTraceType` is what makes the switch exhaustiveness-checked — the call site already narrows to that type after its `if (!traceType) continue` guard, so it cost nothing. The next member added to that union is now a compile error here, same as in `buildLayer`.

**On the test gap** — you're right that asserting on `mapSeriesType` alone is what let this through, and I've added the end-to-end tests. I checked they actually catch it by reverting the fix and re-running:

```
✕ gives a step series the same stamped selector a line series gets
    expect(received).toBeDefined()
    Received: undefined
✕ gives a step-area series a selector too
```

They go through `anyChartToMaidr` and assert a step layer's `selectors` is defined *and equal to the line layer's*, so the two can't drift apart again.

**`StepTrace.description` recomputation** — leaving as is. `transitionIndices` is precomputed because `getRotorFilterUnits()` and `moveToRotorFilter()` read it on the navigation path; `description` is read only when the user opens the modal, so precomputing it would trade constructor work on every chart for savings on an occasional action. Happy to change it if you'd rather have the consistency.

`npm run lint:fix`, `npm run type-check` and `npx jest --selectProjects unit` clean — **1408 passed, 112 suites**. Prior commit was green across all 14 CI checks including E2E (chromium); this one is re-running now.

---

jooyoungseo:
Correction to my previous comment: the AnyChart selector fix is **`43799f2`**, not `9e1a5c1` — I quoted a SHA that doesn't exist in this branch. Nothing else in that comment changes.

---

jooyoungseo:
Both confirmed and fixed in `6e05d83`. Verified each against the code first — the second one in particular, since a wording claim is easy to assert and hard to be sure about.

**1. The intersection chord.** You're right, and the sharpest part of your framing is that it contradicts this branch's own claim. I wrote "audio is inherited unchanged"; `audio.ts:249` was the one place where it wasn't.

The reasoning holds up: `LineTrace` computes `intersections` from the *data* — two series sharing an exact `(x, y)` sample — not from path geometry, so a step trace's state carries them identically. `StepTrace.state` spreads `...baseState`, so they arrive. Only the bare-string gate stopped them. And `supportsIntersectionMode()` being a virtual method is exactly why the *navigation* half picked up STEP for free while the audio half didn't — a good illustration of why the string-literal checks are the fragile ones.

I kept an explicit allowlist rather than dropping the check entirely as you suggested. You're correct that `intersections.length > 1` already implies multiple series share the point, and `LineTrace` is the only producer today — but an allowlist says *which* traces this is meant for, and the next trace to set `intersections` should opt in deliberately rather than inherit a chord by accident. What I did take from the suggestion is comparing against `TraceType.LINE` / `TraceType.STEP` instead of `'line'`, so a renamed enum member breaks the build. There is already dead code in this repo from exactly that pattern — `highContrast.ts` tests for `'histogram'` when the enum value is `'hist'`.

**2. "Line 1 of 2".** Confirmed at `describe.ts:880`. The noun is now a parameter defaulting to `Line`, and the STEP branch passes `Series`:

> `Series 1 of 3, Position is 3 of 10`

I extended `test/command/announce-position.test.ts` rather than adding a file, since the harness was already there — four cases: the verbose and terse step wordings, that an authored group name still wins over the ordinal in terse mode, and a guard that a multiline chart still says `Line 1 of 3`. That last one is the regression I'd otherwise have risked introducing while fixing this.

You're also right about *why* both survived: the example and the e2e spec are both single-series, so neither of these paths is reachable from them. The multi-series model tests I added earlier checked `groupCount` and `plotType` but stopped short of the announcement strings, which is the gap you identified.

`npm run lint:fix`, `npm run type-check`, `npx jest --selectProjects unit` — **1412 passed, 112 suites**. The preceding commit was green on all 14 CI checks including E2E (chromium); this one is re-running.

---

jooyoungseo:
Thanks — and on the one open question, I traced it rather than leaving it as a gut-check. **The trace-level gating is deliberate, and narrowing it to the active row would introduce a real bug.**

`RotorNavigationService` resolves the active unit *by label*, on every keypress:

```ts
private getActiveFilterUnit(mode: string): RotorFilterUnit | null {
  return activeTrace.getRotorFilterUnits().find(unit => unit.label === mode) ?? null;
}
```

and the caller falls through when that returns null (`rotor.ts:247-256`):

```ts
const filterUnit = this.getActiveFilterUnit(mode);
if (filterUnit)
  return this.moveFilter(filterUnit, 'left');
// …otherwise falls through to compare-mode movement
const moved = activeTrace.moveLeftRotor(this.getCompareType());
```

So if `getRotorFilterUnits()` were row-dependent, a user who selected Transitions on series 2 and then moved to a flat series 1 would find the unit no longer resolves — and their left/right would silently perform **compare-mode** navigation instead, while the rotor still says "Transitions". A mode that quietly does something else is worse than a mode that reports a boundary.

On whether the inert case reads as broken: it isn't silent. `moveToRotorFilter` calls `notifyRotorBounds()` when no transition matches, and `moveFilter` maps that to an announced boundary message — so the user hears a bounds announcement rather than nothing. Same contract as intersection mode, and the same trace-level granularity `Candlestick` uses for its trend units.

No change made, since the behaviour the alternative would produce is worse. I've noted the reasoning here rather than in the code because it's about the *service's* resolution contract, not something visible at the `StepTrace` call site — happy to add it as a comment on `getRotorFilterUnits()` if you'd rather it live next to the code.

Final state: **13/14 CI green on `6e05d83`, including E2E (chromium)**; only `claude-review` still running. 1412 unit tests, 112 suites.

---

PR #724, opened 2026-08-04
fix(plotly): read normalized bar values from calcdata

### LINKED ISSUE #720: fix(plotly): normalized bar charts announce raw values, not the normalized ones drawn
A Plotly bar chart with `barnorm: 'percent'` is drawn as percentages but announced with the raw input numbers. MAIDR classifies it correctly as `stacked_normalized_bar` and then reads out values that contradict that classification and the chart on screen.

## Reproduction

```html
<script src="https://cdnjs.cloudflare.com/ajax/libs/plotly.js/3.1.1/plotly.min.js"></script>
<script src="https://cdn.jsdelivr.net/npm/maidr@3.75.1/dist/maidr.js"></script>
<div id="chart"></div>
<script>
  Plotly.newPlot('chart', [
    { type: 'bar', name: 'East', x: ['Q1', 'Q2'], y: [120, 185] },
    { type: 'bar', name: 'West', x: ['Q1', 'Q2'], y: [90, 140] },
  ], {
    barmode: 'relative',
    barnorm: 'percent',
    xaxis: { title: { text: 'Quarter' } },
    yaxis: { title: { text: 'Share' } },
  });
</script>
```

Enter, then Right:

```
Quarter is Q1, Share is 120, Level is East
```

Expected roughly `Share is 57.1` — East is 120 of 210 at Q1.

## What Plotly actually drew

Read off the same chart:

```js
gd._fullLayout.barnorm        // 'percent'
gd._fullLayout.yaxis.range    // [0, 105.26]
gd.calcdata[0][0].y           // 57.14285714285714
gd.calcdata[0][0].s           // 57.14285714285714
```

The bar is at 57.14 on a 0–105 axis. Nothing on screen shows 120. A sighted reader sees a share; a screen reader user is told the unnormalized total.

## Control

The same shape without `barnorm` is correct, which places the fault precisely:

```
barmode: 'stack', no barnorm
  calcdata[0][0].y === 120, raw === 120
  announced: "Quarter is Q1, Revenue is 120, Level is East"   ✓
```

Raw and computed agree for a plain stack, so the existing behaviour is right there. They diverge only under `barnorm`, and that is the one case that is wrong.

## Cause

[`extractSegmentedBarLayer`](https://github.com/xability/maidr/blob/main/src/adapters/plotly/extractor.ts) reads the raw trace arrays:

```ts
for (const { trace } of barTraces) {
  const x = trace.x;
  const y = trace.y;
  ...
```

Two things make this look like an oversight rather than a decision:

- the function already receives `gd: PlotlyGraphDiv`, so `gd.calcdata` — which holds the normalized values — is in scope and unused
- the caller has *already* consulted `layout.barnorm` to pick the type:

```ts
const type = barnorm === 'percent' || barnorm === 'fraction'
  ? TraceType.NORMALIZED
  : TraceType.STACKED;
```

So the code knows the chart is normalized when choosing how to describe it, and then describes it with numbers that are not.

## Suggested direction

Take the value from `calcdata` for these traces rather than from `trace.y`. `calcdata[i][j].y` (or `.s`) is already what Plotly drew, which keeps this consistent with the rest of the adapter's approach of reporting the rendered figure rather than recomputing it. Worth checking `barnorm: 'fraction'` in the same pass, and horizontal (`orientation: 'h'`) bars, where the value lives on `x`.

## Impact

Quiet and directional, like #718: the chart looks right, nothing is logged, and only the non-visual output is wrong. `barnorm` exists specifically to say "compare shares, not totals", so a reader is being given the number the author deliberately chose not to present.

Found while testing every MAIDR trace type against Plotly auto-detection, ahead of building a claude.ai skill on top of it. 11 of 13 types came through correctly; this and #718's sibling (violin, filed separately) were the exceptions.

jooyoungseo:
# Pull Request

## Description

A Plotly bar chart with `barnorm: 'percent'` (or `'fraction'`) is drawn as shares, but MAIDR announced the raw input numbers. The chart was classified correctly as `stacked_normalized_bar` and then read out values contradicting that classification and the bars on screen — `Share is 120` where the bar sits at 57.14 on a 0–105 axis.

The bar extractors read the raw `trace.x`/`trace.y` arrays. They already received the graph div, and the caller had already consulted `layout.barnorm` to pick the trace type, so the normalized values plotly computed were in scope and unused.

## Related Issues

Fixes #720

Follow-ups filed from what this work surfaced, both deliberately out of scope here: #725 (computed values announced at full float precision) and #726 (a bar gap sonified as a zero-height bar).

## Changes Made

`src/adapters/plotly/extractor.ts` — both bar paths now take each bar's value from its calcdata entry, which is what plotly actually drew, instead of the raw trace array. The position still comes from the trace array, so category labels are unaffected.

- **Segmented bars** (dodged / stacked / normalized): `extractSegmentedBarLayer` now receives the `SubplotGroup` to reach its per-trace calcdata, matching how `extractMultiBoxLayer` already does it.
- **A lone bar trace** takes the single-layer path instead, so it was missed by the above. Plotly normalizes a single trace against itself — every bar is drawn at 100% — while the raw numbers described three different heights that are not on screen. `extractLayer` already had the calcdata in hand for the histogram case, so the value is threaded one function further into `extractBarLayer`.
- Both paths share one `barPoint` helper. It reads `cd.s`, the bar's own size after normalization, and places it on the axis the orientation calls for. Deciding both together is deliberate: the value's axis and the raw value backing it have to agree, and as two separate ternaries per call site they could drift. `cd.s` is orientation-independent (plotly keeps the position on `cd.p` for both `'v'` and `'h'`), so horizontal bars need no special case. `cd.x`/`cd.y` are deliberately not used: for a stack they hold the running top, cumulative for every trace above the first — reading those would have swapped one wrong number for another.
- The raw trace value remains the fallback when calcdata is unavailable or holds a non-finite size, so charts captured before plotly computes it behave as before.

Grouped bars (`barmode: 'group'`) go through the same helper, so a grouped chart with `barnorm` set reports shares too. Trace-type selection is untouched.

Not applicable — the rendered chart does not change; only the announced values do. The announcements themselves are quoted under Additional Notes.

## Checklist

## Additional Notes

**Verified against real plotly, not just fixtures.** Every claim here was checked against plotly.js 2.35.2 — the version `examples/plotly-*.html` pin — driven through Playwright, and the fixtures mirror what it actually computes:

- `cd.s` carries the normalized segment size while `cd.y` carries the running stack top (100 for the second trace) — which is why the fix reads `s`.
- `barmode: 'group'` with `barnorm` is scaled too, on a `[0, 60.15]` axis.
- A lone trace under `barnorm` gets `s: 100` on every bar.
- Bar `calc()` holds a missing point in place with `s: undefined` (BADNUM) rather than dropping it, so `cd[i]` stays in step with `x[i]`/`y[i]` — the positional lookup the fix relies on.

**End-to-end through the built bundle**, driving the chart by keyboard and reading the `role="alert"` region:

```
stacked + barnorm    Quarter is Q1, Share is 57.14285714285714, Level is East
lone trace + barnorm Quarter is Q1, Share is 100
```

On `main` those read `Share is 120` and `Share is 120` respectively. Both were re-run unchanged after the helper refactor.

**Ten unit tests.** Six fail on `main` — percent, fraction, grouped, horizontal, the lone trace, and the gap-alignment pin. Four are controls that pass either way: an unnormalized stack, an unnormalized lone trace, a chart with no calcdata, and a bar normalized to 0%. The unnormalized-stack case asserts the second series reads 90 rather than the stack top of 210, which is what distinguishes `cd.s` from `cd.y`.

One limit stated rather than papered over: the zero case does **not** distinguish a falsy-check fallback from a nullish one. I checked by swapping `??` for `||` in the source — all tests still passed, because under `barnorm` a drawn 0 implies a raw 0. The test comment says so.

**Bar gaps are deliberately not filtered** the way `extractMultiLineLayer` filters line gaps. Plotly omits a null point from a line's DOM but keeps a zero-height rect for a bar's — 3 `.point` nodes for 3 bar values with a gap, versus 2 markers for the same line data. Dropping it here would slide every later bar's highlight onto its neighbour, which the gap-alignment test catches. The reasoning is recorded on the assertion; the model-side consequence is #726.

Verified locally: `npm run lint:fix`, `npm run type-check`, `npm run build`, `npm test` (1390 unit + 57 esm, all passing). The E2E suite has no plotly fixture, so it does not exercise this path — the browser verification above stands in for it, and a plotly e2e fixture would be a worthwhile future addition.

### DIFF
--- a/src/adapters/plotly/extractor.ts
+++ b/src/adapters/plotly/extractor.ts
@@ -898,7 +898,7 @@ function extractLayer(
       return extractScatterLayer(trace, id, title, selectors, axes, gd);
 
     case TraceType.BAR:
-      return extractBarLayer(trace, id, title, selectors, axes);
+      return extractBarLayer(trace, calcdata, id, title, selectors, axes);
 
     case TraceType.HEATMAP:
       return extractHeatmapLayer(trace, id, title, selectors, axes, gd);
@@ -989,8 +989,30 @@ function extractScatterLayer(
 // Bar
 // ---------------------------------------------------------------------------
 
+/**
+ * Reads the value plotly actually drew for one bar. Shared by the single-trace
+ * and segmented bar extractors, which face the same `barnorm` question.
+ *
+ * `cd.s` is the bar's own size after `barnorm` has been applied, so it is the
+ * percentage or fraction on screen rather than the raw input number. It is
+ * orientation-independent — plotly keeps the position on `cd.p` for both
+ * vertical and horizontal bars. `cd.x`/`cd.y` are deliberately not used: for
+ * stacked bars they hold the running top of the stack, not the segment.
+ *
+ * Falls back to the raw trace value when calcdata is unavailable (a chart
+ * captured before plotly computed it) or holds a non-finite size.
+ */
+function drawnBarValue(
+  cd: PlotlyCalcData | undefined,
+  raw: string | number,
+): string | number {
+  const size = cd?.s;
+  return typeof size === 'number' && Number.isFinite(size) ? size : raw;
+}
+
 function extractBarLayer(
   trace: PlotlyTrace,
+  calcdata: PlotlyCalcData[],
   id: string,
   title: string | undefined,
   selectors: string | undefined,
@@ -1010,7 +1032,8 @@ function extractBarLayer(
     // vertical, which already matches AbstractBarPlot's per-orientation reading
     // (value from `point.x` when HORIZONTAL, from `point.y` otherwise). No swap
     // is needed — and the plotly x/y axes already line up with the layer axes.
-    data.push({ x: x[i], y: y[i] });
+    const value = drawnBarValue(calcdata[i], isHorizontal ? x[i] : y[i]);
+    data.push(isHorizontal ? { x: value, y: y[i] } : { x: x[i], y: value });
   }
 
   if (data.length === 0)
@@ -1426,26 +1449,6 @@ function extractCandlestickLayer(
 // ---------------------------------------------------------------------------
 // Segmented bars (dodged / stacked / normalized)
 // ---------------------------------------------------------------------------
-
-/**
- * Reads the value plotly actually drew for one bar.
- *
- * `cd.s` is the bar's own size after `barnorm` has been applied, so it is the
- * percentage or fraction on screen rather than the raw input number. It is
- * orientation-independent — plotly keeps the position on `cd.p` for both
- * vertical and horizontal bars. `cd.x`/`cd.y` are deliberately not used: for
- * stacked bars they hold the running top of the stack, not the segment.
- *
- * Falls back to the raw trace value when calcdata is unavailable (a chart
- * captured before plotly computed it) or holds a non-finite size.
- */
-function drawnBarValue(
-  cd: PlotlyCalcData | undefined,
-  raw: string | number,
-): string | number {
-  const size = cd?.s;
-  return typeof size === 'number' && Number.isFinite(size) ? size : raw;
-}
 
 /**
  * Combines multiple plotly bar traces into a single MAIDR segmented bar layer.
--- a/src/adapters/plotly/extractor.ts
+++ b/src/adapters/plotly/extractor.ts
@@ -990,24 +990,37 @@ function extractScatterLayer(
 // ---------------------------------------------------------------------------
 
 /**
- * Reads the value plotly actually drew for one bar. Shared by the single-trace
- * and segmented bar extractors, which face the same `barnorm` question.
+ * Builds one bar point, taking the value plotly actually drew from calcdata
+ * and putting it on the axis the orientation calls for. Shared by the
+ * single-trace and segmented bar extractors, which face the same `barnorm`
+ * question and must place the value the same way.
  *
- * `cd.s` is the bar's own size after `barnorm` has been applied, so it is the
- * percentage or fraction on screen rather than the raw input number. It is
- * orientation-independent — plotly keeps the position on `cd.p` for both
- * vertical and horizontal bars. `cd.x`/`cd.y` are deliberately not used: for
- * stacked bars they hold the running top of the stack, not the segment.
+ * The value comes from `cd.s`, the bar's own size after `barnorm` has been
+ * applied, so it is the percentage or fraction on screen rather than the raw
+ * input number. `cd.s` is orientation-independent — plotly keeps the position
+ * on `cd.p` for both vertical and horizontal bars. `cd.x`/`cd.y` are
+ * deliberately not used: for stacked bars they hold the running top of the
+ * stack, not the segment.
  *
- * Falls back to the raw trace value when calcdata is unavailable (a chart
- * captured before plotly computed it) or holds a non-finite size.
+ * Plotly stores the bar value on `x` for horizontal bars and on `y` for
+ * vertical, which already matches AbstractBarPlot's per-orientation reading
+ * (value from `point.x` when HORIZONTAL, from `point.y` otherwise). No swap is
+ * needed — and the plotly x/y axes already line up with the layer axes. The
+ * raw value on that same axis is the fallback, used when calcdata is
+ * unavailable (a chart captured before plotly computed it) or holds a
+ * non-finite size.
  */
-function drawnBarValue(
+function barPoint(
   cd: PlotlyCalcData | undefined,
-  raw: string | number,
-): string | number {
+  x: string | number,
+  y: string | number,
+  isHorizontal: boolean,
+): BarPoint {
   const size = cd?.s;
-  return typeof size === 'number' && Number.isFinite(size) ? size : raw;
+  const drawn = typeof size === 'number' && Number.isFinite(size) ? size : undefined;
+  return isHorizontal
+    ? { x: drawn ?? x, y }
+    : { x, y: drawn ?? y };
 }
 
 function extractBarLayer(
@@ -1028,12 +1041,7 @@ function extractBarLayer(
   const data: BarPoint[] = [];
 
   for (let i = 0; i < len; i++) {
-    // Plotly stores the bar value on `x` for horizontal bars and on `y` for
-    // vertical, which already matches AbstractBarPlot's per-orientation reading
-    // (value from `point.x` when HORIZONTAL, from `point.y` otherwise). No swap
-    // is needed — and the plotly x/y axes already line up with the layer axes.
-    const value = drawnBarValue(calcdata[i], isHorizontal ? x[i] : y[i]);
-    data.push(isHorizontal ? { x: value, y: y[i] } : { x: x[i], y: value });
+    data.push(barPoint(calcdata[i], x[i], y[i], isHorizontal));
   }
 
   if (data.length === 0)
@@ -1483,10 +1491,7 @@ function extractSegmentedBarLayer(
     const series: SegmentedPoint[] = [];
 
     for (let i = 0; i < len; i++) {
-      // Plotly stores the value on `x` for horizontal bars and on `y` for
-      // vertical, matching AbstractBarPlot's per-orientation reading — no swap.
-      const value = drawnBarValue(cd[i], isHorizontal ? x[i] : y[i]);
-      series.push(isHorizontal ? { x: value, y: y[i], z } : { x: x[i], y: value, z });
+      series.push({ ...barPoint(cd[i], x[i], y[i], isHorizontal), z });
     }
 
     data.push(series);
--- a/src/adapters/plotly/extractor.ts
+++ b/src/adapters/plotly/extractor.ts
@@ -441,14 +441,14 @@ function buildSubplotLayers(
     const barnorm = layout.barnorm ?? '';
 
     if (barmode === 'group') {
-      const layer = extractSegmentedBarLayer(barTraces, TraceType.DODGED, xLabel, yLabel, gd);
+      const layer = extractSegmentedBarLayer(barTraces, group, TraceType.DODGED, xLabel, yLabel, gd);
       if (layer)
         layers.push(layer);
     } else if (barmode === 'stack' || barmode === 'relative') {
       const type = barnorm === 'percent' || barnorm === 'fraction'
         ? TraceType.NORMALIZED
         : TraceType.STACKED;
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Added the grouped-bar coverage in 7ffdbbd. Before writing it I checked the claim against a real chart rather than the plotly source I had reasoned from, using plotly.js 2.35.2 — the version `examples/plotly-*.html` pin — driven through Playwright.

`barmode: 'group'` with `barnorm: 'percent'`:

```
value axis range: [0, 60.15037593984962]
 trace 0 (East, raw y=[120, 80])
   pt0 {"p":0,"s":57.14285714285714,"b":0,"y":57.14285714285714}
 trace 1 (West, raw y=[90, 70])
   pt0 {"p":0,"s":42.857142857142854,"b":0,"y":42.857142857142854}
```

So grouped bars are scaled, and the axis tops out at 60 rather than 120 — the same divergence as the stacked case this PR fixes, and it was equally unannounced. The new case pins `TraceType.DODGED` with those values; it fails on `main` alongside the other three.

The same run confirmed the two readings the fix depends on:

- **`cd.s` vs `cd.y`.** For the stack, trace 1 at Q1 has `s: 42.857…` but `y: 100` — the running stack top. Taking `.y`, which the issue offered as an alternative, would have swapped the raw total for a different wrong number on every trace above the first.
- **Horizontal bars.** `s` carries the size, `x` carries the running total, and `y` carries the *numeric position* (`0`, `1`) rather than the category string. The category therefore has to keep coming from the trace array, as it does. The horizontal fixture now includes those positions, so the assertion that `Q1` survives has something to fail against — previously it would have passed even if the code had read `cd.y`.

Full suite green after the change: 1386 unit + 57 esm, plus `lint:fix`, `type-check`, and `build`.

---

jooyoungseo:
Good question to raise — it's the one assumption the fix rests on that nothing was exercising. I could test it here (plotly.js 2.35.2 driven through Playwright), so rather than reason from the source: **bar `calc()` does not drop points.** Pinned in 21eb51b.

Two series with a `null` mid-series in one and a missing value in the other:

```
raw array lengths : [4,4]
calcdata lengths  : [4,4]
rendered bars     : [4,4]
 trace 0 (East, y=[120, null, 60, 40])
   cd[0] {"p":0,"s":57.14285714285714}
   cd[1] {"p":1}                          <- held in place, s is BADNUM
   cd[2] {"p":2,"s":100}
   cd[3] {"p":3,"s":66.66666666666667}
```

The gap keeps its slot and its position, carrying `s: undefined` — plotly's BADNUM — so `cd[i]` stays in step with `x[i]`/`y[i]`. Unlike scatter, bar's `calc()` builds `serieslen` entries up front and writes BADNUM into the gaps rather than compacting. Same result for `orientation: 'h'`.

That also means the two failure modes are already separated: a gap hits the `typeof`/`isFinite` guard and falls back to the raw value (`null` here, unchanged from before this PR — the old code passed `y[i]` through identically), while the surrounding points keep their own shares.

The new test pins both halves. Its sharpest assertion is `data[0][2].y === 100`: East is the whole stack at Q3 because West is absent there, and a one-entry shift would report 66.67 instead — plausible-looking and wrong, which is the failure mode you described. It fails on `main` along with the other four.

On the typing note: you're right that `PlotlyTrace.x`/`.y` are `(number | string)[]` and won't admit the `null` that plotly's runtime data actually carries. I've kept that to a single narrow cast in the test with a comment, rather than widening the type here — that ripples into every consumer that indexes these arrays and assigns into `BarPoint`, which is well outside this fix. Worth its own issue if you'd like me to open one.

Suite green: 1387 unit + 57 esm, plus `lint:fix`, `type-check`, and `build`.

---

jooyoungseo:
Confirmed the single-trace gap and fixed it here rather than deferring it — f008feb. Checked against plotly.js 2.35.2 first:

```
barmode: 'stack', barnorm: 'percent', one trace, y = [120, 80, 40]
  calcdata: [{p:0,s:100}, {p:1,s:100}, {p:2,s:100}]
```

Plotly normalizes a lone trace against itself, so all three bars are drawn at identical full height while the announcement read 120, 80, 40 — three different heights that are not on screen. Degenerate chart, but the discrepancy is louder than the multi-trace case rather than quieter.

The fix was smaller than expected: `extractLayer` already had the calcdata in hand for the histogram branch, so it was one more argument into `extractBarLayer` and a reuse of `drawnBarValue`, which moved up to the bar section it now shares. Two tests added — the normalized lone trace, and an unnormalized control.

End-to-end through the built bundle, driving by keyboard and reading the `role="alert"` region:

```
stacked + barnorm     Quarter is Q1, Share is 57.14285714285714, Level is East
lone trace + barnorm  Quarter is Q1, Share is 100
```

Both read `Share is 120` on `main`. Nine tests now; six fail on `main`, three are controls. Suite green: 1389 unit + 57 esm, plus `lint:fix`, `type-check`, and `build`. PR description updated.

One thing that only surfaced end-to-end, which I have **not** changed: the value is announced at full precision — `57.14285714285714`, sixteen digits spoken aloud. Normalized charts now always produce non-integers, so this went from rare to guaranteed on this path, though any non-integer data behaved this way before. It looks like formatting-layer work rather than extractor work: no adapter emits `AxisFormat` today, and every adapter that computes percentages (highcharts' `p.percentage`, amcharts, vega) faces the same question, so a plotly-only patch would set an inconsistent precedent. Happy to open a follow-up issue for it if you agree that's the right home.

---

jooyoungseo:
Took the helper suggestion — 7838d25. Worth doing for a reason beyond the duplication: each call site carried *two* ternaries that had to agree, one choosing the axis the drawn value lands on and one choosing the raw value that backs it. If those ever drifted apart, a horizontal chart would fall back to the category label as its value. `barPoint` now decides both together, and the segmented caller spreads the result to add its `fill`:

```ts
series.push({ ...barPoint(cd[i], x[i], y[i], isHorizontal), z });
```

Behaviour unchanged: the nine existing cases pass, and the announcements from a rebuilt bundle are identical for both the stacked and lone-trace charts.

One correction to your note about the `Number.isFinite` guard, since I tried to pin it and could not. I added a zero case, then swapped `??` for `||` in the source to check the test would catch it — **all 38 passed either way**. Under `barnorm` a drawn 0 implies a raw 0, so both operators answer 0 and no fixture can separate them. The guard is still the right shape, but the distinction is unobservable here rather than test-covered, and the test now says so rather than implying coverage it does not have:

> Note this does not distinguish a falsy-check fallback from a nullish one — under `barnorm` a drawn 0 implies a raw 0, so both would answer 0 here. What it pins is that the zero survives extraction at all.

Suite green: 1390 unit + 57 esm, plus `lint:fix`, `type-check`, and `build`.

The float-precision follow-up is still worth filing, as you and the previous review both noted — I'm holding until the maintainer confirms the formatting layer is the right home, since it would set the first adapter-emitted `AxisFormat` precedent and the same question applies to every adapter that computes percentages.

---

jooyoungseo:
Checked the gap analysis against a running chart. One part of it is wrong, and the suggested fix would introduce a bug — but there's a real finding underneath it, so thank you for pushing on this. Recorded the reasoning in 9bb72dc.

**A gapped bar does not announce `null`.** Driving a plain stacked chart with `y: [120, null, 40]` through the built bundle:

```
point 1: Quarter is Q1, Revenue is 120, Level is East
point 2: Quarter is Q2, Revenue is missing, Level is East
point 3: Quarter is Q3, Revenue is 40, Level is East
```

The text service already renders the null as `missing`. No `null`, no `NaN`, no fabricated `0` in the announcement.

**Filtering nulls the way the line extractor does would break bars.** The two conventions differ because the DOM differs, which I hadn't appreciated until measuring it:

```
bar,  y = [120, null, 40] -> 3 .point nodes
  path[1] = "M192.5,320V320H332.5V320"   <- zero-height rect at the baseline
line, y = [120, null, 40] -> 2 markers
```

Plotly omits a null point from a line's DOM but keeps a zero-height rect for a bar's. `extractMultiLineLayer` skips gaps *because* the DOM omits them — its own comment says "skipping keeps indices aligned with the DOM". Applying that here would invert the effect: the data would be one shorter than the rendered bars, sliding every later highlight onto its neighbour. That is precisely the off-by-one the gap-alignment test exists to catch, so adopting the suggestion would have made that test fail.

I've noted this on the assertion rather than leaving the asymmetry looking accidental, since the next person comparing the two extractors will reasonably wonder.

**The real finding**, which stands: `AbstractBarPlot` builds `barValues` with `Number(point.y)`, so a gap becomes `0` there — and that array feeds `min`/`max`, sonification, and extrema navigation, where the text path's `missing` handling does not reach. For `[120, null, 40]` the row's min is `0` rather than `40`. Pre-existing and untouched by this PR — the null passthrough is byte-identical to `data.push({ x: x[i], y: y[i] })` before it — but a genuine gap in a path that has no equivalent of `missing`. Worth its own issue; I'd rather not fold a model-layer change into an adapter fix.

That makes two follow-ups now queued behind this PR — this one and the unrounded precision — both waiting on the maintainer's call. Suite green at 38 in this file, 1390 unit + 57 esm overall.

---

jooyoungseo:
Both follow-ups are now filed, so neither gets lost:

- **#725** — computed percentages announced at full float precision (`57.14285714285714`). Filed against the formatting layer rather than this adapter, since no adapter emits `AxisFormat` today and highcharts/amcharts/vega-lite all compute values with the same exposure.
- **#726** — a bar gap is sonified as a zero-height bar. `Number(null) === 0` in `AbstractBarPlot`'s `barValues` feeds `min`/`max`, sonification, and extrema, so the audio reports a measured zero where the text correctly says `missing`. Filed against the model, and it carries the warning not to "fix" it by filtering nulls in the extractors — plotly keeps a zero-height rect in the bar DOM, so that would misalign every later highlight.

The `BarPoint` typing note is the same thread as #726: the type says `number | string` while a gap really can be `null`, which is why the gap test needs a cast. Whichever way #726 resolves — excluding gaps from the numeric summaries, or representing them so audio can tell absent from zero — should settle the type at the same time, rather than widening it here and leaving the model still reading `0`.

Agreed on a plotly e2e fixture being the right long-term answer to the coverage gap. The browser verification quoted in the description is a stand-in, not a substitute, and it only runs when someone thinks to run it.

---

PR #727, opened 2026-08-04
fix(format): round announced values to two decimals by default

### LINKED ISSUE #725: fix: computed percentages are announced at full float precision
# Bug Report

## Description

A value MAIDR computes rather than reads verbatim is announced at full float precision, so a screen reader speaks sixteen digits for a single bar. Normalized charts make this unavoidable: a share is almost never a round number.

Surfaced while verifying #724 end to end. That PR is the reason the case is now guaranteed rather than occasional — before it, a normalized plotly chart announced the raw integers, wrongly — but the formatting behaviour itself predates it and is not plotly-specific.

## Steps to Reproduce

1. Render a plotly bar chart with `barnorm: 'percent'`:

```js
Plotly.newPlot('chart', [
  { type: 'bar', x: ['Q1', 'Q2'], y: [120, 80], name: 'East' },
  { type: 'bar', x: ['Q1', 'Q2'], y: [90, 70], name: 'West' },
], { barmode: 'stack', barnorm: 'percent',
     xaxis: { title: { text: 'Quarter' } }, yaxis: { title: { text: 'Share' } } });
```

2. Activate MAIDR and press <kbd>Enter</kbd>, then <kbd>→</kbd>.

## Actual Behavior

```
Quarter is Q1, Share is 57.14285714285714, Level is East
```

## Expected Behavior

Something a person can listen to — `Share is 57.1`, or whatever precision the project settles on.

## Notes

The scope is wider than one adapter, which is why this is filed rather than patched inside #724.

- No adapter emits `AxisFormat` today. The formatting system (`src/util/format.ts`, `AxisConfig.format`) is fed from author-supplied maidr JSON, so auto-detection adapters have no precedent for emitting one.
- Every adapter that computes a value rather than reading it has the same exposure — `p.percentage` in `src/adapters/highcharts/adapter.ts`, and the amcharts and vega-lite normalized paths.
- Fixing it in one adapter would set the precedent for all of them, so it wants a single deliberate decision: whether adapters may emit `AxisFormat`, or whether the text service should apply a default precision to computed values.

Rounding the extracted data instead would be the wrong layer — braille, sonification, and extrema all read the same numbers, and the value is not wrong, only its rendering.

## Environment

- Browser: Chromium (via Playwright)
- Version: `main` at 7a138dc, plus #724
- Chart library: plotly.js 2.35.2 (the version `examples/plotly-*.html` pin); not plotly-specific

jooyoungseo:
# Pull Request

## Description

A value MAIDR computes rather than reads verbatim is announced at full float precision. A `barnorm` share came out as:

```
Quarter is Q1, Share is 57.14285714285714, Level is East
```

Sixteen digits spoken aloud for two digits of meaning.

`defaultFormat` was `` `${value}` ``, so any value without an explicit `AxisFormat` reached the screen reader exactly as JavaScript stringifies it. Every adapter that derives a percentage rather than reading one has the same exposure — plotly's `barnorm`, highcharts' `p.percentage`, the amcharts and vega-lite normalized paths — which is why the default belongs here rather than in any one of them.

## Related Issues

Fixes #725

## Changes Made

`src/util/format.ts` — `defaultFormat` now rounds a fractional number to two decimal places.

- **Two places** matches what the Go To Extrema dialog already displays (`src/ui/components/GoToExtrema.tsx` uses `toFixed(2)`), so the two surfaces agree rather than each picking their own precision.
- **Integers and strings are untouched.** `120` stays `120`, `Q1` stays `Q1` — no trailing `.00` appears where nothing was fractional.
- **Nothing is padded.** `0.5` stays `0.5`, not `0.50`, because the result is re-parsed as a number rather than left as a `toFixed` string. A value that rounds up to a whole number therefore reads as one: `99.999` becomes `100`.
- **A small value is not rounded away.** `0.000123456` would become `0` at two places, which is a different claim about the chart than the truth. Those fall back to three significant digits — `0.000123`. Below ~1e-6 that returns an exponent, but so does plain stringification, so nothing changes notation because of this.
- **Non-finite values skip rounding**, leaving `NaN` for `FormatUtil.wrapFormat` to render as `missing`. `Infinity` is announced as `Infinity`, unchanged — `wrapFormat` maps only `NaN`.

An explicit `AxisFormat` still takes priority, so an author who needs four decimals or a currency still gets exactly that. `defaultFormat`'s identity as an export is preserved.

`e2e_tests/specs/histogram.spec.ts` builds its expected announcement through `defaultFormat` rather than stringifying the raw datum, since the iris bin edges carry float artefacts (`1.5900000000000003`) that are now announced as `1.59`. Importing the real formatter rather than copying its rounding keeps the expectation tied to the app's contract.

Not applicable — this changes spoken text, quoted below.

## Checklist

## Additional Notes

**Rounding stops at the announcement.** Sonification, braille, and extrema navigation all read the underlying numbers from the model, not the formatted string — `AbstractBarPlot.audio` uses `barValues`, `braille` uses `values`/`min`/`max`. So a bar's pitch and its braille cell are unchanged by this; only the spoken text is shorter. Rounding the extracted data instead would have changed all four modalities, which is why this sits in the formatting layer.

**Verified end to end** on the issue's own chart, driving the built bundle by keyboard and reading the `role="alert"` region:

```
before   Quarter is Q1, Share is 57.14285714285714, Level is East
after    Quarter is Q1, Share is 57.14, Level is East
```

**Fourteen unit tests**, across two files that had no coverage before:

- `test/util/format.test.ts` — the shortening, untouched integers and strings, no padding, whole-number rounding, the small-value fallback and its exponent, an explicit `AxisFormat` still winning, and the `missing` wrapper.
- `test/service/formatter.test.ts` — the same behaviour through `FormatterService`'s resolve → wrap → apply chain, which is the path every chart actually takes: the default on an unformatted axis, an explicit format winning while its sibling axis keeps the default, arrays as boxplot outliers arrive, a missing value, and an unknown layer falling back.

**On the choice of two decimals.** There is no way at this layer to tell deliberate precision from computed noise, so this is a default rather than a detection, and it reaches beyond the `barnorm` case in #725: any axis without an explicit `AxisFormat` now has its spoken value rounded, including data whose precision is deliberate. Two places is the same answer the extrema dialog already gives, and the `AxisFormat` escape hatch already exists for anything else — that is what it is for.

**Known, pre-existing, not changed here:** `FormatterService.hasCustomFormatter` compares `layerFormatters[axis] !== defaultFormat`, but `resolveAxisFormat` always stores a fresh `wrapFormat(...)` closure, so it returns `true` for every layer it knows about — measured, not inferred. Both `goToExtremaViewModel` call sites that gate on it therefore run values through the formatter regardless, despite comments saying otherwise. Benign here (extrema labels get the same two decimals as the announcement) but the gate does not do what it says. Worth its own issue rather than a drive-by fix inside a formatting change.

`toFixed`'s binary-rounding quirk (`1.005` → `1`) is inherited, not introduced — `GoToExtrema.tsx` already carries the same exposure, and diverging here would make the two surfaces disagree.

Verified locally: `npm run lint:fix`, `npm run type-check`, `npm run build`, `npm test` (1405 unit + 57 esm, all passing), plus the histogram e2e spec against a matching build (23/23).

### DIFF
--- a/e2e_tests/specs/histogram.spec.ts
+++ b/e2e_tests/specs/histogram.spec.ts
@@ -51,16 +51,30 @@ function getHistogramDisplayValue(layer: MaidrLayer | undefined, index: number):
   const dataPoint = layer.data[index];
 
   if (dataPoint && 'xMin' in dataPoint && 'xMax' in dataPoint) {
-    const xMin = defaultFormat(dataPoint.xMin as number);
-    const xMax = defaultFormat(dataPoint.xMax as number);
-    return `${xMin} through ${xMax}`;
+    return `${announced(dataPoint.xMin)} through ${announced(dataPoint.xMax)}`;
   }
 
   if (dataPoint && 'x' in dataPoint) {
-    return defaultFormat(dataPoint.x as number);
+    return announced(dataPoint.x);
   }
 
   throw new Error(`Data point at index ${index} has invalid format`);
+}
+
+/**
+ * Renders one datum the way MAIDR announces it.
+ *
+ * `MaidrLayer['data']` is a wide union, so narrowing rather than asserting
+ * keeps this honest about a point whose field is neither a number nor a
+ * string — `defaultFormat` only accepts those two.
+ *
+ * @param value - A raw field from a data point
+ * @returns The value as it reaches the screen reader
+ */
+function announced(value: unknown): string {
+  return typeof value === 'number' || typeof value === 'string'
+    ? defaultFormat(value)
+    : String(value);
 }
 
 test.describe('Histogram', () => {
--- a/e2e_tests/specs/histogram.spec.ts
+++ b/e2e_tests/specs/histogram.spec.ts
@@ -1,6 +1,7 @@
 import type { Page } from '@playwright/test';
 import type { Maidr, MaidrLayer } from '../../src/type/grammar';
 import { expect, test } from '@playwright/test';
+import { defaultFormat } from '../../src/util/format';
 import { HistogramPage } from '../page-objects/plots/histogram-page';
 import { TestConstants } from '../utils/constants';
 import { extractMaidrData } from '../utils/maidr-data';
@@ -24,6 +25,15 @@ async function setupHistogramPage(
 
 /**
  * Safely extracts the display value from a histogram data point
+ *
+ * Values are put through {@link defaultFormat}, the same formatter the
+ * announcement goes through, rather than stringified raw. The iris bin edges
+ * in the fixture carry float artefacts — `1.5900000000000003` — and MAIDR
+ * announces the rounded `1.59`. Importing the real formatter rather than
+ * copying its rounding keeps this expectation tied to the app's contract: a
+ * change in default precision updates the spec with it, while landing on the
+ * wrong bin still fails.
+ *
  * @param layer - The histogram layer containing data points
  * @param index - Index of the data point to extract value from
  * @returns The formatted string value suitable for display comparison
@@ -41,13 +51,13 @@ function getHistogramDisplayValue(layer: MaidrLayer | undefined, index: number):
   const dataPoint = layer.data[index];
 
   if (dataPoint && 'xMin' in dataPoint && 'xMax' in dataPoint) {
-    const xMin = String(dataPoint.xMin);
-    const xMax = String(dataPoint.xMax);
+    const xMin = defaultFormat(dataPoint.xMin as number);
+    const xMax = defaultFormat(dataPoint.xMax as number);
     return `${xMin} through ${xMax}`;
   }
 
   if (dataPoint && 'x' in dataPoint) {
-    return String(dataPoint.x);
+    return defaultFormat(dataPoint.x as number);
   }
 
   throw new Error(`Data point at index ${index} has invalid format`);
--- a/src/util/format.ts
+++ b/src/util/format.ts
@@ -48,6 +48,11 @@ export const defaultFormat: FormatFunction = (value: number | string): string =>
 
   // Rounding erased the value (0.00012 -> 0). Announcing `0` for something that
   // is not zero is worse than the extra digits, so keep it visible instead.
+  //
+  // The check catches the negative case as well, and relies on it: a small
+  // negative rounds to `-0`, and `-0 !== 0` is false in IEEE-754, so it lands
+  // here rather than being announced as a signed zero. Comparing against
+  // `Object.is(rounded, 0)` instead would let `-0` through.
   return `${Number(value.toPrecision(DEFAULT_SIGNIFICANT_DIGITS))}`;
 };
 
--- a/src/util/format.ts
+++ b/src/util/format.ts
@@ -6,14 +6,49 @@ import type { AxisFormat, FormatFunction, FormatType } from '@type/grammar';
 export type FormattableValue = number | string | (number | string)[];
 
 /**
+ * Decimal places the default format keeps for a fractional number. Matches the
+ * precision the Go To Extrema dialog already displays values at.
+ */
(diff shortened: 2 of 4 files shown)


jooyoungseo:
Both review rounds addressed. f3c4834 and 27afbc8.

**`FormatterService` now has tests** (it had none). Five cases through the real resolve → wrap → apply chain rather than the pure function alone: the default applied to an unformatted axis, an explicit `AxisFormat` winning while the sibling axis still gets the default, arrays as boxplot outliers arrive, a missing value still reading `missing`, and an unknown layer falling back.

**Exponential notation is pinned.** Worth stating what I found: below ~1e-6 the fallback returns an exponent, but so does plain stringification — `String(1.234e-7)` is already `'1.234e-7'`. Only the mantissa shortens, so nothing switches notation because of this change. The test says so.

**The `as number` casts are gone.** You were right that they claimed something the union cannot back. Narrowed with a `typeof` check instead, which needs no cast since `defaultFormat` accepts `number | string` anyway.

Two things I am *not* changing, with reasons:

- **`toFixed`'s float quirk** (`1.005` → `"1"`). Pre-existing JS behaviour, and `GoToExtrema.tsx` already carries the same exposure through its own `toFixed(2)`. Diverging here would make the two surfaces disagree, which is worse than the quirk.
- **`hasCustomFormatter` looking dead.** Your reading is right and I confirmed it: `resolveAxisFormat` always returns a fresh `wrapFormat(...)` closure, so `layerFormatters[axis] !== defaultFormat` is true for every known layer, formatted or not. That predates this PR and this PR does not change it — the identity of the `defaultFormat` export is untouched. It wants its own issue rather than a drive-by fix inside a formatting change; happy to file one.

On the **global behaviour change**: agreed, and it is worth being plain that this reaches beyond #725. Any axis without an explicit `AxisFormat` now has its spoken value rounded, including data whose precision is deliberate. The commit body carries that into the changelog. The `AxisFormat` escape hatch is the answer for anyone who needs more, and it already exists.

One process note for anyone re-running this locally: the e2e suite loads `dist/maidr.js` over `file://`, so it tests whatever was built last. I got two spurious histogram failures from a `dist/` left over from another branch before rebuilding. With a matching build, all 23 histogram specs pass.

Suite green: 1405 unit + 57 esm, plus `lint:fix`, `type-check`, and `build`.

---

jooyoungseo:
Your `hasCustomFormatter` reading is correct, and it contradicts an earlier review on this PR that concluded the opposite — so I measured it rather than pick a side:

```
hasCustomFormatter(layer with no format)  = true
hasCustomFormatter(unknown layer)         = false
```

`resolveAxisFormat` always returns a fresh `wrapFormat(...)` closure, so the stored function is never reference-equal to the `defaultFormat` export. The gate is `true` for every layer the service knows about, formatted or not. An earlier review said this made `goToExtremaViewModel` "unaffected" by the change; that is wrong. Both call sites — `formatTargetLabels` and `getAvailableXValueOptions` — do now run values through the rounding formatter, despite comments there saying labels pass through unchanged when no format is configured.

The outcome is benign and arguably what you want: Go To Extrema labels get the same two decimals as the announcement, and the dialog's own `toFixed(2)` on `target.value` is a separate path, so nothing is rounded twice into something wrong. But the gate does not do what it says, and this PR is what makes that visible rather than invisible.

Not fixing it here — it is untouched pre-existing logic and the fix is a judgement call between repairing the reference check and deleting the branch, which does not belong inside a formatting change. Happy to open an issue for it.

On the `toFixed` binary-rounding quirk: agreed it is inherited rather than introduced, and `GoToExtrema.tsx` already carries the same exposure. Diverging here would make the two surfaces disagree, which is worse than the quirk.

---

PR #728, opened 2026-08-04
fix(bar): keep a gap in bar data out of pitch, braille, and extrema

### LINKED ISSUE #726: fix: a gap in a bar chart is sonified as a zero-height bar
# Bug Report

## Description

A missing point in a bar chart is announced correctly as `missing`, but every other modality treats it as a bar of value zero. `AbstractBarPlot` builds its numeric array with `Number(point.y)`, and `Number(null)` is `0`:

https://github.com/xability/maidr/blob/main/src/model/bar.ts#L30-L36

That array feeds `min`/`max`, and through them sonification and extrema navigation. So the text layer knows the data is absent while the audio layer reports a real measurement of zero — the two modalities disagree about what is on screen, and only the spoken one is right.

Found while reviewing #724, which does not touch this: the null passthrough there is byte-identical before and after that fix.

## Steps to Reproduce

1. Render a bar chart with a gap:

```js
Plotly.newPlot('chart', [
  { type: 'bar', x: ['Q1', 'Q2', 'Q3'], y: [120, null, 40], name: 'East' },
  { type: 'bar', x: ['Q1', 'Q2', 'Q3'], y: [90, 70, 20], name: 'West' },
], { barmode: 'stack',
     xaxis: { title: { text: 'Quarter' } }, yaxis: { title: { text: 'Revenue' } } });
```

2. Activate MAIDR, press <kbd>Enter</kbd>, then <kbd>→</kbd> across the first series.

## Actual Behavior

The announcements are right:

```
Quarter is Q1, Revenue is 120, Level is East
Quarter is Q2, Revenue is missing, Level is East
Quarter is Q3, Revenue is 40, Level is East
```

But `barValues` for that row is `[120, 0, 40]`, so the row's minimum is `0` rather than `40`. The gap sounds like a measured low point, and extrema navigation can land on it as the minimum.

## Expected Behavior

A gap should be excluded from the numeric summaries, or represented so the audio and extrema paths can tell "absent" from "zero" — consistent with what the text layer already does.

## Notes

**Do not fix this by filtering nulls out of the bar extractors.** `extractMultiLineLayer` does exactly that for lines, and its comment explains why it is safe there — "skipping keeps indices aligned with the DOM, which omits null points". Bars are the opposite case. Measured against plotly.js 2.35.2:

```
bar,  y = [120, null, 40] -> 3 .point nodes
  path[1] = "M192.5,320V320H332.5V320"   <- zero-height rect, still in the DOM
line, y = [120, null, 40] -> 2 markers
```

Plotly keeps a zero-height rect for a bar's gap and omits a line's. Dropping the point in the extractor would leave the data one shorter than the rendered bars and slide every later highlight onto its neighbour. The gap-alignment test added in #724 covers that.

So the fix belongs in the model rather than the adapters — which also means it applies to every chart library, not just plotly.

## Environment

- Browser: Chromium (via Playwright)
- Version: `main` at 7a138dc, plus #724
- Chart library: plotly.js 2.35.2 for the reproduction; the model-layer behaviour is library-independent

jooyoungseo:
# Pull Request

## Description

A missing bar was announced correctly as `missing` and treated as a measurement of zero by every other modality. `AbstractBarPlot` built its numeric array with `Number(point.y)`, and `Number(null)` is `0`, so the gap:

- set the row's minimum, and through it the range every other bar's pitch is scaled against,
- sounded like a real bar at the bottom of the chart,
- could be reached as the row's extreme,
- and rendered in braille as the **tallest** glyph, because `0` was the only blank case and every band comparison a gap loses falls through to the top band.

The text layer knew the data was absent while the audio layer reported a real zero. The two modalities disagreed about what was on screen, and only the spoken one was right.

## Related Issues

Fixes #726

## Changes Made

**The core fix**

- **`src/model/bar.ts`** — a gap reads as `NaN` rather than `0`, via a `toBarValue` helper, and `min`/`max` are computed from finite values only. A gap is not a measurement, so it cannot set a range. A blank cell counts as a gap for the same reason `null` does: `Number('')` and `Number('  ')` are also `0`.
- **`src/service/audio.ts`** — a non-finite magnitude sounds the empty tone, the same cue an out-of-bounds move already uses. Interpolating it would have handed the oscillator a non-finite frequency. Deliberately not scoped to bar traces.
- **`src/service/braille.ts`** — a non-finite value renders blank, alongside the existing zero case. Without it a gap fell through to `⠉`, drawing the missing bar as the highest one on the chart.

**Reaching the paths the first commit missed.** Review found four, two of which were regressions this branch introduced:

- **`SegmentedTrace.createSummaryLevel`** summed gaps straight into the Total row, making that total `NaN`. NaN spreads: it seeded `MathUtil.minMax`, and `safeMin`/`safeMax` over every row use `Math.min`/`Math.max`, which propagate it. **One missing segment gave the whole chart a NaN pitch range**, and every real bar in it a non-finite frequency — the bug this PR set out to fix, reintroduced. Now only measured segments are summed, and a category is a gap in the total only when every segment in it is one.
- **`SegmentedTrace.mapToSvgElements`** reads `barValues` directly for its `skipZeros` DOM alignment, where the base class recomputes `Number(point.y)`. A gap used to be `0` there and matched; as `NaN` it stopped matching, so a gapped cell would have taken a rect belonging to a later bar and shifted every highlight after it in the row. The old reading is restored — this path decides which rendered element belongs to which datum, not what a value means.
- **`description`** in all three `AbstractBarPlot` subclasses now shares a `rangeStats()` helper, so an all-gap chart reports `missing` rather than the `±Infinity` `safeMin`/`safeMax` answer an empty set with. Sharing it is the point: two of the three override the stats block, and the guard existed for one commit before drifting was possible again.
- **`getExtremaTargets`**, in both classes that define it, offers no extreme for an all-gap row. `indexOf(±Infinity)` returns `-1`, and the target would have carried `pointIndex: -1` and moved the cursor off the row.

Not applicable — the change is to sound, braille, and navigation.

## Checklist

## Additional Notes

**Gaps are still not filtered out of the data**, and that is deliberate. `extractMultiLineLayer` drops null points for lines because plotly omits them from a line's DOM — its comment says "skipping keeps indices aligned with the DOM". Bars are the opposite case: plotly renders a zero-height rect and keeps it, measured at 3 `.point` nodes for 3 bar values with a gap versus 2 markers for the same line data. Dropping the point here would slide every later bar's highlight onto its neighbour, which the gap-alignment test added in #724 catches.

**Twenty-one unit tests**, across three files:

- `test/model/barGaps.test.ts` — range exclusion, a gap reporting no magnitude while the text still reads `missing`, blank and whitespace-only cells, an all-gap chart describing itself as `missing` for both `BarTrace` and `Histogram`, extrema reaching the true extreme and offering none for an all-gap row, the segmented Total row's NaN spread and its sums and announcement, and two that drive a real five-rect DOM to assert which rect the cursor highlights past a gap and that a gap and a real zero land on the same one.
- `test/service/braille.test.ts` — a gap renders as a blank cell between `⠉` and `⣀` rather than as the tallest bar.
- `test/service/audio.emptyTone.test.ts` — a non-finite magnitude plays the five-harmonic empty cue; a magnitude of zero still plays a single pitched tone.

Several are controls that pass either way — an unmodified row, a real zero, a stack with no gaps totalling `[210, 150]`, a histogram with ordinary bin counts — pinning that this only changes what a gap does.

**Corrections to earlier framing in this PR, kept rather than quietly edited away:**

- The all-gap `description` case **did not** announce the word `Infinity`, as an earlier revision of this description and the commit message for it both claimed. `Description.tsx`'s `isDisplayable` already filters non-finite numbers, and nothing else consumes `DescriptionState.stats`, so the Min and Max rows simply vanished from the dialog. Reporting `missing` is still better than a row silently disappearing, but the severity was "the stat quietly goes away", not "the user is told Infinity".
- A review held that the Total row announces the literal string `NaN`. It does not: driven in a browser, a Total whose every segment is a gap announces `Quarter is Q1, Revenue is missing, Level is Sum`. `TextService.formatSingleValue` has no missing-value mapping of its own, but it delegates to `FormatterService`, which wraps every formatter in `FormatUtil.wrapFormat` — and that maps `NaN` exactly as it maps `null`. Pinned with tests anyway, since nothing had asserted the summary row's text.

**Two limits worth stating rather than leaving to be discovered:**

- **This reaches only adapters that preserve `null` through to the point.** Highcharts filters null points out (`adapter.ts:1038`) and Chart.js coerces them to `0` (`extractor.ts:632`), so a gap in those charts still behaves as it did. #726 is fixed for the model and for plotly, not universally.
- **`BarPoint` stays typed as `number | string`.** Widening it to admit the `null` that really flows through is honest, and I tried it — it produces 9 errors cascading into `TextState`, `BrailleState`, and `ExtremaTarget` and their consumers. That is a type-accuracy refactor with no behavioural content, better as its own change than smuggled into a sonification fix.

Verified locally: `npm run lint:fix`, `npm run type-check`, `npm run build`, `npm test` (1413 unit + 57 esm, all passing).

### DIFF
--- a/src/model/bar.ts
+++ b/src/model/bar.ts
@@ -20,14 +20,18 @@ import { MovableGrid } from './movable';
  * that every other bar's pitch is scaled against. `NaN` keeps it absent, which
  * is what the text layer already announces it as.
  *
- * An empty string counts as a gap for the same reason: `Number('')` is also
- * `0`, so a hand-authored figure with a blank cell would land in the same trap.
+ * A blank cell counts as a gap for the same reason: `Number('')` and
+ * `Number('  ')` are also `0`, so a hand-authored figure with an empty cell
+ * would land in the same trap.
  *
  * @param raw - The value from the point, on whichever axis carries magnitude
  * @returns The magnitude, or `NaN` when the bar is a gap
  */
 function toBarValue(raw: string | number | null | undefined): number {
-  if (raw === null || raw === undefined || raw === '') {
+  if (raw === null || raw === undefined) {
+    return Number.NaN;
+  }
+  if (typeof raw === 'string' && raw.trim() === '') {
     return Number.NaN;
   }
   return Number(raw);
--- a/src/model/bar.ts
+++ b/src/model/bar.ts
@@ -145,20 +145,36 @@ export abstract class AbstractBarPlot<T extends BarPoint> extends AbstractTrace 
   }
 
   /**
+   * The chart's min and max description stats.
+   *
+   * A chart of nothing but gaps has no range at all, and `safeMin`/`safeMax`
+   * answer an empty set with positive and negative Infinity. Report that the
+   * way every other modality reports an absent value rather than announcing an
+   * infinity.
+   *
+   * Shared so `SegmentedTrace`, which replaces the whole stats block rather
+   * than extending it, cannot drift back to announcing an infinity.
+   *
+   * @returns The min and max stats, in that order
+   */
+  protected rangeStats(): DescriptionState['stats'] {
+    const chartMin = MathUtil.safeMin(this.min);
+    const chartMax = MathUtil.safeMax(this.max);
+    return [
+      { label: 'Min value', value: isMeasured(chartMin) ? chartMin : 'missing' },
+      { label: 'Max value', value: isMeasured(chartMax) ? chartMax : 'missing' },
+    ];
+  }
+
+  /**
    * Gets the description state for the bar plot trace.
    * @returns The description state containing chart metadata and data table
    */
   public get description(): DescriptionState {
     const isVertical = this.orientation === Orientation.VERTICAL;
-    // A chart of nothing but gaps has no range at all, and safeMin/safeMax
-    // answer an empty set with ±Infinity. Report it the way every other
-    // modality reports an absent value rather than announcing an infinity.
-    const chartMin = MathUtil.safeMin(this.min);
-    const chartMax = MathUtil.safeMax(this.max);
     const stats: DescriptionState['stats'] = [
       { label: 'Number of bars', value: this.points[0].length },
-      { label: 'Min value', value: isMeasured(chartMin) ? chartMin : 'missing' },
-      { label: 'Max value', value: isMeasured(chartMax) ? chartMax : 'missing' },
+      ...this.rangeStats(),
     ];
 
     if (this.points.length > 1) {
--- a/src/model/bar.ts
+++ b/src/model/bar.ts
@@ -20,6 +20,9 @@ import { MovableGrid } from './movable';
  * that every other bar's pitch is scaled against. `NaN` keeps it absent, which
  * is what the text layer already announces it as.
  *
+ * An empty string counts as a gap for the same reason: `Number('')` is also
+ * `0`, so a hand-authored figure with a blank cell would land in the same trap.
+ *
  * @param raw - The value from the point, on whichever axis carries magnitude
  * @returns The magnitude, or `NaN` when the bar is a gap
  */
@@ -147,10 +150,15 @@ export abstract class AbstractBarPlot<T extends BarPoint> extends AbstractTrace 
    */
   public get description(): DescriptionState {
     const isVertical = this.orientation === Orientation.VERTICAL;
+    // A chart of nothing but gaps has no range at all, and safeMin/safeMax
+    // answer an empty set with ±Infinity. Report it the way every other
+    // modality reports an absent value rather than announcing an infinity.
+    const chartMin = MathUtil.safeMin(this.min);
+    const chartMax = MathUtil.safeMax(this.max);
     const stats: DescriptionState['stats'] = [
       { label: 'Number of bars', value: this.points[0].length },
-      { label: 'Min value', value: MathUtil.safeMin(this.min) },
-      { label: 'Max value', value: MathUtil.safeMax(this.max) },
+      { label: 'Min value', value: isMeasured(chartMin) ? chartMin : 'missing' },
+      { label: 'Max value', value: isMeasured(chartMax) ? chartMax : 'missing' },
     ];
 
     if (this.points.length > 1) {
--- a/src/model/bar.ts
+++ b/src/model/bar.ts
@@ -9,6 +9,37 @@ import { Svg } from '@util/svg';
 import { AbstractTrace } from './abstract';
 import { MovableGrid } from './movable';
 
+/**
+ * Reads one bar's magnitude, keeping a gap distinguishable from a zero.
+ *
+ * A chart library reports a missing bar as `null` — plotly renders a
+ * zero-height rect and keeps the point, so the gap arrives here rather than
+ * being dropped. `Number(null)` is `0`, which makes an absent bar
+ * indistinguishable from one genuinely measured at zero: it sounds like a real
+ * low point, it can be reached as the row's minimum, and it pulls the range
+ * that every other bar's pitch is scaled against. `NaN` keeps it absent, which
+ * is what the text layer already announces it as.
+ *
+ * @param raw - The value from the point, on whichever axis carries magnitude
+ * @returns The magnitude, or `NaN` when the bar is a gap
+ */
+function toBarValue(raw: string | number | null | undefined): number {
+  if (raw === null || raw === undefined || raw === '') {
+    return Number.NaN;
+  }
+  return Number(raw);
+}
+
+/**
+ * Reports whether a bar value is a real measurement rather than a gap.
+ *
+ * @param value - A magnitude from `barValues`
+ * @returns True when the value can take part in a range or a comparison
+ */
+function isMeasured(value: number): boolean {
+  return Number.isFinite(value);
+}
+
 export abstract class AbstractBarPlot<T extends BarPoint> extends AbstractTrace {
   protected readonly movable: Movable;
 
@@ -29,13 +60,15 @@ export abstract class AbstractBarPlot<T extends BarPoint> extends AbstractTrace 
 
     this.barValues = points.map(row =>
       row.map(point =>
-        this.orientation === Orientation.VERTICAL
-          ? Number(point.y)
-          : Number(point.x),
+        toBarValue(
+          this.orientation === Orientation.VERTICAL ? point.y : point.x,
+        ),
       ),
     );
-    this.min = this.barValues.map(row => MathUtil.safeMin(row));
-    this.max = this.barValues.map(row => MathUtil.safeMax(row));
+    // A gap is not a measurement, so it must not set the row's range. Left in,
+    // it drags the minimum to 0 and every other bar's pitch along with it.
+    this.min = this.barValues.map(row => MathUtil.safeMin(row.filter(isMeasured)));
+    this.max = this.barValues.map(row => MathUtil.safeMax(row.filter(isMeasured)));
     this.highlightValues = this.mapToSvgElements(layer.selectors as string);
     this.movable = new MovableGrid<T>(this.points);
   }
--- a/src/model/bar.ts
+++ b/src/model/bar.ts
@@ -33,10 +33,13 @@ function toBarValue(raw: string | number | null | undefined): number {
 /**
  * Reports whether a bar value is a real measurement rather than a gap.
  *
+ * Exported for `SegmentedTrace`, which builds a summary row from these values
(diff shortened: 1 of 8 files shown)


jooyoungseo:
You found a real one, and it was worse than the bug this branch started on. Fixed in 0c1f5f8. I reproduced it before touching anything, since a claim this consequential deserved measuring rather than trusting:

```
stacked, two series, gap in East at Q1:
  freq.min = NaN   freq.max = NaN   raw = NaN
  barValues rows = [[NaN, 80], [90, 70], [NaN, 150]]   <- Total row poisoned
```

Exactly the chain you described. `createSummaryLevel` summed the gap straight in, `MathUtil.minMax` seeded on the NaN, and `safeMin`/`safeMax` over `this.min`/`this.max` use `Math.min`/`Math.max`, which propagate it. One missing segment gave **every real bar on the chart** a non-finite frequency — the bug I opened this branch to fix, reintroduced through a path the fix did not reach. Thank you for catching it before merge.

The fix sums the measured segments only, and marks a category a gap in the total only when every segment in it is one. `summaryValues` is filtered before its range is taken. Q1 in the repro now totals 90 — the one real segment — rather than NaN.

**The all-gap row is fixed too**, which Copilot raised independently on `bar.ts:71`. Both `BarTrace.getExtremaTargets` and `SegmentedTrace.getExtremaTargets` now return no targets when the row's range is non-finite. You were both right that `indexOf(±Infinity)` returns `-1` and the target would have carried `pointIndex: -1`; worth noting this is a regression my first commit introduced, since an all-gap row used to be all zeros and `indexOf(0)` found index 0.

Four new tests, three of which fail against the previous commit on this branch:

- the chart's pitch range stays finite with a gapped segment
- a category with a gap totals its measured segments
- a category with no measured segment stays a gap in the total
- an all-gap row offers no extreme

Plus a control — a stack with no gaps still totals `[210, 150]` exactly as before.

On the `mapToSvgElements` note: agreed it should stay as it is. Its `Number(point.y) !== 0` is asking a different question — whether the chart library omitted a zero-height rect from the DOM — and a gap renders as a rect that *is* present, so routing it through `isMeasured` would change DOM alignment rather than clarify it. Left alone deliberately.

`Histogram` also extends `AbstractBarPlot`, but it has no summary row and its points come from bin counts rather than nullable series data, so it has no equivalent path.

Suite green: 1401 unit + 57 esm, plus `lint:fix`, `type-check`, and `build`.

---

jooyoungseo:
Both confirmed, both fixed in 6f5f4b4. The second one was the more serious of the two, and you were right to flag it as a maybe rather than assert it.

**1. `description` — confirmed.** `segmented.ts:228-229` had the unguarded `safeMin`/`safeMax`. Rather than copy the guard into the override, I moved it into a shared `rangeStats()` on `AbstractBarPlot` that both call, so the subclass cannot drift back the same way a third time.

**2. `mapToSvgElements` — confirmed, and it is a regression this branch introduced.** All six `skipZeros` checks read `barValues` directly, where the base class recomputes `Number(point.y)`. That is why the base class stayed correct and the subclass did not: `Number(null)` is still `0` there, while `barValues` now holds `NaN`.

The right resolution is to keep the *old* reading, which is what I did — a gap counts as possibly-omitted exactly as a zero does. This path decides which rendered element belongs to which datum; it is not about what a value means, and my change had no business altering it.

The two DOM tests drive a real five-rect document rather than reasoning about it. The default order walks each category through the series in reverse, so with the gap standing aside the row reads `West e0, East e1 | West e2, East gap | West e3, East e4`. Take a rect for the gap instead and East at Q3 runs past the end of the list — **the cell the user is standing on highlights nothing at all**. That is the failure your note predicted, and it is what the test catches.

One wrinkle worth recording for whoever writes the next DOM test here: `Svg.selectAllElements` inserts hidden clones, so a second trace built against the same document sees more rects than the chart has. The helper installs a fresh document per call, and the comment says why.

All three new tests fail against the previous commit. Suite green: 1408 unit + 57 esm, plus `lint:fix`, `type-check`, and `build`.

On the `AudioService` guard not being bar-scoped: agreed, and deliberate. A non-finite `freq.raw` from any trace would have reached `MathUtil.interpolate` and produced a non-finite oscillator frequency, so the guard is right wherever it comes from. No other trace can currently produce one, so nothing else changes behaviour today.

---

jooyoungseo:
This one does not reproduce — I drove it in a browser before concluding, since the reasoning was specific enough to be worth checking rather than arguing with. Pinned in 25a70a7 regardless.

A stacked chart where Q1 has no measured segment in either series, navigating up to the summary row:

```
Quarter is Q1, Revenue is missing, Level is East
Quarter is Q1, Revenue is missing, Level is West
Quarter is Q1, Revenue is missing, Level is Sum     <- the Total row
```

`missing`, not `NaN`. The step the analysis stops one short of: `TextService.formatSingleValue` has no missing-value mapping of its own, correctly, but it delegates to `FormatterService` whenever one exists — and `FormatterService.resolveAxisFormat` wraps every formatter in `FormatUtil.wrapFormat`, which maps `NaN` to `missing` exactly as it maps `null`. So the synthetic Total's `NaN` lands on the same path a raw `null` does.

The `String(value)` fallback you found is real but only reachable when `TextService` has no formatter at all, which the controller never constructs it without. That is a test/headless shape, not a user-facing one.

Two tests added anyway, because you were right that nothing asserted the summary row's text: a Total with no measured segment is non-finite (which is what routes it to `missing`), and a Total that does have measured segments still reports their sum — 90 for a category where one series is present and the other is a gap.

Suite green: 1410 unit + 57 esm.

Agreed on the rest of your read, including that `isDomOmittable` keeping the old gap-equals-zero behaviour is the right call for a path that aligns elements rather than interpreting values.

---

jooyoungseo:
Right on `Histogram`, and worth taking rather than deferring — b5f7c8a.

`histogram.ts:27-28` still called `safeMin`/`safeMax` directly. That is the third `AbstractBarPlot` subclass, and it overrides the stats block exactly the way `SegmentedTrace` does. Having introduced `rangeStats()` specifically so an override could not drift back to announcing an infinity, and then applied it to one of the two subclasses that override it, is worse than not having factored it out at all — the helper's whole justification is that the next reader shouldn't have to notice.

Practical severity is nil today, as you say: plotly's precomputed bin counts are always real numbers, so a gap cannot reach a histogram row. The test says so in as many words, so nobody deletes it later thinking it guards something reachable.

Also took the whitespace nit. `Number('  ')` is `0` as surely as `Number('')` is, so the rule is now "a blank cell" rather than "an empty string" — one condition, and it makes the docstring honest. And added the note that the `AudioService` guard is deliberately not bar-scoped, since you're right that the PR description frames it in bar terms and the placement is broader on purpose.

Both new tests fail against the previous commit. Suite green: 1413 unit + 57 esm.

---

jooyoungseo:
You caught me overstating a symptom, and you're right. Checked before conceding:

`Description.tsx:85` — `isDisplayable` returns `Number.isFinite(value)` for numbers, and `:214` filters the stats list through it. Nothing else in the codebase consumes `DescriptionState.stats`. So pre-fix, the Min and Max rows for an all-gap chart **silently vanished from the dialog**; the word `Infinity` never reached anyone.

My commit message and the PR description both said the chart "described its minimum as `Infinity`". That was wrong. The fix is still worth having — a row reading `missing` tells a blind reader the data is absent, where a row that isn't there tells them nothing and looks identical to a chart type that has no such stat — but the severity was "the stat quietly goes away", not "the user is told Infinity". I've corrected the PR description and left the correction visible rather than editing it away, since the commit message is already pushed and cannot be quietly fixed.

On the other two:

**The `isMeasured` DRY note** — I'd rather not. Both call sites are literally `Number.isFinite`, and `isMeasured` earns its name inside `bar.ts` by meaning "this cell is a measurement rather than a gap". Importing it into `AudioService` would make a service that deliberately knows nothing about trace types depend on `@model/bar` to ask whether a number is finite, and would imply the two checks must move together when they needn't. The precedent you cite (`text.ts` importing `focusedSubplotTitle`) is a domain concept with no stdlib equivalent; this isn't. If the bar-side predicate ever grows past `Number.isFinite`, that is the moment to reconsider.

**The array `raw` path** — agreed, and out of scope. `freq.raw` is `number[]` for box, violin, and multi-line intersections, and a non-finite element inside one would reach `MathUtil.interpolate` unguarded. No bar trace produces an array `raw`, so nothing here changes that, and none of those types can currently emit a non-finite element. Worth its own issue if that ever stops being true rather than a speculative guard now.

---

PR #730, opened 2026-08-04
refactor(format): drop the formatter gate that could never close

### LINKED ISSUE #729: fix: hasCustomFormatter always returns true, and its callers' fallback is unreachable
# Bug Report

## Description

`FormatterService.hasCustomFormatter` decides whether a layer's axis has an author-supplied `AxisFormat` by comparing the stored function against the `defaultFormat` export:

https://github.com/xability/maidr/blob/main/src/service/formatter.ts#L126-L133

But `resolveAxisFormat` never stores `defaultFormat` itself — it stores `FormatUtil.wrapFormat(...)`, a fresh closure built on every call:

```ts
private resolveAxisFormat(axisFormat?: AxisFormat): FormatFunction {
  const baseFormat = FormatUtil.resolveFormat(axisFormat);
  return FormatUtil.wrapFormat(baseFormat);
}
```

So the reference comparison can never match, and the method reports `true` for every layer it knows about, formatted or not.

## Steps to Reproduce

```ts
const service = new FormatterService({
  id: 'chart',
  subplots: [[{ layers: [{
    id: 'L',
    type: TraceType.BAR,
    axes: { x: { label: 'X' }, y: { label: 'Y' } },   // no format anywhere
    data: [],
  }] }]],
});

service.hasCustomFormatter('L', 'y');
```

## Actual Behavior

```
hasCustomFormatter(layer with no format) = true
hasCustomFormatter(unknown layer)        = false
```

Only the unknown-layer path — which returns early before the comparison — answers `false`.

## Expected Behavior

Either the check reports what it claims, or it goes away and its callers stop pretending to branch on it.

## Impact

Both call sites are in `goToExtremaViewModel.ts`, and both document a fallback that never runs:

- `formatTargetLabels` (:225) — *"When no custom formatter is configured the labels pass through unchanged."*
- `getAvailableXValueOptions` (:284) — *"pass through unchanged when the active layer has no custom x formatter."*

Until #727 this was invisible: `defaultFormat` was `` `${value}` ``, so running a value through it or skipping it produced the same string. #727 gave `defaultFormat` real behaviour — it rounds to two decimals — so the branch now decides something, and the branch that runs is not the one the comments describe. The outcome happens to be the desirable one (Go To Extrema labels get the same rounding as the announcement, so the two surfaces agree), which is exactly why this needs settling deliberately rather than being left to a comparison that has never worked.

Note also that `test/state/viewModel/goToExtremaViewModel.test.ts` covers the unreachable branch by stubbing `hasCustomFormatter: () => false`. That test passes and proves nothing about production, which is the second reason this went unnoticed.

## Environment

- Version: `main` at 3fbd1b9
- Not browser- or chart-library-specific

jooyoungseo:
# Pull Request

## Description

`FormatterService.hasCustomFormatter` asked whether a layer's axis had an author-supplied `AxisFormat` by comparing the stored function against the `defaultFormat` export. But `resolveAxisFormat` never stores `defaultFormat` — it stores `FormatUtil.wrapFormat(...)`, a fresh closure built on every call. The comparison could not match, so the method answered `true` for every layer the service knew about.

Measured before changing anything:

```
hasCustomFormatter(layer with no format) = true
hasCustomFormatter(unknown layer)        = false
```

Only the unknown-layer path answers `false`, and it returns early before the comparison ever happens.

## Related Issues

Fixes #729

## Changes Made

Both callers live in `goToExtremaViewModel.ts`, and both documented a pass-through that never ran:

- `formatTargetLabels` — *"When no custom formatter is configured the labels pass through unchanged."*
- `getAvailableXValueOptions` — *"pass through unchanged when the active layer has no custom x formatter."*

**Removed the gate rather than repairing it.** Repairing it would mean making the false branch reachable — and that branch is the wrong behaviour now. Until #727, `defaultFormat` was `` `${value}` ``, so calling it and skipping it produced the same string and the bug was invisible. #727 gave it real behaviour: it rounds a long float to two decimals. Running it is what keeps a Go To Extrema label in step with the announcement for the same point; skipping it would put `57.14285714285714` in the dialog while the announcement says `57.14`.

So the callers now format whenever they have a formatter and a layer id to look it up by — which is what has always happened in practice. `hasCustomFormatter` had no other callers and `FormatterService` is constructed only by `controller.ts`, so nothing public changes.

Two things review asked for on the way, both inside functions this PR already rewrites:

- The `this.formatter!` non-null assertions in both callers are gone, replaced by a narrowed local `const` that holds across the closure — `.claude/rules/typescript.md` asks for exactly that.
- `getAvailableXValueOptions`'s JSDoc still promised a fallback "when no custom x formatter is configured" after that stopped being what decides it. It now names what does: no formatter injected, or no layer id.

Not applicable.

## Checklist

## Additional Notes

**Behaviour is unchanged for every case production reaches.** The gate's false branch required `hasCustomFormatter` to answer false, which it only did for a layer id the service has never seen. Both callers source that id from the same place — `getAvailableXValueOptions` through `activeLayerId()`, `formatTargetLabels` from `state.layerId` at the `toggle()` call site — so neither can be handed an id the service has not seen. If one ever were, the label now rounds instead of being stringified raw, which is the more consistent of the two.

**A test went with the method.** `test/state/viewModel/goToExtremaViewModel.test.ts` had a case stubbing `hasCustomFormatter: () => false` and asserting the raw-label path. It passed, and proved nothing — it pinned a branch production cannot enter, which is the second reason this went unnoticed for so long.

**Six tests in its place**, all built on a real `FormatterService` rather than a stub's idea of one:

- `getAvailableXValueOptions` on a layer with no `AxisFormat` — `57.14285714285714` becomes `57.14`, while `2` and `3` come through untouched.
- `formatTargetLabels`, reached through `toggle()` — a long float rounds in the label (`Max Bar at 57.14285714285714` → `Max Bar at 57.14`), a value the formatter leaves alone keeps its label (`Q1`), and an author-supplied `fixed`/4 format still wins (`57.1429`). This caller had the same dead branch removed and was asserted nowhere until the second commit: its only prior coverage was the audio-cue test, which passes no formatter and no targets.
- In `test/service/formatter.test.ts`, the property that makes unconditional formatting safe: a configured and an unconfigured axis both produce a sensible string, so there is nothing a caller would need to branch around.

**Reading the results against `main`:** six cases in the ViewModel file fail there, but only the behavioural ones are evidence. The rest fail because the stubs no longer define `hasCustomFormatter` and `main` still calls it.

**Noticed, not changed:** `getFormatter` returns the bare `defaultFormat` for an unknown layer rather than a `wrapFormat`-wrapped one, so that path renders `NaN` as `"NaN"` instead of `"missing"`. It predates this change and is unreachable through the callers above; worth its own issue if anyone can find a way in.

Verified locally: `npm run lint:fix`, `npm run type-check`, `npm run build`, `npm test` (1483 unit + 57 esm, all passing).

### DIFF
--- a/src/service/formatter.ts
+++ b/src/service/formatter.ts
@@ -117,22 +117,6 @@ export class FormatterService implements Disposable {
   }
 
   /**
-   * Checks if a layer has a custom formatter for the specified axis.
-   *
-   * @param layerId - The ID of the layer
-   * @param axis - The axis type ('x', 'y', or 'z')
-   * @returns True if a custom formatter is configured
-   */
-  public hasCustomFormatter(layerId: string, axis: AxisType): boolean {
-    const layerFormatters = this.formatters.get(layerId);
-    if (!layerFormatters) {
-      return false;
-    }
-    // Check if the formatter is not the default
-    return layerFormatters[axis] !== defaultFormat;
-  }
-
-  /**
    * Formats a value (single or array) using the formatter for the specified layer and axis.
    *
    * This is the primary method for formatting values in the application.
--- a/src/state/viewModel/goToExtremaViewModel.ts
+++ b/src/state/viewModel/goToExtremaViewModel.ts
@@ -220,10 +220,15 @@ export class GoToExtremaViewModel extends AbstractViewModel<GoToExtremaState> {
 
   /**
    * Format extrema target labels by replacing raw xValues with formatted ones.
-   * When no custom formatter is configured the labels pass through unchanged.
+   *
+   * Every layer is formatted, not only those with an author-supplied
+   * `AxisFormat`: the default formatter rounds a long float to two decimals,
+   * which is what the announcement says, and a dialog label that disagreed with
+   * the announcement for the same point would be worse than either. A value the
+   * formatter leaves alone is detected below and passes through untouched.
    */
   private formatTargetLabels(targets: ExtremaTarget[], layerId: string): ExtremaTarget[] {
-    if (!this.formatter || !this.formatter.hasCustomFormatter(layerId, 'x')) {
+    if (!this.formatter) {
       return targets;
     }
 
@@ -279,9 +284,10 @@ export class GoToExtremaViewModel extends AbstractViewModel<GoToExtremaState> {
     }
 
     const layerId = this.activeLayerId();
-    // Same gate the extrema target labels use (formatTargetLabels): pass through
-    // unchanged when the active layer has no custom x formatter.
-    if (!this.formatter || layerId === null || !this.formatter.hasCustomFormatter(layerId, 'x')) {
+    // Same rule the extrema target labels follow (formatTargetLabels): format
+    // whenever there is a formatter and a layer to look it up by, so these
+    // labels round the way the announcement does.
+    if (!this.formatter || layerId === null) {
       return rawValues.map(value => ({ value, label: String(value) }));
     }
 
--- a/src/state/viewModel/goToExtremaViewModel.ts
+++ b/src/state/viewModel/goToExtremaViewModel.ts
@@ -228,7 +228,8 @@ export class GoToExtremaViewModel extends AbstractViewModel<GoToExtremaState> {
    * formatter leaves alone is detected below and passes through untouched.
    */
   private formatTargetLabels(targets: ExtremaTarget[], layerId: string): ExtremaTarget[] {
-    if (!this.formatter) {
+    const formatter = this.formatter;
+    if (!formatter) {
       return targets;
     }
 
@@ -236,7 +237,7 @@ export class GoToExtremaViewModel extends AbstractViewModel<GoToExtremaState> {
       if (target.xValue === undefined) {
         return target;
       }
-      const formatted = this.formatter!.formatSingleValue(target.xValue, layerId, 'x');
+      const formatted = formatter.formatSingleValue(target.xValue, layerId, 'x');
       const raw = String(target.xValue);
       if (formatted === raw) {
         return target;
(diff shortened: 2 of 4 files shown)


jooyoungseo:
All four points taken — 6eadf4a.

**The coverage gap was the real one.** I changed two callers and asserted one. `formatTargetLabels` only ever reached the test suite through the audio-cue case, which passes no formatter and `getExtremaTargets() => []`, so the branch removal was pinned nowhere — exactly the failure the replaced test embodied, repeated on the other method. Three cases now drive it through `toggle()` with a real `FormatterService`:

- a long float rounds for a layer with no `AxisFormat` (`Max Bar at 57.14285714285714` → `Max Bar at 57.14`)
- a value the formatter leaves alone keeps its label untouched (`Q1`)
- an author-supplied format still wins on the same path (`57.1429` with `fixed`/4)

**The unknown-`layerId` argument does hold for both**, and you were right that the write-up only covered one. `formatTargetLabels` takes its `layerId` from `state.layerId` at the `toggle()` call site — the same `TraceState` source `activeLayerId()` reads — so neither caller can be handed an id the service has not seen.

**Non-null assertions gone.** A local `const formatter = this.formatter` keeps the narrowing across the closure, which is what `.claude/rules/typescript.md` asks for ("avoid non-null assertions; narrow instead"). Those predate this PR, but both are inside functions it already rewrites, so leaving them would have been a choice.

**The stale JSDoc was mine.** `getAvailableXValueOptions` still promised a fallback "when no custom x formatter is configured" after that stopped being what decides it. It now says what actually decides: no formatter injected, or no layer id.

One note on reading the test results: against `main`, six cases in that file fail rather than three. Only the behavioural ones are meaningful — the others fail because the stubs no longer define `hasCustomFormatter` and `main` still calls it. Worth knowing before anyone takes the count as evidence.

Suite green: 1483 unit + 57 esm, plus `lint:fix`, `type-check`, and `build`.

---

PR #746, opened 2026-08-05
feat(adapters): bind a step chart as a step trace, not a line

### LINKED ISSUE #744: feat(plotly): map line.shape to the step trace type
A Plotly step chart arrives as an ordinary line. `StepTrace` landed in #723 and works through the maidr JSON path, but the Plotly adapter has no way to reach it, so a chart Plotly drew as stairs is described as if its values were interpolated between samples.

## Reproduction

```html
<script src="https://cdnjs.cloudflare.com/ajax/libs/plotly.js/3.1.1/plotly.min.js"></script>
<script src="https://cdn.jsdelivr.net/npm/maidr@3.75.1/dist/maidr.js"></script>
<div id="chart"></div>
<script>
  Plotly.newPlot('chart', [{
    type: 'scatter',
    mode: 'lines',
    line: { shape: 'hv' },
    x: [1, 2, 3, 4],
    y: [10, 14, 9, 17],
  }], {
    xaxis: { title: { text: 'Time' } },
    yaxis: { title: { text: 'Level' } },
  });
</script>
```

Announced as `single line`, and navigation reads
`Time is 1, Level is 10, Group is trace 0`.

Verified the chart really is a step chart on the same page — `_fullData[0]` is
`{ type: 'scatter', mode: 'lines', line: { shape: 'hv' } }`, which Plotly draws
piecewise-constant. Same result for `shape: 'vh'` and `shape: 'hvh'`.

## Where

`mapScatterMode` decides the type from `mode` alone, so `line.shape` never
reaches it:

```ts
function mapScatterMode(mode?: string): TraceType {
  if (!mode)
    return TraceType.SCATTER;
  if (mode.includes('lines'))
    return TraceType.LINE;
  if (mode.includes('markers'))
    return TraceType.SCATTER;
  return TraceType.SCATTER;
}
```

It is called as `mapScatterMode(trace.mode)` from `mapTraceType(trace)`, which
does have the whole trace in hand.

## What is lost

Not just a label. `StepTrace` exists because piecewise-constant data is
navigated and described differently, and all of it is currently unreachable
from Plotly:

- the **transition rotor**, which restricts movement to the points where the
  level actually changes — the interesting points in a step chart, and tedious
  to find one sample at a time
- **ordinal level names**, so a reader hears "REM" rather than "3"
- the **step convention** itself, which `STEP_DIRECTION_LABEL` spells out as
  "value holds until the next x value, then jumps" and so on

That last one is the part a sighted reader gets for free from the shape of the
stairs, and the one most likely to be misread: told it is a line, a reader
reasonably assumes the value moved gradually between samples, when it did not.

## Mapping

Plotly's `line.shape` is the same idea as `StepDirection`, so this looks close
to one-to-one:

| `line.shape` | meaning | `StepDirection` |
|---|---|---|
| `hv` | hold, then jump at the next x | `hv` |
| `vh` | jump at this x, then hold | `vh` |
| `hvh`, `vhv` | jump midway between x values | `mid` |
| `linear`, `spline`, absent | interpolated | not a step — stays `LINE` |

Passing the trace rather than just `mode` would be enough for the lookup.
Worth deciding explicitly what `hvh` and `vhv` should be: both jump mid-interval
but differ in where the horizontal segments sit, and `mid` may or may not be the
right reading for each.

## Context

Found while re-testing every MAIDR trace type against Plotly auto-detection
after #722 and #724 landed. Those two both check out — a violin now binds as two
layers and navigates (`Group is A, minimum Value is 1`), a `barnorm` chart now
announces `Share is 57.14` against the 57.142857 Plotly drew, and an unsupported
chart no longer re-warns in a loop. Step is the one type that a user can draw in
Plotly and have described as something it is not.

jooyoungseo:
# Pull Request

## Description

`StepTrace` landed in #723 and works through the maidr JSON path, but no adapter that reads a chart library's own settings could reach it. A chart the library drew as stairs was described as a line — telling a reader the value moved gradually between samples when it holds and then jumps — and the transition rotor, the ordinal level names, and the spoken step convention were all unreachable.

This fixes that in every adapter that introspects the library and could therefore be wrong: Plotly (the reported case), Chart.js, Highcharts, Vega-Lite, Victory, and amCharts.

Each mapping was read off the drawing code rather than the docs, since the naming does not line up across libraries — Chart.js `stepped: 'after'` and d3's `curveStepAfter` are opposite conventions.

| Library | Setting | → `stepDirection` |
|---|---|---|
| Plotly | `line.shape: 'hv' \| 'vh' \| 'hvh'` | `hv` / `vh` / `mid` |
| Plotly | `line.shape: 'vhv'` | step, no direction (see below) |
| Chart.js | `stepped: 'before' \| 'after' \| 'middle'` (or `true`) | `hv` / `vh` / `mid` |
| Highcharts | `step: 'left' \| 'center' \| 'right'` (or `true`) | `hv` / `mid` / `vh` |
| Vega-Lite | `interpolate: 'step-after' \| 'step-before' \| 'step'` | `hv` / `vh` / `mid` |
| Victory | `interpolation="stepAfter" \| "stepBefore" \| "step"` | `hv` / `vh` / `mid` |
| amCharts | `StepLineSeries` | step, no direction |

**The `vhv` decision the issue asked for.** `hvh` is `mid` — it holds each sample's own value and jumps at the midpoint, exactly what `mid` describes. `vhv` is not: it rises at `x[i]`, runs flat at the *mean* of `y[i]` and `y[i+1]`, then rises again at `x[i+1]`. No MAIDR convention describes that, and `mid` would actively mislead, promising a jump midway between x values where `vhv` jumps at the x values themselves. It binds as a step — the data is piecewise constant and the transition rotor is the navigation it wants — with no direction named, which is what `stepDirection` being optional is for.

amCharts is the same reasoning from the other end: it positions the staircase from the axis cell rather than reporting a convention, so naming one would be a guess. This follows the precedent the AnyChart step layers already set.

**Splitting.** A layer announces one `stepDirection` for all of its series, so a chart mixing conventions splits into one layer per convention rather than merging and describing one of them wrongly. In Chart.js that also meant each layer now reports which datasets back its rows, so highlights on a mixed chart still route to the right series.

**Not changed, and why.** Recharts, D3 and Google Charts are declarative: the caller tells MAIDR the chart type, so MAIDR reports exactly what it was told and nothing is currently wrong. Supporting step there means adding a new value to a public config API, which is a separate change. Frappe Charts has no step option at all.

## Related Issues

Closes #744

## Changes Made

- **`adapters/plotly`** — pass the whole trace to `mapScatterMode` and read `line.shape`; group step traces by convention; extend the step selector to reuse the line one.
- **`adapters/chartjs`** — read `dataset.stepped`, falling back to `elements.line` as Chart.js resolves it; split line datasets into a line layer and one step layer per convention; thread per-layer dataset indices through both the single-subplot and axis-stacked-panel paths so highlight routing follows.
- **`adapters/highcharts`** — read `series.options.step` (which Highcharts resolves from `plotOptions`), including the legacy boolean.
- **`adapters/vegalite`** — read `interpolate` off the mark def for `line` and `area`; sibling-layer coalescing now requires matching type *and* direction.
- **`adapters/victory`** — read the `interpolation` prop; a function-valued one (a custom curve factory) stays a line, since nothing about it is inspectable.
- **`adapters/amcharts`** — classify `StepLineSeries` as its own kind, merge those series into a step layer, and give the navigation map its own series list so a chart with both keeps its highlights aligned.
- **docs** — a step row in each adapter's supported-type table, plus notes for the two adapters whose mapping needs one.
- **`examples/plotly-step.html`** — the three Plotly shapes side by side.

## Checklist

## Verification

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass — 1532 unit tests across 116 suites plus the ESM project, with 47 new cases covering the six adapters.

**`E2E (chromium)` runs in CI on this PR** and has been green on each commit. (An earlier version of this description said the E2E suite had not been run; that was wrong and is corrected here. What could not run was the suite *in my local environment*, where the browser cannot reach external hosts through the egress proxy — `curl` can, Chromium gets `ERR_CONNECTION_RESET`. Serving the built bundles over localhost with the chart libraries vendored works fine, which is how the checks below were done.)

Beyond CI and the unit tests, the two riskiest parts were verified against real renders:

**Chart.js direction mapping — measured from rendered pixels**, since it is the one mapping with no well-documented external convention to check against. A two-point dataset with everything but the line switched off, reading back the canvas `ImageData` to find where the line sits between the points:

| `stepped` | just after p0 | before mid | after mid | just before p1 | → |
|---|---|---|---|---|---|
| `before` | old | old | old | old | **`hv`** |
| `after` | **new** | new | new | new | **`vh`** |
| `middle` | old | old | **new** | new | **`mid`** |
| `true` | old | old | old | old | **`hv`** |
| `false` | between | between | between | between | not a step |

**Chart.js highlight routing — driven in Chromium** against the built `dist/chartjs.js` with a deliberately interleaved chart (`[Trend, Before(step), After(step), Trend 2]`), reading `chart.getActiveElements()`: the line layer's rows route to datasets **0** and **3**, and the two step layers to **1** and **2**. Row 1 landing on dataset 3 is the case that would break without the dataset-index threading.

**Plotly end to end** against the built `dist/maidr.js`: all three shapes announce `type: step` (a plain line still announces `multiline`); the Transitions rotor jumps 8 → 10 → 13 → 15 on a held series, exactly the level changes; and the spoken convention differs correctly per shape — `hv` "holds until the next x value, then jumps", `vh` "jumps at the current x value, then holds", `hvh` "jumps midway between x values".

---

https://claude.ai/code/session_01Hv1rkb9z7AVZescynitfGd

### DIFF
--- a/docs/amcharts.md
+++ b/docs/amcharts.md
@@ -93,8 +93,11 @@ amCharts 5 renders to an HTML5 `<canvas>`, so there are no per-element SVG nodes
 | Stacked Bar | multiple `ColumnSeries` | `stacked: true` |
 | 100% Stacked (Normalized) | multiple `ColumnSeries` | `stacked: true` + `valueYShow: "valueYTotalPercent"` |
 | Line (single & multi-series) | `LineSeries` | line series class |
+| Step (single & multi-series) | `StepLineSeries` | step-line series class |
 | Histogram | `ColumnSeries` | value X axis + `openValueXField` bin edges |
 | Heatmap | `ColumnSeries` | category X **and** category Y axes + `value` field |
+
+A `StepLineSeries` is piecewise constant — the value is held and then jumps — so it maps to MAIDR's step trace rather than to a line, and is announced and navigated as a step plot. amCharts positions the staircase from the axis cell rather than reporting a step convention, so the adapter emits no `stepDirection` and MAIDR's description does not name one.
 
 > Box plots, candlestick, scatter, violin, and smooth/regression layers are **not** supported by the amCharts binder. amCharts 5 has no dedicated scatter or box series, and there is no reliable runtime signal to distinguish a scatter (hidden-stroke `LineSeries`) from a normal line chart.
 
--- a/docs/chartjs.md
+++ b/docs/chartjs.md
@@ -73,6 +73,7 @@ MAIDR's Chart.js adapter is a standard Chart.js plugin:
 | Stacked Bar | `'bar'` with `scales.x.stacked` / `scales.y.stacked` | — | [Stacked bar](examples.html) |
 | Dodged Bar | `'bar'` with multiple datasets (no stacking) | — | [Dodged bar](examples.html) |
 | Line | `'line'` | — | [Line chart](examples.html) |
+| Step | `'line'` with `stepped` on the dataset (or `elements.line`) | — | [Line chart](examples.html) |
 | Scatter | `'scatter'` | — | [Scatter plot](examples.html) |
 | Box Plot | `'boxplot'` | `@sgratzl/chartjs-chart-boxplot` | [Box plot](examples.html) |
 | Candlestick | `'candlestick'` | `chartjs-chart-financial` + a date adapter | [Candlestick](examples.html) |
--- a/docs/highcharts.md
+++ b/docs/highcharts.md
@@ -105,7 +105,7 @@ import { createHighchartsSync, highchartsToMaidr } from 'maidr/highcharts';
 |------------|---------------------------|---------|
 | Bar | `bar`, `column` | [highcharts-bar.html](highcharts-bar.html) |
 | Line | `line`, `spline`, `area`, `areaspline` | [highcharts-line.html](highcharts-line.html) |
-| Step | any of the above + `step: 'left' \| 'center' \| 'right'` | [highcharts-line.html](highcharts-line.html) |
+| Step | `line`, `spline`, `area`, `areaspline` + `step: 'left' \| 'center' \| 'right'` | [highcharts-line.html](highcharts-line.html) |
 | Scatter | `scatter` | [highcharts-scatter.html](highcharts-scatter.html) |
 | Box Plot | `boxplot` | [highcharts-box.html](highcharts-box.html) |
 | Heatmap | `heatmap` (requires `modules/heatmap.js`) | [highcharts-heatmap.html](highcharts-heatmap.html) |
--- a/docs/highcharts.md
+++ b/docs/highcharts.md
@@ -105,6 +105,7 @@ import { createHighchartsSync, highchartsToMaidr } from 'maidr/highcharts';
 |------------|---------------------------|---------|
 | Bar | `bar`, `column` | [highcharts-bar.html](highcharts-bar.html) |
 | Line | `line`, `spline`, `area`, `areaspline` | [highcharts-line.html](highcharts-line.html) |
+| Step | any of the above + `step: 'left' \| 'center' \| 'right'` | [highcharts-line.html](highcharts-line.html) |
 | Scatter | `scatter` | [highcharts-scatter.html](highcharts-scatter.html) |
 | Box Plot | `boxplot` | [highcharts-box.html](highcharts-box.html) |
 | Heatmap | `heatmap` (requires `modules/heatmap.js`) | [highcharts-heatmap.html](highcharts-heatmap.html) |
--- a/docs/plotly.md
+++ b/docs/plotly.md
@@ -63,6 +63,7 @@ For dynamically-created charts (SPAs, notebooks), a `MutationObserver` watches f
 | Bar | `type: 'bar'` | [Bar chart](examples.html) |
 | Scatter | `type: 'scatter'`, `mode: 'markers'` | [Scatter plot](examples.html) |
 | Line | `type: 'scatter'`, `mode: 'lines'` | [Line chart](examples.html) |
+| Step | `type: 'scatter'`, `mode: 'lines'`, `line: { shape: 'hv' \| 'vh' \| 'hvh' \| 'vhv' }` | [Step chart](examples.html) |
 | Box Plot | `type: 'box'` | [Box plot](examples.html) |
 | Violin Plot | `type: 'violin'` | [Violin plot](examples.html) |
 | Heatmap | `type: 'heatmap'` | [Heatmap](examples.html) |
@@ -71,6 +72,17 @@ For dynamically-created charts (SPAs, notebooks), a `MutationObserver` watches f
 | Grouped Bar | `barmode: 'group'` + multiple bar traces | [Grouped bar](examples.html) |
 | Stacked Bar | `barmode: 'stack'` + multiple bar traces | [Stacked bar](examples.html) |
 | Subplots / Facets | multiple `xaxis`/`yaxis` pairs, `layout.grid`, or Plotly Express facets | [Subplots](examples.html) |
+
+**Notes on chart-type detection:**
+
+- A line trace whose `line.shape` is one of the step-wise shapes is piecewise
+  constant — the value is held and then jumps — so it maps to MAIDR's step
+  trace rather than to a line, and is announced and navigated as a step plot
+  (including the Transitions rotor, which moves only between the points where
+  the level changes). `hv` and `vh` carry across to MAIDR's `stepDirection`
+  unchanged, and `hvh` becomes `mid`. `vhv` binds as a step but names no
+  convention: its flat segments sit at the mean of the two levels rather than
+  at either one, which none of MAIDR's three conventions describes.
 
 ## Code Examples
 
--- a/docs/vegalite.md
+++ b/docs/vegalite.md
@@ -95,6 +95,7 @@ Because Vega-Lite renders **asynchronously** through `vegaEmbed()`, the adapter 
 | `bar` | `color`/`fill`, `stack: null` or `false` | Dodged (grouped) bar | [vegalite-binddodged.html](https://github.com/xability/maidr/blob/main/examples/vegalite-binddodged.html) |
 | `bar` | `color`/`fill`, `stack: 'normalize'` | Normalized stacked bar | [vegalite-bindnormalized.html](https://github.com/xability/maidr/blob/main/examples/vegalite-bindnormalized.html) |
 | `line`, `area` | — | Line | [vegalite-bindline.html](https://github.com/xability/maidr/blob/main/examples/vegalite-bindline.html) |
+| `line`, `area` | `interpolate: 'step'`, `'step-before'`, `'step-after'` | Step | [vegalite-bindline.html](https://github.com/xability/maidr/blob/main/examples/vegalite-bindline.html) |
 | `point`, `circle`, `square`, `tick` | — | Scatter | [vegalite-bindscatter.html](https://github.com/xability/maidr/blob/main/examples/vegalite-bindscatter.html) |
 | `rect` | — | Heatmap | [vegalite-bindheatmap.html](https://github.com/xability/maidr/blob/main/examples/vegalite-bindheatmap.html) |
 | `boxplot` | — | Box plot (vertical & horizontal) | [vegalite-bindbox.html](https://github.com/xability/maidr/blob/main/examples/vegalite-bindbox.html) |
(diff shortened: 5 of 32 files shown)


jooyoungseo:
Acted on all three findings — pushed in 44f686a.

**1. Lockfile churn — dropped.** You were right about the cause: `npm install` regenerated it under a different npm than the project uses. `package-lock.json` is now byte-identical to the base commit.

**2. `?? 0` fallback — invariant recorded.** Added a comment noting that an out-of-range position cannot arise while an extractor fills `LocalDatasetIndices` by walking the very datasets it was handed, and why it stays a fallback rather than a throw: misrouting one highlight beats taking the chart's whole accessibility layer down.

**3. Highlight routing — verified in a real render, and it holds.** This was the right thing to push on, so I drove it in Chromium against the built `dist/chartjs.js` with a deliberately interleaved mixed chart (`[Trend, Before(step), After(step), Trend 2]`), reading `chart.getActiveElements()` — what Chart.js is actually told to highlight:

| MAIDR layer | row | announced | → Chart.js dataset |
|---|---|---|---|
| 0 (line) | 0 | `Group is Trend` | **0** |
| 0 (line) | 1 | `Group is Trend 2` | **3** |
| 1 (step, `hv`) | 0 | `Group is Before` | **1** |
| 2 (step, `vh`) | 0 | `Group is After` | **2** |

Row 1 of the line layer landing on dataset **3**, not 1, is exactly the case that would have broken without the dataset-index threading — the old "all datasets, in order" assumption would have highlighted the step series there.

While in the browser I also confirmed the Plotly path end to end against the built `dist/maidr.js`, since the issue was about what a reader actually hears:

- All three shapes announce `type: step`; a plain line still announces `multiline` (control).
- The **Transitions rotor** works from an auto-detected chart: on staffing `2,2,5,5,5,3,3,6,6`, Right jumps hour 8 → **10** → **13** → **15** — precisely the level changes, skipping the held samples.
- The **spoken convention** differs correctly per shape, which is the part of #744 most likely to mislead:
  - `hv` → "value holds until the next x value, then jumps"
  - `vh` → "value jumps at the current x value, then holds"
  - `hvh` → "value jumps midway between x values"

No page errors in any run.

One correction to the PR description while I'm here: I said E2E couldn't run here. The Playwright *suite* still can't (the browser cannot reach external hosts through this environment's egress proxy — `curl` can, Chromium gets `ERR_CONNECTION_RESET`), but serving the built bundles over localhost with the chart libraries vendored locally works fine, which is how the above was done. So the highlight-routing check you asked for before merge is done; the full E2E suite still hasn't been.

On the repeated `stepDirectionOf` + `STEP_DIRECTION_BY_*` shape — agreed it's four near-identical mappings, and I left them separate deliberately: the inputs differ in kind (a boolean union, a mark-def field, a React prop, a dataset field), the semantics are per-library rather than shared, and the direction table is the part most worth reading next to the code that uses it. Worth revisiting if a fifth arrives.

---

jooyoungseo:
Thanks — took the JSDoc note, and it turned out to be more than style preference. Pushed in 470b0a8.

I checked it empirically rather than assuming. `jsdoc/require-param` is **not** enforced here (I stripped a block and lint still passed), so both forms were legal — but the convention in these files is lopsided. Counting private helpers with a JSDoc block:

| file | with `@param` | summary only |
|---|---|---|
| `plotly/extractor.ts` | 5 | 26 |
| `highcharts/adapter.ts` | 3 | 17 |
| `victory/converters.ts` | 1 | 16 |

and the `@param` entries in each were largely the functions this PR added. So these files reserve the tag blocks for exported functions, and my private helpers were the outliers — which makes it a genuine deviation from "preserve the surrounding patterns", not a matter of taste. Dropped the tag blocks and kept every line that explains something the signature doesn't: why a function-valued Victory `interpolation` stays a line, why amCharts emits no `stepDirection`, what each library's step values actually mean.

Worth noting one interaction for future reference: `jsdoc/check-param-names` makes this all-or-nothing per function — partially documenting params is what fails, which is what pushed me into the full blocks initially.

On the other two notes:

- **`?? 0` fallback** — agreed on all counts, and thanks for restating it as a deliberate silent-failure path; that framing is exactly why it's worth having in the comment rather than only in a commit message.
- **E2E** — resolved since your review was written. **`E2E (chromium)` runs in CI on this PR** (it was in progress on the previous commit and is re-running now), so it isn't relying on manual verification. Separately I drove the highlight routing in Chromium against the built bundles — the interleaved mixed chart routes MAIDR rows to Chart.js datasets 0/3 (line layer) and 1/2 (the two step layers), which is the case that would break without the dataset-index threading. Details in [my earlier comment](https://github.com/xability/maidr/pull/746#issuecomment-5195744498).

Lint, type-check and the 537 adapter tests are green locally after the change; it touches comments only.

---

jooyoungseo:
Both notes acted on — 917c3ed.

## Chart.js: verified against a real render

You were right to single this out as the one mapping without an external convention to check against, and right that source-reading isn't the same as seeing it. So I measured it from **rendered pixels** rather than from `_steppedLineTo` again: a two-point dataset `[10, 90]`, everything except the line switched off, then reading back the canvas `ImageData` and asking at which y the line sits at four x positions between the points. That's the same thing a sighted reader sees, and it's independent of how Chart.js issues its canvas calls.

| `stepped` | just after p0 | before mid | after mid | just before p1 | → |
|---|---|---|---|---|---|
| `before` | old | old | old | old | **`hv`** |
| `after` | **new** | new | new | new | **`vh`** |
| `middle` | old | old | **new** | new | **`mid`** |
| `true` | old | old | old | old | **`hv`** (= `before`) |
| `false` | between | between | between | between | interpolated, not a step |

("old" = still `y[0]`, "new" = already `y[1]`.) All five match what the adapter emits, including the legacy `true`. The inversion relative to d3 is real and is handled the right way round: Chart.js `'after'` jumps *immediately*, which is d3's `stepBefore`.

Two notes on getting there, in case anyone repeats it: hooking `CanvasRenderingContext2D.lineTo` doesn't work well — gridlines swamp the signal, and gating on `beforeDatasetsDraw`/`afterDatasetsDraw` records nothing because the constructor only *schedules* the render. Reading pixels after two `requestAnimationFrame`s sidesteps both.

## `Partial<Record<…>>` — agreed, aligned

Not just cosmetic: `Record<string, StepDirection>` claims *every* string maps to a direction, so the lookup's type said `StepDirection` while the value is `undefined` for anything outside the handful of keys. The callers already treat it as optional, so the type was simply wrong. All four now match Plotly's `Partial`. Type-check passes; no call site needed a change, which is itself the evidence they were already handling the miss.

## Coverage gap you flagged

Fair, and now narrower than when you wrote it: **`E2E (chromium)` runs in CI on this PR** (green on the previous commit), so the adapters aren't resting on mapping tables alone. On top of that, Chart.js now has the pixel-level check above, and Plotly has both `examples/plotly-step.html` and a browser run confirming the announcement, the Transitions rotor (jumps 8 → 10 → 13 → 15 on a held series), and the three distinct spoken conventions. Highcharts, Vega-Lite and Victory remain verified against d3's documented curve semantics plus their unit tests — the d3-backed three being the ones you noted map onto a well-documented external convention.

Lint, type-check and the 537 adapter tests are green locally.

---

jooyoungseo:
Thanks — the docs catch is a good one, fixed in 43ecbd2.

**`docs/highcharts.md`** — you're right that "any of the above" sat under the Bar row as much as the Line row, and a reader could reasonably have tried `step` on a bar series. Now names the four types explicitly:

```
| Step | `line`, `spline`, `area`, `areaspline` + `step: 'left' \| 'center' \| 'right'` | … |
```

**On E2E** — this is the third review to raise it, and the cause was my own PR description, so I've fixed it at the source rather than replying a third time. `E2E (chromium)` **does** run in CI here and has been green on every commit. The original wording was wrong: what couldn't run was the suite *in my local environment*, where the browser can't reach external hosts through the egress proxy. The description now says so plainly and carries the verification details, so the next reader isn't sent down this path again.

Your specific worry — regressions on existing non-stepped multi-dataset Chart.js charts from the `LocalDatasetIndices` change — is the right one to have, and it's covered from two directions beyond CI:

- The interleaved browser run: a **plain** line layer with datasets 0 and 3 routes its MAIDR rows to 0 and 3, not 0 and 1. That is the non-stepped path going through the new code.
- `singleSubplotDatasetIndices` and `layerDatasets` both fall back to the previous behaviour when no layer declares indices, so a chart with no stepped dataset takes exactly the old path — one line layer, all datasets, in order.

**`?? 0`** — noted, and I agree the silent-fallback risk is worth remembering. It's documented at the call site with the invariant it depends on, so if a future refactor does break that invariant, the comment is where someone will look.

All 14 checks are green on the previous commit except E2E, which was still running when I last looked; this commit is docs-only.

---

PR #752, opened 2026-08-06
fix(box): describe every box's range and outliers, not one chart-wide pair

jooyoungseo:
# Pull Request

## Description

Pressing `d` on a box plot opened a Chart Description whose summary claimed a single `Min` and a single `Max` for the whole chart. On a plot with several boxes that reads as though the chart had one minimum and one maximum, with nothing saying which box either belongs to. The data table underneath was also incomplete: it listed Min/Q1/Q2/Q3/Max but silently dropped the lower and upper outliers, even though both are sections the user can navigate to and hear announced.

There was a second, quieter inconsistency: the summary's `Min`/`Max` came from the trace-wide range used for sonification, which folds outliers in. So a chart with a low outlier showed `Min: -40` in the summary while the table's own Min column for that same group showed `5`.

## Related Issues

None filed — reported directly.

## Changes Made

`src/model/box.ts`, `BoxTrace.description`:

- **Summary extremes are now attributed to a group.** With more than one box, the summary reads `Lowest minimum: 1 (B)` and `Highest maximum: 40 (C)` instead of a bare `Min`/`Max` pair. A single-group box plot has nothing to disambiguate, so it keeps plain `Minimum: 2` / `Maximum: 10`. A group whose whisker end is missing is skipped rather than poisoning the range; if no group has one, the row reads `missing`, matching how the bar trace reports an unmeasurable range.
- **Summary extremes are whisker ends**, not the outlier-inclusive audio range, so they agree with the Minimum and Maximum columns of the table directly below.
- **The data table gained the outlier columns.** Headers and cells both walk `this.sections`, so the table now covers every navigable section — `Group, Lower outlier(s), Minimum, 25%, 50%, 75%, Maximum, Upper outlier(s)` — and the two can't drift apart if the section list changes. This also aligns the column names with the section names the user hears while navigating, and with how `ViolinBoxTrace` already builds its table. A group with no outliers renders a blank cell. Horizontal box plots keep their table in navigation order (groups are reversed at construction so navigation starts at the lower-left box).

`test/model/boxDescription.test.ts` (new, 8 tests): group attribution with several boxes, the single-group form, whisker-vs-outlier range, the all-missing and partially-missing cases, group count/names, the outlier columns, and horizontal row order.

n/a — dialog content change; the assertions in the new test file show the exact before/after strings.

## Checklist

## Additional Notes

Verification run locally, all green: `npm run lint:fix`, `npm run type-check`, `npm run build`, and `npm test` (119 unit suites / 1576 tests, plus 6 ESM suites / 61 tests). Documentation was left alone — the description dialog's contents are not spelled out in the docs site.

`ViolinBoxTrace` (`src/model/violinBox.ts`) carries the same single `Min`/`Max` summary shape. It is left untouched here to keep this change to the box plot that was reported; violin plots produce no outliers, so only the extreme-attribution half would apply to it. Happy to follow up in a separate PR if you want it aligned.

---
_Generated by [Claude Code](https://claude.ai/code/session_01S1NsuQkNcN8TCEwEmDHg2r)_

### DIFF
--- a/src/model/box.ts
+++ b/src/model/box.ts
@@ -132,19 +132,29 @@ export class BoxTrace extends AbstractTrace {
     const stats: DescriptionState['stats'] = [
       { label: 'Number of groups', value: this.points.length },
       { label: 'Group names', value: groupNames },
-      { label: 'Min', value: this.min },
-      { label: 'Max', value: this.max },
+      ...this.rangeStats(),
     ];
 
-    const headers = ['Group', 'Min', 'Q1', 'Q2', 'Q3', 'Max'];
-    const rows: (string | number)[][] = this.points.map(p => [
-      p.z,
-      p.min,
-      p.q1,
-      p.q2,
-      p.q3,
-      p.max,
-    ]);
+    // Both the headers and the cells walk `this.sections`, so every section the
+    // user can navigate to — the outliers included — gets a column, and the two
+    // stay aligned if the section list ever changes.
+    const headers = ['Group', ...this.sections];
+    const isHorizontal = this.orientation === Orientation.HORIZONTAL;
+
+    const rows: (string | number)[][] = this.points.map((point, pointIdx) => {
+      const sectionValues = this.sections.map((_, sectionIdx) => {
+        const value = isHorizontal
+          ? this.boxValues[pointIdx]?.[sectionIdx]
+          : this.boxValues[sectionIdx]?.[pointIdx];
+        // Outlier sections hold an array of values; a group with none renders
+        // as a blank cell rather than an empty-looking `[]`.
+        if (Array.isArray(value)) {
+          return value.join(', ');
+        }
+        return value ?? '';
+      });
+      return [point.z, ...sectionValues];
+    });
 
     return {
       chartType: this.getChartTypeLabel(),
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Thanks — notes 1 and 2 are addressed in b469a74. Note 3 needs no action; details below.

**1. `grouped` drift.** Fair, and the failure mode you describe is the right one to guard against. Rather than passing the flag down, `extremeStat` now takes both label choices and makes the check once, so the label and the value format come from a single evaluation and there is no second call site to keep in sync:

```ts
this.extremeStat(
  { single: 'Minimum', grouped: 'Lowest minimum' },
  p => p.min,
  (candidate, current) => candidate < current,
)
```

**2. Tie-breaking.** Written down in the JSDoc and covered by a new test (`names the first group in navigation order when boxes tie`). Worth locking in rather than just documenting, since several boxes bottoming out at the same floor is the ordinary case, not a corner one. Note that "first in `this.points`" is first in *navigation* order for both orientations — a horizontal box plot reverses its points at construction, so the named group is always the first one the user reaches.

**3. `NaN` announced as "NaN".** This one doesn't reproduce — the value never reaches the accessibility tree as a string. `DataTable` renders every cell through `formatCell`, and `isDisplayable` rejects non-finite numbers (`src/ui/component/Description.tsx:85-104`):

```ts
if (typeof value === 'number')
  return Number.isFinite(value);
...
function formatCell(value: unknown): string {
  if (!isDisplayable(value))
    return '';
  return String(value);
}
```

So `formatCell(NaN)` returns `''` and the cell renders blank. `value ?? ''` in the model only needs to survive as far as that filter; it isn't the last line of defence. The existing test `marks a range as missing when no group carries the whisker end` exercises the `NaN` path through the model.

Re: the sandbox note — `lint:fix`, `type-check`, `build`, and `test` were all re-run locally after this push and are green (1577 unit tests across 119 suites, plus 61 ESM tests across 6 suites). CI on this PR is the real check, agreed.

---

PR #753, opened 2026-08-06
feat(audio): play the menu open/close cue for every dialog

jooyoungseo:
# Pull Request

## Description

Pressing `G` opens the Go To Extrema modal with a short rising two-note tick, and dismissing it plays the falling counterpart. No other dialog made a sound. This gives every dialog the same open/close cue.

## Related Issues

None.

## Changes Made

The cue lived in `GoToExtremaViewModel`, called by hand on `toggle()` and `hide()`, which is why only that one modal had it. Every dialog changes focus through `DisplayService.toggleFocus`, so the cue moves to `DisplayViewModel` and is keyed off the focus transition instead: entering a dialog scope plays the open cue, leaving one plays the close cue.

Keying it off the transition rather than off each dialog covers every dismissal path — Escape, the close button, the backdrop, re-pressing the shortcut, choosing an entry — without each dialog having to remember to play it. Focus-out stays silent, because disposal clears the focus without emitting a change, preserving the behaviour the Go To Extrema cue already had.

Dialogs that now sound:

| Dialog | Shortcut |
| --- | --- |
| Go To Extrema | `g` (unchanged behaviour) |
| Help | `Ctrl+/` |
| Chat | `?` |
| Settings | `Ctrl+,` |
| Chart description | `d` |
| Command palette | `Ctrl+Shift+P` |
| Candlestick reference picker | `Ctrl+Shift+L` |

Braille (`b`) and review (`r`) stay silent: they render as an inline text field beside the chart rather than as an overlay, so a dialog cue would misdescribe them. The plot navigation scopes — `TRACE`, `SUBPLOT`, `GRID_CELL`, and `CANDLESTICK_DELTA` — are not dialogs and are silent too; `CANDLESTICK_DELTA` keeps its own enter/exit arpeggio.

Files:

- `src/type/event.ts` — one `MODAL_SCOPES` set, with `DIALOG_SCOPES` derived from it by excluding the two inline panels, so a scope cannot be added to one list and forgotten in the other.
- `src/service/display.ts` — `toggleFocus` tests membership in `MODAL_SCOPES` instead of the hand-rolled chain of comparisons that restated the same list. Behaviour unchanged.
- `src/state/viewModel/displayViewModel.ts` — the previous-focus tracking and `playDialogCue()`; takes `AudioService`.
- `src/state/viewModel/goToExtremaViewModel.ts` — drops the two hand-rolled cue calls and the now-unused `AudioService` dependency.
- `src/controller.ts` — rewires the two constructors.
- `src/service/audio.ts` — JSDoc on `playMenuOpenTone`/`playMenuCloseTone` now describes the shared cue rather than the Go-To modal. No behaviour change.

Not applicable — the change is audible, not visual. It was verified by instrumenting `AudioContext.prototype.createOscillator` in Chromium and driving `examples/barplot.html` against the built bundle, recording each oscillator's frequency at `start()`:

| Action | Oscillators played |
| --- | --- |
| Help open (`Ctrl+/`) | `[660, 990]` |
| Help close (`Escape`) | `[990, 660]` |
| Description open (`d`) | `[660, 990]` |
| Description close (`Escape`) | `[990, 660]` |
| Go To Extrema open (`g`) | `[660, 990]` |
| Go To Extrema close (`Escape`) | `[990, 660]` |
| Braille toggle (`b`, then `b`) | `[]` |

Rising on open, falling on close, identical across dialogs; Go To Extrema unchanged; braille silent, as intended.

## Checklist

- [x] I have performed a self-review of my own code and ensured it follows the project&#39;s coding standards.

## Additional Notes

`test/state/viewModel/displayViewModel.test.ts` is new: it drives focus events through a `DisplayService` stub and asserts the open cue for each of the seven dialog scopes, the close cue on leaving each, silence for the navigation scopes and for braille/review, silence on the first event of a focus session, no double cue when a scope is re-reported, a single open cue when one dialog replaces another, and silence after `dispose()`. A parity case asserts the production `DIALOG_SCOPES` equals the list those cases enumerate, so a scope added to one and not the other fails rather than going uncovered.

`test/state/viewModel/goToExtremaViewModel.test.ts` kept its coverage of the same paths, re-pointed at the scope transitions that now drive the cue.

Verified locally: `npm run lint:fix`, `npm run type-check`, `npm run build`, and `npm test` (1590 unit + 61 ESM tests, all passing). CI is green on all 14 checks including `E2E (chromium)`.

---
_Generated by [Claude Code](https://claude.ai/code/session_018L5NutBmFmMxUeVWGYE9sn)_

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -164,12 +164,11 @@ export class Controller implements Disposable {
       store,
       this.goToExtremaService,
       this.context,
-      this.audioService,
       this.formatterService,
     );
     this.reviewViewModel = new ReviewViewModel(store, this.reviewService);
     this.descriptionViewModel = new DescriptionViewModel(store, this.descriptionService);
-    this.displayViewModel = new DisplayViewModel(store, this.displayService);
+    this.displayViewModel = new DisplayViewModel(store, this.displayService, this.audioService);
     this.helpViewModel = new HelpViewModel(store, this.helpService);
     this.settingsViewModel = new SettingsViewModel(store, this.settingsService);
 
--- a/src/service/audio.ts
+++ b/src/service/audio.ts
@@ -906,17 +906,19 @@ export class AudioService implements Observer<PlotState>, Disposable {
   }
 
   /**
-   * Plays the "menu open" cue — a short rising two-note tick — when the Go-To
-   * modal opens. A navigational affordance, so it plays in any audio mode
-   * except OFF (mirroring playWarningToneIfEnabled).
+   * Plays the "menu open" cue — a short rising two-note tick — when a dialog
+   * opens: Go To Extrema, help, settings, chat, the chart description, the
+   * command palette, or the candlestick reference picker all share it. A
+   * navigational affordance, so it plays in any audio mode except OFF
+   * (mirroring playWarningToneIfEnabled).
    */
   public playMenuOpenTone(): void {
     this.playMenuTone(MENU_OPEN_FREQUENCIES);
   }
 
   /**
-   * Plays the "menu close" cue — a short falling two-note tick — when the Go-To
-   * modal is dismissed. Plays in any audio mode except OFF.
+   * Plays the "menu close" cue — a short falling two-note tick — when a dialog
+   * is dismissed. Plays in any audio mode except OFF.
    */
   public playMenuCloseTone(): void {
     this.playMenuTone(MENU_CLOSE_FREQUENCIES);
--- a/src/service/display.ts
+++ b/src/service/display.ts
@@ -2,7 +2,7 @@ import type { Context } from '@model/context';
 import type { TextService } from '@service/text';
 import type { Disposable } from '@type/disposable';
 import type { Event, Focus } from '@type/event';
-import { Emitter, Scope } from '@type/event';
+import { Emitter, MODAL_SCOPES, Scope } from '@type/event';
 import { Constant } from '@util/constant';
 import { Stack } from '@util/stack';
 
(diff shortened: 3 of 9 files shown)


jooyoungseo:
Addressed the review in `e72dd39`.

**The duplicated scope list** was the substantive one, and it was correct: `DIALOG_SCOPES` restated the modal list `DisplayService.toggleFocus` already kept as a chain of comparisons, so a dialog added later had to be registered in both places or silently lose either its cue or its mode-toggle handling. There is now one `MODAL_SCOPES` set in `src/type/event.ts`; `toggleFocus` tests membership in it, and `DIALOG_SCOPES` is derived from it by excluding braille and review. The exclusion is the only thing stated twice, and it is stated where the reason for it lives.

**The JSDoc wording** was inaccurate as flagged. The rule is that the cue describes the scope being *entered*, so moving straight from one dialog to another sounds a single open cue rather than a close then an open. The comment now says that, and notes it is a rule for the case rather than a description of one that occurs — as the review of `SCOPED_KEYMAP` found, no shortcut reaches a second dialog from inside the first today.

**The test's own scope list** now has the drift guard rather than being replaced by an import: a new first case asserts the production `DIALOG_SCOPES` equals the list the cases enumerate. Importing the set outright would have made the `test.each` cases agree with it by construction, which is the coverage the comment was worried about losing. A scope added to one and not the other now fails that case instead of going untested.

Not changed: `playDialogCue` calling `AudioService` directly. That is the pattern `chatViewModel` and `textViewModel` already follow, and reworking it would be a change to the convention rather than to this PR.

`Build Project` failed on `7e8fba8` for an infrastructure reason, not the code — `Failed to resolve action download info. Error: Service Unavailable`, before checkout ran. The new push re-runs it.

Verified locally on `e72dd39`: `lint:fix`, `type-check`, `build`, and `npm test` all clean (1590 unit + 61 ESM). E2E was not run locally; CI covers it.

---

jooyoungseo:
Both review passes noted they could only verify by reading, so here is the cue confirmed in a real browser rather than in jsdom.

Driving `examples/barplot.html` against the built bundle from `e72dd39` in Chromium, with `AudioContext.prototype.createOscillator` instrumented to record each oscillator's frequency at `start()`:

| Action | Oscillators played |
| --- | --- |
| Help open (`Ctrl+/`) | `[660, 990]` |
| Help close (`Escape`) | `[990, 660]` |
| Description open (`d`) | `[660, 990]` |
| Description close (`Escape`) | `[990, 660]` |
| Go To Extrema open (`g`) | `[660, 990]` |
| Go To Extrema close (`Escape`) | `[990, 660]` |
| Braille toggle (`b`, then `b`) | `[]` |

Rising on open, falling on close, identical across dialogs; Go To Extrema unchanged from what it did before this PR; braille silent, as intended. That is the whole behavioural claim of the PR, observed rather than inferred.

The instrumentation was a throwaway script run locally, not committed — the repository state is still the two commits above.

On CI: `E2E (chromium)` on run `31115695526` was cancelled 16 seconds in by `The runner has received a shutdown signal`, before it finished resolving actions. Same infrastructure flake as the earlier `Build Project` failure on `7e8fba8`, and unrelated to the diff. I will re-run it once the workflow finishes.

---

PR #755, opened 2026-08-06
fix(violin): attribute the described extremes to the violin they come from

jooyoungseo:
# Pull Request

> **#752 has merged and this is now rebased onto `main`** — the two commits it stacked on are gone from this branch's history, and the diff is this PR's own change only.

## Description

`ViolinBoxTrace.description` carried the same single `Min`/`Max` summary the box plot shed in #752 — one pair taken across every violin, so a chart of five read as though it had one minimum and one maximum, with nothing saying which violin either belonged to. This applies the same group attribution, with two differences the violin's own shape forces.

## Related Issues

Follow-up to #752, requested after that review.

## Changes Made

**`src/model/boxExtremes.ts` (new).** The `extremeStat` helper moves out of `BoxTrace` now that a second caller exists — the project rule is no abstraction before that point, and this is that point. It keeps the behavior #752 pinned down: NaN ends skipped, `missing` when no group has one, group name only when there is more than one group, earliest group wins a tie. `isLower`/`isHigher` come along as the two `beats` rules.

**`src/model/violinBox.ts`.** `rangeStats()` replaces the `Min`/`Max` pair, with one violin-specific rule: **only sections the violin actually draws get a row**. The minimum is unconditional in `buildViolinSections()`, but the maximum rides on `showExtrema` — and naming a maximum the user cannot navigate to would describe a section that is not on screen. Today's code silently reports the Q3-derived range as `Max` in that configuration.

**`src/model/boxExtremes.ts`, unnamed groups.** A guard the box plot never needed. Verified against `examples/violin.html` in a browser and the summary came out `Lowest minimum: 326 (undefined)` — violin layers carry their group under `fill` and leave `z` unset, which is also why their data table has always shown a blank Group column. An unnamed group now falls back to the bare value.

**Name trimming** (2nd commit, from Copilot's review). `hasName()` judged the name by its trimmed form but printed the raw `z`, so a padded name announced as `25 (  A  )` — and for a screen reader that padding reads as a pause. Reading the name and testing it are now one step.

**`test/model/violinBoxDescription.test.ts` (new, 8 tests):** multi-violin attribution, single-violin form, the max row dropped under `showExtrema: false` and kept under `true`, all-missing ends, the unnamed-group fallback, padded-name trimming, and group count / section list.

Driven with Playwright against `examples/violin.html` built from this branch:

```
Summary
  Number of groups: 5
  Sections: Minimum, 25%, 50%, 75%, Maximum
  Lowest minimum: 326        ← was "326 (undefined)" before the name guard
  Highest maximum: 14171
```

`examples/boxplot-vertical.html` and `examples/boxplot-horizontal.html` were re-driven after the helper extraction and still name their groups (`21.957… (Group 3)`), confirming the move is behavior-preserving for #752.

## Checklist

## Additional Notes

Re-verified after the rebase onto `main`: `npm run lint:fix`, `npm run type-check`, and `npm test` all green (1651 unit tests across 125 suites, plus 61 ESM tests across 6 suites).

Earlier red checks on this branch were two separate causes, both now resolved. A GitHub Actions outage (`Failed to resolve action download info. Error: Service Unavailable`) took out three jobs before any code ran; and `claude-review` then failed with `fatal: couldn't find remote ref claude/maidr-boxplot-d-key-description-t7oef8` — it restores its trusted config from the PR's base branch, and this PR's base was deleted when #752 merged. The rebase onto `main` fixes that.

**Out of scope, found while verifying:** the violin data table's Group column is blank because the layer data uses `fill` where `BoxPoint` declares `z`. That is a schema mismatch between the producer and `src/type/grammar.ts`, not a description bug, and fixing it likely means a change on the py-maidr side too — so this PR only stops the summary from reading out `undefined`. Worth its own issue.

### DIFF
--- a/src/model/box.ts
+++ b/src/model/box.ts
@@ -8,6 +8,7 @@ import { Constant } from '@util/constant';
 import { MathUtil } from '@util/math';
 import { Svg } from '@util/svg';
 import { AbstractTrace } from './abstract';
+import { extremeStat, isHigher, isLower } from './boxExtremes';
 import { MovableGrid } from './movable';
 
 /**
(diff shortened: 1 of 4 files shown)


jooyoungseo:
Copilot's catch on `boxExtremes.ts:60` was right and is fixed in 9ea1f19.

`hasName()` judged the name by its trimmed form but the value printed the raw `winner.z`, so a padded name announced as `25 (  A  )` — and for a screen reader that padding is not invisible, it reads as a pause. Reading the name and testing it are now one step:

```ts
function groupName(point: BoxPoint): string | null {
  const name = point.z as string | undefined;
  if (typeof name !== 'string') {
    return null;
  }
  const trimmed = name.trim();
  return trimmed === '' ? null : trimmed;
}
```

Covered by `trims a padded group name rather than announcing the padding`.

On the other review's non-blocking note: the `showExtrema: true` test does re-cover the default path, since `showExtrema` defaults to true. Keeping it deliberately — the pair reads as one statement about the option (`false` drops the row, `true` keeps it), and it pins the default's meaning so a future flip of that default fails a test instead of silently changing what the dialog reports.

Re-verified after the change: `lint:fix`, `type-check`, `build`, and `test` all green (1585 unit tests across 120 suites, plus 61 ESM tests across 6 suites).

---

PR #757, opened 2026-08-06
fix(description): round the numbers the chart description reads out

jooyoungseo:
# Pull Request

> **#755 has merged and this is now rebased onto `main`** — the commits it stacked on are gone from this branch's history, and the diff is this PR's own change only.

## Description

The description dialog printed whatever float the trace was holding. The box plot example describes a whisker as `21.957700280519678` — seventeen digits to listen through for two digits of meaning — and the histogram's bin edges were no better.

Announcements have been rounded to two decimals since #727, but the dialog never was. So the same value read one way when you navigated to it and another when you asked for a description of it.

## Related Issues

Found while verifying #752 in a browser; raised there and split out at the maintainer's request.

## Changes Made

**`src/service/description.ts`.** `DescriptionService` applies the same `defaultFormat` to every stat value and every table cell on the way out. It is a pure util, so no `FormatterService` plumbing or controller change is needed. Only what the dialog shows is shortened, and only on the way out — the trace keeps its own numbers, and sonification, braille, and extrema read those directly.

Two things had to move for it to reach everything:

- **Multi-value cells travel unjoined.** A box plot's outliers were joined into a string by the trace, so they arrived as opaque text no formatter could get inside. Traces now hand the values over as an array and the service rounds each one before joining. `DescriptionState.dataTable.rows` widens to carry `number[]`; the service returns the narrower `DisplayDescriptionState`, so the view still never sees an array.
- **Non-finite numbers stay numbers.** `defaultFormat(NaN)` returns the literal text `"NaN"`, and `Description.tsx`'s `isDisplayable` blanks a non-finite *number* but would happily print that *string*. Guarding this is what keeps the blank cell blank.

**`src/model/boxExtremes.ts`.** The group-attributed summary value (`21.96 (Group 3)`) is composed into a string in the model, past where the service can round what is inside it, so it rounds the number with the same `defaultFormat` before composing. A test caught this — the value was still going out at full precision.

**Second commit, from Copilot's review.** The array branch ran its result through `join`, which coerced a non-finite entry straight back into the `"NaN"` text the scalar guard exists to avoid. A joined cell cannot hand a blank back for one of its entries, so the entry is dropped instead — an outlier that isn't a number isn't an outlier. Two docstrings the code had outgrown were corrected in the same commit.

**Tests:** `test/service/descriptionRounding.test.ts` (new, 7 tests) — a computed float rounded, outliers rounded before joining, a non-finite outlier dropped, integers left unpadded, the non-finite scalar guard, summary rows rounded with the group name intact, and label strings untouched. `test/model/boxDescription.test.ts` updated for the unjoined cells.

Driven with Playwright against the examples built from this branch:

```
examples/boxplot-vertical.html
  Lowest minimum: 21.96 (Group 3)          was 21.957700280519678 (Group 3)
  Group 3 | -20, -10, -9.8, 6.06, 14.74 | 21.96 | 66.08 | 81.79 | 99.38 | 148.95
                                          was -9.795014876280863, 6.057387303065189 …

examples/histogram.html                     bin edges, previously full floats
  1.44 | 33 | 1.29 | 1.59

examples/barplot.html                       integers stay unpadded
  Sat | 87
```

## Checklist

## Additional Notes

Re-verified after the rebase onto `main`: `npm run lint:fix`, `npm run type-check`, `npm run build`, and `npm test` all green — 1658 unit tests across 126 suites, plus 61 ESM tests across 6 suites.

Earlier red checks on this branch were a GitHub Actions outage, not this diff. Five jobs across three workflows died at `Getting action download info` with `Service Unavailable`, before checkout — `main` was hit the same way (the **Push on main** run at 15:56 and a **PR Preview Deploy** at 16:02), so it was repo-wide. I also confirmed the one check that could plausibly have been real, `Lint Commit Messages`, passes: CI's exact command `commitlint --from=HEAD~1 --to=HEAD` exits 0 locally.

**A deliberate limit.** This applies `defaultFormat`, not the per-layer `FormatterService`. The formatter is per-axis, and `DescriptionState` has no column-to-axis mapping — a bar chart's table is `[x, y]` and a heatmap's values are `z`, so picking one axis for the whole table would apply, say, a currency format to a category column. Rounding is axis-agnostic and cannot produce a wrong number, only a less decorated one. Wiring authored `AxisFormat`s through would mean giving the table per-column axes; worth doing, but a larger change than this one and easy to get wrong in the direction of printing something false.

### DIFF
--- a/src/model/box.ts
+++ b/src/model/box.ts
@@ -142,16 +142,13 @@ export class BoxTrace extends AbstractTrace {
     const headers = ['Group', ...this.sections];
     const isHorizontal = this.orientation === Orientation.HORIZONTAL;
 
-    const rows: (string | number)[][] = this.points.map((point, pointIdx) => {
+    const rows: DescriptionState['dataTable']['rows'] = this.points.map((point, pointIdx) => {
       const sectionValues = this.sections.map((_, sectionIdx) => {
         const value = isHorizontal
           ? this.boxValues[pointIdx]?.[sectionIdx]
           : this.boxValues[sectionIdx]?.[pointIdx];
-        // Outlier sections hold an array of values; a group with none renders
-        // as a blank cell rather than an empty-looking `[]`.
-        if (Array.isArray(value)) {
-          return value.join(', ');
-        }
+        // Outlier sections hold an array, which travels on unjoined so the
+        // description service can round each value before joining them.
         return value ?? '';
       });
       return [point.z, ...sectionValues];
--- a/src/model/boxExtremes.ts
+++ b/src/model/boxExtremes.ts
@@ -1,5 +1,6 @@
 import type { BoxPoint } from '@type/grammar';
 import type { DescriptionState } from '@type/state';
+import { defaultFormat } from '@util/format';
 
 /**
  * Labels for a range row, one for each shape the summary can take.
@@ -56,7 +57,14 @@ export function extremeStat(
   );
   const value = valueOf(winner);
   const name = grouped ? groupName(winner) : null;
-  return { label, value: name === null ? value : `${value} (${name})` };
+  if (name === null) {
+    // A bare number travels on for DescriptionService to round.
+    return { label, value };
+  }
+  // Composing the group name onto the value makes this a string, past the
+  // point where the service can round what is inside it, so the number is
+  // rounded here with the same `defaultFormat` the service would have used.
+  return { label, value: `${defaultFormat(value)} (${name})` };
 }
 
 /**
--- a/src/model/violinBox.ts
+++ b/src/model/violinBox.ts
@@ -164,14 +164,11 @@ export class ViolinBoxTrace extends AbstractTrace {
     const headers = ['Group', ...this.sections];
     const isHorizontal = this.orientation === Orientation.HORIZONTAL;
 
-    const rows: (string | number)[][] = this.points.map((point, pointIdx) => {
+    const rows: DescriptionState['dataTable']['rows'] = this.points.map((point, pointIdx) => {
       const sectionValues = this.sections.map((_, sectionIdx) => {
         const value = isHorizontal
           ? this.boxValues[pointIdx]?.[sectionIdx]
           : this.boxValues[sectionIdx]?.[pointIdx];
-        if (Array.isArray(value)) {
-          return value.join(', ');
-        }
         return value ?? '';
       });
       return [point.z, ...sectionValues];
--- a/src/service/description.ts
+++ b/src/service/description.ts
@@ -1,8 +1,33 @@
 import type { Context } from '@model/context';
 import type { DisplayService } from '@service/display';
-import type { DescriptionStat, DescriptionState } from '@type/state';
+import type { DescriptionStat, DescriptionState, DisplayDescriptionState } from '@type/state';
 import { AbstractTrace } from '@model/abstract';
 import { Scope } from '@type/event';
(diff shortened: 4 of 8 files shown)


jooyoungseo:
All four addressed in 8b720b5. The first was a real bug — good catch.

**1. Non-finite entry in an array cell** (`description.ts:30`). Correct, and I confirmed it: `[1.23, NaN, 5].join(', ')` gives `"1.23, NaN, 5"`, so the array branch handed back the very text the scalar guard exists to avoid. The two suggested options aren't equivalent, though — mapping a non-finite entry to `''` would leave a hole in the list (`"-2.56, , -1"`), and a joined cell can't hand a blank back for just one entry the way the scalar path can. So the entry is dropped:

```ts
if (Array.isArray(value)) {
  return value
    .filter(entry => Number.isFinite(entry))
    .map(entry => defaultFormat(entry))
    .join(', ');
}
```

An outlier that isn't a number isn't an outlier. Pinned by `drops a non-finite outlier instead of joining the text "NaN" in`: `[-2.5551, NaN, -1]` → `'-2.56, -1'`, and an all-non-finite list → `''`, which the dialog renders blank.

**2. "Only the announcement is shortened"** (`:109`). Fair — I'd carried that phrasing over from `defaultFormat`'s docstring, where "announcement" means the live-region text this doesn't touch. Now: *"Only what the dialog shows is shortened, and only on the way out — the trace keeps its own numbers. Sonification, braille, and extrema read those directly and never pass through here."* The second half is still accurate; those paths read the trace's values, not `DescriptionState`.

**3. `roundCell` docstring** (`:22`). Right, and it was loose in two directions — an array doesn't come back "unchanged" either. Rewritten to state each branch, including that `defaultFormat` returns a string untouched.

**4. `number[]` vs "multi-value cell"** (`state.ts:406`). Taking the narrow-the-wording option rather than widening the type. `(string | number)[]` has no producer today, and adding it would mean `roundCell` carrying a branch for a shape nothing emits — against this repo's "no abstraction before a second caller exists" rule. The comment now says the array form is numeric on purpose and why.

Verified after the change: `lint:fix`, `type-check`, `build`, and `test` all green (1658 unit tests across 126 suites, plus 61 ESM tests across 6 suites).

---

jooyoungseo:
Flagging the red `Copilot Instructions In Sync` on `8b720b5` so it isn't mistaken for something in this diff: it is a GitHub Actions platform failure, not a check result. The job never got as far as checking out the repo —

```
Prepare all required actions
Getting action download info
Failed to resolve action download info. Error: Service Unavailable
Retrying in 15.667 seconds
Failed to resolve action download info. Error: Service Unavailable
Retrying in 19.488 seconds
##[error]Service Unavailable
```

`main` is hit too — the **Push on main** run at 15:56 and a **PR Preview Deploy** at 16:02 both failed the same way — so this is repo-wide and not specific to either open PR. It's also been intermittent for the last ~45 minutes, which is why jobs here have been sitting queued and some pass while their neighbours fail.

Nothing to fix in code; this check touches `.claude/` → `.github/` sync and this PR changes neither. I re-ran the failed jobs once (GitHub rejected a second attempt with `This workflow is already running`, so the run is retrying on its own) and will re-run again once the platform settles. Verified locally on `8b720b5` in the meantime: `lint:fix`, `type-check`, `build`, and `test` all green — 1658 unit tests across 126 suites, plus 61 ESM tests across 6 suites.

---

PR #762, opened 2026-08-08
perf(build): ship React's production build

jooyoungseo:
# Pull Request

## Description

Every MAIDR release so far has shipped **React's development build**.

`process.env.NODE_ENV` resolved to `undefined` for as long as `define` carried only the `process.env` prefix entry — rollup rewrote it to `({}).NODE_ENV`, which is `undefined`, which is `!== 'production'`, which is React's development path. Nothing failed, so nothing surfaced it. The visible tell was a console notice on any built example:

```
Download the React DevTools for a better development experience
```

That is ~300 kB of warnings, invariant checks and DevTools hooks that no consumer of a released bundle can act on.

#759 deliberately kept the value at `undefined`. It was fixing the rolldown switch that made the flag visible in the first place, and swapping which React ships as a side effect of a dependency bump would have buried a behavioural change inside a build-tool change. This is that change, on its own.

## Related Issues

Follow-up from #759, where this was found and deferred.

## Changes Made

- `vite.config.ts` and `scripts/build.js`: `'process.env.NODE_ENV'` → `JSON.stringify('production')`
- `test/scripts/buildDefineNodeEnv.test.ts`: pin the **value**, not just the key

### Size

| | before | after | change |
|---|---|---|---|
| `dist/maidr.js` | 1,666,923 B | 1,366,082 B | **−300 KB (−18%)** |
| gzipped | 475,955 B | 394,686 B | **−81 KB (−17%)** |

Measured on the same machine, same command (`node scripts/build.js core`), toggling only the define value.

### Why the test changed

The existing regression test asserted the *key* was present, because #759's failure was a hard crash — the bundle threw `process is not defined` and MAIDR never loaded. This failure mode is the opposite: `undefined` loads fine and works, it just quietly carries the wrong React. A key-only assertion would stay green through a regression here, so the test now also pins the value to `"production"`.

Not applicable — the observable difference is the absence of the DevTools console notice.

## Checklist

## Additional Notes

Verified locally:

| Check | Result |
|---|---|
| `npm run type-check` | pass |
| `npm run lint` | pass |
| `npm test` | **127/127 suites, 1663 tests** |
| `npm run build` | pass |
| **`npx playwright test --project=chromium`** | **330/330 passed** |
| Console of a built example | DevTools notice gone; `window.maidr` still initialises |

**The e2e run is the gate here, not the unit suite.** React's production build removes development-only warnings and the extra checks behind them, so a component relying on dev-mode behaviour would break at runtime while every unit test stayed green — the unit suite renders components but never loads the built bundle. 330/330 unchanged is the evidence that nothing depended on it.

### Risk

This changes which React runs in the published bundle. The upside is size and a quieter console; the downside is that dev-only warnings no longer fire for downstream integrators who were (perhaps unknowingly) relying on seeing them. That seems clearly right for a published artefact, but it is a behavioural change and worth a maintainer's eye rather than a rubber stamp — which is why it is separate from #759 instead of folded into it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FA4dYokG4c7fkRj45XQvJx)_

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -370,8 +370,9 @@ export function createViteConfig(config) {
     },
     // `process.env.NODE_ENV` is spelled out alongside `process.env` because
     // rolldown only substitutes a `define` key that matches the whole member
-    // expression; see the matching comment in vite.config.ts.
-    define: { 'process.env': {}, 'process.env.NODE_ENV': 'undefined' },
+    // expression, and it is `production` because these are the bundles that get
+    // published; see the matching comment in vite.config.ts.
+    define: { 'process.env': {}, 'process.env.NODE_ENV': JSON.stringify('production') },
     resolve: { alias: config.aliases },
   };
 }
--- a/test/scripts/buildDefineNodeEnv.test.ts
+++ b/test/scripts/buildDefineNodeEnv.test.ts
@@ -20,6 +20,12 @@ import { pathToFileURL } from 'node:url';
  * focused, but found body". This asserts the config directly so the next
  * bundler swap fails here instead of there.
  *
+ * The value is asserted too, not just the key. While the prefix entry was the
+ * only one here the flag resolved to `undefined`, which is falsy for React's
+ * purposes -- so the published bundles shipped its development build without
+ * anything failing. Both halves of that need pinning: the key so the bundle
+ * loads at all, the value so it loads the right React.
+ *
  * `scripts/build.js` is plain ESM JavaScript and `allowJs` is false, so it
  * cannot be imported from a ts-jest test directly. Each case runs in a real
  * node subprocess with `--input-type=module`, as in buildOutputFilenames.test.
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Both actionable findings are fixed in `f66fea1`. Checking them out led somewhere further than the review put it, so writing up what I found.

**On finding 3 (the "roughly 90 kB" figure).** Fixed — that was my number, and it disagreed with my own measurement table. The comment no longer carries a figure at all; the PR body has the measured one, and a number duplicated in a code comment is a number that goes stale.

**On finding 2 (`vite.config.ts` not wired to a dev server).** Correct, and it goes further than "not wired to a dev server" — **nothing builds through that file at all**:

- `scripts/build.js:347` passes `configFile: false`, so the library build never merges the root config. This is the path that produces every published bundle.
- Every example names its own config (`--config examples/*/vite.config.ts`).
- No npm script invokes `vite` bare, which is the only other way the root config would get auto-loaded.

So the root `vite.config.ts` is not consumed by any build path today. That means my `define` change there is inert — worth being explicit about, since the PR body implies both files matter equally. **The measured 300 kB / 81 kB saving comes entirely from `scripts/build.js`**, and it stands: I measured with `node scripts/build.js core`, which is the path that ships.

I kept the change in the root config rather than dropping it, and tightened the test to pin `production` there too. An unused config is exactly the kind that drifts, and the failure it would reintroduce cost an e2e run to find the first time. Better it be correct in advance than correct-looking. The comments in both files now say this plainly instead of asserting a role the file does not have.

**One thing I did not do:** the root `vite.config.ts` being dead config is a real finding, but deleting it is a separate question from this PR — `test/scripts/publishedPackage.test.ts` reads it, and whether it is meant as a template for future builds is a maintainer's call, not mine to make inside a bundle-size change.

**On finding 1** — thanks for independently checking `grep -rn NODE_ENV src/`. That is the assertion the whole PR rests on and I had only reasoned about it from React's side.

Re-verified after the change: lint, type-check, and **127/127 suites / 1664 tests** (one more than before — the new `production` assertion on the root config). The 330/330 e2e result is unaffected; this commit touches comments and a test only.

---

PR #767, opened 2026-08-09
feat(pie): support pie charts across the core and every adapter

jooyoungseo:
# Pull Request

## Description

Adds pie chart as a first-class trace type: `TraceType.PIE`, a `PieTrace` model, and pie support in all eleven chart-library adapters. A pie is now navigable, sonified, brailled, highlighted and announced like any other trace.

This is the foundation of a three-repo change. `src/model/factory.ts` throws on an unregistered trace type and that throw propagates out of `new Figure(...)`, so **this PR must merge before** the py-maidr and r-maidr pie PRs can render anything.

## Related Issues

None — this implements pie support from scratch.

## Changes Made

**Model and grammar**

- `PiePoint { x: string | number; y: number }` — flat `PiePoint[]`, never nested. `y` is numeric (unlike `BarPoint`) because it is both the sonified magnitude and the percentage numerator.
- `TraceType.PIE = 'pie'`, plus entries in the two total `Record<TraceType, …>` maps (`CHART_TYPE_LABEL`, `IS_ORIENTED`).
- `src/model/pie.ts` — `PieTrace extends AbstractTrace` **directly**, not `AbstractBarPlot`. The latter's constructor bakes an orientation into its values, audio panning, text axes and description; a pie has none, so all of that would have to be undone again.
- Slices are wrapped in a single row so `MovableGrid` navigates them — with one row its own bounds checks make up and down out of bounds, so no suppression code is needed.

**Design decisions worth a reviewer's eye**

- **Percentage is derived, not authored.** Computed once in the constructor from the slice values, so it cannot disagree with the values it is a share of. There is deliberately no `percentage` field on the wire.
- **Sonification pitches the raw value, not the percentage.** Percentage is a monotone rescale, so pitch ordering is identical either way; the raw value keeps audio consistent with braille and the reported min/max.
- **Gaps follow the bar convention.** A non-finite slice is excluded from the total and the range and announced as missing, rather than collapsing into a real zero. A zero total reports `0%`, never `NaN%`.
- `findNearestPoint` hit-tests wedge geometry via `isPointInFill` rather than bounding boxes — wedge boxes overlap heavily, so a box test picks the wrong slice.

**Services**

- Braille reuses the existing `BarBrailleEncoder`, registered for `PIE`. (An unregistered encoder fails *silently* at the `encoders.has` guard, not loudly.)
- Text puts the share on the existing optional `z` slot: "Fruit is Apples, Units is 30, Percentage is 25.0%".
- High contrast: pie joins stacked/dodged in the pattern-fill branch, so adjacent same-hue wedges stay distinguishable.

**Adapters** — recharts, victory, chartjs, plotly, highcharts, vegalite, d3 (new `binders/pie.ts`), google-charts, anychart, amcharts, frappe. Two adapters needed structural work rather than a new branch: amcharts was XY-only and never *found* a `PieChart`, and anychart's `anychart.pie()` exposes no series API so it is detected chart-level. Chart.js is canvas-based and gained an SVG wedge overlay — a bounding box would have covered three neighbouring slices.

**Docs, examples, tests** — nine adapter pie examples plus a vanilla `examples/pie.html`, all registered in `scripts/build-site.js`; `docs/BRAILLE.md`, `docs/SCHEMA.md`, `docs/llms.txt`, README and the ten adapter doc pages; unit tests for the model, services and every adapter; and an e2e spec with page object.

`.claude/rules/model.md`'s "Registering a new trace type" checklist was incomplete and partly wrong (it omitted the two total `Record` maps and the braille encoder map, and said `AbstractTrace` is generic when it is not) — corrected, with the `.github/` counterpart regenerated via `npm run sync:copilot`.

n/a — accessibility behaviour is covered by the e2e assertions on announcements and braille output.

## Checklist

## Additional Notes

Verification actually run on this branch:

| Gate | Result |
|---|---|
| `npm run build` | all 12 builds pass |
| `npm test` | 135 suites / 1779 tests + 6 suites / 61 tests, all passing |
| `npx playwright test` (chromium) | **350 passed**, including 20 new pie tests |

One note on the e2e run: the sandbox ships chromium-1194 while `@playwright/test` 1.59.1 pins 1217, and firefox/webkit are absent, so the suite was run against the installed chromium via a temporary config that was **not** committed. Firefox and WebKit were therefore not exercised here and will be covered by CI.

Two spots where I made a judgement call the spec did not settle:

- A gap slice announces `Units is NaN` in the value slot. This is pre-existing behaviour shared with `BarTrace` (which announces `Revenue is null`), not something pie introduced; fixing it belongs in `TextService.formatSingleValue` and would benefit every trace. Filed as follow-up rather than fixed here.
- Position announces the generic "Position is 3 of 5" rather than "Slice 3 of 5". Happy to change if you'd prefer pie-specific wording.

---

### DIFF
--- a/.claude/rules/model.md
+++ b/.claude/rules/model.md
@@ -18,8 +18,8 @@ element is active without knowing which level that is.
 
 Trace implementations live directly in `src/model/` — `bar.ts`, `line.ts`,
 `scatter.ts`, `box.ts`, `heatmap.ts`, `histogram.ts`, `candlestick.ts`,
-`segmented.ts`, `smooth.ts`, `violin.ts`. There is no `src/model/trace/`
-subdirectory.
+`segmented.ts`, `smooth.ts`, `violin.ts`, `pie.ts`. There is no
+`src/model/trace/` subdirectory.
 
 ## Rules
 
@@ -30,7 +30,9 @@ subdirectory.
   bounds check consistent with the data shape you actually index into.
 - **Never import from `src/service/`, `src/state/`, or `src/ui/`.** The model
   notifies; it does not call.
-- **One trace class per file**, extending `AbstractTrace<T>`.
+- **One trace class per file**, extending `AbstractTrace`. It is not generic —
+  the point type lives on the subclass's own `points` field, not in a type
+  parameter.
 
 ## Strategy pattern: the four modalities
 
@@ -50,12 +52,35 @@ returned state consistent with the other traces.
 
 ## Registering a new trace type
 
-1. Add the class in `src/model/`, extending `AbstractTrace<YourPoint>`.
-2. Implement `moveOnce()`, `isMovable()`, and the four modality accessors.
-3. Add the `case` to `TraceFactory.create()` in `src/model/factory.ts` —
-   an unregistered type throws at runtime.
-4. Add the type to `TraceType` and the point shape to `src/type/grammar.ts`.
-5. Confirm the audio, text, braille, and highlight services handle it.
+A trace type is registered in six places, then verified across the modality
+services. `TraceType` keys a total `Record` in two of them, so those two fail
+the build if you forget; the rest fail at runtime, or — for braille — not at
+all. `PieTrace` is the most recent worked example; grep `TraceType.PIE` to see
+every one of them at once.
+
+1. In `src/type/grammar.ts`: add the `TraceType` member, the point shape
+   (`PiePoint`), and that point to the `MaidrLayer.data` union.
+2. Add the class in `src/model/`, extending `AbstractTrace` — which is **not**
+   generic. Assign `movable` (usually a `MovableGrid` over your points, which
+   supplies `moveOnce()` and `isMovable()`), and implement the abstract
+   accessors: `audio`, `braille`, `text`, `description`, `dimension`,
+   `highlightValues`, `values`, `supportsExtrema`, and `findNearestPoint`.
+3. Add the `case` to `TraceFactory.create()` in `src/model/factory.ts` — an
+   unregistered type throws `Invalid trace type`, and the throw propagates out
+   of `new Figure(...)`, so nothing renders at all.
+4. Add the display name to `CHART_TYPE_LABEL` in `src/model/abstract.ts`. It is
+   a `Record<TraceType, string>`, so omitting it is a compile error.
+5. Answer `IS_ORIENTED` in `src/util/orientation.ts` — also a total `Record`,
+   and also a compile error to omit. `true` means the trace resolves
+   `layer.orientation` and swaps its main and cross axes; a pie is `false`,
+   because slices sit around a circle rather than along an axis.
+6. Register a braille encoder in the `encoders` Map in `src/service/braille.ts`.
+   **An unregistered type is silent**, not a crash: the guarded early return
+   leaves braille users with an empty display and no error anywhere. Reuse an
+   existing encoder when the state shape matches — a pie's single row of
+   magnitudes is the `BarBrailleEncoder`'s input exactly.
+7. Confirm the audio, text, highlight, and high-contrast services handle it, and
+   document the braille encoding in `docs/BRAILLE.md`.
 
 ## Ownership-aware highlight disposal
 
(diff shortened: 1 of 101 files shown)


jooyoungseo:
Thanks for the review. Pushed 9a36cec.

**`docs/SCHEMA.md` indentation** — fixed. The cause was a copy-paste from the `line`/`step` examples, whose data is an array of arrays and so nests one level deeper; the pie block now matches the `dodged_bar` example and its own `axes` sub-block.

**Two items I'm deliberately not fixing here**, and I agree with your reasoning on both:

- The gap-slice `Units is NaN` wording is shared with `BarTrace` (`Revenue is null`), so the fix belongs in `TextService.formatSingleValue` and would benefit every trace. Cross-cutting; follow-up issue.
- Generic `Position is X of Y` stays, for the announcement-grammar consistency you note.

**Negative values** — also follow-up. Worth noting the binding PRs took different positions on this, which is itself worth a decision: r-maidr's ggplot2 path reports `ymax - ymin` (the magnitude the wedge actually subtends, so a negative datum silently emits its absolute value), while py-maidr raises matplotlib's own error. I'd rather settle that once, across all three, than pick a side per-repo now.

**One thing your review prompted that turned out to be a real defect.** Auditing the adapter DOM assumptions that were never exercised against a live library, I found AnyChart's `findPieWedgeLayer` fell back to the whole `<svg>` whenever no `ac_layer_*` group held an arc path — i.e. *precisely* when the wedge-layer assumption has failed. The caller then stamped the first N arc-shaped paths found anywhere, so a legend marker or a rounded frame could be labelled as a slice. That's a silent mislabel, which is the one failure mode that matters for accessibility: the announcement and the highlight would disagree with no error anywhere.

It now returns `null` in that case and the caller warns and stamps nothing. The structure-free fallback (a chart with no `ac_layer_*` groups at all) is kept, matching `findCandlestickPathLayer`.

Same commit also fixes a filter disagreement: `buildPieLayer` dropped non-finite rows while the stamper counted unfiltered rows, so a pie with a null-valued row expected one more wedge than it emitted — a spurious warning at best, an off-by-one highlight at worst. Both sides now share one predicate. Two unit tests added.

**amCharts** I audited and left alone, because it already degrades safely at every point I could find — worth recording where, since it isn't obvious:

- Wrong `PieChart.className` → `findCharts` returns `[]` → explicit named throw at `adapter.ts:102` / `binder.tsx:294`. Loud, never a mislabel.
- Wrong `PieSeries.className` → `classifySeriesKind` defaults to `'bar'`, but `extractBarPoints` reads `categoryX`/`valueY` while an am5percent item carries `category`/`value`, so every row is skipped and `noSupportedDataError()` throws. The in-code comment worrying it becomes "a bar chart of its slices" is wrong — it can't.
- Missing `dataItem.get('slice')` → guarded by `slice?.globalBounds?.()` → no rect pushed, `overlay.clear()`. No throw, right data item, only geometry missing — same posture as the existing `columnRect`/`pointRect` paths.
- Index alignment is exact: `filterPieItems` and `extractPiePoints` use the same `Number.isFinite(Number(v))` predicate, so the two lists cannot drift.

**What I could not verify, stated plainly:** neither adapter's wedge-DOM assumption has been checked against the real library. This sandbox has no network — loading `examples/amcharts-pie.html` gives `ERR_CONNECTION_RESET` and `am5 is not defined`. What I changed is the blast radius, not the assumption. The [PR preview](https://pr-767.maidr-preview.pages.dev) has both examples if someone with network access wants to confirm the wedges highlight; that's the one manual check I'd want before merge.

Two known limitations I'd rather log than paper over: an amCharts pie in a **multi-panel** root never highlights (a pie has no `plotContainer`, so `readPlotBounds` returns null and `binder.tsx:135` suppresses the overlay — single pies are fine), and whether `anychart-base.min.js` bundles the pie module is unknowable offline (it fails loudly if not, and only affects that example page).

Verification after this commit: `lint:fix` clean, `type-check` clean, `npm test` **1781 + 61 passing**, `npm run build` all 12 targets, and the full Playwright suite **350 passed** against the installed chromium.

---

Issue #770, opened 2026-08-09
fix(amcharts): a pie in a multi-panel root never highlights


jooyoungseo:
Found while auditing the amCharts adapter during #767. Not a regression — it falls out of pre-existing panel logic meeting the new pie path.

## The problem

`readPlotBounds` (`src/adapters/amcharts/geometry.ts:20-22`) returns `null` for a chart with no `plotContainer`. An am5percent `PieChart` has none — it has no plot area, which is correct for a pie. `binder.tsx:135` then suppresses the overlay whenever plot bounds are missing.

Net effect: a **single** amCharts pie highlights normally, but a pie sitting in a multi-panel root gets no visual highlight at all. Navigation, audio, braille and announcements are unaffected.

The suppression is right for XY panels — without plot bounds you cannot clip an overlay to the correct panel, and drawing outside it is worse than drawing nothing. It just does not fit a chart type that legitimately has no plot container.

## Suggested direction

Give the pie path its own bounds source rather than loosening the XY guard. A pie's wedges have `globalBounds()`, so the panel rectangle can be derived from the union of its slices, or from the chart container itself.

## Why it matters

Highlighting is the modality low-vision users with residual sight rely on. Losing it silently in one layout is exactly the kind of gap that does not show up in a test asserting announcements.

jooyoungseo:
Verified against the real library at last — amcharts5 **5.20.1** from npm, bundled with esbuild and driven in Chromium. The npm package ships ES modules with no browser bundle, which is why this sat unverified; bundling it is the step that was missing.

**Two things in this issue are wrong, and the second one got implemented.**

## 1. It is not limited to multi-panel

> a **single** amCharts pie highlights normally, but a pie sitting in a multi-panel root gets no visual highlight at all

No amCharts pie has ever highlighted, in any layout. The multi-panel suppression at `binder.tsx:135` is only the first of two places the highlight dies. The second is `sliceRect` (`overlay.ts:194`), which reads the same geometry and is on the path for a single chart too. A single pie also draws nothing.

## 2. The suggested direction was the bug

> A pie's wedges have `globalBounds()`, so the panel rectangle can be derived from the union of its slices

That is exactly what `readSliceBounds` was written to do — and it is why the fix did not work. A `Slice` reports a **degenerate point at its own centre**:

```json
{ "bounds": { "left": 225, "top": 200, "right": 225, "bottom": 200 },
[...]

PR #773, opened 2026-08-09
fix: give blank axis labels a name, and stamp only the wedges a pie fills

jooyoungseo:
# Pull Request

## Description

Two defects, both found the same way: by building the bundle from `main` and driving the emitted pages in a real browser. Nothing in the pie work across the three repos had ever been rendered — every assertion was at the schema or SVG level — and both of these are invisible until it is.

## Related Issues

Refs xability/maidr#770 (the amCharts/AnyChart verification reviewers asked for twice) and xability/py-maidr#323.

### 1. Every Base R chart announced its points with the nouns missing

Heard by a screen reader:

```
 is Apples,  is 30, Percentage is 30.0%
 is A,  is 3
```

`AbstractTrace` used `??`:

```ts
this.xAxis = layer.axes?.x?.label ?? DEFAULT_X_AXIS;
```

A producer with no axis name to give writes one of two spellings of "none", and `??` catches only one. `undefined` gets the default; an **empty string is a value** and passes straight through. r-maidr emits `label: ""` because `graphics::pie()`, `barplot()`, `hist()` and `boxplot()` genuinely have no labels to report — so this is every Base R chart whose author did not pass `xlab`/`ylab`, which is the common case, and not pie-specific at all.

A blank name is worse than a generic one: `"X is Apples"` at least tells the listener which axis is being read.

Fixed in the model rather than per-binding, so it holds for hand-authored MAIDR JSON too. **After:** `X is Apples, Y is 30, Percentage is 30.0%`.

A producer-side follow-up — Base R emitting something better than the generic pair, the way py-maidr's pie uses `"Category"`/`"Value"` — belongs in r-maidr. This is the renderer-side floor that stops the announcement being broken regardless of producer.

### 2. AnyChart pie stamped candidates that were not data

Running the real library (anychart 8.13, vendored from npm, no CDN) surfaced two things static reading could not:

**Every stamper was tried on every chart.** Harmless for XY, but a pie has no series API — `getSeriesCount` is not a function on one — so the bar and line stampers each threw and warned before the pie stamper did the real work. Two console warnings on every correctly rendered pie, in exactly the place someone debugging a genuine stamping failure would look. Stampers now dispatch on what the chart is, which also collapses the duplicated try/catch ladder at both call sites.

**AnyChart draws every slice twice** — once filled in the slice's colour, once with `fill="none"` for the stroke. Both sit in the same layer, both carry an arc command, so a four-slice pie offered **eight** candidates and the count guard fired on a chart that was working:

```
[maidr/anychart] Expected 4 pie wedges but found 8 after filtering.
```

Taking the first N in DOM order picked the fills only because AnyChart emits them first. That is an ordering accident, not a guarantee — were it ever to emit outlines first, every highlight would land on an invisible path while the announcement carried on naming the right slice. That is the silent mislabel this lookup exists to prevent, so correctness here should not rest on emission order.

Browser-verified, Chromium, no CDN:

| | Before | After |
|---|---|---|
| Base R pie | `" is Apples,  is 30, Percentage is 30.0%"` | `"X is Apples, Y is 30, Percentage is 30.0%"` |
| Base R barplot | `" is A,  is 3"` | `"X is A, Y is 3"` |
| AnyChart pie | 8 candidates, 2 spurious warnings | 4 wedges `0-0`–`0-3`, 4 distinct fills, **no warnings** |

## Checklist

No documentation change: both restore behaviour the docs already describe.

## Additional Notes

| Gate | Result |
|---|---|
| `npm test` | 1806 + 61 passing (was 1800 + 61) |

Six new tests in `test/model/blankAxisLabel.test.ts`: label present, axes absent entirely, empty string, whitespace, a legitimately short label (`"n"`, `"0"` — neither should fall back), and per-axis rather than all-or-nothing fallback. The existing 42 AnyChart tests still pass.

**Also verified while I had the harness up**, since none of it had been checked before: matplotlib pie, ggplot2 `coord_polar` pie, Base R `pie()` and a mixed bar+pie figure all render, announce, braille (`⠤⠉⣀`, matching the unit test) and highlight, with no console errors. Values survive matplotlib's normalisation — `30/50/20`, not `0.3/0.5/0.2` — which was the central design bet of the py-maidr work and had never been confirmed at runtime.

**Still not verified:** amCharts. Its npm package ships ES modules rather than a browser bundle, so unlike AnyChart it cannot be dropped into the example page without a bundling step. xability/maidr#770's multi-panel pie highlight remains unconfirmed against the real library.

### DIFF
--- a/src/adapters/anychart/converters.ts
+++ b/src/adapters/anychart/converters.ts
@@ -2036,6 +2036,50 @@ function findPieWedgeLayer(svg: SVGElement): Element | null {
  *
  * On any other chart type this is a no-op.
  */
+/**
+ * Stamps the highlight attributes this chart's own kind needs.
+ *
+ * Every stamper used to be tried on every chart, each in its own `try`. That
+ * is harmless for an XY chart — a bar stamper finds no line series and
+ * returns — but a pie carries no series API at all: `getSeriesCount` is not a
+ * function on one, so the bar and line stampers each threw and warned before
+ * the pie stamper did the real work. Two console warnings on every correctly
+ * rendered pie, which is noise in exactly the place someone debugging a
+ * genuine stamping failure would look.
+ *
+ * Asking what the chart is costs one call and makes the warnings mean
+ * something again: past this point, a warning is a stamper failing at a job
+ * that was actually its own.
+ *
+ * @param chart - The AnyChart instance being bound
+ * @param svg - The rendered SVG to stamp
+ * @param stampPrefix - Panel token prefix, empty for a single chart
+ */
+function stampChartAttributes(
+  chart: AnyChartInstance,
+  svg: SVGElement,
+  stampPrefix = '',
+): void {
+  const stampers: [string, typeof stampPieAttributes][] = isPieChart(chart)
+    ? [['pie', stampPieAttributes]]
+    : [
+        ['bar', stampBarAttributes],
+        ['line', stampLineAttributes],
+        ['scatter', stampScatterAttributes],
+        ['box', stampBoxAttributes],
+        ['heatmap', stampHeatmapAttributes],
+        ['candlestick', stampCandlestickAttributes],
+      ];
+
+  for (const [kind, stamp] of stampers) {
+    try {
+      stamp(chart, svg, stampPrefix);
+    } catch (err) {
+      console.warn(`[maidr/anychart] Failed to stamp ${kind} attributes:`, err);
+    }
+  }
+}
+
 function stampPieAttributes(
   chart: AnyChartInstance,
   svg: SVGElement,
@@ -2075,6 +2119,18 @@ function stampPieAttributes(
     // `fill-opacity`, so any value below 1 is the overlay.
     const fillOpacityAttr = path.getAttribute('fill-opacity');
     if (fillOpacityAttr !== null && Number.parseFloat(fillOpacityAttr) < 1)
+      continue;
+    // Skip the wedge outlines. AnyChart draws every slice twice — once filled
+    // in the slice's colour, once with `fill="none"` for the stroke — so a
+    // four-slice pie offers eight arc paths. Both halves sit in the same
+    // layer and both carry an arc command, and only the fill is the datum.
+    //
+    // Taking the first N in DOM order happens to pick the fills, because
+    // AnyChart emits them first. That is an ordering accident, not a
+    // guarantee: were it ever to emit outlines first, every highlight would
+    // land on an invisible path while the announcement carried on naming the
+    // right slice — the silent mislabel this whole lookup exists to avoid.
+    if (path.getAttribute('fill') === 'none')
       continue;
     candidates.push(path);
   }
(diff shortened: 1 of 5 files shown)


jooyoungseo:
Both reviews addressed in `e33680b` and `792da9a`. Three findings were real; two I checked and found do not hold. Gates re-run after the changes: `lint:fix`, `type-check`, `build` clean; **1817 + 61 passing** (was 1806 + 61).

## Acted on

**Missing z-axis coverage — correct, and it led to a real bug.**

The fallback shipped covering x, y and z, but only x and y were tested. Adding the z cases sent me to check who else reads a z label, and `LineTrace` does *not* go through `this.z`:

```ts
private get groupLabel(): string {
  return this.layer.axes?.z?.label ?? TYPE;   // TYPE = 'Group'
}
```

It resolves from the layer directly, against a trace-specific default rather than the generic `"Level"`. So the fallback never reached it, and a line chart whose producer wrote `z.label: ""` still announced its group with the noun missing — through `currentGroup` (`line.ts:148`) and the text state's `zLabel` (`line.ts:320`). `named` is now exported and used there.

Eight new cases: four for `this.z`, four for the line's group label. **The two blank/whitespace line cases fail against `?? TYPE`** — I reverted the fix to confirm rather than assume.

**No test for the AnyChart fix — correct.** `appendPieWedges` now optionally emits the stroke-only twin, so the eight-candidates-for-four-slices case is reproducible. It is parametrised over **both** emission orders — the fill-first one AnyChart uses today, and the reverse — because the point of filtering on `fill="none"` rather than taking the first N in DOM order is precisely that correctness must not rest on that order; a test that only reproduced today's order would not pin the thing the fix is for. A second test asserts a working pie bind logs nothing at all.

Verified both bite: **reverting the outline filter fails 3 tests, reverting the dispatch fails 2.**

**JSDoc narrating history — fair.** Trimmed to the durable why (two spellings of "none", `??` catches one; whitespace counts as blank) and dropped the reproduction of the old broken string. I could not find the guidance quoted from `CLAUDE.md` in this repo — the point stands on its own, so I acted on it regardless.

## Checked, and these do not hold

**The `??` pattern at the three other sites is not the same defect.** I read each before deciding, and in every case the blank is already handled — differently, but deliberately:

- `plot.ts:138-139` — `DEFAULT_FIGURE_AXIS` is `''`. The fallback *is* the empty string, so routing it through `named()` is a literal no-op: a figure has no generic axis name to fall back to by design.
- `abstract.ts:403` and `plot.ts:135` (subplot and figure title) — every consumer goes through `isAuthoredTitle`, which already opens with `title.trim() !== ''`. A blank title is rejected as unauthored on exactly the same footing as the placeholder, and its docstring names itself the single source of truth for the rule "so it cannot drift across copies". Adding a second, earlier blank check would put a copy of that rule back.

Figure `subtitle`/`caption` are the one place the argument would apply, and they are neither axis labels nor titles — outside this PR either way.

**The scope finding is mistaken about the PR.** The title is *"give blank axis labels a name, **and stamp only the wedges a pie fills**"*, and section 2 of the body — "AnyChart pie stamped candidates that were not data" — is entirely about `4f7cc6e`, with the before/after in the results table. The squashed changelog entry carries both halves, which is the outcome the finding was protecting.

## On the gates

Both reviews noted they could not execute `npm test` in the sandbox. Everything above was run locally, and each new test was checked against the pre-fix code rather than only against the fix.

---

PR #775, opened 2026-08-09
fix(amcharts): measure a pie wedge instead of believing its reported box

### LINKED ISSUE #774: amCharts pie highlight never draws: a Slice reports a degenerate globalBounds()
Verified against the real library — amcharts5 **5.20.1** from npm, bundled with esbuild, driven in Chromium. This is the check #770 asked for and could not previously run, because the amCharts package ships ES modules rather than a browser bundle.

**Everything except the visual highlight works.** The gap is highlight-only.

## What works

A two-panel root, one `PieChart` per panel (Fruit 30/50/20, Veg 40/60):

```
lobby -> subplot 1   Subplot 1 of 2: This is a pie plot. Press 'ENTER' to select this subplot.
enter                Entered subplot 1 of 2, pie plot.
slice 1              Label is Apples, Value is 30, Percentage is 30.0%
slice 2              Label is Bananas, Value is 50, Percentage is 50.0%
slice 3              Label is Cherries, Value is 20, Percentage is 20.0%
braille              ⠤⠉⣀

-> subplot 2         Subplot 2 of 2: This is a pie plot. Press 'ENTER' to select this subplot.
enter                Entered subplot 2 of 2, pie plot.
slice 1              Label is Carrots, Value is 40, Percentage is 40.0%
slice 2              Label is Peas, Value is 60, Percentage is 60.0%
braille              ⣀⠉
```

Values arrive raw (30, not 0.3), percentages are right, multi-panel navigation moves between panels correctly, and both braille strings match the unit-test encodings. No console errors.

## What is broken

The overlay container is created and never receives a box — on any slice, in either panel:

```
1 container(s), boxes: NONE
```

queried by the adapter's own `[data-maidr-amcharts-overlay]` / `[data-maidr-amcharts-highlight]` attributes.

## Root cause

`readSliceBounds` (`src/adapters/amcharts/geometry.ts:55`) documents its premise as:

> Each slice is a sprite with `globalBounds()`, and the union of those boxes is the rectangle the pie occupies

**That premise does not hold.** Probing every slice directly:

```json
{
  "bounds": { "left": 225, "top": 200, "right": 225, "bottom": 200 },
  "extra":  { "w": 0, "h": 0,
              "radius": 160, "arc": 108, "startAngle": -90,
              "dispW": { "left": 10000000, "top": 10000000,
                         "right": -10000000, "bottom": -10000000 } }
}
```

Every slice returns a **degenerate point at the pie's centre** — `left === right`, `top === bottom`, `width() === height() === 0`. The display object's local bounds are amCharts' empty sentinel (`±10000000`), i.e. no bounds were ever recorded: a `Slice` paints through a custom draw callback rather than through primitives that feed the bounds accumulator.

Measured after `datavalidated`, then again after 3s plus a viewport resize and another 1.5s. It is not a layout-timing artifact.

Consequences, in order:

1. `readSliceBounds` unions zero-area boxes → a **zero-area** panel rectangle.
2. `applyHighlight` (`binder.tsx:133`) takes that as `plotBounds` — non-null, so the multi-panel suppression does not fire.
3. `sliceRect` (`overlay.ts:194`) reads the same `globalBounds()`, so the per-slice rect is degenerate too.
4. `intersectRect` collapses it → `rects` empty → `overlay.clear()`.

So there are **two** independent uses of the same false premise. Fixing only the panel bounds would still leave every highlight a zero-size dot at the centre.

## The geometry is available

The slice does not report a box, but it reports everything needed to compute one:

| Field | Value | Meaning |
|---|---|---|
| `globalBounds()` | `225, 200` | the centre, and it is **correct** — 900px wide root, two 450px halves, so 225 and 675 |
| `radius` | `160` | outer radius |
| `arc` | `108` | sweep in degrees — 30/100 × 360, exactly right |
| `startAngle` | `-90` | 12 o'clock |

A wedge's bounding box follows from centre, inner/outer radius, and the angular span: the four arc endpoints, plus each cardinal angle (0°/90°/180°/270°) that falls inside the sweep, plus the centre when `innerRadius` is 0.

## Not a regression

This path has never worked with the real library — the adapter's pie support shipped verified only against fakes, whose `globalBounds()` returned whatever the test set. The existing amCharts unit tests pass precisely because their fake slices report real boxes, which is the one thing amCharts does not do.

Tests for the fix should use a slice that reports what amCharts actually reports: a degenerate `globalBounds()` plus `radius`/`arc`/`startAngle`.

## Reproduction

```js
import * as am5 from '@amcharts/amcharts5';
import * as am5percent from '@amcharts/amcharts5/percent';
import { bindAmCharts } from 'maidr/amcharts';

const root = am5.Root.new('chartdiv');
const chart = root.container.children.push(am5percent.PieChart.new(root, {}));
const series = chart.series.push(am5percent.PieSeries.new(root, {
  categoryField: 'name', valueField: 'value',
}));
series.data.setAll([{ name: 'Apples', value: 30 }, { name: 'Bananas', value: 50 }]);
bindAmCharts(root);

// after layout:
series.dataItems[0].get('slice').globalBounds();  // -> a point, not a box
```

Refs #770.

jooyoungseo:
# Pull Request

## Description

Closes #774. Refs #770 — this is what that issue asked for, now that the verification is possible.

The amCharts pie highlight **never drew**. Not misplaced, not intermittent: the overlay container was created and never received a single box, on any slice, in any panel. Everything else worked, which is why nothing caught it.

## Related Issues

Closes #774. Refs #770.

### How it was found

#770 asked for verification against the real library, twice, and it could not run: amcharts5 ships ES modules rather than a browser bundle, so there was no CDN drop-in. Bundling it with esbuild made it possible. Two `PieChart` panels in one root, amcharts5 **5.20.1** from npm, driven in Chromium.

**Everything except the highlight was already correct:**

```
lobby -> subplot 1   Subplot 1 of 2: This is a pie plot. Press 'ENTER' to select this subplot.
enter                Entered subplot 1 of 2, pie plot.
slice 1              Label is Apples, Value is 30, Percentage is 30.0%
slice 2              Label is Bananas, Value is 50, Percentage is 50.0%
slice 3              Label is Cherries, Value is 20, Percentage is 20.0%
braille              ⠤⠉⣀
-> subplot 2         Subplot 2 of 2: This is a pie plot. Press 'ENTER' to select this subplot.
slice 1              Label is Carrots, Value is 40, Percentage is 40.0%
slice 2              Label is Peas, Value is 60, Percentage is 60.0%
braille              ⣀⠉
```

Raw values, correct percentages, multi-panel navigation moving between panels, and both braille strings matching the unit-test encodings. Queried by the adapter's own attributes, the highlight was:

```
1 container(s), boxes: NONE
```

### Root cause

`readSliceBounds` documented its premise:

> Each slice is a sprite with `globalBounds()`, and the union of those boxes is the rectangle the pie occupies

**It does not hold.** Probing every slice:

```json
{ "bounds": { "left": 225, "top": 200, "right": 225, "bottom": 200 },
  "extra":  { "w": 0, "h": 0, "radius": 160, "arc": 108, "startAngle": -90,
              "dispW": { "left": 10000000, "right": -10000000, ... } } }
```

A degenerate **point at the wedge's own centre**. The display object's local bounds are amCharts' empty sentinel — nothing was ever recorded, because a `Slice` paints through a draw callback rather than through primitives that feed the bounds accumulator. Measured after `datavalidated`, then again after 3s plus a viewport resize; always a point.

Two consequences, from **two independent uses of the same premise**:

1. `readSliceBounds` unioned zero-area boxes → a zero-area panel rectangle. Non-null, so the binder's multi-panel suppression never fired.
2. `sliceRect` read the same `globalBounds()`, so each slice's own rect was degenerate too, and `intersectRect` clipped it to nothing.

Fixing only the first would have left every highlight a dot at the centre.

### The fix

The centre it reports is correct, and the extent is in the settings, so the box is recoverable. `wedgeBounds` walks the wedge's corners — both arc ends at each radius — and adds **every cardinal angle the sweep crosses**, which is where the arc rather than its endpoints reaches furthest out. A wedge with no inner radius gets its centre from the inner corners at radius zero; a full turn short-circuits to the circle.

The cardinal step is the part that is easy to get wrong: for a wedge spanning −45° to 45°, both endpoints sit at x = 135.36 while the arc bulges to 150. A box from endpoints alone stops short of the widest part of the wedge.

`sliceExtent` prefers that and falls back to a reported box **with area**. So a build whose slices do report real boxes keeps working, and a degenerate box can never be believed again — the area test is what makes the fallback safe rather than a way back into the bug.

Re-run against the real library after the fix:

| Slice | Highlight box |
|---|---|
| Apples (30%) | `225px/40px 160 × 209.443` |
| Bananas (50%) | `65px/150.557px 312.169 × 209.443` |
| Cherries (20%) | `72.831px/40px 152.169 × 160` |
| Carrots (40%, panel 2) | `675px/40px 160 × 289.443` |
| Peas (60%, panel 2) | `515px/40px 254.046 × 320` |

Checked by hand rather than eyeballed. Apples is 108° from 12 o'clock, centre (225, 200), r = 160:

- arc ends (225, 40) and (377.2, 249.4)
- cardinal 0° at (385, 200)
- centre (225, 200)

→ left 225, top 40, width 160, height 209.44. That is the browser's output to the digit. Panel 2's centre lands at x = 675, which is right for a 900px root split into two 450px halves.

## Checklist

No documentation change: this restores behaviour the docs already describe.

## Additional Notes

| Gate | Result |
|---|---|
| `npm test` | **1827 + 61** passing (was 1817 + 61) |

**Why the existing tests could not have caught this.** All 42 amCharts tests passed before and after. Their fake slices report real boxes — the one thing amCharts does not do — so they exercised a shape the library never produces. `test/adapters/amcharts/wedgeBounds.test.ts` models what it actually reports: a degenerate `globalBounds()` plus `radius` / `innerRadius` / `startAngle` / `arc`. Ten cases, including the cardinal-crossing bulge, a donut wedge that must *not* swallow the centre, a sweep expressed past a full turn, and an XY column that must return `null` rather than gain a wedge box somewhere the data is not.

The three existing `readSliceBounds` cases still pass unchanged — they now document the reported-box fallback rather than the premise.

**Noticed while verifying, not fixed here:** returning to the figure lobby leaves the last slice's box on screen, and it stays there while moving to another panel. `applyHighlight` is the only thing that clears the overlay and it runs on navigation events, so there is no hook for leaving a subplot — structural, and it applies to every chart type this adapter supports rather than to pie. It was simply never observable for pie before, since no box was ever drawn. Filed separately rather than folded in.

---

### DIFF
--- a/src/adapters/amcharts/geometry.ts
+++ b/src/adapters/amcharts/geometry.ts
@@ -47,9 +47,9 @@ function numberSetting(slice: AmSprite, key: string): number | null {
 /**
  * The box a sprite reports, when it reports one with area.
  *
- * The area test is what makes this safe as a fallback: a `Slice` answers a
- * degenerate point, and unioning those is the bug in #774. Kept for any build
- * whose slices do report a real box but no radius to measure instead.
+ * The area test is what makes this safe as a fallback: a slice that reports a
+ * point rather than a box is rejected here rather than believed. Kept for any
+ * build whose slices do report a real box but no radius to measure instead.
  */
 function reportedBox(slice: AmSprite): AmBounds | null {
   const box = slice.globalBounds?.();
@@ -131,8 +131,14 @@ export function wedgeBounds(slice: AmSprite): AmBounds | null {
     }
   }
 
-  // Both radii at the arc ends; only the outer radius at a cardinal, since the
-  // inner point lies in the same direction and so is never further out.
+  // Both radii at the arc ends, but only the outer radius at a cardinal in
+  // between. A cardinal is where cos or sin is at a critical point, so it is
+  // the angle that governs one extreme of the box — and there the larger
+  // radius always reaches further in that direction. Sampling the inner
+  // radius at an interior cardinal could therefore only ever repeat a bound
+  // the outer one already set. At the sweep's two ends there is no such
+  // critical point, so the inner corners are their own extremes and do need
+  // sampling: that is what keeps a donut wedge from swallowing the centre.
   const xs: number[] = [];
   const ys: number[] = [];
   const add = (angle: number, r: number): void => {
--- a/src/adapters/amcharts/geometry.ts
+++ b/src/adapters/amcharts/geometry.ts
@@ -36,16 +36,140 @@ export function readPlotBounds(chart: AmChart): AmBounds | null {
   return null;
 }
 
+/** Cardinal directions, where a circle reaches its extreme in x or y. */
+const CARDINALS = [0, 90, 180, 270];
+
+function numberSetting(slice: AmSprite, key: string): number | null {
+  const value = slice.get?.(key);
+  return typeof value === 'number' && Number.isFinite(value) ? value : null;
+}
+
 /**
- * Read a pie panel's bounds as the union of its wedges' global boxes.
+ * The box a sprite reports, when it reports one with area.
+ *
+ * The area test is what makes this safe as a fallback: a `Slice` answers a
+ * degenerate point, and unioning those is the bug in #774. Kept for any build
+ * whose slices do report a real box but no radius to measure instead.
+ */
+function reportedBox(slice: AmSprite): AmBounds | null {
+  const box = slice.globalBounds?.();
+  if (!box || ![box.left, box.right, box.top, box.bottom].every(Number.isFinite)) {
+    return null;
+  }
+  const hasArea = box.left !== box.right && box.top !== box.bottom;
+  return hasArea ? box : null;
+}
+
+/**
+ * The extent of one pie wedge: its own geometry, or the box it reports.
+ *
+ * Prefers {@link wedgeBounds} so a real amCharts slice is measured rather than
+ * believed, and falls back to a reported box with area for a sprite that
+ * carries no radius to measure.
+ */
+export function sliceExtent(slice: AmSprite): AmBounds | null {
+  return wedgeBounds(slice) ?? reportedBox(slice);
+}
+
+/**
+ * The bounding box of one pie wedge, computed from what a `Slice` reports.
+ *
+ * A `Slice` paints through a draw callback rather than through primitives that
+ * feed amCharts' bounds accumulator, so `globalBounds()` answers a *degenerate
+ * box at the wedge's own centre* — `left === right`, `top === bottom`, and
+ * `width()`/`height()` both zero. Verified against amcharts5 5.20.1: measured
+ * after `datavalidated`, again after a resize, always a point. Every highlight
+ * derived from it collapsed and the overlay drew nothing at all (#774).
+ *
+ * The centre it reports is correct, though, and the wedge's extent is in its
+ * settings, so the box is recoverable: walk the wedge's corners — both arc
+ * ends at each radius — and add every cardinal angle the sweep crosses, which
+ * is where the arc, not its endpoints, reaches furthest out. A wedge with no
+ * inner radius includes its centre, which the inner-radius corners supply at
+ * radius zero.
+ *
+ * @param slice - The `Slice` sprite behind one pie data item
+ * @returns Its box in root-relative CSS px, or `null` when the sprite reports
+ *   no usable centre or radius (an XY chart's sprites, or a pie asked before
+ *   its first layout).
+ */
+export function wedgeBounds(slice: AmSprite): AmBounds | null {
+  const box = slice.globalBounds?.();
+  if (!box || ![box.left, box.right, box.top, box.bottom].every(Number.isFinite)) {
+    return null;
+  }
+  const cx = (box.left + box.right) / 2;
+  const cy = (box.top + box.bottom) / 2;
+
+  const radius = numberSetting(slice, 'radius');
+  if (radius === null || radius <= 0) {
+    return null;
+  }
+  const inner = Math.max(numberSetting(slice, 'innerRadius') ?? 0, 0);
+  const start = numberSetting(slice, 'startAngle') ?? 0;
+  const arc = numberSetting(slice, 'arc') ?? 360;
+
+  // A full circle needs no angular reasoning, and asking for it would make
+  // every cardinal a candidate anyway.
+  if (Math.abs(arc) >= 360) {
+    return { left: cx - radius, top: cy - radius, right: cx + radius, bottom: cy + radius };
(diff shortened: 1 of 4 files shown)


jooyoungseo:
Both suggestions taken, in `a55973a`. `npm test` → **1830 + 61** (was 1827 + 61); lint, type-check and build clean.

**The cardinal invariant.** Your reasoning is right and it is exactly the thing a future edit would "fix" into redundancy, so it is now written down at the call site rather than left to be re-derived:

> A cardinal is where cos or sin is at a critical point, so it is the angle that governs one extreme of the box — and there the larger radius always reaches further in that direction. Sampling the inner radius at an interior cardinal could therefore only ever repeat a bound the outer one already set. At the sweep's two ends there is no such critical point, so the inner corners are their own extremes and do need sampling: that is what keeps a donut wedge from swallowing the centre.

Stating both halves, since the asymmetry is the confusing part — the reason interior cardinals *don't* need it is the same reason the endpoints *do*.

**Direct `sliceExtent` cases.** Three added. The one worth having is the precedence case you named — a sprite reporting a radius **and** a real box:

```ts
globalBounds: () => ({ left: 0, top: 0, right: 999, bottom: 999 }),
get: key => ({ radius: 50, startAngle: -90, arc: 90 })[key],
// -> [499.5, 449.5, 549.5, 499.5], the wedge around the box's centre
```

Nothing pinned that before, so a refactor could have swapped the order and every test would still have passed while highlights silently became 999×999 panel-sized boxes. Plus the fallback case (box, no radius) and the refusal case (degenerate box, no radius) so all three branches are stated directly rather than inferred through `readSliceBounds`.

**On the issue-number nit** — took it partly. Dropped `#774` from `reportedBox`'s docstring, where the load-bearing content is just "a slice can report a point, so test for area". Kept it on `wedgeBounds` and on the test helper: that one records an external library's undocumented behaviour, which nobody can re-derive from this repo and which the issue holds the measurements for. A number that points at evidence outside the codebase seems worth more than one pointing at history inside it.

**Untouched `globalBounds()` call sites** — agreed, and that was deliberate: `columnRect` and the bullet geometry are XY sprites drawn from primitives, which is the class that does report real boxes. Narrowing the fix to the slice path is why the other 42 amCharts tests never moved.

---

PR #777, opened 2026-08-09
fix(model)!: tell the navigate callback when the selection ends

### LINKED ISSUE #776: amCharts highlight stays on screen after leaving a subplot, including over the wrong panel
Noticed while verifying #774 against the real library. Not caused by that fix — exposed by it.

## Behaviour

In a two-panel amCharts root, after navigating to a data point and pressing <kbd>Escape</kbd>:

```
slice 3          Label is Cherries, Value is 20, Percentage is 20.0%
                 highlight: 72.831px/40px 152.169x160

back to lobby    Returned to figure overview, subplot 1 of 2.
                 highlight: 72.831px/40px 152.169x160     <- unchanged
-> subplot 2     Subplot 2 of 2: This is a pie plot. ...
                 highlight: 72.831px/40px 152.169x160     <- still panel 1's wedge
enter            Entered subplot 2 of 2, pie plot.
                 highlight: 72.831px/40px 152.169x160     <- still panel 1's wedge
```

The box only moves once a data point inside panel 2 is selected. Until then it sits over **panel 1**, while the announcements are describing panel 2.

For a sighted low-vision user driving by keyboard, that is the highlight pointing at one panel while the screen reader names another — the mislabel the highlight exists to prevent, one level up.

## Cause

`applyHighlight` (`src/adapters/amcharts/binder.tsx`) is the only thing that clears the overlay, and it runs on navigation **events**. Leaving a subplot for the figure lobby produces no target to draw, but also no event that reaches the binder, so the last box drawn simply stays.

## Not pie-specific

The overlay is shared by every kind the adapter supports, and nothing about the clearing path is kind-dependent — an XY panel's column box should linger the same way. Pie is only where it became *visible*, because before #774 no pie box was ever drawn to linger.

## Fix shape

The binder needs a signal for "no data point is selected" — the lobby — and should clear on it, the same way `applyHighlight` already clears when `targets.length === 0`. Worth checking at the same time whether the Chart.js adapter, which uses the same overlay approach, has the same gap.

Refs #774, #770.

jooyoungseo:
# Pull Request

> [!IMPORTANT]
> **This must be squashed with the `BREAKING CHANGE:` footer intact, or it will not release at all.**
>
> I originally wrote here that the `!` in `fix(model)!:` is what cuts the major. **That is wrong for this repository**, and the failure is silent. `.releaserc` configures `@semantic-release/commit-analyzer` with no `preset`, so the default **angular** preset applies, and its header pattern does not admit `!`. Measured against the installed version:
>
> | Commit message | Release |
> |---|---|
> | `fix(a)!: x` | **null — no release at all** |
> | `fix(a)!: x` + `BREAKING CHANGE:` footer | **major** |
> | `fix(a): x` + `BREAKING CHANGE:` footer | **major** |
> | `fix(a): x` | patch |
>
> So the footer is the only thing that produces a major, and a `!`-marked subject is *worse* than an unmarked one — it drops the commit from analysis entirely. The footer is reproduced verbatim at the bottom of this description so it survives whichever source your squash draws its body from. Repo-wide fix tracked in #778.

## Description

Closes #776.

Leaving a subplot makes the **figure** active rather than a trace, so the trace observers behind `onNavigate` went quiet and said nothing about having stopped. A canvas adapter has no other signal, so it kept the last point's box on screen — and carried it to whichever panel the user moved to next, outlining a chart the box does not belong to.

## Related Issues

Closes #776. Found while verifying #774 in a real browser.

## Changes Made

`Controller.registerNavigateCallback` attached its observer to traces only:

```ts
subplot.traces.forEach(traceRow => traceRow.forEach((trace) => {
  trace.addObserver(observer);
}));
```

There was no way to express "nothing is selected" at all — `NavigateCallback` could only ever say *where* the cursor is, never that it is nowhere. So the figure now reports `null` when it becomes the active element, and both canvas adapters clear on it.

Recording the `null` matters as much as clearing on it: both adapters keep a `lastActive` that the resize hook replays, so without it the next resize would put the stale box back.

**Not amCharts-specific.** Chart.js takes the same route through `createHighlightCallback` (`plugin.tsx:328`), so a multi-panel Chart.js figure stranded its highlight identically. Only amCharts surfaced it, because #774 meant its pie never drew a box to be stranded in the first place. Both adapters are fixed here.

An adapter could not have fixed this alone — `onNavigate` is its entire surface, and observing scope from the model or a service would cross the boundary `.claude/rules/architecture.md` draws.

## Breaking change

`NavigateCallback` becomes:

```ts
export type NavigateCallback = (info: { layerId: string; row: number; col: number } | null) => void;
```

A callback that dereferences its argument without checking will throw on leaving a subplot. TypeScript consumers get a compile error, which is the point; JavaScript consumers get a runtime one.

Blast radius is the two in-repo adapters, both updated here, plus anyone who set `onNavigate` while constructing MAIDR data programmatically — the field is documented as non-serializable and programmatic-only, so it is not reachable from JSON-authored charts.

The alternatives were a second `onNavigateClear?` hook or an `active: boolean` field on the payload; both are additive and neither says what happened as directly. **The choice of the breaking option was the maintainer's**, made explicitly after the trade-offs were laid out in #776.

Chromium, amcharts5 5.20.1, two pie panels. Before → after, at the moment of returning to the lobby:

```
before:  back to lobby  | highlight: 72.831px/40px 152.169px x 160px   <- Cherries, still outlined
after:   back to lobby  | highlight: NONE
```

and it stays `NONE` through moving to panel 2 and entering it, until a slice there is selected — at which point the boxes appear at x=675 / 515, in the right-hand half where panel 2 actually is.

## Checklist

Documentation: the `NavigateCallback` JSDoc now states what `null` means and why a consumer drawing an overlay has to act on it.

## Additional Notes

| Gate | Result |
|---|---|
| `npm test` | **1834 + 61** (was 1830 + 61) |

`test/controller/navigateCallbackClear.test.ts` drives moves through `Context` exactly as a keypress does, with the observers registered the same way `Controller.registerNavigateCallback` registers them — the `Controller` itself needs a live DOM, a Redux store and every service to construct, and the contract under test is between the model's observers and the callback.

**Verified they bite:** with the figure observer removed, the two `null` cases fail and the two position cases pass. The position cases are guards that the change did not cost the existing behaviour, not reproductions — saying so rather than implying all four reproduce the bug.

Two of my initial expectations in that test were wrong and the code was right: the cursor sits *before* the first point on entering a subplot, and before the first panel at the lobby, so the first move lands on index 0 rather than stepping past it. The test now documents that rather than asserting around it.

BREAKING CHANGE: `NavigateCallback` receives `null` when no data point is active. A callback that dereferences its argument without checking will throw on leaving a subplot. Consumers are the in-repo Chart.js plugin and amCharts binder, both updated here, plus anyone who set `onNavigate` when constructing MAIDR data programmatically.

### DIFF
--- a/src/adapters/amcharts/binder.tsx
+++ b/src/adapters/amcharts/binder.tsx
@@ -157,7 +157,7 @@ function applyHighlight(
 function createHighlightCallback(
   navMap: NavMap,
   getOverlay: () => HighlightOverlay | null,
-  recordActive: (event: NavEvent) => void,
+  recordActive: (event: NavEvent | null) => void,
 ): NavigateCallback {
   return (event) => {
     try {
@@ -165,6 +165,14 @@ function createHighlightCallback(
       const overlay = getOverlay();
       if (!overlay)
         return;
+      // `null` is the cursor leaving a subplot for the figure lobby: nothing
+      // is selected, so nothing should be outlined. Recording it also matters
+      // beyond this call -- the resize hook replays `lastActive`, and would
+      // otherwise put the stale box back on the next resize.
+      if (!event) {
+        overlay.clear();
+        return;
+      }
       applyHighlight(overlay, navMap, event);
     } catch {
       // Ignore highlight errors (e.g., during teardown or before layout).
--- a/src/adapters/chartjs/plugin.tsx
+++ b/src/adapters/chartjs/plugin.tsx
@@ -172,19 +172,25 @@ function createHighlightCallback(
   maps: TargetMaps,
   layerDatasetIndices: LayerDatasetIndices,
   getOverlay: () => HighlightOverlay | null,
-  recordActive: (event: NavEvent) => void,
+  recordActive: (event: NavEvent | null) => void,
 ): NavigateCallback {
   return (event) => {
     try {
       recordActive(event);
-      const targets = resolveActiveTargets(
-        layers,
-        maps,
-        layerDatasetIndices,
-        event.layerId,
-        event.row,
-        event.col,
-      );
+      // `null` is the cursor leaving a subplot for the figure lobby: nothing
+      // is selected, so no target is active. `applyHighlight` clears on an
+      // empty list, and recording the `null` keeps the resize hook from
+      // replaying the stale point.
+      const targets = event === null
+        ? []
+        : resolveActiveTargets(
+            layers,
+            maps,
+            layerDatasetIndices,
+            event.layerId,
+            event.row,
+            event.col,
+          );
       applyHighlight(chart, getOverlay(), targets);
     } catch {
       // Silently ignore highlight errors (e.g., after chart destruction)
@@ -317,7 +323,7 @@ function initMaidrForChart(chart: ChartJsChart): void {
 
   // Record the latest MAIDR navigation event on the per-chart binding so
   // the resize hook can replay it after Chart.js re-lays out the canvas.
-  const recordActive = (event: NavEvent): void => {
+  const recordActive = (event: NavEvent | null): void => {
     const b = chartBindings.get(chart);
     if (b)
       b.lastActive = event;
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -3,7 +3,7 @@ import type { AppStore } from '@state/store';
 import type { Disposable } from '@type/disposable';
 import type { Maidr, NavigateCallback } from '@type/grammar';
 import type { Observer } from '@type/observable';
-import type { FigureState, TraceState } from '@type/state';
+import type { TraceState } from '@type/state';
 import { Context } from '@model/context';
 import { Figure } from '@model/plot';
 import { AudioService } from '@service/audio';
@@ -584,6 +584,6 @@ export class Controller implements Disposable {
     // move, so it is what reports the selection ending.
     this.figure.addObserver({
       update: () => callback(null),
-    } as Observer<FigureState>);
+    });
   }
 }
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -3,7 +3,7 @@ import type { AppStore } from '@state/store';
 import type { Disposable } from '@type/disposable';
 import type { Maidr, NavigateCallback } from '@type/grammar';
 import type { Observer } from '@type/observable';
-import type { TraceState } from '@type/state';
+import type { FigureState, TraceState } from '@type/state';
 import { Context } from '@model/context';
 import { Figure } from '@model/plot';
 import { AudioService } from '@service/audio';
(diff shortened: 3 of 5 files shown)


jooyoungseo:
Cast dropped in `084891a` — you were right, `type-check` stays green without it, and the now-unused `FigureState` import went with it.

Thanks for tracing `exitSubplot` → `Figure.notifyStateUpdate` and both resize-replay short-circuits independently rather than taking the description's word for it. The `if (!ov || !lastActive) return` / `if (!active) return` pair is exactly why recording the `null` mattered as much as clearing on it.

## One correction, and it is the important part of this PR

> Commit message is correctly `fix(model)!:` with a `BREAKING CHANGE:` footer — matches `.claude/rules/git-workflow.md`

It matches the rule, but **the rule is wrong**, and I only found that because this PR forced the question. `.claude/rules/git-workflow.md` says:

> `BREAKING CHANGE:` in the footer, or `!` after the type → major release

The second clause does not hold here. Measured against the installed `@semantic-release/commit-analyzer` by calling `analyzeCommits` directly:

| Commit message | Release |
|---|---|
| `feat!: x` | **null** |
| `fix(a)!: x` | **null** |
| `feat: x` | minor |
| `fix: x` | patch |
| `fix(a)!: x` + `BREAKING CHANGE:` footer | major |
| `fix(a): x` + `BREAKING CHANGE:` footer | major |

`.releaserc` sets no `preset`, so the default **angular** preset applies and its header pattern (`/^(\w*)(?:\((.*)\))?: (.*)$/`) does not admit `!`. A marked subject fails to parse, `type` comes out null, no rule matches — so a `!`-only breaking change ships with **no release at all**, not a wrong one. Silent.

commitlint accepts `!` (it is built on the conventionalcommits parser), which is why this passed review and would then have done nothing: the two ends of the pipeline disagree about the same syntax.

This repository squash-merges, so the release commit is composed from the PR title and whichever body the repo is configured to use. The footer is now reproduced verbatim at the end of the PR description so it survives either setting. The `!` stays in the title deliberately: if the footer were ever lost, `!` yields *no release* — visible — where an unmarked subject would yield a quiet **patch** carrying a breaking change. A missing release is noticed; a wrong one is not.

Tracked repo-wide in **#778**, with the recommendation being to set `preset: "conventionalcommits"` so the documented rule becomes true, rather than editing the documentation down to the behaviour.

## On the sandbox

Noted, and worth saying plainly rather than letting it read as a formality: every gate quoted in the description was run locally, and each new test was checked against the pre-fix code rather than only against the fix. With the figure observer removed, the two `null` cases fail and the two position cases pass — the latter are guards that nothing regressed, not reproductions.

---

PR #779, opened 2026-08-10
fix(ci): make the `!` breaking marker actually reach the release

### LINKED ISSUE #778: A `!`-marked breaking commit produces no release at all, and the documented rule says it produces a major
`.claude/rules/git-workflow.md` states:

> - `BREAKING CHANGE:` in the footer, or `!` after the type → major release

The second half is not true here, and it does not merely fail to work — **a `!`-marked subject drops the commit from release analysis entirely**, so a breaking change lands with no version bump and no changelog entry.

## Measured

Against the installed `@semantic-release/commit-analyzer`, calling `analyzeCommits` directly:

| Commit message | Release |
|---|---|
| `feat!: x` | **null** |
| `feat(a)!: x` | **null** |
| `fix!: x` | **null** |
| `fix(a)!: x` | **null** |
| `feat: x` | minor |
| `fix: x` | patch |
| `fix(a)!: x` + `BREAKING CHANGE:` footer | major |
| `fix(a): x` + `BREAKING CHANGE:` footer | major |

## Why

`.releaserc` configures the analyzer with no `preset`:

```json
"plugins": [
  "@semantic-release/commit-analyzer",
  …
]
```

so the default **angular** preset applies, whose header pattern is `/^(\w*)(?:\((.*)\))?: (.*)$/` — no `!`. A subject carrying one fails to parse, `type` comes out null, no release rule matches, and the commit contributes nothing. The `BREAKING CHANGE:` footer still works because it is read from the body as a note, independently of the header.

So the failure mode is the worst available one: the marker that *looks* most explicit is the one that silently suppresses the release.

## Why it bites harder with squash merges

Squashing composes the release commit from the PR title and whatever body the repository is configured to use. A PR titled `fix(scope)!: …` whose body does not carry the footer produces a squash commit that releases **nothing** — the change ships to `main` and no version goes out. Nobody sees an error; there is simply no release.

This nearly happened on #777, which is a real breaking change to `NavigateCallback`. Its title carries `!` because `.claude/rules/git-workflow.md` says that is sufficient. It is not, and the footer had to be reproduced in the PR description to make the squash safe.

## Two ways to fix, and they are not equivalent

**1. Make `!` work (recommended).** Configure the analyzer with the conventionalcommits preset, which does admit `!`:

```json
[
  "@semantic-release/commit-analyzer",
  { "preset": "conventionalcommits" }
]
```

This makes the documented rule true rather than adjusting the documentation down to the behaviour, and `!` in a PR title then survives a squash on its own — which is what anyone reading Conventional Commits will expect. It needs `conventional-changelog-conventionalcommits` as a devDependency, and it changes changelog section headings, so the diff is not purely mechanical.

Note that **commitlint already accepts `!`** (`@commitlint/config-conventional` is built on the conventionalcommits parser), which is exactly why this passes review and then quietly does nothing at release time: the two halves of the pipeline disagree about the same syntax.

**2. Correct the documentation.** Drop the `!` clause from `.claude/rules/git-workflow.md` and say plainly that the footer is the only breaking marker. Cheaper, but leaves a trap: `!` remains *accepted* by commitlint and *silently fatal* to the release, and every contributor who knows Conventional Commits will keep reaching for it.

I would take (1), and add a check that fails a PR whose title carries `!` while its body has no footer — the combination that produces silence.

Happy to implement whichever is preferred; flagging rather than choosing, since it changes release behaviour for the whole repository.

jooyoungseo:
# Pull Request

## Description

Closes #778.

`.claude/rules/git-workflow.md` has documented `!` after the type as producing a major release the whole time. It never did here — and the failure is silent in the worst available way: a `!`-marked subject does not merely fail to *signal* breaking, it fails to **parse**, so the commit contributes nothing at all. No release, no changelog entry.

Measured against the installed plugins by calling them directly:

| Commit message | Before | After |
|---|---|---|
| `feat!: x` | **null** | major |
| `fix!: x` | **null** | major |
| `feat(scope)!: x` | **null** | major |
| `fix(scope)!: x` | **null** | major |
| `fix: x` + `BREAKING CHANGE:` footer | major | major |
| `feat: x` | minor | minor |
| `fix: x` / `perf: x` | patch | patch |
| `docs:` / `chore:` / `refactor:` | null | null |

## Related Issues

Closes #778. Surfaced by #777, a real breaking change whose title carried `!` because the rules file says that is sufficient — the footer had to be reproduced in the PR description to make the squash safe.

## Changes Made

`.releaserc.json` set no `preset`, so both commit-parsing plugins fell back to the angular defaults, whose header pattern is `/^(\w*)(?:\((.*)\))?: (.*)$/` — no `!`. A marked subject yields no `type`, no release rule matches, and the commit is dropped.

**commitlint accepts `!`** (it is built on the conventionalcommits parser), so the two ends of the pipeline disagreed about one character and nothing before the release would have caught it.

**Both plugins, not just the analyzer.** The notes generator parses commits with its own copy of the options, and dropped a `!` commit from the release notes even in a release cut by some *other* commit. Confirmed by running it both ways:

```
default          -> mentions the breaking commit: false
with parserOpts  -> mentions the breaking commit: true
```

**`parserOpts` rather than the preset swap I proposed in #778.** Two reasons the recommendation changed once I tried it:

1. Switching to `preset: "conventionalcommits"` **did not work** — still `null`. The version resolvable here is `conventional-changelog-conventionalcommits@7`, while `commit-analyzer` parses with `conventional-commits-parser@^6` (its own dependency is `conventional-changelog-angular@^8`). The preset is a generation behind what would consume it, so fixing it that way means adding and pinning a dependency.
2. A preset swap also restructures the changelog headings — a much larger diff, and a visible one, for what is a one-character parsing fix.

Nothing else moves. Squashed merges compose the release commit from the PR title, so the rules file now also says a breaking change needs its `!` **in the title**, not only in a commit inside the branch.

n/a

## Checklist

`.claude/rules/git-workflow.md` now explains *why* `!` works — it is the `parserOpts`, and removing them silently reverts to no-release — and `.github/instructions/git-workflow.instructions.md` is regenerated via `npm run sync:copilot`.

## Additional Notes

| Gate | Result |
|---|---|
| `npm test` | **1834 + 73** (was 1834 + 61) |

`test/scripts/releaseBreakingMarker.esm-test.ts` — 12 cases — reads the real `.releaserc.json` and runs the real plugins, so it fails if someone drops the `parserOpts`, upgrades to a preset that stops honouring `!`, or adds a commit-parsing plugin without them. It lives in `test/scripts/` per the convention for config that has to stay in step with the build, and is an `esm-test` because both plugins are ESM-only — importing them from a `.test.ts` fails with `Cannot use import statement outside a module`.

**Verified it bites:** against `main`'s `.releaserc.json` all 12 fail.

Six of the cases assert the *unchanged* behaviour — `feat`, `fix`, `perf`, `docs`, `chore`, `refactor`. Admitting `!` widens a regex every commit passes through, so the cases that matter most are the ones that must not move.

`test/scripts/semanticRelease.d.ts` declares the two functions the test calls; neither plugin ships types and neither has a DefinitelyTyped package.

---


jooyoungseo:
Both plugins are now explicit `devDependencies` (`439403e`). Your reasoning is the right one: the test's whole purpose is pinning their parsing behaviour, so depending on someone else's hoisting to make the import resolve is the wrong footing.

Declared at the versions already resolved — `^13.0.1` and `^14.1.0` — so `npm ci` installs exactly what it installed before. `npm ls` now shows them as direct and deduped against semantic-release's own copies:

```
maidr@3.75.1
+-- @semantic-release/commit-analyzer@13.0.1
+-- @semantic-release/release-notes-generator@14.1.0
`-- semantic-release@25.0.9
  +-- @semantic-release/commit-analyzer@13.0.1 deduped
  `-- @semantic-release/release-notes-generator@14.1.0 deduped
```

**One thing worth flagging, because it nearly rode along.** Running `npm install` produced a **93-line lockfile deletion** unrelated to the change: this npm version strips the `libc` fields from optional platform binaries (`["glibc"]` / `["musl"]`). Those fields are how npm picks the right prebuilt binary on Alpine versus glibc, so shipping their removal inside a "declare two devDependencies" commit would have been a quiet cross-platform install change hidden in noise.

So the lockfile was restored and the two entries added directly instead. Net churn is **2 lines**, and `npm ci --dry-run` succeeds, which is what proves package.json and the lockfile still agree.

## On the duplicated regex

Agreed, and worth stating why it stays duplicated rather than reading as an oversight: semantic-release gives each plugin its own options object and there is no shared-anchor mechanism in `.releaserc.json`. The duplication is the format's, not a choice. `releaseBreakingMarker.esm-test.ts` reads **both** plugin entries separately and asserts against each — the analyzer cases via `optionsFor('@semantic-release/commit-analyzer')`, the notes case via `optionsFor('@semantic-release/release-notes-generator')` — so drift between the two fails, and so does deleting either options block entirely (`optionsFor` throws with a message saying why the plugin needs them).

`optionsFor()` re-reading the file per call: agreed it is negligible, and leaving it makes each case independent of the others' ordering.

Gates re-run after the change: lint, type-check clean; **1834 + 73** passing.

---

PR #781, opened 2026-08-10
fix(pie): announce a negative slice as the chart draws it

### LINKED ISSUE #771: Decide one negative-value policy for pie across maidr, py-maidr and r-maidr
The three pie PRs (maidr#767, py-maidr#322, r-maidr#104) each landed a different answer to the same question, and it should be settled once rather than drift further.

A negative slice is not merely unusual — it breaks the invariant the whole trace rests on. It skews `total`, and can push another slice's share above 100%, so the announced percentages stop summing to anything meaningful.

## Where each binding currently stands

| Binding | Behaviour |
|---|---|
| **maidr** (JS) | `PiePoint.y` is documented as "negative values are not meaningful", but nothing guards it. A negative value flows through into `total` and the percentages. |
| **py-maidr** | Raises matplotlib's own `ValueError("Wedge sizes 'x' must be non negative values")`. In practice matplotlib raises first anyway. |
| **r-maidr** (ggplot2) | Reports `ymax - ymin`, i.e. the magnitude the wedge actually subtends — so a negative datum silently emits its absolute value. |
| **r-maidr** (Base R) | Defers to `graphics::pie()`, which rejects negatives itself. |

So today: Python rejects, ggplot2 silently absolutises, and the renderer accepts whatever arrives.

## The options

1. **Producers reject; renderer trusts.** What PIE_SPEC currently says. Requires closing the ggplot2 hole.
2. **Renderer defends.** Treat a negative slice as a gap in `PieTrace` — it already has gap machinery (excluded from total and range, announced as "missing"). Robust against any producer, including hand-authored JSON, which no producer-side rule can cover.
3. **Both.** Producers reject where their library gives them a natural place to; renderer treats a negative as a gap as a backstop.

I lean toward **3**. The renderer is the only layer that sees every producer, and hand-authored MAIDR JSON has no producer at all — but a library that already errors (matplotlib, `graphics::pie`) should keep erroring, because failing at the call site is far more useful to the author than a silently-dropped slice.

Whatever is chosen should land in PIE_SPEC and in all three repos together.

jooyoungseo:
# Pull Request

## Description

Closes #771.

A negative slice was read as a gap — announced `missing`, kept out of the total and the range. That protected every *other* slice's percentage, which was the point of #772. But the decision was made without checking whether the slice is drawn, and **it is**.

## Related Issues

Closes #771. Follows #772 (which introduced the gap treatment) and xability/r-maidr#118 (which stopped the producer laundering the sign, so the value now arrives intact).

## Changes Made

`geom_col` + `coord_polar` is not a pie — it is a **stacked bar bent into a circle**, where a value below the baseline is ordinary. ggplot2 has nothing to object to, and lays `c(60, -40, 30)` out like this:

```
      ymin  ymax  extent    drawn share
A       30    90      60          46.2%
B      -40     0      40          30.8%
C        0    30      30          23.1%
total span 130
```

Against that, MAIDR announced **60 at 66.7%, missing, 30 at 33.3%** — two slices where three are drawn, and not one figure in common with what is on screen. Two people at the same screen could not discuss the same chart. For an accessibility tool that is a worse failure than the corrupted percentages the gap treatment was avoiding.

So:

- **The value keeps its sign** — `Units is -40`.
- **The share is of the circle as drawn** — divided by the sum of the absolute values, which is exactly the span a producer lays a pie out over. Dividing by the signed total (50) would announce shares no slice occupies, summing to 260%.
- **The caveat lives in the chart description**, not in the per-slice announcement, which is read on every move and already at the length where extra words cost more than they carry.

**On where the caveat went.** I had proposed announcing it once on entering the trace. Checking the code first showed that does not work: the entry cue is built by `TextService.subplotEntryText` from the figure lobby, and **a single-panel pie is never entered from a lobby** — the caveat would have been heard only in multi-panel figures. The chart description is the one summary every layout can reach.

**Nothing changes for the other producers**, both of which refuse a negative before MAIDR sees it — verified rather than assumed:

```
matplotlib   ValueError: Wedge sizes 'x' must be non negative values
graphics::pie  'x' values must be positive
```

**An all-positive pie is untouched.** With no negatives Σ|v| = Σv, so every existing chart reports exactly what it reported before — pinned by a test.

Chromium, against a real ggplot2 chart rendered through r-maidr with this build swapped in:

```
f is A, Units is 60, Percentage is 46.2%
f is B, Units is -40, Percentage is 30.8%
f is C, Units is 30, Percentage is 23.1%
```

and `d`:

```
Number of slices: 3
Min value: -40
Max value: 60
Total: 50
Note: This pie contains a negative value, so a slice cannot be a share of the
      total. Percentages are shares of the circle as drawn, out of 130.
```

The three percentages are the three ggplot2 draws, to the decimal. No console errors.

## Checklist

## Additional Notes

| Gate | Result |
|---|---|
| `npm test` | **1835 + 73** (was 1834 + 73) |

The four existing tests under *"pie slices with a negative value"* stated the old policy directly — `expect(shareAtSlice(trace, 1)).toBe('missing')`, `expect(audio.freq.min).toBe(30)`. They were **rewritten to state the new contract, not deleted**, so the file still documents what happens to a negative rather than falling silent about it. A fifth was added pinning that an all-positive pie is unchanged, which is the regression that would matter most.

**Not a breaking change under this repo's release config**, and worth saying why rather than leaving it to inference: it alters announced output, but `NavigateCallback`-style API surface is untouched, so `fix:` is the honest type. The percentages a user hears for a chart containing a negative *do* change — that is the fix.

`Total` deliberately stays the signed sum (50) rather than becoming the share basis (130). It is the arithmetic total of the data, which is a true and useful number; the note is what reconciles it with the percentages.

---

### DIFF
--- a/src/model/bar.ts
+++ b/src/model/bar.ts
@@ -27,7 +27,7 @@ import { MovableGrid } from './movable';
  * @param raw - The value from the point, on whichever axis carries magnitude
  * @returns The magnitude, or `NaN` when the bar is a gap
  */
-function toBarValue(raw: string | number | null | undefined): number {
+export function toBarValue(raw: string | number | null | undefined): number {
   if (raw === null || raw === undefined) {
     return Number.NaN;
   }
--- a/src/model/pie.ts
+++ b/src/model/pie.ts
@@ -24,17 +24,18 @@ const MISSING_TEXT = 'missing';
  * absent, which is what {@link isMeasured} and the text layer already read as
  * a gap. Mirrors `toBarValue` in `bar.ts`, for the same reasons.
  *
- * A negative slice is read as a gap too. {@link PiePoint.y} documents that a
- * negative value is not meaningful in a pie, and nothing upstream is obliged to
- * enforce that: left in, it subtracts from the total every other slice's share
- * is divided by, so a pie of 60, -40 and 30 announces its first slice as 120%
- * and the percentages stop summing to anything. Reading it as a gap is the
- * treatment this trace already has for a slice with nothing to report — out of
- * the total, out of the range, announced as missing — which tells the reader
- * about the one bad slice instead of quietly corrupting every other one.
+ * A negative slice keeps its sign. It used to be read as a gap, on the grounds
+ * that a negative is not meaningful in a pie — but the producer that emits one
+ * actually draws it: ggplot2's `geom_col` + `coord_polar` is a stacked bar bent
+ * into a circle, where a value below the baseline is ordinary, and it lays out
+ * `c(60, -40, 30)` as three wedges of 46.2%, 30.8% and 23.1%. Announcing that
+ * middle wedge as missing left a screen-reader user hearing two slices where a
+ * sighted reader sees three, with no percentage in common — the two cannot
+ * discuss the same chart. MAIDR conveys the chart; it does not correct it
+ * (#771).
  *
  * @param raw - The value from the slice
- * @returns The magnitude, or `NaN` when the slice is a gap
+ * @returns The value, or `NaN` when the slice is a gap
  */
 function toSliceValue(raw: number | string | null | undefined): number {
   if (raw === null || raw === undefined) {
@@ -43,9 +44,7 @@ function toSliceValue(raw: number | string | null | undefined): number {
   if (typeof raw === 'string' && raw.trim() === '') {
     return Number.NaN;
   }
-  const value = Number(raw);
-  // `NaN < 0` is false, so an unparseable value falls through unchanged.
-  return value < 0 ? Number.NaN : value;
+  return Number(raw);
 }
 
 /**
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Both findings were right, and the first is a real bug. Fixed in `f7ce960`.

## 1. The unrounded basis — reproduced

Your trace through `roundCell` → `defaultFormat` is exactly it: the stat is a finished **string**, so `defaultFormat`'s non-number branch hands it back untouched with the raw sum inside. Your example, run:

```
shareBasis raw -> 35.599999999999994
in the note    -> "...shares of the circle as drawn, out of 35.599999999999994."
```

Worth naming why that is worse here than it would be elsewhere: this sentence exists *because* the percentages and the total disagree. Reconciling them with a number nobody can trust undoes the only thing the note is for. Now formatted through `defaultFormat`, the same path the stats beside it take, so it reads `35.6` and matches them.

And you were right that nothing here would have caught it — every case used integers. A non-integer case is added, and **it bites**: with the raw interpolation restored, `1 failed, 35 passed`.

## 2. The duplicate — confirmed byte-for-byte

```
$ diff <(sed -n '/^function toBarValue/,/^}/p' src/model/bar.ts) \
       <(sed -n '/^function toSliceValue/,/^}/p' src/model/pie.ts)
1c1
< function toBarValue(raw: string | number | null | undefined): number {
---
> function toSliceValue(raw: number | string | null | undefined): number {
```

Only the name and the union's order. Your framing is the sharp part: the negative clamp *was* the reason for the separate copy, and removing it is what closes the gap — so the cleanup is not incidental to this PR, it is a consequence of it. `toBarValue` is exported and used directly. No new coupling: `isMeasured` already came from that module.

One thing I did not let go with it — the JSDoc explaining **why a negative keeps its sign** was the most valuable part of the deleted function. It moved to the call site rather than being dropped, since a future reader meeting `toBarValue` in `pie.ts` will reasonably wonder why a pie parses values the way a bar does.

## On the sandbox

Noted. For what it is worth, the reason both findings landed is that they are the kind static reading catches and a green suite does not — the suite was green through both.

`npm test` → **1836 + 73** (was 1835 + 73), lint and type-check clean.

**#782 is stacked on this branch** and will pick both up when this merges.

---

Issue #780, opened 2026-08-10
A pie sounds like a bar chart: panning is linear, and nothing says where a slice sits on the circle


jooyoungseo:
Navigation works, and the values are right. What is missing is the **circle** — the one thing that makes a pie a pie is absent from every modality except the visual highlight.

Verified in Chromium against a real chart (amcharts5 5.20.1, two pie panels), not read off the source.

## What works today

```
→   Label is Apples, Value is 30, Percentage is 30.0%
→   Label is Bananas, Value is 50, Percentage is 50.0%
←   Label is Apples, Value is 30, Percentage is 30.0%
←   No more data to display        (stops at the first slice)
↑ ↓ No more data to display        (a pie is one-dimensional; deliberate)
```

Left/right step one slice at a time in drawing order, both ends are bounded, braille encodes the magnitudes, and pitch tracks the value. All correct.

## What is missing

**1. Panning is linear, so a pie is spatially indistinguishable from a bar chart.**

`src/model/pie.ts`:

```ts
panning: { x: this.col, y: 0, rows: 1, cols: this.sliceValues[this.row].length },
```

The pan sweeps left-to-right by slice *index*. Slice 1 is hard left, the last slice is hard right — exactly what a bar chart does. A listener sweeping a pie hears a straight line.

A pie has real geometry available for free: each slice's angular midpoint follows from the running total of the shares. Panning by `sin(angle)` would put 12 o'clock in the centre, 3 o'clock hard right, 6 o'clock back to centre, 9 o'clock hard left — the sweep would go out and come back, which is what a circle *is*. It would also make the wrap from the last slice to the first audible as a return rather than a jump.

**2. Position is ordinal only.**

```
p  ->  Position is 1 of 3
```

`AnnouncePositionCommand` has no `TraceType.PIE` branch, so a pie falls through to the generic count. Nothing conveys where the slice sits on the dial or how far it sweeps.

The gap is narrower than it first looks — in a pie the angle *is* the value, so "Percentage is 30.0%" already tells you the arc is 30% of the circle. What it does not give is the **orientation**: where the slice starts. That is what a sighted colleague is describing when they say "the big slice on the right", and there is currently no way to connect the two readings of the same chart.

## Suggested direction

- **Pan by angle** rather than by index, in `PieTrace.audio`. Self-contained, and the biggest gain per line changed: it is the difference between hearing a list and hearing a circle.
[...]

PR #782, opened 2026-08-10
feat(pie): pan a slice by where it sits on the dial

jooyoungseo:
# Pull Request

> [!NOTE]
> **Stacked on #781** — the base branch is `claude/pie-negative-share-jh7gl6`, not `main`. The angles are computed against the same absolute-value basis #781 introduces, and building them separately would have meant two PRs each adding their own copy of that sum, then conflicting. Merge #781 first; this retargets to `main` on its own. The diff against #781 is `src/model/pie.ts` and its test only.

## Description

Refs #780.

A pie sounded like a bar chart. `panning.x` was the slice **index** against `cols = number of slices`, so the sweep ran hard left to hard right in a straight line. The one thing that makes a pie a pie was present in the visual highlight and in no other modality.

## Related Issues

Refs #780. Raised from the question of how a screen-reader user is meant to know where they are in the circle — the honest answer being that they could not.

## Changes Made

`AudioService` reads the pan as:

```ts
interpolate(panning.x, 0, panning.cols - 1, -1, 1)
```

so `cols: 2` makes the mapping `2x - 1`, and `x = (sin θ + 1) / 2` lands the pan on `sin θ` exactly:

| Slice midpoint | θ | pan |
|---|---|---|
| 12 o'clock | 0 | 0 — centre |
| 3 o'clock | π/2 | +1 — hard right |
| 6 o'clock | π | 0 — centre again |
| 9 o'clock | 3π/2 | −1 — hard left |

θ is the slice's angular midpoint — everything swept before it, plus half of itself — clockwise from 12 o'clock. **A sweep of the slices now goes out and comes back**, which is the difference between hearing a circle and hearing a row.

The clearest case is a single slice: by index it panned **hard left**, and it is now **centre**, because one slice fills the dial and its midpoint is 6 o'clock.

A gap consumes no arc, having contributed nothing to the basis, so the slices either side sit exactly where they would if it were absent from the data.

## Scope

Only the panning half of #780. The other half — announcing the clock position on `p` — is left out deliberately: `AnnouncePositionCommand` reads `TraceState`, which carries no angular information, so it needs either a grammar addition or a new field, and that is a design decision rather than a line change. Filed thinking in #780.

Also recorded there, and worth repeating because it bounds this change too: **the payload states no start angle.** Clockwise-from-12 is the convention of every producer wired up so far, not something `PiePoint` declares. A producer laying a pie out counter-clockwise would pan the wrong way round, and nothing here would know.

n/a — the change is audible, not visual. The cardinal positions are pinned by test rather than by listening.

## Checklist

## Additional Notes

| Gate | Result |
|---|---|
| `npm test` | **1838 + 73** (was 1835 + 73 on #781) |

The existing `pans across the slices as a single row` asserted the old mapping literally (`{ x: 1, y: 0, rows: 1, cols: 3 }`). It was rewritten rather than deleted, and three cases added around it, chosen so the arithmetic is checkable by eye rather than by trusting a decimal:

- **two equal slices** → midpoints at 3 and 9 o'clock → pan exactly `+1` then `−1`;
- **one slice** → midpoint at 6 o'clock → back to centre;
- **four equal slices** → symmetric pairs, rising right then mirroring left;
- **a gap** → the slices around it land where they would without it.

**Verified all four bite:** with the index mapping restored, `4 failed, 34 passed`.

`feat:` rather than `fix:` — nothing was broken, and a listener who has learned the old left-to-right sweep will notice the change.

---

### DIFF
--- a/src/model/pie.ts
+++ b/src/model/pie.ts
@@ -49,6 +49,39 @@ function toPercentage(value: number, basis: number): string {
 }
 
 /**
+ * Where each slice sits on the dial, in radians clockwise from 12 o'clock.
+ *
+ * The angle a slice occupies is its share of the circle, so the midpoint of
+ * slice *i* is everything before it plus half of itself. Producers lay a pie
+ * out from 12 o'clock going clockwise — matplotlib, `graphics::pie`, ggplot2's
+ * `coord_polar`, amCharts and AnyChart all do — so that is the origin used
+ * here. The payload states no start angle, so a producer that ever laid one
+ * out differently would pan the wrong way round; that is a grammar question,
+ * noted in #780.
+ *
+ * A gap occupies no angle, having contributed nothing to the basis. It still
+ * gets a midpoint — wherever the sweep has reached — so panning it is
+ * meaningful rather than `NaN`.
+ *
+ * @param values - Every slice's value, `NaN` for a gap
+ * @param basis - The sum of the absolute values of the measured slices
+ * @returns One angle per slice, in the same order
+ */
+function toMidAngles(values: readonly number[], basis: number): number[] {
+  if (basis === 0) {
+    // Nothing is drawn, so every slice sits at 12 o'clock and pans centre.
+    return values.map(() => 0);
+  }
+  let swept = 0;
+  return values.map((value) => {
+    const share = isMeasured(value) ? Math.abs(value) / basis : 0;
+    const mid = swept + share / 2;
+    swept += share;
+    return mid * 2 * Math.PI;
+  });
+}
+
+/**
  * Whether a screen-space point falls inside an element's filled geometry.
  *
  * `isPointInFill` works in the element's own user space, so the point has to be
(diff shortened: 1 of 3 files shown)


jooyoungseo:
Both notes taken, in `8391d9f`. Also rebased onto #781's latest, which picked up the two fixes from its own review.

**The class doc** said "N slices read as one row, left and right" — true of the navigation, no longer true of the sound. It now separates the two, because they genuinely differ now: *"one row to navigate and a circle to listen to."*

**The `cols: 2` contract** is the better of the two catches, and I took it further than a note near `Panning` alone. The problem is that `cols` *reads* as a column count from `audio.ts`, so `cols: 2` looks like a lie rather than a deliberate range. `AudioState.panning` now says what the service actually needs — a position and a range to read it against — and names the pie as the case that uses it that way:

> `x` and `cols` are usually a column index and a column count, which pans a trace left to right. They are not required to be: what the service needs is a position and the range to read it against. `PieTrace` supplies `cols: 2` and a fractional `x` so the pan follows a slice around the dial rather than along a row — a circle sweeps out and back, which an index cannot express.

Agreed on not adding a dedicated field: `Panning` serves several trace types generically and there is no second caller for a bespoke one. Documenting the existing contract is the cheaper half and the part that was actually missing.

Thanks for re-deriving the trigonometry rather than taking the table's word for it — particularly the `basis === 0` short-circuit and the gap's `share = 0` not shifting the slices after it, which are the two places this would have failed quietly.

`npm test` → **1839 + 73**, lint and type-check clean.

---

PR #783, opened 2026-08-10
docs: document the cdnjs mirror now that the listing is live

jooyoungseo:
# Pull Request

## Description

The `cdnjs/packages` pull request merged and [the library](https://cdnjs.com/libraries/maidr) is now serving, so the URLs `cdnjs/README.md` deliberately kept out of the docs — they would have 404ed — can go in. This is the "Once it is listed" follow-up that file wrote for itself.

## Related Issues

Follow-up to the cdnjs submission added in #713.

## Changes Made

**`README.md`** — the cdnjs URL, placed next to the reason someone would reach for it: a sandboxed embedding context whose `script-src` names `cdnjs.cloudflare.com` cannot load maidr from jsDelivr, however well jsDelivr works everywhere else. It also states the two constraints a reader would otherwise discover the hard way — the URL is version-pinned because cdnjs serves no floating alias, and only `maidr.js` is mirrored, not the adapter bundles the guides use.

**`cdnjs/README.md`** — "Once it is listed" becomes "It is listed", recording what was verified and what is still worth watching.

### Where this departs from the checklist, and why

The checklist said to add the URLs to `README.md` **and `docs/`**. I did the README only.

cdnjs serves no floating alias — every mention is a hard-coded version string. There are nine core `dist/maidr.js` sites across `docs/` (anychart, d3, frappe, google-charts, highcharts ×2, plotly, vegalite), so following it literally means ten version strings to bump at each release instead of one, in a repo that has just spent a release cycle on the cost of stale URLs.

Those pages are also mostly about adapter bundles — `anychart.mjs`, `chartjs.js`, `highcharts.js` and the rest — which cdnjs does **not** mirror, by design, since the manifest keeps its file globs narrow. Putting a cdnjs URL on a page's `maidr.js` line while its adapter line stays on jsDelivr invites exactly the 404 the narrow mirror was chosen to avoid. Saying "only `maidr.js` is mirrored" once, in the README, seemed better than implying otherwise nine times.

Happy to fan it out across `docs/` if you would rather have the letter of the checklist.

## Verification

Checked against the live mirror rather than assumed — each of these was a way the listing could have been quietly wrong:

| Check | Result |
|---|---|
| `maidr.min.js` exists | **Yes** — 1,866,505 bytes, and genuinely minified (`!function(e){"function"==typeof define…`) rather than a copy of `maidr.js` at 1,893,803. cdnjs generates it as their CONTRIBUTING.md commits to, so the manifest's `filename` resolves and does not need swapping for `maidr.js`. |
| `maidr-math.css` exists | **Yes** — 367,928 bytes, KaTeX rules intact, byte-identical to npm's. This is the entry `cdnjs/README.md` calls easy to leave out and expensive to get wrong: `src/util/katex.ts` resolves it against the URL `maidr.js` was loaded from, so a page served from cdnjs would render LaTeX in AI chat responses unstyled if the mirror lacked it. |
| `maidr.css` | 406-byte placeholder, matching what the build emits. |
| `autoupdate` working | Ten versions backfilled (3.68.0 → 3.75.1), so it is reading npm rather than pinning the submitted version. |
| Both URLs in the README | HTTP 200. |

## Checklist

- [ ] I have tested the changes locally following `ManualTestingProcess.md` — **not run**: this is a Markdown-only change touching no source, and `node_modules` is not installed in this environment. `test/cdnjs/manifest.test.ts` pins `cdnjs/maidr.json` against `package.json` and `scripts/build.js`, none of which this touches. CI covers it.
- [ ] I have added appropriate unit tests, if applicable — not applicable.

## Additional Notes

Prose in `README.md` is unwrapped to match its neighbours; `cdnjs/README.md` stays wrapped to match its own. Worth re-checking after the next release that <https://cdnjs.com/libraries/maidr> picks it up unaided — if it does not, the `autoupdate` block needs another pull request to `cdnjs/packages`.

---
_Generated by [Claude Code](https://claude.ai/code/session_0195xCq4AQ6273n8yWCGv3t1)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,14 @@ To use maidr, follow these steps:
    </html>
    ```
 
+   maidr is also served by [cdnjs](https://cdnjs.com/libraries/maidr), which matters when a page cannot reach jsDelivr: sandboxed embedding contexts allow a fixed set of CDN hosts, and a page whose `script-src` names `cdnjs.cloudflare.com` cannot load maidr from jsDelivr however well jsDelivr works everywhere else.
+
+   ```html
+   <script src="https://cdnjs.cloudflare.com/ajax/libs/maidr/3.75.1/maidr.min.js"></script>
+   ```
+
+   cdnjs serves no floating alias, so that URL names a version and stays on it until you change it — [the library page](https://cdnjs.com/libraries/maidr) lists the current one. Only `maidr.js` is mirrored there, not the per-chart-library adapter bundles the guides below use; those load from jsDelivr.
+
 3. **Add your data**: Define the maidr JSON schema for your plot. See the [Data Schema](docs/SCHEMA.md) documentation for the full schema structure, object properties, and data formats for each plot type.
 
 ## Data Schema
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ To use maidr, follow these steps:
 
    That URL names a version, because cdnjs serves no floating alias: it stays on 3.75.1 until someone changes it, and [the library page](https://cdnjs.com/libraries/maidr) lists the current release.
 
-   Only `maidr.js` is mirrored on cdnjs. The per-chart-library adapter bundles the guides below use — `anychart.mjs`, `chartjs.js` and the rest — are not, so those examples load from jsDelivr.
+   `maidr.js` is the only script mirrored on cdnjs, which decides where each guide below can load from. The [D3](docs/d3.md) and [Plotly](docs/plotly.md) guides use it on its own, so they work from cdnjs unchanged. The rest also load a per-chart-library adapter bundle — `anychart.mjs`, `chartjs.js`, `highcharts.js` and so on — and those are not mirrored, so those examples load from jsDelivr.
 
 3. **Add your data**: Define the maidr JSON schema for your plot. See the [Data Schema](docs/SCHEMA.md) documentation for the full schema structure, object properties, and data formats for each plot type.
 
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ To use maidr, follow these steps:
 
    That URL names a version, because cdnjs serves no floating alias: it stays on 3.75.1 until someone changes it, and [the library page](https://cdnjs.com/libraries/maidr) lists the current release.
 
-   `maidr.js` is the only script mirrored on cdnjs, which decides where each guide below can load from. The [D3](docs/d3.md) and [Plotly](docs/plotly.md) guides use it on its own, so they work from cdnjs unchanged. The rest also load a per-chart-library adapter bundle — `anychart.mjs`, `chartjs.js`, `highcharts.js` and so on — and those are not mirrored, so those examples load from jsDelivr.
+   `maidr.js` is the only script mirrored on cdnjs, which decides how much of a guide below cdnjs can serve. The [Plotly](docs/plotly.md) guide needs nothing else from maidr — Plotly charts are detected without help — so swapping in the URL above covers its whole maidr side. Every other guide also loads a per-chart-library adapter bundle (`d3.js`, `anychart.mjs`, `chartjs.js`, `highcharts.js` and so on) to bind the chart, and those are not mirrored, so they load from jsDelivr. The charting library itself is separate again: each guide names its own CDN, and most are on cdnjs too.
 
 3. **Add your data**: Define the maidr JSON schema for your plot. See the [Data Schema](docs/SCHEMA.md) documentation for the full schema structure, object properties, and data formats for each plot type.
 
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ To use maidr, follow these steps:
 
    That URL names a version, because cdnjs serves no floating alias: it stays on 3.75.1 until someone changes it, and [the library page](https://cdnjs.com/libraries/maidr) lists the current release.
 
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Both notes addressed in `84279e6`.

**Version drift (1).** I went looking for the release checklist to add the line to, and there isn't one — `CONTRIBUTING.md` and the git-workflow rules cover commit conventions, and semantic-release owns the version from there. So the step is recorded in `cdnjs/README.md` instead, in a new "After each release" section beside the autoupdate check that was already there. That file's whole job is cdnjs upkeep, so it is where the next person looking at this listing will be.

Worth stating why I did not do the obvious stronger thing. A test pinning the README's version to `package.json` is exactly the pattern this repo already uses for the manifest — but semantic-release *commits* the version bump, so the test would go red on `main` after every release until someone edited the README. Making it safe means teaching the release job to rewrite the URL, which is a pipeline change rather than a doc one and does not belong in this PR. The reasoning is written down next to the step so whoever picks it up is not starting from scratch.

The failure mode is worth naming too, since it is the quiet kind: a stale pin does not 404. The URL keeps resolving and keeps serving an older build, so nothing anywhere reports a problem.

**Nit (2).** Split into two paragraphs — the version-pin behaviour, then the mirror's scope, with the adapter bundles named (`anychart.mjs`, `chartjs.js`) rather than gestured at.

Thanks for checking the `maidr-math.css` claim against `src/util/katex.ts` rather than taking it from the description — that resolution behaviour is the whole reason the file has to be in the mirror, and it is the one thing here that would fail silently.

---

jooyoungseo:
Correcting something I got wrong in this PR's own reasoning, pushed as `fcd4c33`.

I justified keeping cdnjs out of `docs/` partly on "those pages are mostly about adapter bundles cdnjs does not mirror". That is true of five guides and **false of two**. Checking what each actually loads:

| Guide | maidr files loaded | Works from cdnjs? |
|---|---|---|
| `d3.md` | `maidr.js` | **yes** |
| `plotly.md` | `maidr.js` | **yes** |
| `anychart.md` | `maidr.js`, `anychart.mjs` | no |
| `frappe.md` | `maidr.js`, `frappe.js` | no |
| `google-charts.md` | `maidr.js`, `google-charts.js` | no |
| `highcharts.md` | `maidr.js`, `highcharts.js`, `highcharts.mjs` | no |
| `vegalite.md` | `maidr.js`, `vegalite.js` | no |
| `chartjs.md`, `amcharts.md` | adapter only | no |

The D3 and Plotly guides load `maidr.js` on its own, and both D3 and Plotly are themselves on cdnjs — so a page under a `script-src` allowing only `cdnjs.cloudflare.com` can follow those two unchanged.

The README sentence said the adapter bundles "the guides below use" are not mirrored, which reads as though no guide can use cdnjs. It now names both sides: D3 and Plotly work from cdnjs, the rest also pull an adapter and stay on jsDelivr. Those readers — locked-down sandboxes — are precisely who the listing exists for, so that was the wrong direction to be imprecise in.

The scope decision itself is unchanged, and this is why: adding a cdnjs URL to the five adapter guides would advertise a route that dead-ends at the unmirrored adapter, and adding one to D3 and Plotly would buy two more pinned versions to bump per release for readers who now have an accurate sentence pointing them at the one URL in the README. Still happy to fan it out if you would rather.

---

jooyoungseo:
You're right, and the way I got it wrong is worth stating plainly: **my scan pattern excluded digits.** I matched filenames with `[a-z-]+`, so `dist/d3.js` never appeared in the results I then reasoned from — and I presented that output as evidence in the previous comment. The D3 guide does load the adapter, exactly as you say.

Rescanned every guide with a digit-aware pattern:

```
docs/amcharts.md        amcharts.js
docs/anychart.md        anychart.mjs maidr.js
docs/chartjs.md         chartjs.js
docs/d3.md              d3.js maidr.js        ← the one I missed
docs/frappe.md          frappe.js maidr.js
docs/google-charts.md   google-charts.js maidr.js
docs/highcharts.md      highcharts.js highcharts.mjs maidr.js
docs/plotly.md          maidr.js
docs/vegalite.md        maidr.js vegalite.js
```

D3 is the only entry that changes, and **Plotly is the only guide whose maidr side is the mirrored file alone.** Fixed in `ae2ac89`.

While rewriting it I found a second overclaim of my own: "works from cdnjs unchanged". The Plotly guide loads plotly.js from `cdn.plot.ly`, so a reader confined to cdnjs has to swap that URL as well — the guide is not usable *unchanged*, only its maidr side is fully covered. The sentence now scopes itself to the maidr side and says the charting library is a separate question, since each guide names its own CDN.

Two rounds of this claim now, both caught rather than shipped. The underlying reason it was fragile is that the fact lives in nine files and the README asserts it from memory — worth a test pinning "which guides load only `maidr.js`" if it ever earns one, though I'd rather not add that on the strength of a sentence.

---

jooyoungseo:
Right that it was unverified — I asserted it from impression in a paragraph where everything else had been checked, which is the part worth not repeating. Dropped in `a4d1437`.

I did check it before deciding, and the result argues for removal more strongly than either of us guessed. Querying cdnjs for each library the nine guides load:

| Library | cdnjs | |
|---|---|---|
| D3 | ✅ | |
| Plotly.js | ✅ | |
| Chart.js | ✅ | listed as `Chart.js` — a lowercase query 404s, which is how a check like this misleads |
| Frappe Charts | ✅ | |
| Highcharts | ✅ | |
| Vega-Lite | ✅ | |
| AnyChart | ❌ | genuinely absent |
| Google Charts | — | `gstatic.com/charts/loader.js`, a dynamic loader, not a mirrorable bundle |
| amCharts | ⚠️ | **listed, at 3.21.15** — the guide uses amCharts 5 |

Two corrections to your reading. **amCharts is on cdnjs**, so its own CDN in the guide is not evidence of absence — but it is frozen at v3 against a guide written for v5, which is worse than absent for a reader who trusts a "yes". And **AnyChart**, which you did not flag, is the one that is actually missing.

So "most" was defensible at 6 of 9 and still the wrong thing to write: every short phrasing lands wrong on amCharts in one direction or the other. The aside was never load-bearing either — what a reader needs is that the charting library is a separate choice from maidr's, which the sentence now says without guessing where each one lives.

That is three rounds of imprecision in this one paragraph, all caught before merge. The pattern is consistent: I was writing summary claims about nine files from impression instead of from a check, in prose that reads authoritative. The paragraph now asserts only what was verified.

---

PR #784, opened 2026-08-10
feat(pie): announce where a slice sits on the dial

### LINKED ISSUE #780: A pie sounds like a bar chart: panning is linear, and nothing says where a slice sits on the circle

jooyoungseo:
# Pull Request

> [!NOTE]
> **No longer stacked.** #781 and #782 have both landed, and this is now rebased onto `main` as a single commit — `src/command/describe.ts` and its test, 183 insertions across 2 files. The earlier note about a base of `claude/pie-circular-panning-jh7gl6` no longer applies.

## Description

Closes #780.

`p` answered *which* slice — "Position is 1 of 3" — and never *where*. A pie's whole shape is where its slices are, so that is the one thing the ordinal cannot carry. A sighted colleague describing the same chart says "the big slice on the right", and there was no reading that could meet that.

It now says:

```
Position is 1 of 4, from 12 o'clock to 3 o'clock
```

## Related Issues

Closes #780. Completes the second half of that issue; #782 did the panning.

## Changes Made

**No state grew a field.** The angles come from the braille row's magnitudes — the same route `announceBoxplotPosition` already takes for *its* trace-specific data (`state.braille as BoxBrailleState`), so this follows an established pattern rather than widening a shared type. A slice's arc is its share of the sum of the absolute values, accumulated in order: the circle **as drawn**, matching what the percentages are already of.

**Twelve o'clock is now a stated convention** rather than an assumption nobody had written down. I raised in #780 that the payload declares no start angle and a counter-clockwise producer would be announced wrongly; the decision is to fix the origin at 12 o'clock clockwise. It is what matplotlib, `graphics::pie`, ggplot2's `coord_polar`, amCharts and AnyChart all do, and the grammar carries nothing that could contradict it. Recorded in the method's own documentation so a future producer that breaks the convention has something to break against.

**Clock positions, rounded to the hour.** Degrees would be more precise and less useful — clock positions are what people say out loud, and this exists to make two readings of one chart meet. The percentage already carries the exact width, so precision here would be duplicated and harder to hear.

A slice too thin to span an hour reads as a point — `at 12 o'clock` — rather than as a range from a position to itself.

Chromium, a four-quarter ggplot2 pie rendered through r-maidr with this build swapped in:

```
f is A, Units is 25, Percentage is 25.0%
    p -> Position is 1 of 4, from 12 o'clock to 3 o'clock
f is B, Units is 25, Percentage is 25.0%
    p -> Position is 2 of 4, from 3 o'clock to 6 o'clock
f is C, Units is 25, Percentage is 25.0%
    p -> Position is 3 of 4, from 6 o'clock to 9 o'clock
f is D, Units is 25, Percentage is 25.0%
    p -> Position is 4 of 4, from 9 o'clock to 12 o'clock
```

No console errors.

## Checklist

## Additional Notes

| Gate | Result |
|---|---|
| `npm run lint:fix` | clean, no files changed |
| `npm run build` | all builds pass (135.1s) |
| `npm test` | **1846 + 73** (main is at 1839 + 73 after #782) |

Re-run after the rebase onto `main`, not carried over from the stacked run.

Seven cases in `test/command/announce-position.test.ts`, built on the file's existing mock harness. Four equal slices throughout, so the arithmetic is checkable by eye rather than by trusting a decimal:

- each quarter lands on its own quarter of the dial, and the sweep **accumulates** rather than restarting per slice;
- the last slice ends at **12, not 0** — the modulo case that would otherwise announce an hour that does not exist;
- a slice thinner than an hour reads as a point;
- a **negative** slice occupies its arc like any other, so the slices after it do not shift (the case #781 makes reachable);
- terse mode says only where it is;
- an all-zero pie falls back to the ordinal rather than dividing by a basis of zero.

**Verified they bite:** with the `TraceType.PIE` dispatch removed, `6 failed, 12 passed` — the seventh passes either way, since the fallback is what the generic path already produced.

### DIFF
--- a/src/command/describe.ts
+++ b/src/command/describe.ts
@@ -879,10 +879,14 @@ export class AnnouncePositionCommand extends AnnounceCommand {
     if (Math.round((end - start) * 12) >= 12) {
       // A slice can be the whole dial -- a pie of one category, or one
       // category at 100%. Both ends then round to 12, and calling that a
-      // point would be the exact opposite of what it is. A slice within half
-      // an hour of the full turn reads the same way, since at this
-      // granularity it is the circle; the percentage carries the remainder.
-      where = 'the whole circle';
+      // point would be the exact opposite of what it is.
+      //
+      // A slice within half an hour of the full turn rounds the same way
+      // without being the same thing: the slices beside it still occupy arc,
+      // and "1 of 3, the whole circle" would contradict itself in one
+      // sentence. Only a slice that is the entire basis gets the plain
+      // reading.
+      where = end - start >= 1 ? 'the whole circle' : 'nearly the whole circle';
     } else if (startHour === endHour) {
       // A slice too thin to span an hour reads as a point rather than as a
       // range from a position to itself.
--- a/src/command/describe.ts
+++ b/src/command/describe.ts
@@ -607,10 +607,6 @@ export class AnnouncePointCommand extends AnnounceCommand {
 }
 
 /**
- * Command to announce the current position in the chart.
- * Formats output based on text mode (terse/verbose) and chart type.
- */
-/**
  * Turns a fraction of the way round the dial into a clock hour.
  *
  * Twelve o'clock is both the origin and the full turn, so a fraction of 0 and
@@ -624,6 +620,10 @@ function toClockHour(fraction: number): number {
   return hour === 0 ? 12 : hour;
 }
 
+/**
+ * Command to announce the current position in the chart.
+ * Formats output based on text mode (terse/verbose) and chart type.
+ */
 export class AnnouncePositionCommand extends AnnounceCommand {
   /**
    * Creates an instance of AnnouncePositionCommand.
@@ -829,10 +829,6 @@ export class AnnouncePositionCommand extends AnnounceCommand {
   }
 
   /**
-   * Announces position for segmented bar charts (stacked, normalized, dodged).
-   * Shows column position and level information.
-   */
-  /**
    * Announces where a pie slice sits on the dial, as clock positions.
    *
    * "Position is 2 of 3" says which slice; it does not say where it is, and a
@@ -879,11 +875,21 @@ export class AnnouncePositionCommand extends AnnounceCommand {
 
     const startHour = toClockHour(start);
     const endHour = toClockHour(end);
-    // A slice too thin to span an hour reads as a point rather than as a
-    // range from a position to itself.
-    const where = startHour === endHour
-      ? `at ${startHour} o'clock`
-      : `from ${startHour} o'clock to ${endHour} o'clock`;
+    let where: string;
+    if (Math.round((end - start) * 12) >= 12) {
+      // A slice can be the whole dial -- a pie of one category, or one
+      // category at 100%. Both ends then round to 12, and calling that a
+      // point would be the exact opposite of what it is. A slice within half
+      // an hour of the full turn reads the same way, since at this
+      // granularity it is the circle; the percentage carries the remainder.
+      where = 'the whole circle';
+    } else if (startHour === endHour) {
+      // A slice too thin to span an hour reads as a point rather than as a
+      // range from a position to itself.
+      where = `at ${startHour} o'clock`;
+    } else {
+      where = `from ${startHour} o'clock to ${endHour} o'clock`;
+    }
 
     if (this.textService.isTerse() || this.textService.isOff()) {
       this.textViewModel.update(where);
@@ -892,6 +898,10 @@ export class AnnouncePositionCommand extends AnnounceCommand {
     }
   }
 
+  /**
+   * Announces position for segmented bar charts (stacked, normalized, dodged).
+   * Shows column position and level information.
+   */
   private announceSegmentedBarPosition(state: NonEmptyTraceState, x: number, cols: number): void {
     const level = state.text.z?.value ?? '';
     const position = x + 1;
--- a/src/command/describe.ts
+++ b/src/command/describe.ts
@@ -5,7 +5,7 @@ import type { HighlightService } from '@service/highlight';
 import type { TextService } from '@service/text';
 import type { BrailleViewModel } from '@state/viewModel/brailleViewModel';
 import type { TextViewModel } from '@state/viewModel/textViewModel';
-import type { BoxBrailleState, LineBrailleState, NonEmptyTraceState } from '@type/state';
+import type { BarBrailleState, BoxBrailleState, LineBrailleState, NonEmptyTraceState } from '@type/state';
 import type { Command } from './command';
 import { focusedSubplotTitle } from '@model/plot';
 import { Scope } from '@type/event';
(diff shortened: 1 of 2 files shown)


jooyoungseo:
All three taken in `b44f26d`. The one you filed as non-blocking turned out to be the real defect; the two blocking ones were cosmetic. Worth saying which way round that went.

**The single-slice case is a bug, and I fixed it rather than noting it.** A slice filling the dial starts and ends at 12, so the point branch announced *the entire circle as a single position* — the exact inversion of what it is, in the one modality this feature exists to provide. And it is not rare: a pie of one category, or one category at 100%, is an ordinary chart.

Checking it, the same failure extends further than the exact case. A slice of 199/200 spans 11.94 hours; both ends still round to 12, so it read `at 12 o'clock` too. So the guard belongs on the **span**, not on the two hours happening to match:

```ts
if (Math.round((end - start) * 12) >= 12) {
  where = 'the whole circle';
} else if (startHour === endHour) {
  where = `at ${startHour} o'clock`;
} else {
  where = `from ${startHour} o'clock to ${endHour} o'clock`;
}
```

At hour granularity a slice within half an hour of the full turn *is* the circle, and the percentage already carries the remainder — the same reasoning that made the whole announcement hour-rounded in the first place. Two tests, and both bite: forcing the branch false gives `2 failed, 1847 passed`.

**Both JSDoc findings were right.** `toClockHour` went in between the class doc and `export class AnnouncePositionCommand`, and the pie method went in between the segmented-bar doc and `announceSegmentedBarPosition`. Two real declarations left undocumented and two comments sitting over the wrong symbol. `toClockHour` is module-level so it cannot move inside the class as suggested; it now sits above the class doc instead, which puts every doc back beside what it describes.

**The gap case is now covered.** You were right that the all-zero test does not reach it — a gap is `NaN`, not `0`, and it takes a different path through `magnitude()`. The assertion is that the slice after a gap lands where it would if the gap were absent, checked against an actual gapless pie rather than against a number I would have had to compute myself:

```ts
const gapped = createCommand(pieState([1, Number.NaN, 1, 1], 2));
const gapless = createCommand(pieState([1, 1, 1], 1));
// both → 'from 4 o'clock to 8 o'clock'
```

**On the duplicated share computation** — agreed, and I am leaving it. It is the `announceBoxplotPosition` precedent, and the alternative is widening a shared state type for one trace. But your framing of the risk is the right one, so it is worth being precise about what would break: if `PieTrace`'s basis changed, the pan and the announcement would disagree — the sound would place a slice somewhere the words did not. That is a silent divergence between two modalities, which is the kind this project cares about most. Not worth a refactor today with one caller, worth a follow-up issue if a second appears.

| Gate | Result |
|---|---|
| `npm run lint:fix` | clean (rewrote `NaN` → `Number.NaN` in the new test) |
| `npm test` | **1849 + 73** (was 1846 + 73) |

---

jooyoungseo:
Took the non-blocking one too, in `f8f8fa6` — your `[1150, 1, 1]` example is the argument. "Position is 1 of 3, the whole circle" contradicts itself inside one sentence, and a listener who then presses right and lands on slice 2 has been told something that cannot be true.

I had the threshold right and the wording wrong. The span check is what stops a full slice reading as a point, and that part stands. But "the whole circle" is a claim about the *pie*, and the span only knows about the *slice*. So only a slice that is the entire basis gets the plain reading now:

```ts
where = end - start >= 1 ? 'the whole circle' : 'nearly the whole circle';
```

The qualified form is true in every case the threshold catches — a 97.9% slice really is nearly the whole circle — and it still answers the orientation question rather than turning into a measurement, which was the point of rounding to hours in the first place.

`>= 1` is exact where it needs to be, not approximately: a lone measured slice divides its own magnitude by a basis that is that magnitude plus zeros, so `end` is `1` exactly. That case is worth having, because the reviewer-flagged reading is not the only near-full one — a gap or a zero beside a full slice draws nothing, so `[100, NaN, 0]` genuinely *is* the whole dial and keeps the plain reading. Three tests now:

- `[100]` → `Position is 1 of 1, the whole circle`
- `[100, NaN, 0]` → `Position is 1 of 3, the whole circle` — the rest is undrawn
- `[199, 1]` → `Position is 1 of 2, nearly the whole circle`
- `[1150, 1, 1]` → `Position is 1 of 3, nearly the whole circle` — yours

**Verified the distinction bites:** collapsing it back to an unconditional `'the whole circle'` gives `2 failed, 1849 passed`.

| Gate | Result |
|---|---|
| `npm test` | **1851 + 73** (was 1849 + 73) |

And to close your note on the previous round: CI was green on all 14 checks for `b44f26d` specifically, not only on the earlier commit — the same set is re-running now for this one.

---

Issue #785, opened 2026-08-10
Line trace ignores a per-point label, so a discrete y is announced as its level code


jooyoungseo:
## What happens

A `line` trace whose points carry a per-point `label` announces the numeric `y` and drops the label. The `step` trace, added in #723, reads it.

Same data, same payload shape, two different readings:

| trace | payload point | announced |
|---|---|---|
| `step` | `{"x":"1","y":5,"label":"Awake"}` | `Hour is 1, Sleep stage is Awake` |
| `line` | `{"x":"1","y":5,"label":"Awake"}` | `Hour is 1, Sleep stage is **5**` |

Verified in Chromium against the bundled `maidr.js` from `v4.0.0`, walking both charts with the arrow keys and reading the live region. The `label` is present in both payloads — confirmed in the emitted HTML — so this is the consumer side, not the producer.

## Why it matters

The motivating shape is an **ordinal factor on y**: a hypnogram's sleep stage, a Likert response, a severity grade. `y` has to stay numeric — it drives sonification, braille and the min/max range — so the human-readable name can only travel alongside. That is exactly what `label` is for, and it is what makes the step trace announce "REM" instead of "4".

A reader on a line chart of the same data hears the level code and has no way to recover the name, while a sighted reader sees it on the axis.

## Where it comes from

r-maidr emits the pairing for `geom_step()` (xability/r-maidr#49) and now for `geom_line()` / `geom_path()` too (xability/r-maidr#122). The producer side is done; the line trace needs to read what it is already being sent.

`geom_path()` and `tidyquant::geom_ma()` share the line trace, so they inherit the gap.

## Suggested shape

Whatever the step trace does with `label` when building a point's announcement, generalised to the line trace — ideally on the shared base so any trace with a discrete axis behaves the same way. Points without a `label` should keep announcing the number, which is the right reading for a continuous y.

Worth deciding at the same time whether this belongs on the base trace rather than per-type, since scatter has the same latent case.

PR #787, opened 2026-08-10
ci: tell r-maidr when a new version reaches npm

jooyoungseo:
# Pull Request

## Description

Sending half of a two-repo change. The receiving half is xability/r-maidr#131.

r-maidr ships a copy of this package's built bundle and refreshes it from npm. Left to itself it discovers a release on a **daily poll**, so a fix published here can sit unread by users for up to a day. Publishing is an event this workflow already knows about, so it can say so directly and close the gap to minutes.

#785 is a live example of the cost: the fix is merged, and a reader on a `geom_line()` factor-y chart still hears the level code until a release goes out *and* r-maidr happens to poll.

## Related Issues

No issue — this came out of noticing the delay while tracking #785 through to r-maidr users.

## Changes Made

- **The release step now records the version**, not just whether a release happened. `semantic-release` does not report it back, so it is read off the tag the step was already inspecting: `git describe --tags --exact-match HEAD`, minus the leading `v`.
- **A new notification step** sends a `repository_dispatch` to `xability/r-maidr`:

  ```json
  {"event_type": "maidr-released", "client_payload": {"version": "4.0.1"}}
  ```

  Naming the version lets the receiver fetch that exact one instead of re-resolving `latest` in the window before the dist-tag has propagated. The body is built with `jq --arg` so the version arrives as data rather than as text spliced into a quoted string.

Three deliberate choices:

- **After the publish step, gated on it.** The receiver pulls the tarball straight from the registry, so announcing before the package is actually there would only make it fail.
- **`continue-on-error: true`.** The release is what matters. A missed ping costs at most a day — r-maidr keeps its scheduled refresh precisely as the backstop for this step failing — whereas a red release job costs a manual re-run of a publish that already happened.
- **py-maidr is not notified.** That is not an oversight: it bundles upstream at its *own* release time, so each py-maidr release pins the version it was tested against rather than tracking upstream continuously. Its `update-maidr-js.yml` says so and is `workflow_dispatch`-only. Pinging it would work against that design.

Not applicable — CI-only change.

## Checklist

- [ ] I have added appropriate unit tests, if applicable — not applicable; there is no harness for workflow YAML beyond `actionlint`, which runs in CI and passes.

## Additional Notes

**This needs one manual step before it does anything.** The step reads a secret, `DOWNSTREAM_DISPATCH_TOKEN`, that does not exist yet and can only be created by an admin. Until it is set the step logs that it is unconfigured and exits 0, so releases keep working exactly as they do today and r-maidr keeps using its daily schedule. The change is inert, not broken, in the meantime.

The token needs write access to `xability/r-maidr` **only** — a fine-grained PAT scoped to that one repository with `Contents: write` is the tightest fit — and is stored here, in this repository's secrets. It is deliberately a separate secret from `SEMANTIC_RELEASE_TOKEN` so the publish credential does not also gain cross-repo reach.

Ordering note: `repository_dispatch` only ever triggers the workflow file as it exists on the receiver's default branch, so **r-maidr#131 should land first**. Merging this one first is harmless — without the token it is a no-op anyway.

**Verification.** `actionlint` — the same version CI runs — passes on the changed file. The version-capture shell was exercised under `bash -e`, the Actions default shell, for both branches:

| case | result |
|---|---|
| `git describe` succeeds (`v4.0.1`) | writes `new_release=true` and `version=4.0.1` |
| `git describe` fails (no tag) | writes neither, and does not abort the step under `-e` |

The second case matters: `set -e` is on by default in Actions, and an assignment inside an `if` condition is exempt from it, which is what keeps a no-release run green.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DZocjnGTiRvLDCbcvqYqa3)_

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -59,13 +59,55 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.SEMANTIC_RELEASE_TOKEN }}
         run: |
           npx semantic-release
-          if git describe --tags --exact-match HEAD 2>/dev/null; then
+          # The tag semantic-release just created, when it created one. Also
+          # captured as the version so the downstream notification below can
+          # name it, rather than making the receiver guess from `latest`.
+          if TAG=$(git describe --tags --exact-match HEAD 2>/dev/null); then
             echo "new_release=true" >> "$GITHUB_OUTPUT"
+            echo "version=${TAG#v}" >> "$GITHUB_OUTPUT"
           fi
 
       - name: Publish to npm with trusted publishing
         if: steps.semantic-release.outputs.new_release == 'true'
         run: npm publish --access public --provenance
+
+      # r-maidr ships a copy of this package's built bundle, and refreshes it
+      # from npm. Left to itself it discovers a release on a daily poll, so a
+      # fix can sit unread by users for up to a day after it is published.
+      # Telling it directly closes that gap to minutes.
+      #
+      # Deliberately after the publish step and gated on it: the receiver
+      # fetches the tarball straight from the registry, so announcing before
+      # the package is actually there would just make it fail.
+      #
+      # py-maidr is not notified, and that is not an oversight. It bundles
+      # upstream at *its own* release time so each py-maidr release pins the
+      # version it was tested against, rather than tracking upstream
+      # continuously. Pinging it would work against that design.
+      #
+      # continue-on-error, because the release is the thing that matters
+      # here. A missed ping costs at most a day -- r-maidr keeps its
+      # scheduled refresh precisely as the backstop for this step failing --
+      # whereas a red release job costs a manual re-run of a publish that
+      # already happened.
+      - name: Tell r-maidr the new version is on npm
+        if: steps.semantic-release.outputs.new_release == 'true'
+        continue-on-error: true
+        env:
+          GH_TOKEN: ${{ secrets.DOWNSTREAM_DISPATCH_TOKEN }}
+          VERSION: ${{ steps.semantic-release.outputs.version }}
+        run: |
+          if [ -z "${GH_TOKEN:-}" ]; then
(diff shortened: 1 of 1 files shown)


jooyoungseo:
Thanks for the careful pass — the `tagFormat` cross-check against `.releaserc.json`, the `gh api --input -` POST default, and the `-e`/`if`-assignment reading are all confirmations I'm glad to have.

On the flagged bug, though, I think it's inverted, so I'm not making the change. The claim is that a custom `if:` **replaces** the implicit `success()`. GitHub's documented behaviour is the opposite — the default is applied *unless* you opt out of it:

> A default status check of `success()` is applied unless you include one of these functions.

and, on `failure()`:

> you must still include `failure()` to override the default status check of `success()` that is automatically applied to `if` conditions that don't contain a status check function.

— [Evaluate expressions in workflows and actions → Status check functions](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#status-check-functions)

`if: steps.semantic-release.outputs.new_release == 'true'` contains no status check function, so `success()` is already applied to it. If `npm publish` fails, `success()` is false and the notify step is skipped — which is the gating the description claims, and it is in place. The suggested `success() && steps.semantic-release.outputs.new_release == 'true'` is exactly equivalent to what is already there.

The same rule is what makes the existing `- name: Upload npm logs on failure / if: failure()` step in this file necessary at all: without spelling out `failure()`, that step would inherit the `success()` default and never run on the failure it exists to capture. And it is why the ubiquitous `if: github.ref == 'refs/heads/main'` deploy-step idiom is safe rather than a footgun that ships after red tests.

One thing your comment did make me check that is worth recording: `continue-on-error: true` on the notify step does not weaken this. It governs how *this* step's failure is reported to later steps and to the job, not whether earlier failures gate it.

Taking the comment-length note as read — if that block drifts from the code it explains, it should be corrected rather than kept.

---

Issue #788, opened 2026-08-10
feat: support area, stacked area and streamgraph traces


jooyoungseo:
## Description

There is no `area` trace type. Every adapter that meets an area mark today either drops it or, worse, silently reads it as something it is not.

`src/adapters/vegalite/converters.ts:630` maps both `line` and `area` to `TraceType.LINE`:

```ts
case 'line':
case 'area':
  return stepDirection ? TraceType.STEP : TraceType.LINE;
```

## Why this is more than a missing feature

For a *stacked* area chart the degradation is not lossy, it is wrong. The rendered band top is the cumulative total, while the series' own value is the band height. A line reading announces one number where the chart shows two, and the user has no way to tell which one they just heard. Chart.js `fill: true`, Highcharts `area` / `areaspline` / `arearange`, `plt.stackplot`, and `geom_area` all land in the same place.

## Proposed mapping

- **Single area** — behaves as `line` for navigation, but the trace knows it is a filled band and can announce the value against the baseline.
- **Stacked / normalized area** — the continuous sibling of `SegmentedTrace` (`stacked_bar` / `stacked_normalized_bar`). Same two-level navigation: move along x within a series, move between series on the z axis. Announce the series value and the cumulative total separately.
- **Streamgraph** — stacked area with a non-zero, centered baseline. Only the baseline resolution differs; the navigation model is identical.

Braille reuses `LineBrailleEncoder` for the single case and `BarBrailleEncoder` for the segmented case, matching how the stacked bar family is already wired in `src/service/braille.ts`.

## Scope

- [ ] `TraceType.AREA` (and the stacked variants) in `src/type/grammar.ts`
- [ ] Trace class + `TraceFactory` case in `src/model/factory.ts`
- [ ] Braille encoder registration
- [ ] Vega-Lite adapter: stop folding `area` into `LINE`
- [ ] Chart.js adapter: detect `fill` on a line dataset

## Difficulty

Low-to-medium. No new navigation concept — it reuses the line and segmented models that already exist.

jooyoungseo:
All five scope items are on `main`. Closing — the implementing pull requests never linked here, so this stayed open after the work landed.

| Scope item | Where it is |
|---|---|
| `TraceType.AREA` + stacked variants | `src/type/grammar.ts` — `AREA`, `STACKED_AREA`, `NORMALIZED_AREA` |
| Trace class + factory case | `src/model/area.ts`, `factory.ts:57` |
[...]

PR #815, opened 2026-08-10
feat(area): read an area chart as its own trace type

jooyoungseo:
# Pull Request

## Description

The Vega-Lite adapter folded the `area` mark into `TraceType.LINE`, so every area chart was announced to the user as a line chart:

```ts
// src/adapters/vegalite/converters.ts, before
case 'line':
case 'area':
  return stepDirection ? TraceType.STEP : TraceType.LINE;
```

For a stacked area that is not merely imprecise. The chart draws **two** magnitudes at each sample — the band's height is its own series' value, the band's top edge is the running total of every series — and a line reading collapses them to one number with nothing in the announcement to say which of the two was heard. A reader navigating a stacked revenue chart hears `20` and cannot tell whether that is this product line's revenue or the company's.

This adds `AREA`, `STACKED_AREA` and `NORMALIZED_AREA`, all served by one `AreaTrace`.

## Related Issues

Addresses the core and Vega-Lite halves of #788. Two boxes on that issue are deliberately **not** ticked here and it stays open:

- The Chart.js `fill` detection needs a new field on `ChartJsDataset` and a change to how `extractLineLayers` groups datasets. That is a second producer and a different grouping rule, so it belongs in its own commit rather than bundled here.
- A navigable "Total" row (the continuous sibling of `SegmentedTrace`'s summary level) would need `LineTrace.mapToSvgElements` to tolerate a series with no selector — it currently returns `null` for the whole trace on a length mismatch, which would silently disable highlighting everywhere. Worth doing, but not as a side effect of this change.

## Changes Made

**Model.** `AreaTrace extends LineTrace`, following the `StepTrace` pattern. An unstacked area navigates exactly as a line does and gains only an honest chart-type announcement. The stacked variants recover the second magnitude: the running total and the point's share of it ride alongside the value in a new `TextState.stack`.

Two decisions worth flagging for review:

- **Totals are summed in the model, not taken from the producer.** Charting libraries disagree about whether a stacked layer's `y` is the series value or the already-accumulated edge; asking every adapter to normalise that would put the same arithmetic in each of them.
- **Series are matched by x value, not by column index.** Stacking is only defined when series share their x samples, but a producer may still emit them ragged. Indexing by column would add one series' 2019 value to another's 2020 column and announce the result as that column's total.

**Text.** `TextState.stack` is rendered in **verbose mode only**, reading as `", Total is 30, 40% of it"`. Point-by-point navigation stays exactly as terse as it was — the chart type announcement ("stacked area") is what tells the reader which magnitude `cross` is, so the brief reading did not need to grow.

**Braille.** All three variants use `LineBrailleEncoder`; the profile is each band's own height, matching the audio and the cross value. Documented in `docs/BRAILLE.md` including why the total is deliberately *not* folded in.

**Vega-Lite.** `area` now resolves to the three new types. Vega-Lite stacks an area by default once a series channel is present — `stack` has to be turned *off* to get overlapping bands — so the default branch here is the stacked one, the opposite of `bar` directly above it. Unstacked area layers also join `coalesceSiblingLineLayers`; the stacked variants are excluded, because Vega-Lite stacks within one mark and never across `alt.layer(...)`, so merging them would invent a total that is not drawn.

The stack setting is read without `??`:

```ts
const stack = encoding?.y?.stack !== undefined ? encoding.y.stack : encoding?.x?.stack;
```

Vega-Lite spells "do not stack" as either `false` or `null`, and a nullish coalesce discards the `null` before it can be seen, falls through to the other axis, and reports `undefined` — which reads as "stack by default", the exact opposite of what the spec asked for. My first version used `??` and the test for `stack: null` caught it.

**A pre-existing bug this turned up, left alone:** the `bar` branch above uses that same `??` pattern, so `"stack": null` on a bar mark is currently read as `STACKED` rather than `DODGED`. Not touched here because fixing it changes bar behaviour and belongs in its own change — happy to file it separately.

## Checklist

## Additional Notes

Verification actually run, in full:

| Step | Result |
|---|---|
| `npm run build` | all builds complete |
| `npm test` | **1870 passed** (unit/component) + **73 passed** (ESM), 0 failures |
| `npm run type-check` | see below |

`npm run type-check` reports `tsconfig.json(6,5): error TS5101: Option 'baseUrl' is deprecated`. That is **pre-existing on `main`** — I confirmed it by stashing this branch's changes and re-running on the base commit, where it reproduces identically. `tsc` reports no other errors, so these changes type-check; the config deprecation is a separate problem and I have not touched it.

New tests: `test/model/area.test.ts` (14 cases) and `test/adapters/vegalite/area.test.ts` (10 cases). The registration guard in `test/util/orientation.test.ts` failed on the three new types until they were listed there, which is the guard working as designed.

---

### DIFF
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -178,6 +178,25 @@ In multiline braille displays, all data series are represented simultaneously. H
 
 In single-line braille displays, the user can navigate vertically with the up and down arrow keys to move between lines, and the braille representation updates to show the values for the current line.
 
+## Area plot
+
+An area plot's braille is the line plot encoding above, unchanged: each series
+becomes a height profile, and the fill under the curve is not encoded because
+it carries no value the curve does not already carry.
+
+For a stacked area plot the profile is each band's **own** height, not the
+height of the stack it sits in. That matches what the audio plays and what the
+text announces as the cross-axis value, so the three modalities describe the
+same number. The running total is available in the verbose text announcement
+and in the chart description; it is deliberately not folded into the braille,
+because a profile that silently switched from series values to cumulative ones
+would read as the same shape with different numbers behind it.
+
+### Multiline Displays
+
+As for a line plot: each line of the display is one series, and the user moves
+between series with the up and down arrow keys on a single-line display.
+
 ## Step plot
 
 A step plot's braille is the line plot encoding above, unchanged: each data
--- a/e2e_tests/page-objects/plots/area-page.ts
+++ b/e2e_tests/page-objects/plots/area-page.ts
@@ -1,0 +1,140 @@
+import type { Page } from '@playwright/test';
+import { expect } from '@playwright/test';
+import { TestConstants } from '../../utils/constants';
+import { AreaPlotError } from '../../utils/errors';
+import { BasePage } from '../base-page';
+
+/**
+ * Page object for the stacked area plot page.
+ *
+ * Provides the interactions the area spec needs: reaching the example,
+ * activating MAIDR on it, and reading back the announcement, the braille
+ * output and the text-mode state.
+ */
+export class AreaPlotPage extends BasePage {
+  /**
+   * Selectors for various UI elements
+   */
+  protected override readonly selectors = {
+    notification: `#${TestConstants.MAIDR_NOTIFICATION_CONTAINER} ${TestConstants.PARAGRAPH}`,
+    info: `#${TestConstants.MAIDR_INFO_CONTAINER} ${TestConstants.PARAGRAPH}`,
+    speedIndicator: `#${TestConstants.MAIDR_SPEED_INDICATOR}${TestConstants.AREAPLOT_ID}`,
+    svg: `svg`,
+    // The textarea's id ends with a React `useId()` suffix, so match on the
+    // stable prefix rather than the whole id.
+    braille: `textarea[id^="${TestConstants.BRAILLE_TEXTAREA}"]`,
+    helpModal: TestConstants.MAIDR_HELP_MODAL,
+    helpModalTitle: TestConstants.MAIDR_HELP_MODAL_TITLE,
+    helpModalClose: TestConstants.HELP_MENU_CLOSE_BUTTON,
+    settingsModal: TestConstants.MAIDR_SETTINGS_MODAL,
+    chatModal: TestConstants.MAIDR_CHAT_MODAL,
+  };
+
+  /**
+   * The ID of the area plot
+   */
+  private readonly plotId = TestConstants.AREAPLOT_ID;
+
+  /**
+   * Creates a new AreaPlotPage instance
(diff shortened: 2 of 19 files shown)


jooyoungseo:
Both review rounds are addressed. Summary of what changed and what I verified.

### The `columnTotals` bug — real, fixed in `4c8b49b`

The report was correct, and I reproduced it before fixing it rather than taking it on faith. Two new cases in `test/model/area.test.ts` failed exactly as described:

```
reports the total for the x the cursor is actually on  →  Expected: 40, Received: 10
totals an x that the first series does not carry       →  Expected: 70, Received: 40
```

With `Services` starting at Q2, landing on its own first point reported Q1's total of 10 against a value of 20 — a share of 200%, which no stacked chart can draw and which the divide-by-zero guard had no reason to catch.

`columnTotals` is now a `Map<string, number>` keyed by the x value, looked up with the x under the cursor. Two things fell out of that beyond the fix:

- every x **any** series carries gets an entry, not only those the first series happens to have, so a chart whose later series run past the first can still report the stack over those columns;
- the `.find()` nested in a per-column loop is gone, making it linear rather than quadratic — the performance note from round one.

### The E2E spec — added in `65057a9`

`e2e_tests/specs/area.spec.ts`, plus `examples/area.html` and a page object. The example draws each band the way a renderer actually does — out along the top edge, back along the edge beneath it, closed — so it exercises the 2N-vertex geometry rather than a synthetic polyline.

Writing it drove out two wrong assumptions **of mine**, both of which turned out to be correct product behaviour worth pinning:

- a multi-series trace names its group count, so the instruction text is `stacked area with 2 groups`;
- the braille display is **multi-line** for a multi-series trace, one row per band, newline-separated. That is why a whole-string match against the braille block failed. The spec now asserts per row, which is a stronger check than the one it replaces.

### `defineSvgPathElement` teardown

Left as is. It is copied verbatim from `test/model/stepHighlight.test.ts`, including the `writable: true` that exists so a suite defining its own stub (`test/model/heatmap.test.ts` does) is never blocked by load order. Adding teardown to one of the two would make them diverge for no behavioural gain; if the pattern should change it should change in both, which is not this PR.

### The pre-existing `bar` `??` bug

Filed as #816 rather than bundled here, with the reproduction and the one-line fix.

### Verification actually run

| Step | Result |
|---|---|
| `npx tsc --noEmit` | no errors beyond the pre-existing `TS5101` config deprecation, which reproduces on `main` |
| `npm test` | **1876 passed** + **73 passed** (ESM), 0 failures |
| `npm run build` | all builds complete |
| `npx playwright test --project=chromium` | **358 passed**, full suite |

The E2E run needed a local workaround unrelated to this change: the container ships Chromium 1194 while the pinned Playwright wants 1217, so I aliased the headless shell to run it. CI has the right browser and is the authority there.

---

Issue #806, opened 2026-08-10
feat: support forest plots


jooyoungseo:
## Description

A forest plot shows one effect estimate with its confidence interval per study, usually against a null line, with a pooled summary at the bottom. It is the standard figure of any meta-analysis and cannot be authored today.

## Proposed mapping

Essentially the error-bar trace laid out on a categorical row axis, so it is close to free once that lands. What is specific to a forest plot:

- **A reference line** (no effect at 1.0 or 0.0) that the reader compares every interval against. Whether an interval crosses it *is the result for that study*, so it belongs in the announcement, not just in the drawing.
- **Marker area encodes study weight** — a third magnitude that is otherwise silently lost.
- **The pooled estimate is a different kind of row** and should be announced as such rather than as one more study.
- Heterogeneity statistics printed on the figure are part of how it is read.

## Suggested announcement

"Smith 2019, odds ratio 1.34, 95% CI 0.98 to 1.83, crosses the null, weight 8%".

## Dependency

Error bar / confidence interval support in this repo.

## Difficulty

Low once the interval trace exists.

Issue #805, opened 2026-08-10
feat: support Kaplan-Meier survival curves


jooyoungseo:
## Description

Survival curves are a required figure in a large share of clinical and epidemiological papers. MAIDR cannot express one.

## Proposed mapping

Three parts, two of which already exist:

1. **The curve** — a step function. `StepTrace` covers it as is.
2. **The confidence band** — depends on error band / interval support (see the error-bar issue in this repo).
3. **Censoring marks** — tick marks on the curve indicating subjects lost to follow-up. These are genuinely new: they are annotations at specific x positions rather than data points, and they change how the curve should be interpreted.

Risk tables printed under the axis are part of how the figure is read, and are worth carrying as a companion layer rather than discarding.

## Suggested announcements

- Median survival, which is the number most readers are looking for, as a summary on entry.
- At each step: time, survival probability, and whether a censoring event occurred there.
- With several arms, the separation between curves at a given time.

## Dependency

Largely free once error bands land; filing separately because the censoring marks and the median-survival summary are specific to it.

## Difficulty

Medium, low after the interval work.

Issue #789, opened 2026-08-10
feat: support error bar and confidence interval traces


jooyoungseo:
## Description

MAIDR has no way to express uncertainty. A point estimate with an error bar, a regression band, or a confidence ribbon reads today as a bare point or line — the interval is dropped entirely.

## Why this matters

Uncertainty is not decoration in statistical graphics; it is often the finding. Whether two group means differ is answered by whether their intervals overlap, and that is exactly the part a MAIDR user currently cannot hear. This is arguably the single largest correctness gap in the supported set.

## Proposed mapping

A per-point composite value — the same shape `box` and `candlestick` already use:

```
{ x, y, yMin, yMax }        // or xMin/xMax for a horizontal orientation
```

`BoxExtremes` and the OHLC handling in `src/model/candlestick.ts` establish the pattern: one navigable point that reports several related magnitudes. Audio can sweep the interval or play centre-then-bounds; text announces "mean 4.2, 95% CI 3.8 to 4.6"; braille reuses `BoxBrailleEncoder`'s span idea.

An `errorband` / ribbon variant is the same data attached to a continuous line rather than to discrete categories.

## Producers this unblocks

`plt.errorbar`, `sns.pointplot` (xability/py-maidr#246), `geom_errorbar` / `geom_ribbon` / `geom_pointrange`, Vega-Lite `errorbar` / `errorband`, `VictoryErrorBar`, Highcharts `errorbar`.

## Downstream

Kaplan-Meier survival curves and forest plots both become mostly free once this lands — they are this trace plus a step line and a category axis respectively.

## Difficulty

Low. The composite-value pattern is already proven twice in the codebase.

jooyoungseo:
On `main` and closing — the implementing pull requests never linked here, so this stayed open after the work landed.

`TraceType.ERROR_BAR` (`grammar.ts:1355`), `src/model/errorBar.ts`, the factory case (`factory.ts:85`), `CHART_TYPE_LABEL`, `IS_ORIENTED: true`, and the braille encoder (`braille.ts:1159`) are all in place, with bindings in both py-maidr (`Axes.errorbar`, `sns.pointplot`) and r-maidr (the ggplot2 uncertainty geoms).

[...]

PR #819, opened 2026-08-11
feat(errorbar): read an estimate together with its interval

jooyoungseo:
# Pull Request

## Description

MAIDR had no way to express uncertainty. A point estimate with an error bar, a regression band, a confidence interval — all read as a bare value, with the interval dropped entirely.

That is not a missing detail. Whether two group means differ is answered by whether their intervals overlap, so a reading that names only the estimate drops the finding the chart was drawn to show.

Adds `ERROR_BAR`, carrying `{ x, y, yMin, yMax }`.

## Related Issues

Part of #789. Unblocks #805 (Kaplan-Meier survival curves) and #806 (forest plots) — both are this trace plus a step line and a category axis respectively.

## Changes Made

**Navigation follows `Candlestick`.** The sections are the grid's rows and the samples its columns, so up and down at one x walk the lower bound, the estimate and the upper bound, while left and right walk the samples. That is what lets a reader trace **one** interval rather than rebuilding it from three passes over the chart, and each magnitude names itself through `TextState.section` — the field the box plot and the candlestick already use — so `3.8` and `4.6` at one x can never be mistaken for two samples.

Three decisions worth review:

- **Bounds are absolute positions on the value axis, not offsets from `y`.** Producers disagree about which they hand out — matplotlib's `yerr` is an offset, Vega-Lite's `errorbar` computes bounds — so the schema fixes one and each adapter converts to it. A reader told `0.4` instead of `4.6` is being given a number the chart does not draw anywhere.
- **Audio scales every section against one range, not per row.** The bounds and the estimate are the same quantity on the same axis, so a bound has to *sound* higher than the estimate it sits above. Per-row scaling would put them at the same pitch and erase the interval by ear. There is a test for it.
- **Which sections exist is decided by the data.** A layer whose points carry no `yMin` gets no lower row at all, rather than a row of silence — an empty lane the cursor can enter and hear nothing in reads as a broken chart, not as an absent bound. One-sided intervals and bare estimates both stay navigable.

**Braille** is one row per magnitude via `LineBrailleEncoder` — a profile of lower bounds, of estimates, of upper bounds. Documented in `docs/BRAILLE.md`, including why the interval is deliberately not encoded as a span.

## Checklist

## Additional Notes

**Two bugs the tests caught, both of which shipped as fixes:**

1. **The E2E spec caught a navigation bug.** Entering the trace landed on the *upper bound*, because I had copied `Candlestick`'s `{ row: sections.length - 1 }` option without thinking about what "last section" means here. A reader arriving at a sample is asking what the value is; landing them on the upper bound answers a question they have not asked, with a number that looks like the value until they move. Entry is now the estimate row.

2. **The unit test caught a float artifact.** Interval width is derived by subtraction, so `4.6 - 3.8` is `0.30000000000000071` — sixteen digits a screen reader would spell out, of an artifact the chart does not contain. Cleaned to 12 significant figures rather than a fixed decimal count, because rounding to two would report an interval of `0.003` as zero, which is worse than the noise. There is a case for that too.

**Producers are deliberately not in this PR.** The trace is reachable today by authoring MAIDR JSON directly — the primary API, and what `examples/errorbar.html` uses — but Vega-Lite `errorbar`/`errorband`, matplotlib `plt.errorbar` and `geom_errorbar` emission are each their own change in their own adapter or binding repo. That is the shape #815 → #818 took for area, and it keeps each producer's field-name mapping reviewable on its own. I would rather not ship an extractor for Vega-Lite's compiled `lower_<field>` naming that I cannot execute here to verify.

Verification actually run, in full:

| Step | Result |
|---|---|
| `npx tsc --noEmit` | no errors beyond the pre-existing `TS5101` config deprecation |
| `npm test` | **1918 passed** + **73 passed** (ESM), 0 failures |
| `npm run build` | all builds complete |
| `npx playwright test --project=chromium` | **366 passed**, full suite |

New tests: `test/model/errorBar.test.ts` (14 cases) and `e2e_tests/specs/errorbar.spec.ts` (8 cases) with `examples/errorbar.html`. The registration guard in `test/util/orientation.test.ts` failed on the new type until it was listed there, which is the guard working as designed.

---

### DIFF
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -197,6 +197,28 @@ would read as the same shape with different numbers behind it.
 As for a line plot: each line of the display is one series, and the user moves
 between series with the up and down arrow keys on a single-line display.
 
+## Error bar plot
+
+An error bar layer becomes one braille row per magnitude it draws — the lower
+bounds, the estimates, the upper bounds — each scaled against its own row's
+range, exactly as a multi-series line plot is.
+
+Rows are ordered bottom to top, so moving down the display moves down the
+value axis. A layer whose data carries no lower bound has no lower row: an
+empty lane the reader can enter and feel nothing in reads as a broken chart
+rather than as an absent bound.
+
+The interval itself is not encoded as a span. Feeling the width means moving
+between the rows at one column, which is the same motion that hears it — the
+audio, the announcement and the braille all describe the same three
+magnitudes rather than three different summaries of them.
+
+### Multiline Displays
+
+Each line of the display is one magnitude, so a multi-line display shows the
+whole interval at once. On a single-line display the up and down arrows move
+between the bounds and the estimate.
+
 ## Step plot
 
 A step plot's braille is the line plot encoding above, unchanged: each data
--- a/e2e_tests/page-objects/plots/errorbar-page.ts
+++ b/e2e_tests/page-objects/plots/errorbar-page.ts
@@ -1,0 +1,140 @@
+import type { Page } from '@playwright/test';
+import { expect } from '@playwright/test';
+import { TestConstants } from '../../utils/constants';
+import { ErrorBarPlotError } from '../../utils/errors';
+import { BasePage } from '../base-page';
+
+/**
+ * Page object for the stacked error bar plot page.
+ *
+ * Provides the interactions the area spec needs: reaching the example,
+ * activating MAIDR on it, and reading back the announcement, the braille
+ * output and the text-mode state.
+ */
+export class ErrorBarPlotPage extends BasePage {
+  /**
+   * Selectors for various UI elements
+   */
+  protected override readonly selectors = {
+    notification: `#${TestConstants.MAIDR_NOTIFICATION_CONTAINER} ${TestConstants.PARAGRAPH}`,
+    info: `#${TestConstants.MAIDR_INFO_CONTAINER} ${TestConstants.PARAGRAPH}`,
+    speedIndicator: `#${TestConstants.MAIDR_SPEED_INDICATOR}${TestConstants.ERRORBAR_ID}`,
+    svg: `svg`,
+    // The textarea's id ends with a React `useId()` suffix, so match on the
+    // stable prefix rather than the whole id.
+    braille: `textarea[id^="${TestConstants.BRAILLE_TEXTAREA}"]`,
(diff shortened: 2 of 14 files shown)


jooyoungseo:
You were right that `IS_ORIENTED[ERROR_BAR] = true` and the implementation disagreed. Addressed in 39cf32e5 — but I did neither of the two things the review offered, so let me be specific about why.

Checking against `Candlestick`, which is the precedent here, orientation turns out to be three separate behaviours, and I had only one of them:

**Two were genuine gaps, now fixed.**

1. **Panning** (`candlestick.ts:577-578` swaps `panning.x`/`panning.y`) — I had these hardcoded to `col`/`row`. Stereo position tracks where a point sits *on screen*, so on a horizontally drawn chart the samples run down the page rather than across it, and the audio was panning along an axis the chart does not have.
2. **`mainAxis`/`crossAxis`** (`candlestick.ts:811-812`) — I omitted them entirely. The formatter service falls back to x/y when they are absent, which is silently wrong for a layer whose two axes format differently: a currency estimate gets announced as a bare number and the category gets announced as currency.

**The third — transposing the navigation grid — I deliberately did not do**, and I don't think `BoxTrace`/`Candlestick` support it. `Candlestick.dimension` carries an explicit comment against it:

> Orientation-independent … `AutoplayState` is keyed by direction, so rows must always be the section count. Conditioning on orientation here mis-paces autoplay and mis-clamps `isMovable` bounds.

The same reasoning applies unchanged to error bars: up and down should walk lower bound → estimate → upper bound in *both* orientations, because that is the motion the trace type exists for. Transposing would make up/down walk the samples on a horizontal chart, which both breaks that and hits the autoplay pacing bug the candlestick comment describes. I've copied that reasoning into `ErrorBarTrace.dimension` with the citation, so the next reader finds the answer at the point of confusion rather than in review history.

So the disagreement you found was real, and it resolves toward `IS_ORIENTED` staying `true` — the flag means "this trace resolves `layer.orientation`", and it now does, in the two places where orientation actually changes behaviour.

Added a `describe('horizontal orientation')` block with five cases covering all of it, including one that asserts the grid shape is *preserved* under orientation, with the reasoning inline — so if someone later reads the transposition as an oversight and "fixes" it, that test fails and explains itself.

Verification: 19/19 unit, 8/8 e2e, `type-check`, `lint:fix` and `build` all clean.

---

PR #820, opened 2026-08-11
feat(waterfall): read a bridge as contributions with running totals

### LINKED ISSUE #790: feat: support waterfall charts
## Description

A waterfall chart shows how a starting value reaches an ending value through a sequence of signed contributions. It is a staple of financial and product reporting, and there is no way to author one for MAIDR today.

## Why this is a good fit

Structurally it is a candlestick: each bar carries a *start*, an *end*, and a signed *delta*, plus a category (increase / decrease / total). `src/model/candlestick.ts` and `src/model/candlestickDelta.ts` already model exactly this — a point with an open, a close, and a derived direction — and `CandlestickBrailleEncoder` already encodes it.

The audio design also transfers: the existing bull/bear audio palette maps onto increase/decrease without inventing anything new.

## Proposed mapping

```
{ x: 'Marketing', start: 1200, end: 950, delta: -250, kind: 'decrease' }
{ x: 'Total',     start: 0,    end: 950, delta:  950, kind: 'total' }
```

- Navigation: one column per step, as with `bar`.
- Text: "Marketing, down 250, running total 950".
- Rotor: a filter unit for increases and one for decreases falls out of `RotorFilterUnit`, which candlestick already uses for bullish/bearish.

## Difficulty

Low. Almost entirely reuse — the highest value-per-unit-of-work candidate in the backlog.

jooyoungseo:
Closes #790.

A waterfall shows how a starting value reaches an ending one through a sequence of signed contributions. It is a staple of financial and product reporting, and there was no way to author one for MAIDR.

## The thing that makes it not just a bar chart

Each step draws a bar **floating** between the running total before it and the running total after it. So a step carries two numbers a bar chart would conflate:

- the **contribution** — the bar's *height*
- the **running total** it produced — the bar's *position*

Announcing only one answers half the question the chart was drawn for. "How big was marketing's effect?" is the contribution; "where did that leave us?" is the total. A reader given only the delta is summing in their head across the whole chart; one given only the total never learns what moved it.

Both travel: the contribution in `cross`, the running total in `stack`. `TextState.stack` already existed for stacked traces and is documented as "the running total a stacked point sits inside" — which is exactly a waterfall step, so this reuses it rather than inventing a field. I extended that doc comment to say so, since it previously claimed to be absent on every unstacked trace.

## Why the pitch carries the contribution, not the total

This is the one real design decision, and it goes against the obvious choice (the staircase is what you *see*).

The running totals of a waterfall drift within a narrow band around the current value — in the example fixture, 1200 → 950 → 1430 → 1360. Scaling pitch to those compresses every step into a near-identical tone, and the large movers become indistinguishable from the small ones. The deltas span the full signed range (−250 … 1360), so the profile has shape and the big contributors are audible immediately — which is what a waterfall is read to find.

The same reasoning drives the braille row, so audio, braille and `cross` all describe the **same** number. `test('scales against the signed range of the contributions')` pins it, and the `stats` block carries the starting and ending values so the journey isn't lost.

## `kind`

`increase` / `decrease` / `total`. The opening, closing and any subtotal bars *restate* the running total rather than changing it, and sit on the baseline instead of floating. Without the distinction a subtotal is announced as though it "rose by 950" — a contribution the chart never made. The description counts increases and decreases with the totals excluded for the same reason.

## Registration

All six places from `.claude/rules/model.md`, plus the two that fail loudly if missed:

| # | Place | Done |
|---|---|---|
| 1 | `TraceType.WATERFALL` + `WaterfallPoint` + data union | ✅ |
| 2 | `src/model/waterfall.ts` | ✅ |
| 4 | `CHART_TYPE_LABEL` (total record) | ✅ |
| 5 | `IS_ORIENTED` (total record) — `false` | ✅ |
| 6 | Braille encoder — `BarBrailleEncoder` | ✅ |
| 7 | `docs/BRAILLE.md` | ✅ |

`IS_ORIENTED` is `false`: navigation is one column per step whichever way the bars are drawn, so there is no main/cross axis to swap. `test/util/orientation.test.ts` caught the omission before I got there, as designed.

The braille encoder is the one that fails **silently** — an unregistered type leaves the display blank while text and audio keep working — so the e2e spec asserts the row renders real braille cells, one per step.

## Verification actually run

| Step | Result |
|---|---|
| `npm test` | **1940 passed** (was 1923) + **73 passed** |
| `npx playwright test waterfall.spec.ts` | **8 passed** |
| `npm run build` | clean |

The e2e run confirms end-to-end that the running total is actually announced alongside the contribution, not just present in the state object.

The example (`examples/waterfall.html`) is generated with real floating-bar geometry — each bar's edge meets the next, so the staircase is consistent and the connectors line up.

---

### DIFF
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -289,6 +289,33 @@ space the same way a zero slice does.
 Pie charts use a single-line representation. On multiline braille displays the
 pie appears on the first line and the remaining lines are unused.
 
+## Waterfall chart
+
+A waterfall's braille is a single row of **contributions** — one cell per step,
+scaled against the signed range of the deltas, exactly as a bar chart's row is.
+
+It is the contribution rather than the running total for the same reason the
+audio pitches the contribution: the running totals of a waterfall drift within
+a narrow band around the current value, so a profile built from them would be
+nearly flat across the whole chart and the large movers — the thing a waterfall
+is read to find — would be indistinguishable from the small ones. The deltas
+span the full signed range, so the profile has shape.
+
+A decrease is a negative value, and is encoded against the same range as the
+increases rather than by magnitude alone, so a step down reads lower than a
+step up rather than merely narrower.
+
+The running total is not encoded. It is available in the text announcement
+alongside the contribution, and in the chart description as the starting and
+ending values — the same division as a stacked area plot, and for the same
+reason: a profile that silently switched between two different quantities would
+read as one shape with different numbers behind it.
+
+### Multiline Displays
+
+Waterfall charts use a single-line representation. On multiline braille
+displays the steps appear on the first line and the remaining lines are unused.
+
 ## Multiline Braille Display Support
 
 By leveraging the two-dimensional nature of multiline braille displays, MAIDR can represent multiple lines of a plot simultaneously, allowing users to perceive the distribution of values across all lines at once. This is particularly beneficial for plots with multiple groups or categories, such as grouped boxplots or line plots with multiple lines, and it also applies to scatter plot grid navigation.
--- a/e2e_tests/page-objects/plots/waterfall-page.ts
+++ b/e2e_tests/page-objects/plots/waterfall-page.ts
@@ -1,0 +1,140 @@
+import type { Page } from '@playwright/test';
+import { expect } from '@playwright/test';
+import { TestConstants } from '../../utils/constants';
+import { WaterfallPlotError } from '../../utils/errors';
+import { BasePage } from '../base-page';
+
+/**
+ * Page object for the stacked waterfall plot page.
+ *
+ * Provides the interactions the waterfall spec needs: reaching the example,
+ * activating MAIDR on it, and reading back the announcement, the braille
+ * output and the text-mode state.
+ */
+export class WaterfallPlotPage extends BasePage {
+  /**
+   * Selectors for various UI elements
+   */
+  protected override readonly selectors = {
+    notification: `#${TestConstants.MAIDR_NOTIFICATION_CONTAINER} ${TestConstants.PARAGRAPH}`,
+    info: `#${TestConstants.MAIDR_INFO_CONTAINER} ${TestConstants.PARAGRAPH}`,
+    speedIndicator: `#${TestConstants.MAIDR_SPEED_INDICATOR}${TestConstants.WATERFALL_ID}`,
+    svg: `svg`,
+    // The textarea's id ends with a React `useId()` suffix, so match on the
+    // stable prefix rather than the whole id.
+    braille: `textarea[id^="${TestConstants.BRAILLE_TEXTAREA}"]`,
+    helpModal: TestConstants.MAIDR_HELP_MODAL,
+    helpModalTitle: TestConstants.MAIDR_HELP_MODAL_TITLE,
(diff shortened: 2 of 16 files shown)


jooyoungseo:
The extrema bug is real and well traced — fixed in d7cf0de, with the text-service note addressed in 83283e4.

## The `supportsExtrema` bug

I verified the whole chain before fixing, and it's exactly as described:

```
sets supportsExtrema = true:  bar, candlestick, candlestickDelta, errorBar, heatmap, line, waterfall
overrides getExtremaTargets:  bar, candlestick, candlestickDelta, heatmap, line, segmented
                                                ↑ errorBar and waterfall missing
```

So pressing <kbd>g</kbd> opened the dialog with nothing in it, and selecting a target would have thrown — the base `navigateToExtrema` raises *precisely when* the flag is set, which makes advertising the feature strictly worse than not advertising it.

I implemented the targets rather than clearing the flag, because a waterfall has an unusually clear answer to "where are the extremes": **the biggest mover in each direction**. "What drove the change" is the question the form is read for, and finding it by ear otherwise means walking every step while holding the running maximum in your head.

Totals are excluded, which matters more than it might look: in the example fixture the opening and closing bars carry deltas of 1200 and 1360, dwarfing every real contribution (480, −250, −70). Including them would make "largest increase" mean "the closing balance" on nearly every waterfall and bury the answer the reader wanted — the same exclusion, for the same reason, as the `Largest contribution` stat.

Five cases added, including the two degenerate ones: a chart whose steps all move the same way offers **one** target rather than naming the same bar as both the biggest rise and the biggest fall, and a chart of nothing but totals offers none.

## `isBoxPlotWithSection` — renamed

Confirmed your reading: the two branches differ only in ordering, and waterfall lands on the box-plot one and renders identically *only* because `KIND_LABEL` is already lower case and the branch's `.toLowerCase()` is therefore a no-op. Coincidence, as you said.

Renamed to `announcesSectionBeforeLabel`, which is what it actually discriminates — box plot ("lower quartile Price") against candlestick ("high Price"), with `z` as the separator. Pure rename, no behaviour change; the full suite (1945 + 73) and the box-plot e2e spec confirm it. The waterfall's `KIND_LABEL` now records at its definition *why* the casing has to stay lower, so a future capitalisation isn't silently undone in verbose mode while surviving in terse mode.

## Terse mode — agreed, and left alone

You're right that `formatTerseTraceText` never reads `state.stack`, so my "both travel" claim holds fully only in verbose mode (the default). That's the pre-existing behaviour for stacked areas too, so making terse mode carry it is a cross-cutting change to the text service rather than something waterfall should decide unilaterally. Worth its own issue if the omission is considered wrong; I've kept this PR to the one trace type.

## Follow-up: `errorBar` has the same extrema bug, and it's already on main

Your review turned up that `errorBar.ts` sets `supportsExtrema = true` without the overrides — I introduced that in #819, which has merged, so it's live now: pressing <kbd>g</kbd> on an error bar chart opens an empty dialog today. Fixing it here would mean an unrelated trace riding along in a waterfall PR, so I'm sending it as its own change immediately after this one rather than leaving it in the backlog.

**Verification:** 1945 passed (was 1940) + 73, waterfall e2e 8/8, error-bar and box-plot e2e re-run against the rename (16/16 total), `type-check`, `lint:fix`, `build` clean.

---

PR #822, opened 2026-08-11
feat(wordcloud): read a cloud as terms walked by weight

### LINKED ISSUE #796: feat: support word clouds
## Description

A word cloud encodes one number per term as glyph size. It is the canonical example of a chart that carries real data while being readable only by eye — the underlying frequencies are never written down anywhere on the page.

## Why it is nearly free

Mathematically it *is* a sorted bar chart: a categorical label and a magnitude. The layout is decoration. So the trace is `bar` with a different mark and a sensible default ordering (descending by weight), and `BarBrailleEncoder` applies unchanged.

The only real decisions are presentational:

- default to frequency order rather than the visual layout order, since the spatial arrangement carries no information;
- announce the value alongside the term ("machine, 412 occurrences"), because the visual encoding of magnitude has no textual equivalent at all today.

## Difficulty

Low. Small effort, and it converts a chart type that is currently 100% inaccessible into one that is fully navigable.

jooyoungseo:
Closes #796.

A word cloud is the canonical chart that carries real data while being readable only by eye: each term's weight is drawn as glyph size and written down nowhere on the page. Structurally it's a categorical label and a magnitude, so the reading is a term and its number — "machine, 412" — which is the whole of what the chart encodes and none of what it draws.

## The one part that isn't free

The issue estimated this as "nearly free — it *is* a sorted bar chart". The announcement, the braille and the audio genuinely are. **Navigating in weight order is not**, and it's the reason this extends `AbstractTrace` rather than `AbstractBarPlot`.

`AbstractBarPlot` resolves its highlight elements in **document order** (`Svg.selectAllElements`) inside its own constructor, and pairs `elements[i]` with `points[i]`. It also declares `highlightValues` as `readonly`. So a subclass can neither:

- reassign `highlightValues` after `super()` — it's `readonly` to the declaring class, nor
- permute inside a `mapToSvgElements` override — that override runs *during* `super()`, before any subclass field holding the permutation exists.

Sorting the points there would **announce one term and highlight another**. And nothing in the model would catch it: the text, the audio and the braille would all be correct. Owning the constructor keeps the sort and the permutation together — `byDescendingWeight` returns each term with the index it was authored at, and the resolved elements are reordered by exactly that.

## The test that proves it

The example authors its glyphs in **packing order**, which is what a cloud layout produces and what encodes nothing — `machine` is drawn *second* but is the heaviest term:

```
document order: neural, machine, gradient, tensor, embedding
weight order:   machine, tensor, neural, embedding, gradient
```

A spec asserts the fixture's two orders differ, so the suite can't silently degrade into one that would pass with the sort removed. Then the highlight spec arrows to the first term and checks the highlighted clone's text:

```ts
expect(announcement).toContain('machine');
await expect(page.locator('[id^="maidr-highlight"]').first()).toHaveText('machine');
```

MAIDR highlights by inserting a clone of the drawn element, so the clone's text *is* the glyph lit up on screen. Without the permutation this announces `machine` and highlights `neural`. It passes.

## Registration

| # | Place | |
|---|---|---|
| 1 | `TraceType.WORD_CLOUD` + `WordCloudPoint` + data union | ✅ |
| 2 | `src/model/wordCloud.ts` | ✅ |
| 4 | `CHART_TYPE_LABEL` (total record) | ✅ |
| 5 | `IS_ORIENTED` — `false` | ✅ |
| 6 | Braille encoder — `BarBrailleEncoder` | ✅ |
| 7 | `docs/BRAILLE.md` | ✅ |

`IS_ORIENTED` is `false`: terms are packed into a rectangle, not laid along an axis, and are walked by weight regardless — there's no main/cross axis to swap, same as a pie. `test/util/orientation.test.ts` caught the omission before I did, as designed.

Extrema navigation offers the heaviest and lightest terms (implemented rather than merely advertised — the base `navigateToExtrema` throws precisely when `supportsExtrema` is set), collapsing to one target when every weight is equal.

## Verification actually run

| Step | Result |
|---|---|
| `npm test` | **2050 passed** (was 2032) + **73 passed** |
| `npx playwright test wordcloud.spec.ts` | **7 passed**, including the highlight assertion |
| `npm run build` | clean |

---

### DIFF
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -316,6 +316,26 @@ read as one shape with different numbers behind it.
 Waterfall charts use a single-line representation. On multiline braille
 displays the steps appear on the first line and the remaining lines are unused.
 
+## Word cloud
+
+A word cloud's braille is a single row of **weights**, one cell per term,
+scaled against the row's own range -- the bar encoding above, unchanged.
+
+The row is ordered by weight, heaviest first, which is also the order the
+cursor walks. A cloud's drawn arrangement is chosen to pack glyphs into a
+rectangle and encodes nothing, so a profile in layout order would be an
+arbitrary sequence of heights; in weight order it descends, and the shape of
+that descent is the distribution -- whether one term dominates or many are
+comparable, which is what a cloud is read for.
+
+The terms themselves are not in the braille. They are what the text mode
+announces, and a display of five cells cannot carry five words.
+
+### Multiline Displays
+
+Word clouds use a single-line representation. On multiline braille displays
+the terms appear on the first line and the remaining lines are unused.
+
 ## Multiline Braille Display Support
 
 By leveraging the two-dimensional nature of multiline braille displays, MAIDR can represent multiple lines of a plot simultaneously, allowing users to perceive the distribution of values across all lines at once. This is particularly beneficial for plots with multiple groups or categories, such as grouped boxplots or line plots with multiple lines, and it also applies to scatter plot grid navigation.
--- a/e2e_tests/page-objects/plots/wordcloud-page.ts
+++ b/e2e_tests/page-objects/plots/wordcloud-page.ts
@@ -1,0 +1,140 @@
+import type { Page } from '@playwright/test';
+import { expect } from '@playwright/test';
+import { TestConstants } from '../../utils/constants';
+import { WordCloudPlotError } from '../../utils/errors';
+import { BasePage } from '../base-page';
+
+/**
+ * Page object for the stacked word cloud page.
+ *
+ * Provides the interactions the word cloud spec needs: reaching the example,
+ * activating MAIDR on it, and reading back the announcement, the braille
+ * output and the text-mode state.
+ */
+export class WordCloudPlotPage extends BasePage {
+  /**
+   * Selectors for various UI elements
+   */
+  protected override readonly selectors = {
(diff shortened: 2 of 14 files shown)


jooyoungseo:
Good catch on the type/coercion inconsistency — fixed in 5927b06, though in the opposite direction to the one that looks obvious.

You offered two ways to resolve it: loosen the type, or drop the `Number()` calls. **Dropping them would have introduced a real bug**, so it's worth writing down why.

Subtraction coerces in JS, so the sort comparator would have survived unchanged. But the description's running total is:

```ts
const total = weights.reduce((sum, weight) => sum + weight, 0);
```

`+` on a string concatenates. A cloud of `"10"` and `"2"` would have reported a **total weight of `"0102"`** — a number no test asserted on and no reader could sanity-check.

So the type was the inconsistent half, not the coercion. `WordCloudPoint.y` is now `number | string`, matching `BarPoint.y`, which is that way for exactly this reason: hand-authored JSON and some producers send numbers as strings. The `Number()` calls stay, and the doc comment now says what they're for rather than leaving them looking vestigial.

Pinned by a test that sorts, announces *and* sums string weights — the total is the assertion that matters, since the other two would pass either way.

Thanks also for independently checking the `readonly`/constructor-ordering claim against `AbstractTrace` and confirming the stable-sort guarantee — those were the two things I most wanted a second pair of eyes on.

**Verification:** 2051 passed (was 2050) + 73, word cloud e2e 7/7 including the highlight assertion, type-check, lint and build clean.

---

PR #823, opened 2026-08-11
fix(scatter): keep the rotor cursor and its announcements in step

jooyoungseo:
# Pull Request

## Description

Follow-up to #619, which shipped `POINT_MODE` and scatter's `INTERSECTION_MODE`. Both modes track their own cursor, but three things still read the ROW_COL one — so the modes route keys to the wrong trace, swallow autoplay's first step, and announce a position that never changes.

These were found while auditing #617 (the earlier, now-superseded attempt at the same feature) against `main`. Each is verified live on `main` at `bfb520d7`.

## Related Issues

Follow-up to #619. No issue filed — found by audit.

### 1. Rotor mode desync on a trace swap — dead, silent arrow keys

A rotor mode is `rotorIndex` on `RotorNavigationService` but a **boolean on the trace**, and nothing resynced them when the active trace changed underneath. After <kbd>PageUp</kbd>/<kbd>PageDown</kbd> to another layer, or a live data update rebuilding the figure:

- `getMode()` still returns `POINT NAVIGATION` and `isRotorEnabled()` is still `true`, so `CommandFactory` keeps routing arrows to `RotorNavigationMoveRightCommand`;
- the incoming trace's `isInPointMode` is `false`, so `movePointRight()` hits its guard and returns `false` without notifying;
- `movePoint()` discards that and returns `null`.

The result is **no cursor move, no sonification, no text, no braille, no highlight and no boundary chime** — a screen-reader user gets zero feedback and no way to discover that cycling the rotor would fix it. Under `INTERSECTING POINT NAVIGATION` it is worse than silent: right-arrow announces "no intersection to the right" at *every* point, while <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>→</kbd> falls past the intersection branch into plain column navigation and moves anyway.

`MoveToNextTraceCommand`, `MoveToPrevTraceCommand` and `Controller.updateData` now call `rotor.resetToDataMode()` **while the outgoing trace is still active**, so its flag is cleared with it and the incoming trace lands in data mode. This is the fix `rotor.ts` already described as needed:

> Known limitation: rotorIndex is not reset when the active plot/trace changes […] Resetting the rotor on context change is a broader UX decision tracked separately from this file.

It was hard to reach before — `GRID_MODE` needs a grid config, `INTERSECTION_MODE` needs a stack. #619 added `POINT_MODE` to **every scatter with data**, which is what makes it routine.

`setMode` also clears all three modes before enabling the active one. The state getters resolve intersection before point, so a trace momentarily in both would read the point cursor through the intersection branch.

### 2. `isInitialEntry` swallows autoplay's first tick

`ScatterTrace.moveOnce` tests `isInitialEntry` **before** the mode branches:

```ts
if (this.isInitialEntry) {      // ← taken first, whatever the rotor mode
  this.handleInitialEntry();    // ← resets row/col/mode, NOT pointModeIndex
  this.notifyStateUpdate();     // ← so this re-emits the current point
  return true;
}
```

Manual arrows in point mode go through `RotorNavigationService.movePoint`, which never touches that flag. So autoplay's first tick took the initial-entry branch: it notified without moving — same tone, same announcement, a second time — and the step was lost.

Entering a rotor mode seeds the cursor on a real point, so `setPointMode(true)` / `setIntersectionMode(true)` consume the handshake there. The flag is assigned directly rather than via `handleInitialEntry()`, which would reset `row`/`col`/`mode` and re-anchor the stack intersection mode walks.

### 3. Announce Position could not distinguish points

`AnnouncePositionCommand` reads `audio.panning`, and both modes pinned it:

- **Point mode** emitted `{y: 0, rows: 1}`, so `(0,1)` and `(0,2)` both announce "Column 1 of 2, row 1 of 1" — in the one mode whose entire purpose is walking individual points.
- **Intersection COL** emitted `panY = 0` against `rows = stack.length`, so an N-point stack announces "row 1 of N" at every step.

Each now reports its own cursor. Point mode reports the point's place inside its own x-column; intersection COL reports the stack index. `outOfBoundsState` mirrors both, so the boundary chime describes the same geometry as the tone before it. Stereo placement is untouched — only `panning.x` reaches the oscillator.

`FlatPoint` gains `xIndex` / `yIndexInColumn`, resolved once at construction from the same numbers `xPoints` was grouped from, so the per-keystroke audio path does not scan an array.

## Checklist

### Verification

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass — **2118 tests, 160 suites, 0 failures**.

9 tests added across `test/model/scatter.test.ts`, `test/service/rotor.test.ts` and a new `test/command/traceSwitchRotorReset.test.ts`. 6 of them fail without their fix (verified by reverting each); the other 3 are the rotor `POINT_MODE` dispatch and the bounds/in-bounds consistency lock, neither of which had any coverage before.

### Not in this PR

`moveOnceInGridMode` moves the **grid** cursor and never checks `isInGridCellMode`, so autoplay after <kbd>Enter</kbd> in a grid cell drives the cell selection rather than the points inside it. `isMovable` has the same gap. This predates #619 and is a separate change — reported by @soundaryakp1999 on #617 as item (c).

---
_Generated by [Claude Code](https://claude.ai/code/session_018bssjSyAmytuMCVtFgYEQ6)_

### DIFF
--- a/src/command/factory.ts
+++ b/src/command/factory.ts
@@ -232,9 +232,9 @@ export class CommandFactory {
       case 'EXIT_BRAILLE_AND_SUBPLOT':
         return new ExitBrailleAndSubplotCommand(this.context, this.displayService, this.brailleViewModel, this.candlestickDeltaService, this.subplotCue);
       case 'MOVE_TO_NEXT_TRACE':
-        return new MoveToNextTraceCommand(this.context, this.candlestickDeltaService);
+        return new MoveToNextTraceCommand(this.context, this.candlestickDeltaService, this.rotorService);
       case 'MOVE_TO_PREV_TRACE':
-        return new MoveToPrevTraceCommand(this.context, this.candlestickDeltaService);
+        return new MoveToPrevTraceCommand(this.context, this.candlestickDeltaService, this.rotorService);
 
       case 'TOGGLE_AUDIO':
         return new ToggleAudioCommand(this.audioService);
--- a/src/command/move.ts
+++ b/src/command/move.ts
@@ -435,6 +435,7 @@ export class MoveToNextTraceCommand implements Command {
    * Creates an instance of MoveToNextTraceCommand.
    * @param {Context} context - The context in which the move operation is performed.
    * @param {CandlestickDeltaService} candlestickDeltaService - Deactivates the virtual delta layer before switching.
+   * @param {RotorNavigationService} rotor - Returns the rotor to data mode before stepping, so the incoming trace never inherits a mode it did not enter.
    */
   public constructor(
     context: Context,
@@ -475,6 +476,7 @@ export class MoveToPrevTraceCommand implements Command {
    * Creates an instance of MoveToPrevTraceCommand.
    * @param {Context} context - The context in which the move operation is performed.
    * @param {CandlestickDeltaService} candlestickDeltaService - Deactivates the virtual delta layer before switching.
+   * @param {RotorNavigationService} rotor - Returns the rotor to data mode before stepping, so the incoming trace never inherits a mode it did not enter.
    */
   public constructor(
     context: Context,
--- a/src/command/move.ts
+++ b/src/command/move.ts
@@ -2,6 +2,7 @@ import type { Context } from '@model/context';
 import type { BrailleService } from '@service/braille';
 import type { CandlestickDeltaService } from '@service/candlestickDelta';
 import type { DisplayService } from '@service/display';
+import type { RotorNavigationService } from '@service/rotor';
 import type { BrailleViewModel } from '@state/viewModel/brailleViewModel';
 import type { Command } from './command';
 import type { SubplotCue } from './subplotCue';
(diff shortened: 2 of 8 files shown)


jooyoungseo:
Thanks — you're right on the `setMode` comment, and I've fixed it in `bec0bbf`.

I checked the same path and reached your conclusion: `setGridMode` (`scatter.ts:1282`), `setPointMode` and `setIntersectionMode` all just assign flags, none calls `notifyStateUpdate()`, and `setMode`'s three `notify*` calls run synchronously with no yield between them. There is no observable window. The comment claimed a bug that cannot currently occur.

I kept the three lines but rewrote the justification to say what it actually is — an invariant guard, worth stating because the getters' fixed resolution order makes "never in two modes at once" non-obvious, and because enabling-first would surface exactly that the day a setter starts notifying. The test comment in `rotor.test.ts` had the same overclaim and is corrected to match.

Two notes on the rest:

**Exhaustiveness of the reset call sites** — that's a better argument than the one in my description, and it holds. `getAvailableModes()` (`rotor.ts:428`) only pushes GRID/POINT/INTERSECTION inside `if (activeTrace instanceof AbstractTrace)`, so a non-trace active falls to `DATA_MODE`, which `isDataMode` recognises and `setMode` handles by clearing all three. `swapActiveTrace` and `enterSubplot`/`exitSubplot` are therefore covered without needing their own reset.

**`xIndexOf`'s `?? 0`** — agreed, and that's why the removed `xIndex < 0 ? 0 : xIndex` guard was dead. Worth adding that `xIndexByValue` is built at `scatter.ts:196` from `xValues`, which is `xPoints.map(p => p.x)`, and `flatPoints` is built at `:252` from the same `data[i].x` values — so the map hit is exact, not approximate. The `?? 0` only fires for a malformed layer.

**On the unverified checks** — reasonable not to trust the description. For the record, on `bec0bbf`: `npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass — 2118 tests, 160 suites, 0 failures. The 6 regression tests were each confirmed to fail with their fix reverted; the other 3 cover the rotor `POINT_MODE` dispatch and the bounds/in-bounds consistency lock, neither of which had coverage before.

---

jooyoungseo:
Thanks — `7bda6b2` closes the `@param` gap on both constructors.

I re-ran the full set on `7bda6b2` rather than assuming a JSDoc-only commit is inert: `npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass — **2118 tests, 160 suites, 0 failures**, same as on `bec0bbf`.

**On the duplicate-coordinate note** — deliberately not adding a case for it. Two points at exactly the same `(x, y)` are indistinguishable in every modality this PR touches: same pitch, same pan, same highlight target, same text. `yIndexInColumn` reporting the first match is the only answer that keeps the announcement consistent with what the user actually hears. A test there would assert that two identical inputs produce identical output, which is a tautology rather than a behaviour, and `.claude/rules/testing.md` asks for tests that would have failed before the change — this one passes either way. Worth a separate issue if duplicate-coordinate scatter data ever needs to be *counted* ("3 points here"), but that is a different feature, not a gap in this fix.

Your exhaustiveness argument on the reset call sites is the stronger version of the one in my description, and I've confirmed it independently: `getAvailableModes()` only offers GRID/POINT/INTERSECTION inside `if (activeTrace instanceof AbstractTrace)`, so any non-trace active falls through to `DATA_MODE` — which `isDataMode` recognises and `setMode` handles by clearing all three flags.

---

PR #825, opened 2026-08-11
fix(scatter): let autoplay sweep the points inside an entered grid cell

### LINKED ISSUE #824: fix(scatter): autoplay does nothing inside a grid cell, and would drive the wrong cursor if it fired
## Summary

Reported by @soundaryakp1999 on #617:

> I am not sure if this is related to this pr, but if I press enter in any grid and explore points, autoplay there does not work.

Confirmed on `main` at `20981ec`. It is not related to #617 or #619 — it predates both. There are two independent defects stacked on top of each other.

## Reproduce

1. Open a scatter plot whose layer has `axes.x`/`axes.y` `min`/`max`/`tickStep` (the grid config).
2. Cycle the rotor with <kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>↑</kbd> to `GRID NAVIGATION`.
3. Move to a cell with several points and press <kbd>Enter</kbd>.
4. Press <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>→</kbd>.

**Expected:** autoplay walks the points inside the cell, left to right, and stops at the last one.
**Actual:** nothing happens at all.

## Defect 1 — the autoplay keys are not bound in the grid-cell scope

`Context.enterGridCell()` (`src/model/context.ts:637`) switches the keyboard scope to `Scope.GRID_CELL`, and that scope's keymap binds three keys and nothing else:

```ts
// src/service/keybinding.ts:390
const GRID_CELL_KEYMAP = {
  GRID_CELL_MOVE_LEFT: key('left', 'Navigate Left in Cell', { showInHelp: false }),
  GRID_CELL_MOVE_RIGHT: key('right', 'Navigate Right in Cell', { showInHelp: false }),
  EXIT_GRID_CELL: key('esc', 'Exit Grid Cell', { showInHelp: false }),
} as const;
```

No `AUTOPLAY_*`, no `STOP_AUTOPLAY`, no speed keys — so <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>arrow</kbd> is simply unbound there. Every other navigable scope binds them.

## Defect 2 — even bound, autoplay would move the grid cursor, not the points

`moveOnce` and `isMovable` both branch on `isInGridMode` without ever checking `isInGridCellMode`:

```ts
// src/model/scatter.ts:1032
if (this.isInGridMode && this.gridCells) {
  return this.moveOnceInGridMode(direction);   // moveGridRight/Left/Up/Down
}
```

```ts
// src/model/scatter.ts:1238
if (this.isInGridMode && this.gridCells) {
  switch (target) {
    case 'FORWARD':
      return this.gridCol < this.numGridCols - 1;   // grid bounds, not cell bounds
    ...
```

`moveOnceInGridMode` (`:1103`) dispatches only to `moveGridRight`/`moveGridLeft`/`moveGridUp`/`moveGridDown`. The cell cursor — `cellPointIndex`, moved by `moveCellPointLeft`/`moveCellPointRight` (`:1416`/`:1432`) — is never touched. So fixing defect 1 alone would make autoplay jump cell-to-cell while the user believes they are inside one, silently changing what the arrow keys mean afterwards.

The `autoplay` step-count getter has the matching gap: it reports grid or ROW_COL dimensions, so pacing would be computed from the wrong denominator.

## Proposed fix

1. Bind `AUTOPLAY_FORWARD`/`AUTOPLAY_BACKWARD`, `STOP_AUTOPLAY` and the three speed keys in `GRID_CELL_KEYMAP`. Cell navigation is horizontal-only, so up/down stay unbound — the same shape as intersection mode.
2. Give `moveOnce` and `isMovable` an `isInGridCellMode` branch ahead of the grid one: `FORWARD`/`BACKWARD` walk `moveCellPointRight`/`moveCellPointLeft`, `UPWARD`/`DOWNWARD` report out of bounds.
3. Have the `autoplay` getter report `getCellPointCount()` while in cell mode so the traversal is paced over the points actually being played.

## Accessibility impact

A blind user who enters a cell to explore its points has no way to hear them swept — the one interaction that makes a dense cell legible without stepping every point by hand. Defect 2 is the worse of the two if fixed carelessly: moving the grid cursor while the user believes they are inside a cell desynchronises the mental model with no announcement that the context changed.

jooyoungseo:
# Pull Request

## Description

Fixes the third item @soundaryakp1999 reported on #617 — pressing <kbd>Enter</kbd> into a grid cell and finding autoplay dead. It turned out to be two defects stacked, both predating #619.

## Related Issues

Fixes #824

### 1. The autoplay keys were never bound in the grid-cell scope

`Context.enterGridCell()` switches the keyboard scope to `Scope.GRID_CELL`, and that keymap bound three keys:

```ts
const GRID_CELL_KEYMAP = {
  GRID_CELL_MOVE_LEFT: key('left', …),
  GRID_CELL_MOVE_RIGHT: key('right', …),
  EXIT_GRID_CELL: key('esc', …),
} as const;
```

No `AUTOPLAY_*`, no `STOP_AUTOPLAY`, no speed keys — so <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>arrow</kbd> was unbound and nothing happened. Now bound, horizontally only: a cell's points are one list, and up/down aren't bound for manual movement either.

### 2. Binding it alone would have been worse than the silence

`moveOnce` and `isMovable` both branch on `isInGridMode` without ever checking `isInGridCellMode`, and `moveOnceInGridMode` dispatches only to `moveGridRight`/`moveGridLeft`/`moveGridUp`/`moveGridDown`. So autoplay would have walked the cell **selection** while the user believed they were inside one — silently changing what the arrow keys mean afterwards, with no announcement that the context had moved.

Both now test cell mode ahead of the grid branch and drive `moveCellPointRight`/`moveCellPointLeft`. Vertical movement reports out of bounds instead of falling through.

### 3. Pacing and the closing chime

- `dimension` reports the cell's point count in cell mode, so `AutoplayService` divides its duration budget by the points actually being played rather than by the grid's cells.
- `outOfBoundsState` pans at `cellPointIndex`, so the end-of-sweep chime lands where the last tone did instead of at the cell's position in the grid — the same correction #823 made for point and intersection mode.

## Checklist

### Accessibility

Sweeping is the point of entering a dense cell — without it a blind user has to step every point by hand to learn what is in there. Defect 2 was the more dangerous of the two: moving the grid cursor while the user believes they are inside a cell desynchronises the mental model with nothing announced to correct it.

### Verification

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass — **2126 tests, 161 suites, 0 failures**.

8 tests added in `test/model/scatterGridCellAutoplay.test.ts`. 7 of the 8 fail with the fix reverted (verified); the eighth asserts that vertical autoplay stays *unbound*, which holds either way and is there to keep the horizontal-only contract from drifting.

---
_Generated by [Claude Code](https://claude.ai/code/session_018bssjSyAmytuMCVtFgYEQ6)_

### DIFF
--- a/src/model/scatter.ts
+++ b/src/model/scatter.ts
@@ -820,6 +820,14 @@ export class ScatterTrace extends AbstractTrace implements GridNavigable, PointN
         cols: this.xValues.length,
       };
     }
+    if (this.isInGridCellMode) {
+      // Autoplay divides its duration budget by these, so a cell traversal
+      // must be paced over the cell's points, not the grid's cells.
+      return {
+        rows: 1,
+        cols: Math.max(1, this.cellXPoints.length),
+      };
+    }
     if (this.isInGridMode) {
       return {
         rows: this.numGridRows,
@@ -921,6 +929,23 @@ export class ScatterTrace extends AbstractTrace implements GridNavigable, PointN
    * from before the mode was entered.
    */
   protected override get outOfBoundsState(): TraceEmptyState {
+    // Inside a cell the cursor is cellPointIndex, so the bounds chime has to
+    // pan there — the grid branch below would place it at the cell's position
+    // in the grid instead, on the wrong side of the plot.
+    if (this.isInGridCellMode && this.cellXPoints.length > 0) {
+      return {
+        empty: true,
+        type: 'trace',
+        traceType: this.type,
+        audio: {
+          y: 0,
+          x: this.cellPointIndex,
+          rows: 1,
+          cols: Math.max(1, this.cellXPoints.length),
+        },
+      };
+    }
+
     // Use grid position when in grid mode for correct panning
     if (this.isInGridMode && this.gridCells) {
       return {
@@ -1028,6 +1053,13 @@ export class ScatterTrace extends AbstractTrace implements GridNavigable, PointN
       return true;
     }
 
+    // Cell mode sits inside grid mode, so it has to be tested first: the grid
+    // branch below moves the cell *selection*, which is not what the user is
+    // navigating once they have pressed Enter into a cell.
+    if (this.isInGridCellMode) {
+      return this.moveOnceInGridCellMode(direction);
+    }
+
     // Handle grid mode navigation (used by autoplay and direct calls)
     if (this.isInGridMode && this.gridCells) {
       return this.moveOnceInGridMode(direction);
@@ -1100,6 +1132,33 @@ export class ScatterTrace extends AbstractTrace implements GridNavigable, PointN
    * @param direction - The movement direction
    * @returns True if movement was successful, false if at boundary
    */
+  /**
+   * Handles movement inside an entered grid cell, mapping directions to the
+   * cell's own point cursor. Only left/right are meaningful — a cell's points
+   * are walked as one horizontal list — so up/down report out of bounds
+   * rather than falling through and moving the grid selection underneath a
+   * user who believes they are still inside the cell.
+   *
+   * Used by autoplay, which drives movement through moveOnce; manual arrows
+   * reach the same moveCellPoint* methods through their own commands. Those
+   * methods notify observers or report bounds themselves.
+   * @param direction - The movement direction
+   * @returns True if movement was successful, false at a boundary
+   */
+  private moveOnceInGridCellMode(direction: MovableDirection): boolean {
+    switch (direction) {
+      case 'FORWARD':
+        return this.moveCellPointRight();
+      case 'BACKWARD':
+        return this.moveCellPointLeft();
+      case 'UPWARD':
+      case 'DOWNWARD':
+      default:
+        this.notifyOutOfBounds();
+        return false;
+    }
+  }
+
   private moveOnceInGridMode(direction: MovableDirection): boolean {
     let moved = false;
     switch (direction) {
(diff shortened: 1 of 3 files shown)


jooyoungseo:
Good catch — that was a real slip, fixed in `a970cdd`.

Inserting `moveOnceInGridCellMode` above `moveOnceInGridMode` left the latter's docstring stranded on the new method: it described the wrong function and the old one was left undocumented. The original text is back where it belongs, and since the two names differ by one word I added a line saying which runs when:

```ts
/**
 * Handles movement in grid mode, mapping directions to grid cell selection.
 * Only reached when the user is browsing the grid itself; once they have
 * pressed Enter into a cell, {@link moveOnceInGridCellMode} takes over.
 * …
 */
private moveOnceInGridMode(direction: MovableDirection): boolean {
```

Re-verified on `a970cdd`: `npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass — **2126 tests, 161 suites, 0 failures**.

On the `Math.max(1, this.cellXPoints.length)` note — agreed it's unreachable today, and worth keeping. `dimension` is read by `outOfBoundsState`'s base implementation as well as by autoplay pacing, and a `cols: 0` reaching `MathUtil.interpolate` over `cols - 1` would divide by `-1` rather than throw, so the failure would be a silently mispanned tone rather than a crash. Cheap insurance for something that would be hard to notice.

---

jooyoungseo:
Caught me fair — that was a self-comparison that passed unconditionally. Fixed in `145c24e`, using the same capture-before shape as the sibling test.

I checked whether the corrected assertion is actually load-bearing rather than just assuming it. With the cell-mode branch reverted, `moveOnce('UPWARD')` inside a cell moves the grid row from `{row: 1, col: 1}` to `{row: 2, col: 1}` — so it does catch the defect the test is named for. It went unnoticed because the earlier `expect(trace.moveOnce('UPWARD')).toBe(false)` fails first and short-circuits the test, which is exactly why a passing run couldn't surface it. Both assertions stay: the return value proves the move was refused, the position proves nothing moved underneath the user.

Re-verified on `145c24e`: `npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass — **2126 tests, 161 suites, 0 failures**.

Also confirmed your `STOP_AUTOPLAY: ctrl, left, right` observation — `TRACE_KEYMAP` already uses that horizontal-only shape, so `GRID_CELL_KEYMAP` follows an existing convention rather than inventing one.

---

PR #826, opened 2026-08-11
fix(chartjs): carry a bubble's radius instead of discarding it

### LINKED ISSUE #813: fix(chartjs): a bubble chart is read as a scatter, silently dropping the radius dimension
## Description

`src/adapters/chartjs/extractor.ts:566` routes bubble charts through the scatter extractor:

```ts
case 'scatter':
case 'bubble':
  return extractScatterLayers(chart, pluginOptions);
```

A Chart.js bubble dataset is `{x, y, r}`. `extractScatterLayers` reads `x` and `y`, so `r` — an entire encoded variable — is dropped without any warning. The chart is then described to the user as a scatter plot, so nothing signals that a third dimension existed.

## Impact

The user is not told that data is missing. A bubble chart whose whole point is the size encoding (population, market cap, sample size) is announced as if it only ever had two variables, which is a wrong reading rather than a partial one.

## Expected

Either the radius is carried into the layer and announced, or — until then — the adapter warns that the size channel is not represented, so the omission is at least visible.

## Related

- #601 (3rd dimension scatter as reverb) covers the core-side representation. This issue is the adapter-side data loss, which stands on its own: even without an audio encoding for the third dimension, the value should reach the text and description layers rather than being discarded in the extractor.

jooyoungseo:
# Pull Request

## Description

Closes #813.

A Chart.js bubble datum is `{x, y, r}`. Bubble charts routed through the scatter extractor, which read `x` and `y` and dropped `r`:

```ts
case 'scatter':
case 'bubble':
  return extractScatterLayers(chart, pluginOptions);
```

The size encoding is usually the entire reason a chart is a bubble chart — population, market cap, sample size. So a reader was told about two variables where there were three, with nothing signalling the omission. That is a wrong reading rather than a partial one.

## Changes Made

**Nothing new was needed to carry it**, which is the surprising part. Every piece was already in place and the extractor was the only spot where the value stopped:

| already there | |
|---|---|
| `ScatterPoint` | declares an optional `z: number` |
| `ScatterTrace` | tracks `hasZ`, `minZ`/`maxZ`, per-point `z`, announces it through `textZ()`, and maps it to `zIntensity` |
| `isPointValue()` | already narrows to `{ x: number; y: number; r?: number }` |

So the fix is to stop discarding it:

```ts
points.push(
  typeof point.r === 'number'
    ? { x: point.x, y: point.y, z: point.r }
    : { x: point.x, y: point.y },
);
```

`typeof point.r === 'number'` rather than a truthiness check: `r: 0` is a real radius — a bubble scaled to nothing is still a datum — and `point.r ? …` would drop it into the "no z" branch.

**The z axis is emitted only when a radius was really present.** The same function serves plain scatter, and a `z` label on a chart with no third variable would announce something that is not there — the same class of mistake in the other direction.

**Naming it.** Chart.js has no scale for `r`, so unlike x and y there is no axis title to read the variable's real name off. It reads `Size` — which describes what the reader is being told rather than pretending to name the variable — unless the page says otherwise through the plugin's `axes.z`, which is the only place the real name exists. `MaidrPluginOptions.axes` gained `z?: string` for that.

## Scope

**Corrected after review.** An earlier version of this description said the value "now reaches the text and description layers", implying it stops there. It does not:

```
src/model/scatter.ts:544   zIntensity: this.zIntensityFor([point.z])?.[0],
src/service/audio.ts:449   this.markEchoTail(audio.zIntensity, ...)
```

`zIntensity` reaches `AudioService` and drives the echo train, so **a bubble's radius is audible as of this PR**, not merely readable. That machinery arrived with #619 and this simply feeds it.

The change is **purely additive**: a plain scatter has no `z`, so `hasZ` stays false and no `zIntensity` is emitted. The echo train appears only for charts that previously carried no third dimension at all, and no existing chart changes how it sounds.

#601 remains open for whatever dedicated encoding it has in mind; this is the adapter-side data loss the issue described, which stood on its own.

n/a — the change is in extracted data, pinned by test rather than by eye.

## Checklist

## Additional Notes

| Gate | Result |
|---|---|
| `npm test` | **2039 + 73** (was 2034 + 73) |

Five cases in `test/adapters/chartjs/extractor.test.ts`:

- a bubble carries `r` through as `z`;
- the size dimension is named, so the number is not announced bare;
- the page can name it itself through `axes.z`;
- **a plain scatter still emits no `z` at all** — the case that guards the over-correction;
- a bubble datum *without* `r` stays indistinguishable from a scatter point, since Chart.js tolerates a mixed dataset and inventing a size would be worse than omitting one — and the layer keeps its `z` axis regardless, because the layer owns the axis, not the point.

**Verified they bite:** with the radius discarded again, **4 of the 5 fail**. The fifth passes either way, which is the point of it.

Lint caught one thing worth noting: `DEFAULT_BUBBLE_SIZE_LABEL` was defined below its use and `ts/no-use-before-define` rejected it, so the constant sits above `extractScatterLayers` rather than beside the function it documents.


jooyoungseo:
Both taken, and the second is a correction to something I asserted in the PR body rather than a nit.

**The test gap** — fixed in `00c5050`. You were right that the mixed-dataset case asserted the points and not the axis, and that the other point in that dataset carrying `r: 8` should keep the layer's `z` axis defined. The layer owns the axis, not the point, so one radius-less datum must not suppress the label for everything around it. Implied by the code, untested until now.

**The scope note — you are right and my PR body was wrong.** I wrote that the value "now reaches the text and description layers", implying that is where it stops. It does not. I checked rather than take it on either of our words:

```
src/model/scatter.ts:544   zIntensity: this.zIntensityFor([point.z])?.[0],
src/service/audio.ts:449   this.markEchoTail(audio.zIntensity, ...)
```

`zIntensity` reaches `AudioService` and drives the echo train, so **a bubble's radius is audible as of this PR**, not merely readable. That is a better outcome than I claimed, and understating it in the PR body was a real inaccuracy — someone picking up #601 could have read my scope paragraph and concluded the audio side was untouched here.

Worth adding what I checked next, since "this now makes noise it did not before" deserves a blast-radius answer: it is **purely additive**. A plain scatter has no `z`, so `hasZ` stays false and no `zIntensity` is emitted — the echo train only appears for charts that previously carried no third dimension at all. No existing chart changes how it sounds.

I have corrected the PR description accordingly.

Also confirming your read on `typeof point.r === 'number'`: it is deliberate for exactly the reason you name. `r: 0` is a real radius — a bubble scaled to nothing is still a datum — and `point.r ? ... : ...` would have dropped it into the "no z" branch. Same reasoning as the `Number.isFinite` guards elsewhere in the file.

| Gate | Result |
|---|---|
| `npm test` | **2039 + 73** |

---

jooyoungseo:
Rebuilt on current `main` (`f30c8d9`) before merging. The branch was 30 commits behind — all of tier B and tier C landed since CI last ran here, including the trace-model additions that share `ScatterTrace`'s machinery, so the green ticks above were green against a tree that no longer exists.

Merged cleanly, no conflicts. Re-verified on the merged tree:

| Gate | Result |
|---|---|
| `npm test` | **2605 + 73 passed**, 184 suites |
| `node scripts/test.js -- test/adapters/chartjs/extractor.test.ts` | **116 passed** |

Nothing in the change needed adjusting — `ScatterPoint.z`, `hasZ`/`zIntensityFor` and the layer's `axes.z` are all where they were, and the five bubble cases pass unchanged.

One correction to my own note above, since it is worth being exact: I first ran that verification while switching branches to work on another PR in the same checkout, which invalidated the result. The numbers here are from a clean run with nothing else touching the tree.

---

PR #827, opened 2026-08-11
feat(gauge): read a measure against its range, target and bands

### LINKED ISSUE #797: feat: support gauge and bullet charts
## Description

A gauge shows a single measure against a range; a bullet chart adds a target marker and qualitative bands. Both are everywhere in KPI dashboards and neither can be authored today.

## Proposed mapping

```
{ value: 73, min: 0, max: 100, target: 80, bands: [{to: 50, label: 'poor'}, {to: 75, label: 'ok'}, ...] }
```

A single navigable point. The information the reader needs is relational, so the announcement should carry it: "73 of 100, 7 below target, in the 'ok' band".

## Why bother with a one-point chart

Dashboards are built out of these, and a dashboard where every tile is unreadable is unusable regardless of how good the charts beside it are. The implementation is close to trivial — a single-cell dimension and a text state — so it is a cheap way to make a whole class of pages navigable.

## Difficulty

Very low.

jooyoungseo:
Closes #797.

A gauge draws one measure, and **the number alone is not the reading**. "73" is unanchored. A sighted reader takes the scale from how far along the dial the bar reaches, the classification from which shaded band it ends in, and the shortfall from which side of the target marker it falls on — and none of that is written anywhere a screen reader can reach.

So the announcement carries the relations alongside the value:

| | field | example |
|---|---|---|
| value | `cross` | 73 |
| range | `range` | 0 to 100 |
| target | `stack` | 80 |
| band | `section` | ok |

The description reports the distance to target **by direction** — "Below target by 7" — because that's the question being asked, and a bare `-7` leaves the reader working out which way it points.

## Design notes

**Data is a single object, not an array.** The chart draws exactly one measure; an array of one would describe a shape it does not have. `HeatmapData` is already a non-array member of the `data` union, so there's precedent — and an e2e test asserts `Array.isArray(layer.data) === false` so the shape can't drift.

**Bands are bounded above only.** A band starts where the previous ended, and the first at `min`. Carrying both edges would let a chart declare overlapping or gapped bands the drawing cannot express. They're sorted before reading, so authoring order doesn't matter.

**A value past every band belongs to none.** Bands need not reach `max`, so naming the last one would invent a classification the chart doesn't draw. Tested with a chart whose only band ends at 50 and whose value is 90.

**`supportsExtrema = false`** — deliberately, and this is the third trace this run where that flag mattered. One point has no extremes to navigate between, and the base `navigateToExtrema` throws precisely when the flag is set, so advertising it would be worse than not.

## Registration

| # | Place | |
|---|---|---|
| 1 | `TraceType.GAUGE` + `GaugePoint` + `GaugeBand` + data union | ✅ |
| 2 | `src/model/gauge.ts` | ✅ |
| 5 | `IS_ORIENTED` — `false` | ✅ |
| 6 | Braille encoder — `BarBrailleEncoder`, one cell | ✅ |
| 7 | `docs/BRAILLE.md` | ✅ |

The braille is a single cell scaled against the dial's ends, so a reader running a finger along a dashboard row feels each tile's fill level without leaving braille mode. The target and band aren't encoded — one cell has four height levels and no room for four quantities — and the docs say so rather than leaving it looking like an omission.

## Verification actually run

| Step | Result |
|---|---|
| `npm test` | **2089 passed** (was 2069) + **73 passed** |
| `npx playwright test gauge.spec.ts` | **7 passed** |
| `npm run build` | clean |

The e2e fixture is chosen so a wrong reading can't look right: the value sits *inside* a band rather than on an edge, and is short of the target rather than equal to it.

---

### DIFF
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -336,6 +336,24 @@ announces, and a display of five cells cannot carry five words.
 Word clouds use a single-line representation. On multiline braille displays
 the terms appear on the first line and the remaining lines are unused.
 
+## Gauge and bullet chart
+
+A gauge's braille is a **single cell**, scaled against the dial's own ends
+rather than against the value -- so the cell's dot height is where along the
+dial the measure sits, not merely that a measure exists.
+
+That is the whole of what braille can carry here, and it is worth carrying: a
+reader running a finger along a dashboard row feels each tile's fill level
+without leaving braille mode. The target, the band and the range are not
+encoded, because one cell has four height levels and no room for four
+quantities; they are announced in text instead, which is where a reader asking
+"how far off target" is already looking.
+
+### Multiline Displays
+
+Gauges use a single-line representation. On multiline braille displays the
+measure appears on the first line and the remaining lines are unused.
+
 ## Multiline Braille Display Support
 
 By leveraging the two-dimensional nature of multiline braille displays, MAIDR can represent multiple lines of a plot simultaneously, allowing users to perceive the distribution of values across all lines at once. This is particularly beneficial for plots with multiple groups or categories, such as grouped boxplots or line plots with multiple lines, and it also applies to scatter plot grid navigation.
--- a/e2e_tests/page-objects/plots/gauge-page.ts
+++ b/e2e_tests/page-objects/plots/gauge-page.ts
@@ -1,0 +1,140 @@
+import type { Page } from '@playwright/test';
+import { expect } from '@playwright/test';
+import { TestConstants } from '../../utils/constants';
+import { GaugePlotError } from '../../utils/errors';
+import { BasePage } from '../base-page';
+
+/**
+ * Page object for the stacked gauge page.
+ *
+ * Provides the interactions the gauge spec needs: reaching the example,
+ * activating MAIDR on it, and reading back the announcement, the braille
+ * output and the text-mode state.
+ */
+export class GaugePlotPage extends BasePage {
+  /**
+   * Selectors for various UI elements
+   */
+  protected override readonly selectors = {
+    notification: `#${TestConstants.MAIDR_NOTIFICATION_CONTAINER} ${TestConstants.PARAGRAPH}`,
+    info: `#${TestConstants.MAIDR_INFO_CONTAINER} ${TestConstants.PARAGRAPH}`,
+    speedIndicator: `#${TestConstants.MAIDR_SPEED_INDICATOR}${TestConstants.GAUGE_ID}`,
+    svg: `svg`,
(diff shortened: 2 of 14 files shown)


jooyoungseo:
All three addressed in 053f1b8b.

**🔴 `range` replaces the main-axis value — confirmed, and the diagnosis was exact.**

`formatVerboseTraceText` (`src/service/text.ts:472-482`) is an `else if` chain: `state.range` is checked *before* `state.main.value` and returns instead of it. So the gauge announced

```
Measure is 0 through 100, ok Percent is 73, Target is 80
```

— every number the tests looked for was present, and the measure's name was gone. The bounds were also being formatted through `mainAxisType`, i.e. the category axis's formatter, which is the wrong one for them.

Moved to `z`, the slot for a third quantity, which appends `", Range is 0 to 100"` after the value and formats through the `z` formatter. The announcement is now:

```
Measure is Conversion, ok Percent is 73, Range is 0 to 100, Target is 80
```

`z` also happens to be what `announcesSectionBeforeLabel` keys on, so the band now reads on the candlestick branch rather than the box-plot one — same output here ("ok" is already lower case), but it is no longer relying on that.

There is a comment at the assignment saying why `range` is not the natural home it looks like, since the next person to read this will reach for it.

**🟡 Compare mode.** Added `supportsCompareMode(): boolean { return false; }`. Same reasoning as `supportsExtrema = false` — the rotor was offering "Navigate to Lower Value" on a chart with no other value to reach.

**🔵 Union ordering.** `GaugePoint` moved after `ErrorBarPoint[]`.

**On the test gap, which is the part worth fixing properly.**

You were right about why both suites missed it. The unit tests asserted the `TextState` object, where every field was populated and correct; the e2e asserted the announcement *contained* `'73'`, `'100'`, `'80'`, `'ok'`, all of which survived the regression, just attached to the wrong labels. Neither suite ever looked at the sentence.

So rather than only re-pointing the assertion at `z`, both now assert the rendered text:

- `test/model/gauge.test.ts` gained a `describe('the announcement a reader hears')` that runs the state through `TextService` and asserts the full string, plus a case for a bare gauge with no band or target. A field in the wrong slot now fails on the sentence, not on a state shape that looks fine.
- A companion test asserts `text.range` is `undefined`, with the reason at the assertion — this is a field that must stay empty, and nothing else would say so.
- The e2e now asserts `'Measure is Conversion'`, closing the exact gap you named.

**Verification:** `npm run lint:fix`, `npx tsc --noEmit`, `npm run build` clean; `npm test` 2092 + 73 passing across all projects; `npx playwright test --project=chromium gauge` 7/7.

---

Issue #792, opened 2026-08-10
feat: support lollipop, dot and dumbbell charts


jooyoungseo:
## Description

Three closely related marks, none of them expressible today:

- **Dot plot** — a category and a value, drawn as a point instead of a bar.
- **Lollipop** — the same, with a stem to the baseline.
- **Dumbbell** — two values per category (before / after, or group A / group B) joined by a segment.

## Proposed mapping

Dot and lollipop are `bar` with a different mark; they need no new model at all beyond a distinct type so the description and the chart-type label are honest.

Dumbbell is the interesting one: each row carries a pair and the *gap between them is the message*. That is the same "compare two magnitudes and announce the difference" job `src/model/candlestickDelta.ts` already does, so the delta announcement and its audio treatment can be lifted directly.

```
{ x: 'Denmark', start: 71.2, end: 78.4, delta: 7.2 }
```

## Difficulty

Low.

PR #829, opened 2026-08-11
feat(dumbbell): read a pair per category with the gap between them

jooyoungseo:
Addresses the dumbbell half of #792.

A dumbbell is drawn to be read as a **comparison**, not as two numbers. The segment between each pair of dots is the finding — which rows gained, which lost, and by how much — and it is the one thing a reader cannot recover from hearing the two ends one at a time, because that means holding one number while navigating to the other and subtracting by ear on **every row of the chart**.

So the change travels with both ends:

```
Country is Denmark, 1990 Years is 71.2, Increase is 7.2
Country is Latvia,  2020 Years is 69.5, Decrease is 5.1
```

## Design notes

**The change is derived, not authored.** A drawn segment cannot disagree with the dots it joins, so an authored `delta` would be a second source of truth for a quantity that already has one — and the one a reader would be told is the one the chart did not draw. (The issue's sketch carried it in the data; this is the one place I departed from it.)

**Named by direction rather than signed.** `Increase is 7.2` / `Decrease is 5.1` / `Change is 0`. A bare `-3.1` asks the reader to hear a minus sign and work out which way it points, on every row. The three cases are separately tested because a `delta >= 0` split would announce an unchanged pair as an increase of zero.

**Rows keep the chart's order, not magnitude order.** Unlike an error bar's `lower/value/upper`, a dumbbell's ends have no fixed order on the value axis — a declining row draws its end *below* its start, and a real chart holds both directions at once. Sorting rows by size would put Latvia's finish on the line that holds every other row's start, and the label under the cursor would flip as the reader arrowed sideways.

**The end names live on the layer, not on each point.** `DumbbellData` is an object — as `HeatmapData` and `GaugePoint` already are — so a producer cannot emit rows that disagree about what the chart is comparing. Those names are the content of the comparison: announced as "start" and "end", a chart of 1990 against 2020 tells the reader which dot they are on and not which year it is, which is exactly what the legend gives a sighted reader for free. They default, so a producer without them still emits a chart that reads.

**Extrema rank the biggest mover each way**, not the largest value — the chart is drawn to show which rows moved, and the tallest dot is a fact about the category rather than about the comparison. `supportsExtrema` is `true`, so **both** `getExtremaTargets` and `navigateToExtrema` are overridden; the base throws precisely when that flag is set.

**Braille is two rows, each scaled against its own range**, so the two profiles are comparable as shapes. A shared range would flatten both toward the middle of the combined spread when the ends occupy different parts of the value axis. This is the one trace so far that uses both lines of a multiline display for something.

## Registration

| # | Place | |
|---|---|---|
| 1 | `TraceType.DUMBBELL` + `DumbbellPoint` + `DumbbellData` + data union | ✅ |
| 2 | `src/model/dumbbell.ts` | ✅ |
| 5 | `IS_ORIENTED` — `true` | ✅ |
| 6 | Braille encoder — `LineBrailleEncoder`, two rows | ✅ |
| 7 | `docs/BRAILLE.md` | ✅ |

## Scope

**Dumbbell only.** #792 also asks for dot and lollipop charts, which it correctly describes as needing no new model — but each still costs six registration points, a grammar entry, an example and a spec, and neither carries a design question worth deciding alongside this one. Left open on #792 rather than bundled.

## Verification actually run

| Step | Result |
|---|---|
| `npm test` | **2118 passed** (was 2092) + **73 passed** |
| `npx playwright test --project=chromium dumbbell` | **7 passed** |
| `npm run build` | clean |

The e2e fixture holds a rise, a fall **and** an unchanged pair, and a test asserts it does — a chart where everything rose would let a reading that ignored direction look right. The example is drawn horizontally, which is how a dumbbell usually is, so the orientation path is exercised end to end rather than only in unit tests.

Two failures found this in development, both kept as tests: `test/util/orientation.test.ts` catches an unregistered type via its total-coverage assertion, and the e2e instruction text is `horizontal dumbbell` — the oriented-type prefix — which is correct behaviour my constant had wrong.

---

### DIFF
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -354,6 +354,33 @@ quantities; they are announced in text instead, which is where a reader asking
 Gauges use a single-line representation. On multiline braille displays the
 measure appears on the first line and the remaining lines are unused.
 
+## Dumbbell
+
+A dumbbell's braille is **two rows** — the starting values and the finishing
+ones — one cell per category, with each row scaled against **its own** range.
+
+That per-row scaling is the same treatment a multi-line plot gets, and it is
+what makes the two rows comparable as *shapes*: the point of the chart is
+whether the second row's profile sits differently from the first, and a single
+shared range would flatten both toward the middle of the combined spread when
+the two ends occupy different parts of the value axis.
+
+The rows keep the chart's own order — start first, then end — rather than being
+sorted by magnitude. A dumbbell normally contains rises and falls at once, so
+sorting would put a declining row's finishing value on the line that holds
+every other row's start, and the reader would have no way to know which line
+they were feeling.
+
+The change itself is not encoded. It is what the text announcement carries at
+both ends of every row, and a third line of deltas would read as a third series
+rather than as the gap between the first two.
+
+### Multiline Displays
+
+Dumbbell charts use both lines on a multiline display: the starting values on
+the first and the finishing ones on the second, so the two profiles can be
+compared with one sweep rather than by toggling between rows.
+
 ## Multiline Braille Display Support
 
 By leveraging the two-dimensional nature of multiline braille displays, MAIDR can represent multiple lines of a plot simultaneously, allowing users to perceive the distribution of values across all lines at once. This is particularly beneficial for plots with multiple groups or categories, such as grouped boxplots or line plots with multiple lines, and it also applies to scatter plot grid navigation.
--- a/e2e_tests/page-objects/plots/dumbbell-page.ts
+++ b/e2e_tests/page-objects/plots/dumbbell-page.ts
@@ -1,0 +1,140 @@
+import type { Page } from '@playwright/test';
+import { expect } from '@playwright/test';
+import { TestConstants } from '../../utils/constants';
+import { DumbbellPlotError } from '../../utils/errors';
+import { BasePage } from '../base-page';
+
+/**
+ * Page object for the dumbbell chart page.
+ *
+ * Provides the interactions the dumbbell spec needs: reaching the example,
+ * activating MAIDR on it, and reading back the announcement, the braille
+ * output and the text-mode state.
+ */
+export class DumbbellPlotPage extends BasePage {
+  /**
+   * Selectors for various UI elements
+   */
+  protected override readonly selectors = {
+    notification: `#${TestConstants.MAIDR_NOTIFICATION_CONTAINER} ${TestConstants.PARAGRAPH}`,
+    info: `#${TestConstants.MAIDR_INFO_CONTAINER} ${TestConstants.PARAGRAPH}`,
+    speedIndicator: `#${TestConstants.MAIDR_SPEED_INDICATOR}${TestConstants.DUMBBELL_ID}`,
+    svg: `svg`,
+    // The textarea's id ends with a React `useId()` suffix, so match on the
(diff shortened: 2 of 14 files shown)


jooyoungseo:
Both real. Fixed in a967592 — though the first one needed a different fix than the one suggested, and the second turned out not to be safe to fix here at all.

## 🐛 Extrema label — confirmed, and the suggested fix is also wrong

The bug is exactly as you describe, and the reasoning about `aria-label` is why it mattered enough to get right rather than quickly.

But the proposed fix — condition on the sign, `largest.change >= 0 ? 'increase' : 'decrease'` — produces a *different* wrong answer. On your own two-row example (Δ=-5, Δ=-19), `largest` is the Δ=-5 row, and calling that **"Largest decrease"** names the row that fell *least* as the biggest faller. The label goes from claiming the wrong direction to claiming the wrong row.

The two extremes are the most positive change and the most negative one, and neither is necessarily a rise or a fall. Each has three cases:

| | change > 0 | change = 0 | change < 0 |
|---|---|---|---|
| **most positive** | Largest increase | No change | **Smallest decrease** |
| **most negative** | **Smallest increase** | No change | Largest decrease |

`rankLabel(kind, change)` covers all six. Your example now reads `Smallest decrease at a` / `Largest decrease at b`.

**And `description` had the same defect**, not the correct handling you credited it with — it used `largest.change >= 0 ? 'increase' : 'decrease'`, which is the fix you proposed, so it produced "Largest decrease" for the least-declining row. It goes through `rankLabel` now too.

Three tests added: all-declining, all-rising, and the description's labels. You were right that mixing the directions in every fixture is what hid this — the mixed fixture is the one that makes a wrong reading *look* right here.

## ⚠️ Section case — confirmed, but option (a) is not safe

The inconsistency is real: `text.ts:511` lower-cases, `text.ts:624` does not.

Your suggested fix — drop the `.toLowerCase()`, "safe since `ErrorBarTrace`/`WaterfallTrace`'s constants are already lowercase" — misses the third caller. `BoxplotSection` is **capitalised**:

```ts
MIN: 'Minimum', MAX: 'Maximum', MEAN: 'Mean', Q1: '25%'
```

So that call is load-bearing for box plots, and removing it changes every box plot's verbose announcement from `minimum Value is 3.2` to `Minimum Value is 3.2`. Arguably an improvement — but it is a visible change to a chart type this PR does not touch, and it wants a product call plus an e2e update.

I also considered lower-casing the end labels in the model so the two modes agree, and rejected it: that destroys a distinction the data carries (`Q1 2024` → `q1 2024`) to compensate for one the *formatting layer* chose to drop. The trace should not corrupt its input to work around a downstream decision.

So the comment is corrected to state what actually happens, with a pointer to **#830**, which I filed with the three-caller table, the three options, and the note that the outlier branch at `text.ts:499-503` wants the same treatment.

One thing worth flagging from your own analysis, since it generalises past this PR: dumbbell is the first trace to feed a free-form producer string into `section`, and `text.ts:434-437` already documents that branch as depending on its callers happening to be lower case. That assumption is now load-bearing for data the producer controls, which is what makes #830 worth doing rather than noting.

**Verification:** `npm test` **2121 passed** (was 2118) + 73; `npx playwright test --project=chromium dumbbell` 7/7; `type-check`, `lint:fix`, `build` clean.

---

PR #831, opened 2026-08-11
feat(grammar): let a layer say which one it is

### LINKED ISSUE #828: feat(errorbar): make grouped uncertainty readable — layers cannot be told apart
> **Rewritten.** The original framing — "the flat `ErrorBarPoint[]` blocks grouped uncertainty" — was wrong, and I checked before building on it. Multiple layers per subplot already work, so a binding **can** emit one `error_bar` layer per hue level today with no grammar change. What it cannot do is make them tellable apart. That is the real gap, and it is a smaller one.

## What actually works today

`examples/violin.html` puts two layers (`violin_box` + `violin_kde`) in one subplot, and `TextService.formatLayerSwitchAnnouncement` announces the move between them. On the binding side, `py-maidr`'s `Maidr._merge_plots_by_subplot_position` groups every plot registered at the same `(row, col)` into one subplot's `layers` array — so two `create_maidr` calls on one axes already produce two layers.

Nothing stops a hued point plot from emitting one `error_bar` layer per hue level right now.

## What does not work

`src/service/text.ts:217`:

```ts
let announcement = `Layer ${state.index} of ${state.size}: ${state.plotType || state.traceType} plot`;
```

The announcement is built from the **trace type**. Two `error_bar` layers announce identically:

```
Layer 1 of 2: error_bar plot at Day is Sat, Total bill is 20.4
Layer 2 of 2: error_bar plot at Day is Sat, Total bill is 19.7
```

A reader switching between them hears two different numbers and is never told that the first is Male and the second is Female — which is the whole content of a hue split, and exactly what the legend gives a sighted reader for free.

That is what blocks grouped uncertainty in all three bindings, not the point-array shape.

## Why the obvious fix is not one

`state.title` is already carried on the trace state, so announcing it looks like a one-liner. It is not: `layer.title` in practice holds the **chart's** title, not the layer's identity — `py-maidr` emits the figure title (or `""`) there for every layer of every chart. Preferring it would replace a useful type name with a repeated chart title on every existing multi-layer figure, including `examples/violin.html`.

So a layer needs somewhere to say what *it* is, distinct from what the figure is. Sketches, roughly in order of how much surface they add:

1. **A per-layer name on `MaidrLayer`** — `name?: string`, announced when present, falling back to the current type phrasing. Smallest change; the binding sets it to the hue level.
2. **Reuse the `z` axis.** A hue split already has `axes.z.label` (the legend title, e.g. `sex`) — pair it with a per-layer level to announce `Layer 1 of 2: sex is Male`. More semantically honest, more plumbing.
3. **Widen the data union** to `ErrorBarPoint[][]`, so the series are rows of one trace and the reader arrows between them rather than switching layers. Best for the actual question a grouped error bar chart poses — *do A's and B's intervals overlap at this category?* — because it makes the comparison a keypress instead of a mode change. Largest change: the grid is currently sections × samples, and series would have to become a third axis or be flattened into the rows.

(1) unblocks the bindings immediately. (3) is the better reading and can come later without invalidating it.

## Downstream

- **py-maidr** — `sns.pointplot(..., hue=...)` (#352) drops its intervals and emits a plain multi-line layer. Same for `sns.barplot(..., hue=...)`.
- **r-maidr** — a `geom_errorbar` with a `colour`/`fill` aesthetic has the same shape.

Both are waiting on a way to say which series a layer is, not on a way to carry two series.

jooyoungseo:
Closes #828.

## The problem, restated after checking it

I opened #828 believing the flat `ErrorBarPoint[]` was what blocked grouped uncertainty. **It is not.** Multiple layers per subplot already work — `examples/violin.html` puts two in one subplot, and py-maidr's `_merge_plots_by_subplot_position` already groups every plot registered at the same `(row, col)` into one subplot's `layers`. A binding could emit one `error_bar` layer per hue level today.

What it could not do is make them tellable apart. `src/service/text.ts:217` built the announcement from the **trace type**:

```
Layer 1 of 2: error_bar plot at Day is Sat, Total bill is 20.4
Layer 2 of 2: error_bar plot at Day is Sat, Total bill is 19.7
```

The reader hears two different sets of numbers and is never told that the first is Male and the second Female — the whole content of the split, and what a legend gives a sighted reader at a glance. #828 is rewritten with the corrected framing.

## The change

`MaidrLayer.name` — what *this layer* is — announced in place of the trace type. A layer without one keeps the current wording.

```
Layer 1 of 2: Male at Day is Thur, Total bill is 18.7
Layer 2 of 2: Female at Day is Thur, Total bill is 16.7
```

**Deliberately not `title`.** It looks like the natural home and is the trap: producers put the *figure's* title there for every layer of a figure — py-maidr emits the figure title or `""` — so preferring it would announce the same string at each layer. That is the same bug in a new costume, and a regression for every existing multi-layer figure including `examples/violin.html`.

**Falls back to the type, not to a placeholder.** A figure whose layers genuinely differ in kind (`violin_box` + `violin_kde`) is better served by the type than by "Layer 1 of 2:" with nothing after it. A blank or whitespace-only name is treated as absent for the same reason.

## Why this is the piece worth doing

Grouped uncertainty is the case where reading an interval matters *most* — whether two group means differ is answered by whether their intervals overlap, and a grouped error bar chart poses that comparison once per category. Three bindings are currently unable to describe it:

- **py-maidr** — `sns.pointplot(..., hue=...)` (xability/py-maidr#352) drops its intervals; `sns.barplot(..., hue=...)` the same
- **r-maidr** — `geom_errorbar` with a `colour`/`fill` aesthetic

All three were waiting on a way to say *which series a layer is*, not on a way to carry two series in one.

## `examples/errorbar-grouped.html`

New, and the point of it is that the feature is demonstrable rather than theoretical: two `error_bar` layers in one subplot, one per sex, `PageUp`/`PageDown` between them. The e2e drives exactly the reading the chart exists for — switch layers at one category and hear both groups' intervals for it:

| | Thur |
|---|---|
| Male | 18.7, 16.1 to 21.7 |
| Female | 16.7, 14.2 to 19.5 |

## What is deliberately not here

Arrowing *between* series rather than switching layers — sketch (3) in #828, where the series become rows of one trace. That is the better reading for the overlap question, since it makes the comparison a keypress rather than a mode change, and it is a real design change: the error bar grid is currently sections × samples and series would have to become a third axis or be flattened into the rows. It can land later without invalidating this.

## Verification actually run

| Step | Result |
|---|---|
| `npm test` | **2127 passed** (was 2121) + **73 passed** |
| `npx playwright test --project=chromium errorbarGrouped` | **3 passed** |
| `npm run build` | clean |

Six unit tests cover the naming, the fallback, the blank-name case, and — the one that pins the trap — that the figure title is *not* what gets announced.

---

### DIFF
--- a/docs/react.md
+++ b/docs/react.md
@@ -121,6 +121,9 @@ interface MaidrLayer {
   id: string;              // Unique layer identifier
   type: TraceType;         // Plot type (see TraceType enum)
   title?: string;          // Layer title
+  name?: string;           // What this layer is, e.g. 'Male' — announced on a
+                           // layer switch in place of the trace type, so two
+                           // layers of one kind can be told apart
   selectors?: string | string[]; // CSS selectors for SVG highlight elements
   orientation?: Orientation;     // 'vert' or 'horz' (for bar/box plots)
   axes?: {
--- a/e2e_tests/specs/errorbarGrouped.spec.ts
+++ b/e2e_tests/specs/errorbarGrouped.spec.ts
@@ -1,0 +1,117 @@
+import type { Maidr } from '../../src/type/grammar';
+import { expect, test } from '@playwright/test';
+import { BasePage } from '../page-objects/base-page';
+import { TestConstants } from '../utils/constants';
+import { extractMaidrData } from '../utils/maidr-data';
+import { normalizeText } from '../utils/text';
+
+/**
+ * A grouped error bar chart poses one question — do the two groups' intervals
+ * overlap at this category? — and answering it means reading both groups.
+ *
+ * They are separate layers, so the reader moves between them rather than
+ * arrowing. What makes that readable is each layer saying which group it is:
+ * without a name both announce "error_bar plot" and the reader hears two sets
+ * of numbers with nothing to attach them to.
+ */
+
+/** The example's own page, driven through the shared base helpers. */
+class GroupedErrorBarPage extends BasePage {
+  protected override readonly selectors = {
+    notification: `#${TestConstants.MAIDR_NOTIFICATION_CONTAINER} ${TestConstants.PARAGRAPH}`,
+    svg: `svg`,
+    helpModal: TestConstants.MAIDR_HELP_MODAL,
+    helpModalTitle: TestConstants.MAIDR_HELP_MODAL_TITLE,
+    helpModalClose: TestConstants.HELP_MENU_CLOSE_BUTTON,
+    settingsModal: TestConstants.MAIDR_SETTINGS_MODAL,
+    chatModal: TestConstants.MAIDR_CHAT_MODAL,
+  };
+
+  /** Navigates to the grouped error bar example. */
+  public async navigateToPlot(): Promise<void> {
+    await super.navigateTo('examples/errorbar-grouped.html');
+    await super.verifyPlotLoaded(this.selectors.svg);
+  }
+
+  /** Activates MAIDR on the chart. */
+  public override async activateMaidr(): Promise<void> {
+    await super.activateMaidr(this.selectors.svg, 'errorbar-grouped');
+  }
+
+  /**
+   * Reads the announcement currently on screen.
+   * @returns The announcement text
+   */
+  public override async getInstructionText(): Promise<string> {
+    return super.getInstructionText(this.selectors.notification);
+  }
+
+  /** Moves to the next layer. */
+  public async moveToNextLayer(): Promise<void> {
+    await this.pressKey(TestConstants.PAGE_UP_KEY, 'move to the next layer');
+  }
+}
+
+test.describe('Grouped error bar', () => {
+  let maidrData: Maidr;
+
+  test.beforeAll(async ({ browser }) => {
+    const context = await browser.newContext();
+    const page = await context.newPage();
+
+    try {
+      const plot = new GroupedErrorBarPage(page);
+      await plot.navigateToPlot();
+      await page.waitForSelector(`svg`, { timeout: 10000 });
+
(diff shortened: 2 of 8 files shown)


jooyoungseo:
Both taken, in 3058016.

## `docs/react.md`

Right — that file hand-maintains the `MaidrLayer` interface for producers, and a field they cannot discover is a field that does not exist as far as they are concerned. Added with a note on what it is for.

## The second announcement path

Your trace is correct that it is unreachable today, and I took it anyway — but as a **refactor rather than a second patch**.

`formatSubplotText` and `formatLayerSwitchAnnouncement` compose the same sentence from different states. Patching the fallback into both would leave two copies of the same phrase, which is precisely how they came to disagree in the first place. So both now call one `layerIdentity(state)`, and the reason is written at the helper.

That is not a hypothetical concern here — it is the lesson from #829, merged an hour ago. The extrema labels and the description stats in `DumbbellTrace` drifted apart exactly this way: each built its own `change >= 0 ? 'increase' : 'decrease'`, one was fixed, the other was not, and the review found the survivor. Two copies of a phrase is the shape of that bug.

Worth adding to your note: the invariant that makes this path dead is documented in `test/model/contextStackInvariant.test.ts` and relied on defensively by `AnnounceCommand`, but nothing *enforces* that the two sentences match. Sharing the helper does.

**Verification:** `npm test` 2127 + 73 passing (unchanged — the refactor is behaviour-preserving on the live path, which is the point); `npx playwright test --project=chromium errorbarGrouped` 3/3; `type-check`, `lint:fix`, `build` clean.

---

PR #833, opened 2026-08-11
feat(radar): read categories arranged around a circle

### LINKED ISSUE #800: feat: support radar, spider and polar area charts
## Description

Radar / spider charts and polar area charts are offered by Chart.js, ECharts, amCharts, Highcharts and Plotly, and MAIDR reads none of them. The Vega-Lite adapter is explicit about it — `src/adapters/vegalite/converters.ts:646` returns `null` for an `arc` mark without a `theta` encoding, precisely because a radial plot has no slice model to fall back on.

## Proposed mapping

Structurally a multi-line layer closed into a loop: each spoke is a column, each series is a row.

```
[ {x: 'speed', y: 8, z: 'model A'}, {x: 'range', y: 4, z: 'model A'}, ... ]
```

Navigation, braille and audio all transfer from the multi-line model. What is new is angular position.

## Worth doing together with #780

#780 reports that a pie sounds like a bar chart — panning is linear and nothing says where a slice sits on the circle. That is the same missing capability: MAIDR has no notion of *angle* as a navigable dimension. Building polar position handling once, for pie and radar together, is considerably less work than building it twice, and it removes a known defect while adding a chart family.

## Scope

- [ ] Polar position in the trace state (angle, and where the cursor sits on the circle)
- [ ] Audio panning that follows the angle rather than a linear sweep
- [ ] `TraceType.RADAR` / `POLAR_AREA` + trace classes
- [ ] Apply the same position handling to `PieTrace` (#780)

## Difficulty

Medium.

jooyoungseo:
Closes #800.

Radar, spider and polar area charts are offered by Chart.js, ECharts, amCharts, Highcharts and Plotly, and MAIDR read none of them. The Vega-Lite adapter was explicit about the gap — `converters.ts:646` returns `null` for an `arc` mark without a `theta` encoding.

## Smaller than the issue expected, because half of it already shipped

#800 proposed building polar position as new state, together with #780. **#780 was closed by #784 and is released** — `PieTrace` already pans by angle. So no new state field and no service change: angular position is expressible in the existing `panning` shape, and there is a worked example in the tree. I posted the scope check on #800 before starting.

## The change

The data is a multi-line layer's — each spoke a column, each series a row — so navigation, braille and pitch transfer unchanged and `RadarTrace extends LineTrace` rather than restating them. **No new point shape**; `LinePoint[][]` was already in the data union.

What a circle adds is *where a spoke sits*:

```ts
x: (Math.sin(angle) + 1) / 2,
cols: 2,
```

`AudioService` reads the pan as `interpolate(x, 0, cols - 1, -1, 1)`, so `cols: 2` makes the mapping `2x − 1` and `x = (sin θ + 1) / 2` lands it on `sin θ` exactly. Six spokes then sweep **out and back** rather than hard-left to hard-right, which is the difference between hearing a circle and hearing a row of bars.

The unit test asserts the pan the *service* will compute rather than `panning.x` alone — `cols` is the half of the pair that makes the arithmetic work, and an assertion on `x` would pass for a trace that got it wrong.

**Spokes are evenly spaced**, unlike a pie's slices whose angles follow from their shares. A test doubles one value and asserts no spoke moves.

**`plotType` is overridden** because `LineTrace` names itself `'single line'` / `'multiline'`, and that is what the instruction text and the layer-switch cue announce. A reader told they are on a line plot has been told the wrong chart — pinned in both the unit tests and the e2e.

**Two types, one class** — `RADAR` and `POLAR_AREA` differ in the mark, not in what a reader navigates, the way `area` and `stacked_area` share `AreaTrace`.

## Deliberately not here

**Wrapping from the last spoke back to the first.** The outline is closed and wrapping would arguably read better — but every other trace stops at its bounds, and a reader who has learned that would be silently looped without being told. That is a decision for the navigation model, not something to slip into one trace. There is a test asserting it currently stops, so the day someone decides otherwise the case turns red rather than the behaviour drifting.

**A declared start angle.** The payload does not state where spoke zero sits, and neither does a pie's — the gap #780 flagged. It matters less here: nothing announces an absolute clock position for a radar, so a producer laying its first spoke elsewhere shifts which spoke is heard at centre without making any announcement untrue. Noted on #800 as the thing to settle before any clock-position wording is added.

## Registration

| # | Place | |
|---|---|---|
| 1 | `TraceType.RADAR` + `POLAR_AREA` (data union already covered) | ✅ |
| 2 | `src/model/radar.ts` | ✅ |
| 3 | `TraceFactory.create()` cases | ✅ |
| 5 | `IS_ORIENTED` — `false`, as a pie is | ✅ |
| 6 | Braille encoder — `LineBrailleEncoder`, one row per series | ✅ |
| 7 | `docs/BRAILLE.md` | ✅ |

The braille section says what braille *cannot* carry here and why that is not a loss: a display is a line of cells and an angle has nowhere to go in it, but the angle is carried better by the pan. Braille answers "how do the series compare across the spokes"; audio answers "where on the circle am I".

## Verification actually run

| Step | Result |
|---|---|
| `npm test` | **2143 passed** (was 2127) + **73 passed** |
| `npx playwright test --project=chromium radar` | **5 passed** |
| `npm run build` | clean |

Two tests failed first and were corrected rather than the code: `test/util/orientation.test.ts` caught both new types before I registered them (the total-coverage guard doing its job), and my navigation test asserted `DOWNWARD` reaches the next series when series zero is the *bottom* row — I had asserted my assumption rather than the model, and fixed the test.

---

### DIFF
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -381,6 +381,26 @@ Dumbbell charts use both lines on a multiline display: the starting values on
 the first and the finishing ones on the second, so the two profiles can be
 compared with one sweep rather than by toggling between rows.
 
+## Radar and polar area
+
+A radar's braille is the multi-line encoding unchanged: **one row per series**,
+one cell per spoke, each row scaled against its own range.
+
+What braille cannot carry here is the circle. A display is a line of cells and
+a spoke's angle has nowhere to go in it, so the row reads as the sequence of
+values it is — first spoke to last, in the order the chart declares them.
+
+That is not a loss the way a missing magnitude would be, because the angle is
+carried elsewhere and better: the **stereo position** follows the spoke around
+the dial, so a sweep goes out and comes back. Braille answers "how do the
+series compare across the spokes"; audio answers "where on the circle am I".
+Encoding a rotation into cell heights would answer neither.
+
+### Multiline Displays
+
+Radar charts use one line per series on a multiline display, so several models'
+profiles can be compared with one sweep rather than by toggling between rows.
+
 ## Multiline Braille Display Support
 
 By leveraging the two-dimensional nature of multiline braille displays, MAIDR can represent multiple lines of a plot simultaneously, allowing users to perceive the distribution of values across all lines at once. This is particularly beneficial for plots with multiple groups or categories, such as grouped boxplots or line plots with multiple lines, and it also applies to scatter plot grid navigation.
--- a/e2e_tests/specs/radar.spec.ts
+++ b/e2e_tests/specs/radar.spec.ts
@@ -1,0 +1,132 @@
+import type { Maidr } from '../../src/type/grammar';
+import { expect, test } from '@playwright/test';
+import { BasePage } from '../page-objects/base-page';
+import { TestConstants } from '../utils/constants';
+import { extractMaidrData } from '../utils/maidr-data';
+import { normalizeText } from '../utils/text';
+
+/**
+ * Every braille cell MAIDR can emit lives in the Unicode braille block
+ * (U+2800 to U+28FF), so a display carrying anything else is not braille
+ * output.
+ */
+const BRAILLE_CELL = /^[\u2800-\u28FF]+$/;
+
+/** The example's own page, driven through the shared base helpers. */
+class RadarPlotPage extends BasePage {
+  protected override readonly selectors = {
+    notification: `#${TestConstants.MAIDR_NOTIFICATION_CONTAINER} ${TestConstants.PARAGRAPH}`,
+    svg: `svg`,
+    braille: `textarea[id^="${TestConstants.BRAILLE_TEXTAREA}"]`,
+    helpModal: TestConstants.MAIDR_HELP_MODAL,
+    helpModalTitle: TestConstants.MAIDR_HELP_MODAL_TITLE,
+    helpModalClose: TestConstants.HELP_MENU_CLOSE_BUTTON,
+    settingsModal: TestConstants.MAIDR_SETTINGS_MODAL,
+    chatModal: TestConstants.MAIDR_CHAT_MODAL,
+  };
+
+  /** Navigates to the radar example. */
+  public async navigateToPlot(): Promise<void> {
+    await super.navigateTo('examples/radar.html');
+    await super.verifyPlotLoaded(this.selectors.svg);
+  }
+
+  /** Activates MAIDR on the chart. */
+  public override async activateMaidr(): Promise<void> {
+    await super.activateMaidr(this.selectors.svg, 'radar-models');
+  }
+
+  /**
+   * Reads the announcement currently on screen.
+   * @returns The announcement text
+   */
+  public override async getInstructionText(): Promise<string> {
+    return super.getInstructionText(this.selectors.notification);
+  }
+
+  /**
+   * Reads the current contents of the braille display.
+   * @returns The braille content, whitespace trimmed
+   */
+  public async getBrailleContent(): Promise<string> {
+    const textarea = await this.waitForElement(this.selectors.braille);
+    return (await textarea.inputValue()).trim();
+  }
+}
+
+test.describe('Radar', () => {
+  let maidrData: Maidr;
+
+  test.beforeAll(async ({ browser }) => {
+    const context = await browser.newContext();
+    const page = await context.newPage();
+
+    try {
+      const plot = new RadarPlotPage(page);
+      await plot.navigateToPlot();
+      await page.waitForSelector(`svg`, { timeout: 10000 });
+
+      maidrData = await extractMaidrData(page);
+    } finally {
+      await context.close();
+    }
+  });
+
+  test.beforeEach(async ({ page }) => {
+    await new RadarPlotPage(page).navigateToPlot();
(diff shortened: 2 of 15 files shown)


jooyoungseo:
Both real, both mine, fixed in 19d5fed. Your diagnosis of the shared root cause is the useful part: I overrode `plotType` and stopped there, but **two other sites gate line behaviour on the trace type** rather than on what the trace inherits — so a radar reporting its own type fell out of both while still inheriting the navigation that makes those branches apply.

## 1 — `p` drops the series name

Confirmed by reading the branch. A radar with two or more series sets `groupCount` and `group` (inherited), fails `traceType === LINE || STEP`, and lands on `is2DPlot` → `announce2DPosition`:

```
column 3 of 4, row 2 of 2          <- what a radar got
Series 1 of 3, Model is Model B, Position is 3 of 10   <- what it gets now
```

Losing the series name is the part that matters: on a radar the series *is* the outline being traced, and row/column language describes a grid the chart does not draw.

Added both circular types to the condition, with `'Series'` as the noun. I inverted the noun expression to `traceType === LINE ? 'Line' : 'Series'` rather than adding a third and fourth clause to the `=== STEP` test — a line is the only one of the four that calls its rows lines, and stating it that way means the next trace to extend `LineTrace` gets the right noun by default rather than the wrong one.

Two tests: one asserting the full announcement, one asserting it contains no "Line".

## 2 — the description dialog. Fixed by parameterising, not by copying

You are right that this is the same problem one dialog along, and I should have carried it through.

Rather than override the whole `description` getter — 40 lines duplicated to change four strings, which is precisely the drift risk I have been arguing against elsewhere today — the labels are parameterised on `LineTrace`:

```ts
protected get seriesLabels(): { count; perSeries; names; column } {
  return { count: 'Number of lines', perSeries: 'Points per line',
           names: 'Line names', column: 'Line' };
}
```

**Defaulted to exactly what it emitted before**, so the chart it is named after is unchanged, and `RadarTrace` returns "Number of series" / "Spokes per series" / "Series names" / "Series".

There is a test asserting a plain line chart still says "Number of lines" and "Points per line". A parameterisation whose default silently drifts is worse than the duplication it replaced, and nothing else would have caught that.

## Minor

Thanks for checking `liveData.ts` — agreed the `Array.isArray(layer.data[0])` fallback covers `LinePoint[][]` regardless, so `NESTED_DATA_TYPES` needs no entry. I would rather leave it out than add a line that only looks load-bearing.

**Verification:** `npm test` **2148 passed** (was 2143) + 73; `npx playwright test --project=chromium radar` 5/5; `type-check`, `lint:fix`, `build` clean.

---

jooyoungseo:
Both addressed in `c24d57b`.

**Finding 1 — the stranded doc comment.** Fixed as described: the "Gets the description state for the line trace." block is back directly above `public get description()`, and `seriesLabels` keeps only its own.

**Finding 2 — `intersections` on a radar.** I checked the two things this could mean and they turn out to be different, so it is worth being precise about which one bites.

`state.intersections` is not built by `getSegmentIntersection()`. `findIntersections()` (`src/model/line.ts`) is an exact-match scan — `p.x === currentX && p.y === currentY` — so it means "another series has the same value on this same spoke". On a circle that is still one point: same angle, same radius, same place. The coincidence it reports is real for a radar, not an artefact of reading the data as a straight line. The straight-segment math you are thinking of is `findAllIntersectionsForCurrentLine()`, which feeds rotor navigation, and rotor navigation only visits `intersectionKind === 'point'` results — the exact matches again.

So the payload is not wrong. But the landmine you predicted was there, one level down and worse than a dead computation: `findIntersections()` built its own `panning` literal, `{ x: this.col, cols: this.lineValues[this.row].length }` — the **line's** linear sweep. `RadarTrace` had overridden the `audio` getter, which that literal never goes through. A contributor adding `RADAR` to the `audio.ts:346` gate would have got a chord that swept hard left to hard right on a chart where every other tone goes out and back around the circle: two series meeting at 6 o'clock heard at centre by one route and hard right by the other.

Rather than leave that waiting, I removed the way for the two to disagree. The panning is now one `protected panningFor(row, col)` on `LineTrace`, called by both sites; `RadarTrace` overrides that instead of the whole `audio` getter, so the chord follows the spoke automatically. Two callers existed before the extraction, so this stays on the right side of the "no abstraction before a second caller" rule.

One behavioural note while I was in there: the chord's panning is deliberately the **cursor's** cell for every tone, not each series' own. An intersection is one point several series share, so they have to arrive from the same place — `group: r` is what tells them apart. That matches what the code already did; I pinned it with a test (`the chord arrives from one place, and 'group' tells the tones apart`) because the extraction made it easy to get wrong in passing.

And I took your first option too, since the reasoning is not obvious from the gate: `src/service/audio.ts` now records why radar and polar area are left out of it. Not "the data would be wrong" — it would be right — but that a chord is several tones at once, and a radar reads position by ear, so sounding them together spends the one cue the layout is chosen for. That is a choice about what a radar is for, which is exactly the kind of thing that should not be re-decided by accident.

Two tests added to `test/model/radar.test.ts` (19 total). Full suite green: 2150 + 73 unit/component/esm, 5/5 radar e2e on chromium, `type-check` and `build` clean.

---

PR #834, opened 2026-08-11
feat(gantt): read a schedule as the intervals it is

### LINKED ISSUE #801: feat: support gantt and timeline charts (interval traces)
## Description

Gantt charts, timelines and swimlane diagrams place bars along a time axis, each with a start and an end. MAIDR has no interval concept, so none of them can be authored.

## Why this one is worth the effort

Most unsupported chart types have a workable fallback — a table, a caption, a sentence in the surrounding prose. A project schedule does not. The information is relational (what overlaps what, what blocks what, where the slack is) and grows quadratically with the number of tasks, so alt text stops being viable almost immediately. This is the chart type where the gap between "sighted reader" and "MAIDR user" is widest.

## Proposed mapping

```
{ row: 'Design', start: '2026-01-05', end: '2026-02-14', label: 'Wireframes' }
```

- Rows are lanes; a lane may hold several non-overlapping intervals.
- Audio: duration as tone length is the obvious encoding, with start position driving pitch or pan so that "later" is audibly later.
- Text: "Wireframes, 5 Jan to 14 Feb, 40 days".
- Braille: a span encoding, closer to `BoxBrailleEncoder` than to the bar encoder.

## Follow-on value

An interval primitive also serves broken-axis bars, range bars (`columnrange`, `arearange`), and candlestick-style low/high ranges outside finance.

## Difficulty

Medium — a new point shape and a new audio treatment, but no change to the row/column navigation model.

jooyoungseo:
Closes #801.

## Why this one

Most unsupported chart types have a workable fallback — a table, a caption, a sentence in the surrounding prose. A schedule does not. What a reader wants from one is relational (what overlaps what, what hands over to what, where the slack is) and grows with the square of the task count, so alt text stops being viable almost immediately. #801 called this the chart type where the gap between a sighted reader and a MAIDR user is widest, and that is the reason it was picked next.

## An interval is not a bar

This is the part that decides everything else. A bar has one number and a baseline, and its height is a magnitude a reader hears directly. An interval has **two numbers and no baseline**: its length is a *difference*, audible only if something computes it, and its position is a second fact that a single pitch cannot also carry.

So the two facts are split across two channels:

| Channel | Carries | Why |
|---|---|---|
| Pitch | the **length** | the magnitude the chart is drawn to compare |
| Stereo position | the **start**, as a fraction of the whole axis | "later" is audibly later |
| Announcement | both ends **and** the length | naming one end names an edge of a bar as though it were the bar |

The panning uses the `cols: 2` arithmetic `PieTrace` and `RadarTrace` already use, so a fraction of the axis lands on the pan directly with no new state field.

**A column index would have been actively wrong here**, in a way it is not for a bar chart. The third interval of one lane and the third of another are unrelated in time, so an index-based sweep places two tasks months apart at the same point in the stereo field and two simultaneous tasks at different ones — erasing exactly the overlap the chart exists to show. There is a test for each half of that (`two lanes starting together sound together`, `a column index would have placed them apart`).

## Braille: a length profile, not a span

#801 suggested a span encoding closer to `BoxBrailleEncoder`. I built the length profile instead, and `docs/BRAILLE.md` now carries the argument:

A 40-cell display over a year gives each cell about nine days. A fortnight's task and a three-week one become the same single raised cell; two tasks a week apart begin in the same cell; a milestone shorter than nine days either vanishes or is rounded up to look like a week of work. Every question a schedule is read for is then answered *wrongly* rather than coarsely.

The profile is the opposite trade — coarse about *when*, exact about *how long* — and where an interval sits is already carried by the pan, which is continuous and has the resolution to place it. The two modalities divide the chart between them instead of both approximating the same half.

A lane with nothing booked renders as an **empty row** rather than being skipped, so the reader's line count keeps matching the chart's lane count.

## Data shape

```jsonc
{
  "type": "gantt",
  "data": {
    "unit": "days",
    "points": [
      [{ "x": "Design", "start": 0,  "end": 30,  "label": "Wireframes" }],
      [{ "x": "Build",  "start": 30, "end": 100, "label": "Implementation" }],
      []
    ]
  }
}
```

Nested rather than flat, and an object rather than a bare array, both for reasons the grammar records: a flat list grouped by `x` cannot express a lane with **no** intervals, which is a real statement about a schedule; and `unit` belongs to the chart rather than to any row, so repeating it per point would let a producer emit rows that disagree about what their numbers measure.

`unit` is what makes the length announceable. "40" does not say whether a task runs 40 days or 40 weeks, and the axis format that renders the ends as dates says nothing about the difference between two of them — so the unit travels with the value rather than through the formatter. Omitted, the length is announced bare rather than with a guessed unit.

## One service change

`TextService` now renders `state.crossRange` as a span, exactly the way `state.range` has always replaced the main value for a histogram bin. The field already existed with that documented meaning and was rendered only on the grid path.

Kept as two numbers in the state rather than a pre-joined string specifically so both ends go through the **cross-axis formatter** — which is what turns a schedule's numbers into dates. `test/service/textCrossRange.test.ts` covers that, plus a case asserting that a state without `crossRange` announces exactly as before. Nothing else in the codebase emits one.

## Registration

All six sites plus the docs: `grammar.ts` (`TraceType.GANTT`, `GanttPoint`, `GanttData`, the `data` union), `src/model/gantt.ts`, `factory.ts`, `CHART_TYPE_LABEL`, `IS_ORIENTED` (**true** — a timeline drawn down the page is the ordinary alternative, not an exotic one), the braille encoder map, and `docs/BRAILLE.md`.

## One bug worth naming

The first e2e run caught the orientation handling missing: with the axes taken verbatim, a horizontal chart announced **"Day is Design"** — the lane under the schedule's label and the schedule under the lane's. Both wrong, neither raising anything. Fixed the way `DumbbellTrace` does it, and `which axis is which follows the orientation` now pins both directions in the unit suite, where it should have been caught first.

## Verification

- `npm test` — **2174 + 73 passed**, 168 suites, none failing
- `npx playwright test --project=chromium` — **408 passed** (5 of them new)

---

---


jooyoungseo:
Both real, and the crash is exactly as you traced it. Fixed in `008858b`.

## 1 — the empty-lane crash. Reproduced, and worse than a bad announcement

You flagged this as manual tracing rather than an observed stack trace, so I ran it first. Your reading of the path was right to the line:

```
TypeError: Cannot read properties of undefined (reading 'label')

  > 227 |         value: point.label ? `${point.x}, ${point.label}` : point.x,
```

And your point about the blast radius is the part that matters: `audio`, `braille` and `text` are computed in one object literal in `AbstractTrace.state`, so this did not degrade the announcement — it took down the entire state push. In my own shipped example, on an ordinary ArrowUp.

I took your suggested shape (guard in the trace, not in `MovableGrid`) for the reason you give: a genuinely empty row is unique to this data model, and widening the shared bounds logic would change navigation for every trace to fix one.

Two things about the guard are decisions rather than mechanics:

**It stays a populated state.** Returning `outOfBoundsState` would have been less code and would announce **"No more data to display"** — telling a reader they had left the chart when they have not. The lane is a real row of the schedule; its emptiness is the data. It now reports `Intervals is 0`.

**It pitches `NaN`.** That is how this codebase already says "the point exists to navigate to, it just has no value" — `AudioService` has a documented branch for it, deliberately not scoped to bar traces. Reaching for a number would put a length on a lane that holds none.

### The gap underneath

Fixing the crash surfaced something the crash was hiding: **a lane's name lives on its intervals**, so an empty lane has nowhere to carry one. The first passing run announced my example's empty lane as `Phase is Lane 4`.

So `GanttData.lanes` — optional, one name per lane. A lane that names itself always wins over it, so a producer supplying both cannot make them disagree about a populated lane; a lane that is empty *and* undeclared is named positionally, because a row a reader can reach has to be identifiable even when the chart said nothing about it. `examples/gantt.html` now declares its lane names, which is what makes `Launch` announce as `Launch`.

### Tests

You were right that nothing existing covered it — the description test never navigated, and the e2e braille test read the payload rather than pressing a key. Six new unit cases and one e2e that **arrows onto the lane with the keyboard**, since that is the only path that exercises this. Five of the six fail on the parent commit.

## 2 — terse mode dropped the end. Also real

Confirmed: `formatTerseTraceText` fell through to `cross.value`, which for a gantt is `Number(point.start)`. So the shorter announcement silently named one edge of the bar as though it were the bar — the precise failure the span exists to prevent, in the mode I did not patch.

Same branch as verbose, joined with a dash rather than "through", matching how terse drops every other connective word: `Design, Wireframes, 5-45, 40 days`. Covered by a terse-mode case alongside the verbose ones in `test/service/textCrossRange.test.ts`.

## 3 — the `audio`/`text` inconsistency

Fixed as part of the above, and worth noting the asymmetry was itself the bug: `audio` guarded with `?.[this.col]` and so survived, `text` did not and threw. Both now take the same early return, so there is one answer to "what does this trace do off the end of a lane" rather than two.

## Verification

- `npm test` — **2181 + 73 passed** (up from 2174), 168 suites, none failing
- `npx playwright test --project=chromium gantt` — **6 passed**, including the new keyboard walk onto `Launch`

The earlier `Lint Commit Messages` failure was `self-signed certificate` from `actions/setup-node`, which skipped the five jobs downstream of it; re-run and all 14 checks passed on the parent commit.

---

PR #835, opened 2026-08-11
feat(parallel): pitch every axis against itself

### LINKED ISSUE #794: feat: support parallel coordinates plots
## Description

Parallel coordinates draw one polyline per observation across several vertical axes, one axis per variable. It is a primary tool for multivariate exploration and there is no way to author one for MAIDR.

## Why this is a cheap win

The data is already the shape MAIDR navigates. A multi-line layer is `data: Point[][]` — an array of series, each an array of points. Parallel coordinates is exactly that, with the axis name playing the role of `x` and the observation playing the role of the series:

```
[
  [ {x: 'mpg', y: 0.71, z: 'car A'}, {x: 'hp', y: 0.22, z: 'car A'}, ... ],
  [ {x: 'mpg', y: 0.34, z: 'car B'}, {x: 'hp', y: 0.88, z: 'car B'}, ... ],
]
```

Navigation, braille (`LineBrailleEncoder`) and the row/column model all transfer from the existing multi-line handling. The genuinely new work is small and specific:

- each axis has **its own scale**, so the trace has to hold per-column min/max rather than one pair for the layer, and announce the raw value while sonifying the normalized one;
- axis order is meaningful and worth announcing on entry.

## Why it is worth doing

There is currently no accessible path to multivariate structure at all. Crossing lines between two axes signal negative correlation — a pattern that is visible instantly and, today, completely inaudible.

## Difficulty

Low-to-medium. High value per unit of work.

jooyoungseo:
Closes #794.

## The cheap part, and the part that is the chart

#794 is right that the data is already the shape MAIDR navigates: an observation per row, an axis per column, which is a multi-line layer. Navigation, text, highlighting and the row/column model all transfer by extending `LineTrace`.

The part that does not transfer is the one that decides whether the chart is readable at all. **The columns are not one scale.** A line chart's columns are samples of one quantity; here they are different quantities entirely — miles per gallon beside horsepower beside kerb weight — and nothing about them is comparable except each value's position within its own axis.

`LineTrace` scales against the **current row's** extent. For a series sampling one quantity that is right. For a row that spans every variable, that extent is the range of a horsepower figure and a fuel-economy figure taken together, which is not a range of anything.

Sonified that way a reader hears the *units*, not the data:

| | mpg | hp | weight |
|---|---|---|---|
| Honda Civic | 33 | 65 | 1800 |
| Muscle Coupe | 15 | 230 | 3200 |

Against one 15-to-3200 range, every economy figure sits at the floor of the register and every weight at the ceiling — so the car with the best economy and the worst power sounds indistinguishable from the car with the worst of both.

Scaled per axis, the pitch says where a value sits **on its own axis**. Arrow along the Civic and it is high, high, low; along the Coupe and it is low, low, high. Those two swapping between `mpg` and `hp` is negative correlation, which a sighted reader sees at a glance and which #794 correctly calls "completely inaudible" today. `crossing lines are audible as a swap in pitch` is the test for it.

## Braille normalizes before the encoder sees it

Same reasoning, one layer down. `LineBrailleEncoder` scales each row against that row's own extent — so raw values here would compare a car's economy against its own kerb weight, and the cell heights would encode which variable happens to use bigger numbers.

So the trace normalizes first, and each cell is its value's position on its own axis:

```
⠉⠉⣀    high economy, high power, low weight
⣀⣀⠉    low economy, low power, high weight
```

Two rows whose heights swap between adjacent cells are the same crossing, available by touch. `docs/BRAILLE.md` carries the argument.

## Two smaller things

**Axis order is announced.** #794 asks for it, and it is a real property of the chart rather than of the data: which variables sit next to each other decides which crossings are visible at all, and that is the author's choice. The description reports `Axes, in order` plus each axis's own range.

**The layer-wide Min and Max are dropped.** Inherited from `LineTrace`, where one range *is* the chart. Here they would be the smallest and largest numbers anywhere across axes measuring different things — `15` and `3200`, a pair with no referent — and the description dialog is precisely where a reader goes to learn what the axes are.

**An axis whose values are all equal** places every value at the midpoint rather than at an end. 0 and 1 both claim an extreme the data does not have.

## Registration

All six sites plus the docs: `grammar.ts` (`TraceType.PARALLEL`, reusing `LinePoint[][]` — no new point shape needed, exactly as #794 predicted), `src/model/parallel.ts`, `factory.ts`, `CHART_TYPE_LABEL`, `IS_ORIENTED` (**false** — every column is its own axis, so there is no main and cross to swap), and the braille encoder map. `describe.ts` also gets the type in its multi-series branch, with `Observation` as the noun, so the position announcement keeps naming which observation is being traced instead of falling through to "column 3 of 4, row 2 of 3".

## Verification

- `npm test` — **2168 + 73 passed**, 167 suites, none failing
- `npx playwright test --project=chromium parallel` — **5 passed**

---

---

### DIFF
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -436,6 +436,42 @@ same half of it.
 
 Gantt charts use one line per lane on a multiline display, so several lanes'
 workloads can be compared with one sweep rather than by toggling between rows.
+## Parallel coordinates
+
+One row per observation, one cell per axis, each row scaled from 0 to 1.
+
+The scaling is the whole point, and it happens **before** the encoder sees the
+data. Every other multi-row trace hands the encoder raw values and lets it
+scale each row against that row's own extent, which is right when a row is one
+quantity sampled repeatedly. A parallel coordinates row is not: it holds one
+observation across every variable, so scaling it row-wise would compare a car's
+fuel economy against its own kerb weight, and the cell heights would encode
+which variable happens to use bigger numbers rather than anything about the
+car.
+
+So the trace normalizes first. Each cell is its value's position **on its own
+axis**, which is exactly what the pitch plays and what the chart draws. A row
+of the display is then that observation's profile across the variables, and two
+rows can be compared cell by cell:
+
+```
+⠉⠉⣀    high economy, high power, low weight
+⣀⣀⠉    low economy, low power, high weight
+```
+
+Two rows whose heights swap between adjacent cells are the crossing lines that
+mean negative correlation — the pattern the chart is drawn for, available by
+touch as well as by ear.
+
+An axis whose observations all share one value has no spread to place anything
+within, so every cell on it sits at the midpoint. Both extremes would claim a
+rank the data does not support.
+
+### Multiline Displays
+
+Parallel coordinates use one line per observation on a multiline display, so
+several observations' profiles can be compared with one sweep rather than by
+toggling between rows.
 
 ## Multiline Braille Display Support
 
--- a/e2e_tests/specs/parallel.spec.ts
+++ b/e2e_tests/specs/parallel.spec.ts
@@ -1,0 +1,135 @@
+import type { Maidr } from '../../src/type/grammar';
+import { expect, test } from '@playwright/test';
+import { BasePage } from '../page-objects/base-page';
+import { TestConstants } from '../utils/constants';
+import { extractMaidrData } from '../utils/maidr-data';
+import { normalizeText } from '../utils/text';
+
+/**
+ * Every braille cell MAIDR can emit lives in the Unicode braille block
+ * (U+2800 to U+28FF), so a display carrying anything else is not braille
+ * output.
+ */
+const BRAILLE_CELL = /^[\u2800-\u28FF]+$/;
+
+/** The example's own page, driven through the shared base helpers. */
+class ParallelPlotPage extends BasePage {
+  protected override readonly selectors = {
+    notification: `#${TestConstants.MAIDR_NOTIFICATION_CONTAINER} ${TestConstants.PARAGRAPH}`,
+    svg: `svg`,
+    braille: `textarea[id^="${TestConstants.BRAILLE_TEXTAREA}"]`,
+    helpModal: TestConstants.MAIDR_HELP_MODAL,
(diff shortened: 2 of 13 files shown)


jooyoungseo:
Both fixed in `85b35ee`, and the first one is a better catch than "cosmetic" gives it credit for.

## 1 — two words for one referent

I ran it rather than reading it, since you flagged that you could not:

```
text.z = {"label":"Group","value":"Honda Civic"}
```

So the sentence really is `Observation 1 of 4, Group is Honda Civic` — and you are right that it stands out *because* of the care taken elsewhere. It is the same defect as the one `line.ts:174` describes for `seriesLabels`, one field along: a label announced literally, inherited from a chart type this is not.

Fixed the way you suggest, and by the pattern already there rather than a new one: `groupFallbackLabel` is a `protected` getter that `groupLabel` reads, so `ParallelTrace` overrides it to `'Observation'` exactly as it overrides `seriesLabels`. The `private` getter stays private — only the fallback needed to be reachable.

Worth being precise about one thing, because it is the part that could have been got wrong: **a layer that names its own z axis still wins.** `named(this.layer.axes?.z?.label, this.groupFallbackLabel)` is unchanged in that order, so a producer who said what their series are called has said it. There is a test for that alongside the one for the fallback — overriding into the wrong slot would have passed the first and failed the second.

`AnnounceZCommand` follows from the same getter, so `l z` reads "Z label is Observation" now too.

## 2 — ragged observations

Fair, and the sharper version of the point is that the code says in a comment that it handles this ("a chart whose rows run to different lengths still names every column a cursor can reach") while nothing held it to that. Three cases added, on a fixture whose third car was never measured for weight:

- an axis is scaled by **the observations that reached it** — weight comes out `1800 to 3200` from the two rows that have one, not from three
- the axis names come from the **longest** row rather than the first, asserted on a fixture with the short row reordered to the front, where reading the first row would lose an axis
- every axis a cursor can reach still gets a range in the description

## On the free intersection navigation

Thanks for chasing that down — I had not checked whether `numericX()` resolved the axis names to ordinals, so I would not have claimed it. Good to have it confirmed rather than assumed, since crossings are the thing the chart is read for.

## Verification

- `npm test` — **2173 + 73 passed**, 167 suites
- `npx playwright test --project=chromium parallel` — **5 passed**
- CI was already green on the parent commit: **14/14**

---

PR #836, opened 2026-08-11
feat(bump): sound first place as first place

### LINKED ISSUE #795: feat: support slope graphs and bump charts
## Description

A slope graph connects each category's value at two time points; a bump chart generalises it to many time points and plots **rank** rather than value. Neither is expressible today.

## Proposed mapping

A multi-line layer where the y axis is a rank. The model is already there; what is missing is that a rank axis is inverted (1 is best) and that the interesting event is not a value but a **crossing** — one series overtaking another.

## Suggested design

- Announce rank with its direction made explicit ("rank 3 of 12, up 2 from last period") rather than as a bare number, since a falling pitch meaning an improving rank would be actively misleading.
- Add a rotor filter unit for rank changes, so a user can jump between overtakes instead of scanning every period. `RotorFilterUnit` already supports trace-defined units — see how candlestick registers its bullish/bearish filters.

## Difficulty

Low.

jooyoungseo:
Closes #795.

## A rank is not a magnitude

Structurally a bump chart is a multi-line layer, so navigation, braille and highlighting transfer by extending `LineTrace`. The y axis is what makes it a type of its own: **rank 1 is the best position and the smallest number.**

Sonified as a magnitude, the leader is the lowest note on the chart. A team climbing the table is heard *falling*, on every move, with nothing anywhere in the output to say the reading is upside down — which is exactly the "actively misleading" case #795 names. The bounds go to the audio service the other way round (`min: worstRank, max: bestRank`), so `MathUtil.interpolate` maps first place to the top of the register.

Scaled across the whole table rather than per row, too. Every competitor is ranked in the same table, so their positions are comparable — and `LineTrace`'s per-row bounds would make each competitor's own best round the highest note whether that was first place or eleventh.

## The overtake travels with the position

What a reader wants from a bump chart is who overtook whom, and that is not recoverable from hearing a sequence of ranks one at a time: it means holding the previous period's number while navigating to the next and subtracting by ear, for every competitor.

So each point announces the move alongside the rank, named by direction rather than signed — the dumbbell's argument, which applies with extra force on an axis where the arithmetic already runs backwards:

| situation | announced |
|---|---|
| 3rd → 1st | `Places gained is 2` |
| 1st → 4th | `Places lost is 3` |
| held 2nd | `Change is 0` |
| first period | *nothing* |

The last two rows are a real distinction. Silence in the first period is honest — there is nothing to compare against — so "held its place" has to say something rather than also be silent, or the two become indistinguishable.

## Rotor units for the overtakes

#795 asks for this and it is the right call: overtakes are sparse. A twelve-team table over a season is mostly periods where nothing moved, so a reader scanning every column spends the navigation on the uninteresting part. `Rank gained` and `Rank lost` follow the candlestick's bullish/bearish precedent exactly, filtering along the period axis within the current competitor.

They are **withheld** on a chart where no rank ever moved. Cycling onto a mode whose only possible answer is "none found" is worse than not offering it.

## The description reports both directions

`Climbed furthest` and `Fell furthest` are separate entries, not the larger of the two. Ranks are a permutation, so somebody's gain is somebody else's loss and the two are commonly *the same size* — my own first draft took the larger and the test caught it naming only the fall. Reporting the winner of that tie would drop one of two distinct findings on precisely the charts where both are most clearly present. Each is withheld when nobody moved that way, rather than naming whoever moved least.

The inherited `Min value` / `Max value` are dropped: on a rank axis they say "somebody came first and somebody came last", true of every bump chart ever drawn.

## Braille stays literal, and says so

The encoder is given values, not a direction, so the cells rise with the rank *number* — a row climbing the display is a competitor sliding down the table, the opposite of what the pitch does. Inverting it would mean emitting `n + 1 - rank` in place of the rank, putting numbers in the display that the chart does not contain and that every other reading of the same state would contradict.

So the display stays literal and `docs/BRAILLE.md` states the disagreement outright, since a reader cannot tell from the dots alone. What braille adds is the *shape*: a flat row held its place all season, and two rows that cross swapped positions — legible by touch whichever way the dots run.

## Scope: slope graphs

#795 covers "slope graphs and bump charts". A slope graph of **values** is two points per series on a value axis — a line layer with two samples, which `LineTrace` already reads correctly, with nothing new needed. A slope graph of **ranks** is a two-period bump chart and is covered here. What makes this a type of its own is the rank, not the number of periods, so there is one type rather than two.

## Registration

All six sites plus the docs, plus `describe.ts`'s multi-series branch with `Competitor` as the noun. `groupNameAt` on `LineTrace` becomes `protected` — a subclass naming a series in its own description has to name it the way every other announcement does, and reaching for `points[row][0].z` directly would skip the fallback and report `undefined` for an unnamed series.

## Verification

- `npm test` — **2172 + 73 passed**, 167 suites, none failing
- `npx playwright test --project=chromium bump` — **5 passed**

---

---

### DIFF
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -472,6 +472,33 @@ rank the data does not support.
 Parallel coordinates use one line per observation on a multiline display, so
 several observations' profiles can be compared with one sweep rather than by
 toggling between rows.
+## Bump chart
+
+One row per competitor, one cell per period, each row scaled against its own
+range — the line encoding unchanged.
+
+**The cells rise with the rank number, not with the position.** A row that
+climbs the display is a competitor sliding *down* the table, because rank 1 is
+the best place and the smallest number. That is the opposite of what the pitch
+does, which inverts so first place is the highest note.
+
+The two disagree on purpose, and it is worth knowing which is which. Audio is
+given a direction to invert — the trace hands the service its bounds the other
+way round, and nothing else about the tone changes. Braille is given values,
+and the encoder maps a value to a dot height; inverting those would mean
+emitting `n + 1 - rank` in place of the rank, so the display would carry
+numbers the chart does not contain and every other reading of the same state
+would disagree with it.
+
+So the display stays literal: a low row is a good run, and the announcement and
+the pitch both say so in words and in tone. What braille adds is the *shape* —
+a row that stays flat held its place all season, and two rows that cross swapped
+positions, which is legible by touch whichever way the dots run.
+
+### Multiline Displays
+
+Bump charts use one line per competitor on a multiline display, so a whole
+table's season can be felt at once rather than by toggling between rows.
 
 ## Multiline Braille Display Support
 
--- a/e2e_tests/specs/bump.spec.ts
+++ b/e2e_tests/specs/bump.spec.ts
@@ -1,0 +1,134 @@
+import type { Maidr } from '../../src/type/grammar';
+import { expect, test } from '@playwright/test';
+import { BasePage } from '../page-objects/base-page';
+import { TestConstants } from '../utils/constants';
+import { extractMaidrData } from '../utils/maidr-data';
+import { normalizeText } from '../utils/text';
+
+/**
+ * Every braille cell MAIDR can emit lives in the Unicode braille block
+ * (U+2800 to U+28FF), so a display carrying anything else is not braille
+ * output.
+ */
+const BRAILLE_CELL = /^[\u2800-\u28FF]+$/;
+
+/** The example's own page, driven through the shared base helpers. */
+class BumpPlotPage extends BasePage {
+  protected override readonly selectors = {
+    notification: `#${TestConstants.MAIDR_NOTIFICATION_CONTAINER} ${TestConstants.PARAGRAPH}`,
+    svg: `svg`,
+    braille: `textarea[id^="${TestConstants.BRAILLE_TEXTAREA}"]`,
+    helpModal: TestConstants.MAIDR_HELP_MODAL,
+    helpModalTitle: TestConstants.MAIDR_HELP_MODAL_TITLE,
+    helpModalClose: TestConstants.HELP_MENU_CLOSE_BUTTON,
+    settingsModal: TestConstants.MAIDR_SETTINGS_MODAL,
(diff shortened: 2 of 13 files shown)


jooyoungseo:
All three taken in `3f895b5`. Each of them names a pattern the codebase already set and I had diverged from, which makes them worth doing regardless of severity.

## 1 — `MathUtil.minFrom2D` / `maxFrom2D`

Done. It drops the `.flat()`, matches what `LineTrace` does per row, and removes the call-stack ceiling a spread carries on a large table — a season of a big league is not a small array.

## 2 — precompute the rotor units

Done, in the constructor beside `moves`, with the comment pointing at `Candlestick`'s reasoning since it is the same reasoning: the ranks are fixed at construction and the rotor service asks twice per keystroke.

## 3 — the ragged assumption. This one turned out to be the real one

You marked it "degrades rather than crashes", which is exactly what makes it worth fixing — and having built a fixture that discriminates, it degrades further than the finding suggests. Both halves are wrong, not just one:

| | off row 0's length | correct |
|---|---|---|
| `Led at the end` | Birch — the leader at **R2** | Cedar |
| `Climbed furthest` | Birch, 1 — measured to R2, tied and resolved by index | Cedar, 2 |

So a competitor who joined late or dropped out does not merely lose precision: the chart names the **wrong finisher** and the **wrong biggest climber**, and neither announcement carries any sign that a shorter row was involved.

Two changes rather than the one you suggested. The table's period count now comes from the longest row (`Math.max` over the lengths, as you proposed), *and* `extremeMover` measures each competitor against **its own** last round rather than the table's — otherwise fixing the first would leave a short row comparing its start against a column it does not have.

### A note on the test

My first attempt at the ragged fixture **passed without the fix**, which makes it worthless as a test. Row 0 being short was not enough: the leader at R2 and the leader at R4 happened to be the same competitor, so both readings agreed. The fixture now has three competitors arranged so the two answers genuinely differ, and **2 of its 3 cases fail on the parent commit**. Worth saying plainly, since a green ragged test that proves nothing is worse than no ragged test — it reads as coverage.

## Verification

- `npm test` — **2177 + 73 passed**, 167 suites
- `npx playwright test --project=chromium bump` — **5 passed**
- CI was 14/14 on the parent commit

---

PR #838, opened 2026-08-12
feat(diverging): read a population pyramid by size and side

### LINKED ISSUE #793: feat: support population pyramid (diverging bar) charts
## Description

A population pyramid puts two series back to back across a shared category axis, one growing left and one growing right. Diverging bars more generally (e.g. Likert-scale responses split around a neutral midpoint) have the same shape.

## Proposed mapping

A `dodged_bar` with a signed value axis and a mirrored baseline. `SegmentedTrace` already handles "one category, several series" navigation, so the work is in:

- treating the sign as *direction*, not as a negative magnitude, so the audio does not read the left-hand series as below zero;
- announcing the shared category once and each side's magnitude separately;
- reporting the balance between the two sides, which is the reason the chart is drawn this way.

## Difficulty

Low. Mostly a presentation and announcement layer over `SegmentedTrace`.

jooyoungseo:
Closes #793. Rebased onto current `main`; #837 is merged, so the "merge #837 first" note that was here is gone.

## The sign is a direction, not a magnitude

#793 names this exactly. A producer emits the values as the chart draws them, so the left-hand series is negative. Pitched that way, the biggest bar on the left becomes the **lowest note on the chart** — a cohort of 1,200,000 men sounds smaller than a cohort of 310,000 women, and the pyramid is heard as sagging on one side.

The pitch takes the magnitude and scales from **zero** rather than from the chart's minimum, which on a diverging chart is the largest left-hand bar. Left as-is, the left series would occupy the bottom of the register *in reverse order* and the right series the top, and neither side's bars would be comparable with the other's by ear. With the fix, equal cohorts on opposite sides are indistinguishable by pitch, which is the property the chart is drawn to be read for.

The announcement names the size and the side — `People is 1200, Sex is Men` — which is how a sighted reader takes it in: length from the bar, side from which way it points.

## The balance is the finding, and it is already there

Because the values arrive signed, the summary row `SegmentedTrace` already builds — the sum down a category — **is** the comparison the chart is drawn to make: `(-left) + right` is what one side has over the other.

So it is renamed rather than recomputed. `Sum is -40000` invites a reader to hear a total that came out negative; on this row a minus sign is not a smaller number, it is the other side winning. It announces `Balance is Men ahead`, or `level` at zero rather than naming a winner.

**Which side is ahead is resolved by reading which way the values grow, not by row index.** Nothing obliges a producer to declare the left-hand side first, and a fixed index names the wrong winner for a chart that declares the right-hand side first — a fluent, confident sentence with the two sides swapped. That is the worst failure available here, because the sign is the one clue this trace deliberately removes from the announcement.

A chart that is **not two-sided** keeps the segmented bar's plain `Sum`. "X ahead" is a comparison between exactly two sides, and nothing in the grammar holds a diverging layer to two.

## The highlight follows the reader

`SegmentedTrace` defaults a category's DOM elements to reverse order, because a stacked bar's producers draw its segments bottom-up. A diverging chart is not stacked — the two sides sit either side of a baseline — so a producer emits the left bar then the right one, in the order the data declares them.

The default is now a `groupsRunForward` hook that `DivergingTrace` overrides, rather than `domMapping` patched into the example. The evidence is that the first example authored for this type got it wrong by following the natural order: an author who has to know the stacked bar's convention in order to draw a pyramid will not know it. `domMapping.groupDirection` still overrides in both directions.

**The e2e assertion is the one that matters** — it fails without the fix while the other four pyramid specs pass either way, which is the shape of the defect: audio, text and braille never go through the element mapping, so every announcement stayed correct while the highlight lit the opposite bar.

## An unnamed side is named by the way it grows

`left` / `right` as a fallback, because the sign is exactly what this trace removes from the announcement — leaving an unnamed series with no way to say which half of the chart it is on would be worse than the signed reading it replaces.

## Two things the e2e caught that unit tests did not

**The example had `NaN` for every value.** A horizontal bar layer carries its value on `x` and its category on `y` — the pair `toBarValue` reads — and I authored the example with the vertical convention. Braille rendered blank rather than erroring.

**My first fix for that was wrong in the opposite direction.** Seeing `Age band is -1200`, I concluded the parent puts the value in `main` for a horizontal layer. It does not: the parent swaps *which point field feeds `cross`*, not which half of the announcement carries the length. Reverted, with a horizontal unit test so the ordinary way to draw a pyramid is covered where it should have been.

## Braille stays literal, and says so

The cells encode the **signed** value, so a left-hand row reads low across its whole length — the opposite of what the pitch does. Flipping the sign before encoding would put numbers in the display the chart does not contain. `docs/BRAILLE.md` states the disagreement, since a reader cannot tell from the dots; what braille adds is the silhouette.

## Registration

All six sites plus the docs. `IS_ORIENTED` is **true**, for the same reason the stacked bar it extends is.

## Verification

On the rebase onto current `main`:

- `npm test` — **2296 + 73 passed**, 167 suites
- `npx playwright test --project=chromium pyramid bump funnel hexbin` — **20 passed**

---

### DIFF
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -549,6 +549,31 @@ positions, which is legible by touch whichever way the dots run.
 Bump charts use one line per competitor on a multiline display, so a whole
 table's season can be felt at once rather than by toggling between rows.
 
+## Diverging bar and population pyramid
+
+One row per side plus the balance row, one cell per category -- the bar
+encoding unchanged.
+
+**The cells encode the signed value, so a left-hand row reads low across its
+whole length.** That is the opposite of what the pitch does, which takes the
+magnitude so a bar of the same size sounds the same on either side.
+
+The two disagree on purpose. Audio is given a direction to invert and nothing
+else about the tone changes. Braille is given values, and the encoder maps a
+value to a dot height; flipping the left side's sign before encoding would put
+numbers in the display that the chart does not contain, and every other reading
+of the same state would then contradict it.
+
+So the display stays literal, and what it adds is the *silhouette*: the two
+sides read as a low band and a high band, and the taper of a pyramid -- wide at
+the young bands, narrow at the old -- is legible as a run of dots rising on one
+row while falling on the other.
+
+### Multiline Displays
+
+Diverging charts use one line per side on a multiline display, plus the
+balance, so the two sides can be compared with one sweep.
+
 ## Multiline Braille Display Support
 
 By leveraging the two-dimensional nature of multiline braille displays, MAIDR can represent multiple lines of a plot simultaneously, allowing users to perceive the distribution of values across all lines at once. This is particularly beneficial for plots with multiple groups or categories, such as grouped boxplots or line plots with multiple lines, and it also applies to scatter plot grid navigation.
--- a/e2e_tests/specs/pyramid.spec.ts
+++ b/e2e_tests/specs/pyramid.spec.ts
@@ -1,0 +1,123 @@
+import type { Maidr } from '../../src/type/grammar';
+import { expect, test } from '@playwright/test';
+import { BasePage } from '../page-objects/base-page';
+import { TestConstants } from '../utils/constants';
+import { extractMaidrData } from '../utils/maidr-data';
+import { normalizeText } from '../utils/text';
+
+/**
+ * Every braille cell MAIDR can emit lives in the Unicode braille block
+ * (U+2800 to U+28FF), so a display carrying anything else is not braille
+ * output.
+ */
+const BRAILLE_CELL = /^[\u2800-\u28FF]+$/;
+
+/** The example's own page, driven through the shared base helpers. */
+class PyramidPlotPage extends BasePage {
+  protected override readonly selectors = {
+    notification: `#${TestConstants.MAIDR_NOTIFICATION_CONTAINER} ${TestConstants.PARAGRAPH}`,
+    svg: `svg`,
+    braille: `textarea[id^="${TestConstants.BRAILLE_TEXTAREA}"]`,
+    helpModal: TestConstants.MAIDR_HELP_MODAL,
+    helpModalTitle: TestConstants.MAIDR_HELP_MODAL_TITLE,
+    helpModalClose: TestConstants.HELP_MENU_CLOSE_BUTTON,
+    settingsModal: TestConstants.MAIDR_SETTINGS_MODAL,
+    chatModal: TestConstants.MAIDR_CHAT_MODAL,
+  };
+
+  /** Navigates to the population pyramid example. */
+  public async navigateToPlot(): Promise<void> {
+    await super.navigateTo('examples/pyramid.html');
+    await super.verifyPlotLoaded(this.selectors.svg);
+  }
+
+  /** Activates MAIDR on the chart. */
+  public override async activateMaidr(): Promise<void> {
+    await super.activateMaidr(this.selectors.svg, 'pyramid-population');
+  }
+
+  /**
+   * Reads the announcement currently on screen.
+   * @returns The announcement text
+   */
+  public override async getInstructionText(): Promise<string> {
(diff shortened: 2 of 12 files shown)


jooyoungseo:
Rebased onto `main` now that #837 has merged, so this is a single-commit PR again rather than a stacked one. The dependency note in the description no longer applies.

Nothing changed in the diverging code itself — the rebase resolved two conflicts, both of them "two features adding a section at the same anchor": the gantt and diverging braille sections in `docs/BRAILLE.md`, and the `GANTT` and `DIVERGING` cases in `factory.ts`. Both sides kept.

Re-verified on the new base:

- `npm test` — **2204 + 73 passed**, 170 suites
- `npx playwright test --project=chromium pyramid gantt` — **10 passed**

---

jooyoungseo:
Confirmed, and taken in `f63712ba`. I reproduced it before touching anything, and the trail led somewhere worse than the finding.

## 1 — the highlight mapping. Confirmed exactly as described

A unit test that fails on the parent commit:

```
Expected: "Men"
Received: "Women"
```

Row 0 (Men) resolved to the Women rect, for the reason you traced: no `domMapping`, so `SegmentedTrace.mapToSvgElements` takes the column-major **reverse** branch, and the first rect of each band goes to the last row.

**I fixed the trace rather than the example**, which is a deliberate departure from your suggestion. The reverse default exists because a *stacked* bar's producers draw its segments bottom-up. A diverging chart is not stacked — the two sides sit either side of a baseline rather than on top of one another — so there is no stacking order to inherit.

The evidence is that I wrote the example wrong. An author who has to know the stacked bar's convention in order to draw a pyramid will not know it, and the failure gives them nothing: every announcement stays correct. Nothing shipped reads `diverging_bar` yet, so this is the last moment the default can change for free.

It is a hook (`groupsRunForward`) on `SegmentedTrace` rather than a duplicated mapping walk, and `domMapping.groupDirection` still overrides it in **both** directions — there is a test for a producer that draws right-hand bars first.

## 2 — the balance named the wrong side. Not in the review, found while confirming it

`sideNameAt(value > 0 ? 1 : 0)` assumed row 0 is the left-hand side. Nothing obliges a producer to declare it first. With Women declared first:

```
data:      [Women 400, Men -900]   →  balance -500, so MEN lead by 500
announced: {"label":"Balance","value":"Women ahead"}
```

Reproduced before changing a line. This is the worst failure available on this trace: the sign is the one clue the announcement deliberately removes, so a reader has **nothing left to check a fluent, confident, reversed sentence against**. The side is now resolved by reading which way its values grow, not by index.

## 3 — the two-row assumption. Your reading was right; the fix is to stop claiming

Rather than a comment, a chart that is not two-sided now keeps the segmented bar's plain `Sum`. "X ahead" is a comparison between exactly two sides; a three-way total has no winner to name, and staying quiet is true whatever the shape.

## The e2e assertion, which is the part that matters

```
✘ should light up the bar it just announced, not the one across the axis
✓ (the other four pyramid specs)
```

Without the fix, **one of five fails** — and it is the new one. That is the shape of the defect you identified: audio, text and braille never go through the element mapping, so four specs and every unit test stayed green while the highlight sat on the opposite bar.

It asserts on geometry, not fill. My first attempt used the fill colours and failed for an instructive reason: the highlight clone is recoloured green, so the colour cannot tell the sides apart. The geometry survives, and on a mirrored baseline where a bar *starts* is precisely which side of the axis it is on — men grow left from x = 184, women right from x = 400.

## Also

- `docs/BRAILLE.md` — blank line restored before the heading.
- `widest` — you were right that the balance can never exceed the larger side (one value negative, one positive). Comment rewritten to say what including it actually buys, rather than justifying it with a case that cannot occur.

## Verification

- `npx playwright test --project=chromium pyramid gantt stacked` — **34 passed** (the stacked-bar suite exercises the default path the hook refactored)

The dependency note about #837 in the description is indeed stale; it merged and this branch is rebased onto it.

---

PR #839, opened 2026-08-12
feat(funnel): give the ear the ratio it cannot compute

### LINKED ISSUE #791: feat: support funnel charts
## Description

A funnel chart shows a population shrinking across ordered stages. Widely used in analytics and conversion reporting; unsupported today.

## Proposed mapping

A `bar` whose ordering is meaningful, plus a derived per-stage conversion rate:

```
{ x: 'Visited', y: 10000 }
{ x: 'Signed up', y: 2400 }   // 24.0% of stage 1, 24.0% of previous
```

The value the reader actually wants is rarely the raw count — it is the drop. So the trace should announce both the absolute value and the stage-over-stage retention, and the audio should track the retention so a sharp drop-off is audible as a sharp fall rather than as a gentle slope on a large absolute scale.

## Scope

- [ ] `TraceType.FUNNEL` + trace class
- [ ] Per-stage conversion derived in the model, not the producer
- [ ] Reuse `BarBrailleEncoder`

## Difficulty

Low.

jooyoungseo:
Closes #791.

## The number a reader wants is not the one the chart plots

#791 puts it exactly: "the value the reader actually wants is rarely the raw count — it is the drop." A drop-off is a **ratio between adjacent bars**, and a ratio is precisely what a listener cannot take from two heights heard one at a time.

So the pitch carries the retention. The case for it is stronger than "nicer": on a raw scale the worst stage in a funnel is routinely **not** the biggest fall.

| step | falls | keeps |
|---|---|---|
| Visited → Signed up | 7,600 — three quarters of the axis | **24%** |
| Viewed cart → Purchased | 2,200 — a fifth of the axis | **4.3%** |

The second is the catastrophe; the first is the ordinary top of a funnel. A reader listening to absolute heights hears them the other way round, and the shipped example is built so they genuinely disagree — `the worst stage is the lowest note, which the counts would not give` asserts both halves, including that the raw falls really do rank the other way, so the test cannot pass by coincidence of fixture.

## The count is not lost

Each stage announces the count, the retention, **and** its share of everyone who entered:

> Stage is Signed up, People is 2400, Retained is 24.0%, Entered is 10000, 24.0% of it

The share is the cumulative view the per-stage ratio does not give, and which a reader would otherwise have to multiply out across every stage they had passed. It uses `TextState.stack`'s existing `share` field, which renders as "N% of it" — the same slot a stacked area uses for its running total.

## The entry stage claims neither, and that took two passes

My first version reported the entry stage as `Stage is first` with `Entered is 10000, 100.0% of it`. The e2e caught two problems in one line:

- **`Stage` labelled two different things** — the x-axis label and the z value — in a single sentence. The inverse of the "two words for one referent" defect flagged on #835.
- **"100.0% of it" was vacuous.** The entry stage *is* the population that entered, so that says one number twice and calls one of them a share. And "100% retained" reports a conversion the chart never claimed: nothing converted into the first stage.

Both are gone. The entry stage announces its name and its count and nothing else, and **its silence is the signal** — every other stage carries a retention, so the one that does not is where the funnel starts. The description names it as `Entry stage` for a reader who arrived by rotor rather than by walking.

## Braille keeps the counts

Deliberately not matching the pitch, and `docs/BRAILLE.md` says why. The counts *are* the funnel's silhouette — wide first cell tapering to a narrow last one, with the steep steps legible as the cells where the dots drop several levels at once. Encoding retention instead would put a near-full cell wherever a stage lost nobody, and the display would stop narrowing: a funnel that reads as a rectangle.

The ear is told how bad each step is; the fingers are shown what the whole thing looks like.

## Empty stages

A stage following an empty one has nothing to be a fraction of. Dividing would hand the audio a non-finite frequency and the announcement an infinity, so it answers `NaN` — how this codebase already says a value is absent. A stage that measured **zero** after a non-empty one is a real reading of zero retention, not a gap, and is pitched at the floor rather than sounded as missing.

## Registration

All six sites plus the docs. `IS_ORIENTED` is **true**, for the same reason the bar chart it extends is — a funnel is drawn top-to-bottom as often as left-to-right.

## Verification

- `npm test` — **2204 + 73 passed**, 170 suites, none failing
- `npx playwright test --project=chromium funnel` — **5 passed**

---

---

### DIFF
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -401,6 +401,33 @@ Encoding a rotation into cell heights would answer neither.
 Radar charts use one line per series on a multiline display, so several models'
 profiles can be compared with one sweep rather than by toggling between rows.
 
+## Funnel
+
+A single row of stage counts, scaled against that row's own range -- the bar
+encoding unchanged.
+
+**The cells carry the counts, not the retention the pitch carries.** The two
+deliberately answer different questions, and it is worth knowing which is
+which.
+
+The pitch takes the **ratio** between adjacent stages, because a ratio is what
+a listener cannot compute: two heights heard one at a time do not give a
+percentage, and on a raw scale the worst stage in a funnel is routinely not
+the biggest fall. Braille takes the **counts**, because a display can show a
+shape all at once and that shape *is* the funnel -- a wide first cell tapering
+to a narrow last one, with the steep steps legible as the cells where the dots
+drop several levels at once.
+
+So the ear is told how bad each step is and the fingers are shown what the
+whole thing looks like. Encoding the retention in the cells instead would put
+a near-full cell wherever a stage lost nobody, and the display would no longer
+narrow -- a funnel that reads as a rectangle.
+
+### Multiline Displays
+
+Funnel charts use a single-line representation. On multiline braille displays
+the funnel appears on the first line and the remaining lines are unused.
+
 ## Gantt, timeline and swimlane
 
 A gantt's braille is the multi-line encoding: **one row per lane**, one cell
--- a/e2e_tests/specs/funnel.spec.ts
+++ b/e2e_tests/specs/funnel.spec.ts
@@ -1,0 +1,137 @@
+import type { Maidr } from '../../src/type/grammar';
+import { expect, test } from '@playwright/test';
+import { BasePage } from '../page-objects/base-page';
+import { TestConstants } from '../utils/constants';
+import { extractMaidrData } from '../utils/maidr-data';
+import { normalizeText } from '../utils/text';
+
+/**
+ * Every braille cell MAIDR can emit lives in the Unicode braille block
+ * (U+2800 to U+28FF), so a display carrying anything else is not braille
+ * output.
+ */
+const BRAILLE_CELL = /^[\u2800-\u28FF]+$/;
+
+/** The example's own page, driven through the shared base helpers. */
+class FunnelPlotPage extends BasePage {
+  protected override readonly selectors = {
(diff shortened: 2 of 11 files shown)


jooyoungseo:
Confirmed and fixed in `0507c134`. Reproduced first, on a horizontal fixture:

```
Entry stage     Expected: "Visited"                    Received: 10000
Steepest drop   Expected: "Purchased, 4.3% retained"   Received: "100, 4.3% retained"
```

Your diagnosis was exact, including the part about why nothing caught it: `text.main.value` comes from the parent's already-orientation-aware getter, so the announcement was always right. The description is the *other* place a stage is named, and it is the one summary a reader consults precisely to find out where the funnel starts and where it breaks — so a count there is not a degraded reading, it is an answer to a different question.

Rather than inlining the branch twice, the resolution is a `stageNameAt(stage)` helper. Two call sites already needed it and any future stat naming a stage would be a third, so the orientation question is asked in one place.

## The coverage gap you named is the real finding

You are right that `test/util/orientation.test.ts` asserting membership in the oriented set is not coverage of *being* oriented — it pins the registration, not the reading. `createLayer` now takes an orientation and there is a `HORIZONTAL_STAGES` fixture built by transposing the existing one, so both arrangements read the same data. **Both new cases fail on the parent commit.**

Worth naming, because this is the second time this session: the same class of bug turned up on #838, where a horizontally-drawn pyramid had `NaN` for every value because I authored the example with the vertical point convention. A trace that declares `IS_ORIENTED = true` and then tests only one orientation has tested half of what it claimed, and the untested half fails silently — a plausible number in the right slot, never an error.

## Verification

- `npm test` — **2229 + 73 passed**, 164 suites
- `npx playwright test --project=chromium funnel` — **5 passed**

---

Issue #814, opened 2026-08-10
feat: chart type coverage roadmap — unsupported trace types


jooyoungseo:
## Description

Tracking issue for chart types MAIDR could not express. `TraceType` in `src/type/grammar.ts` defined **15** authorable types when this was filed; it now defines **48**.

**Everything on this roadmap is closed** — all three tiers, both related defects, and everything found along the way.

## The organising principle

Every trace was a `(rows × cols)` grid with a value per cell — see `AbstractPlot.dimension` in `src/model/abstract.ts`. So a candidate's cost was set by how well it folded into that grid, which is what the three tiers measured. Tier A reused the existing model, tier B added a point shape or an axis concept, tier C needed a navigation model MAIDR did not have.

## Tier A — folds into the existing grid

- [x] #788 — area / stacked area (streamgraph deferred; see the note there)
- [x] #789 — error bar / confidence interval
- [x] #790 — waterfall
- [x] #791 — funnel
- [x] #792 — lollipop / dot / dumbbell
- [x] #793 — population pyramid (diverging bar)
- [x] #794 — parallel coordinates
- [x] #795 — slope graph / bump chart
- [x] #796 — word cloud
- [x] #797 — gauge / bullet
- [x] #798 — boxen (letter-value)
- [x] #799 — hexbin

## Tier B — new point shape or axis concept, same navigation model

- [x] #800 — radar / spider / polar area (also closed #780)
- [x] #801 — gantt / timeline (interval traces)
- [x] #802 — contour / filled contour
- [x] #803 — ridgeline (joy) plots
- [x] #804 — mosaic / marimekko
- [x] #805 — Kaplan-Meier survival curves
- [x] #806 — forest plots
- [x] #807 — volcano / manhattan plots

## Tier C — needs a new navigation model

- [x] #808 — treemap (hierarchical navigation)
- [x] #809 — sunburst / icicle
- [x] #810 — sankey / alluvial / chord (graph navigation)
- [x] #811 — network / node-link diagrams
- [x] #812 — choropleth maps

## What tier C actually needed

Less than expected, and the reason is worth recording. `MovableGraph` — already in the codebase for line-chart intersections — navigates an arbitrary graph over `(row, col)` addresses. Every tier C trace is a graph-building step in the model layer over that one class:

| chart | address | what the arrows do |
|---|---|---|
| treemap / sunburst / icicle | `(depth, index-at-depth)` | up = parent, down = first child, left/right = siblings |
[...]

PR #840, opened 2026-08-12
feat(dot): let a dot plot and a lollipop say which chart they are

### LINKED ISSUE #792: feat: support lollipop, dot and dumbbell charts

jooyoungseo:
Closes #792. The dumbbell half shipped in #829; this is the remainder.

## Deliberately no model

#792 is right that "dot and lollipop are `bar` with a different mark; they need no new model at all beyond a distinct type". Both route to `BarTrace` — one category, one value, navigated identically — the way `POLAR_AREA` and `RADAR` both route to `RadarTrace`.

So this PR is registration and naming, and the tests are written to assert exactly that: the naming changes, and **nothing about the reading does** (`announces the category and the value a bar would` compares `text.main`, `text.cross` and `audio.freq` against a bar layer over the same data).

## Why it is not cosmetic

Before this, an author with a lollipop had two options and both were bad:

- declare `bar` — every reader is told the wrong chart type, on a chart that otherwise reads perfectly
- declare `lollipop` — `TraceFactory` throws `Invalid trace type`, the throw propagates out of `new Figure(...)`, and **nothing on the page renders at all**

The chart type is one of the few places a MAIDR user learns what they are looking at. It is worth being true.

## Both are oriented

`IS_ORIENTED` is `true`, since a bar is — and this is the case that would otherwise bite. A **Cleveland dot plot is conventionally drawn with its categories down the page**, so a type that resolved no orientation would announce its axes the wrong way round in the arrangement the chart is most often drawn in. There is a test asserting the two orientations produce different main labels, so `false` here would fail rather than silently mislabel.

## Braille

Both reuse `BarBrailleEncoder`, because they carry a bar's data. `docs/BRAILLE.md` says why the encoding cannot differ: there is no tactile equivalent of "drawn as a point rather than a rectangle", and inventing one would encode the author's styling where the reader expects the data.

## Registration

All six sites plus the docs, for both types.

## Verification

- `npm test` — **2202 + 73 passed**, 170 suites, none failing
- `npx playwright test --project=chromium lollipop` — **4 passed**, including one asserting the announcement contains "lollipop" and **not** "bar"

---

---


jooyoungseo:
All three taken in `43d38a73`.

## 2 — the missing dot example and spec. This is the one that mattered

You are right, and the "real-world risk is low" framing undersells it. Both types route through the same `BarTrace`, so the *reading* was never at risk — but this PR is **entirely registration**, and registration is precisely the class of thing that passes a unit test and fails on the page. Half the deliverable had no end-to-end proof that a rendered chart says the right word.

`examples/dot.html` and `e2e_tests/specs/dot.spec.ts` close it. The example is drawn with its categories down the page, so the spec can assert more than the type name:

```
✓ should declare the layer as a dot plot
✓ should announce itself as a dot plot rather than a bar
✓ should read the categories down the page, not across it
✓ should render one braille row of values
```

The third one is the interesting one. It asserts `Endpoint` and `Milliseconds` land on the right halves of the announcement, not merely that both appear — which is the `IS_ORIENTED = true` claim actually being exercised in the arrangement a Cleveland dot plot is almost always drawn in, rather than asserted at the registration table. **8 passed** across both specs.

## 1 — the ordering

Taken. It is not only convention drift: `.claude/rules/model.md` tells you to find all six registration sites by grepping a `TraceType`, and the six-site checklist is the thing standing between a new type and a silently missing braille encoder. Ordering that a reader can scan is load-bearing for that workflow. `DOT` now follows `DODGED`, and `LOLLIPOP` sits between `LINE` and `NORMALIZED` in both files.

## 3 — `docs/SCHEMA.md`

Out of this PR's scope as you say, but I fixed it rather than leaving it, because the alternative was worse. The list omitted **fifteen** registered types; adding two to it would have made a stale list look maintained, which is how it got fifteen behind. It is now the complete set generated from `TraceType`, points at `src/type/grammar.ts` as the source of truth, and says why `candlestick_delta` is absent — it is built at runtime from a candlestick and a reference line, so it is not something a page declares.

## On the `isPointInBounds` observation

Worth confirming since you flagged it as possibly lucky: it is correct and it is *not* a special case anyone had to write. The zero-tolerance list names `BAR`/`STACKED`/`HISTOGRAM`/`HEATMAP` explicitly, so a new type gets the default point-sized hit radius by not being on it — which is what circular marks want. Nothing to change; recording it so the next reader does not have to re-derive it.

## Verification

- `npm test` — **2225 + 73 passed**, 164 suites
- `npx playwright test --project=chromium dot lollipop` — **8 passed**

---

jooyoungseo:
You are right, and I was wrong — including in my previous comment on this PR, where I described that test as "the `IS_ORIENTED = true` claim actually being exercised." It was not. Fixed in `91235d42`.

I ran the decisive experiment rather than reasoning about it. With `IS_ORIENTED[TraceType.DOT]` set to `false`:

```
✓ should declare the layer as a dot plot
✓ should announce itself as a dot plot rather than a bar
✓ should read the categories down the page, not across it   ← the one that claimed to test it
✓ should render one braille row of values
4 passed
```

All four pass. Your trace was exactly right: `BarTrace` reads `layer.orientation` off the JSON in the constructor, so the main/cross swap happens whatever `IS_ORIENTED` says, and the one string it does reach is the activation announcement's prefix — which the test navigated past before asserting.

Confirmed what that prefix is, rather than assuming:

```
This is a maidr plot of type: horizontal dot. Use Arrows to navigate…
```

So there is now a test that reads the announcement **before** navigating anywhere, for both types. Each fails when its own `IS_ORIENTED` entry is flipped, and only that test does.

The per-point assertions are kept — endpoint-as-category and milliseconds-as-value is worth pinning — but under a name that says what they check rather than one that overclaims.

## The part worth recording

A test that reads like coverage and is not is worse than no test, and this one had a specific way of being harmful: **the unit suite does pin this behaviour** (3 unit tests fail when the entry is flipped). An e2e test claiming to pin it too is how the unit test comes to look redundant and gets deleted by someone tidying up.

This also sharpens the note I left on the roadmap (#814) an hour ago, where I recommended horizontal fixtures for every `IS_ORIENTED = true` trace. That advice was half right: a horizontal fixture catches the *reading* bugs — which is what bit on #838 and #839 — but it does not touch `IS_ORIENTED` itself, because nothing on the per-point path consults it. Those are two different tests, and I had them conflated.

## Verification

- `npx playwright test --project=chromium dot lollipop` — **10 passed**

---

PR #841, opened 2026-08-12
feat(boxen): read a quantile ladder however deep it goes

### LINKED ISSUE #798: feat: support boxen (letter-value) plots
## Description

A boxen / letter-value plot replaces the box plot's fixed five-number summary with a nested sequence of quantiles, so that the tails of a large sample stay legible. `BoxTrace` cannot express it because its point shape is fixed at min / Q1 / median / Q3 / max plus outliers.

## Proposed mapping

Generalise the box point to a variable-depth quantile ladder:

```
{ z: 'setosa', median: 5.0, levels: [{p: 0.25, lo: 4.8, hi: 5.2}, {p: 0.125, lo: 4.6, hi: 5.4}, ...], outliers: [...] }
```

Navigation gains an inward/outward move across levels, which is a natural fit for the rotor rather than for the arrow keys.

## Related

- xability/py-maidr#253 tracks the seaborn (`sns.boxenplot`) side; this issue is the core trace it depends on.

## Difficulty

Low-to-medium — an extension of an existing trace rather than a new model.

jooyoungseo:
Closes #798. Unblocks xability/py-maidr#253 (`sns.boxenplot`), which #798 names as depending on this.

## Depth is the point

A box plot's five-number summary **is** a letter-value ladder with exactly one rung. That fixed depth is why `BoxTrace` cannot express this: the whole purpose of a boxen is that a large sample gets *more* rungs, so its tails stay legible instead of collapsing into one whisker and a scatter of dots.

So `BoxenPoint` carries a variable-depth `levels: LetterValueLevel[]`, where `p` is the **tail probability** — which is how letter-value plots are defined and how the libraries that draw them report it.

## Value order, so the pitch means something

The ladder is walked deepest-lower → median → deepest-upper. That is the box plot's own arrangement generalised, and it is what makes left and right move **monotonically** through the distribution: the pitch rises without exception, so the *rate* it rises at is the signal. Cells that climb gently through the middle and jump at the ends are a heavy tail — the finding a boxen is drawn to make, and the one a box plot's single whisker flattens into a number.

Rungs are **sorted outward** rather than trusted in arrival order. Nothing in the payload says which way round a producer emitted them, and an inward-first producer would otherwise be navigated backwards. There is a test with a deliberately reversed fixture.

## Percentiles, not rung numbers

Each position announces `12.5th percentile`, not `level 3` — one is a number a reader can place, the other is one they have to count back from.

The precision is chosen rather than defaulted: `toPrecision(3)` with trailing zeros trimmed, so common rungs read `25th percentile` rather than `25.0th` while a deep one reads `3.13th percentile` rather than rounding to `0th`. **Rounding to whole percentiles would name the two deepest rungs of a large sample identically** — merging exactly the part of the distribution this chart exists to separate. Both halves are tested.

## Braille is the ladder, not the box glyph

Deliberately **not** `BoxBrailleEncoder`. That one renders five named sections at proportional widths and has nowhere to put a sixth rung: it would truncate a deep ladder or pad a shallow one into looking deeper than it is, and depth is precisely the fact a boxen carries that a box plot cannot.

`LineBrailleEncoder` renders the ladder as the rising profile it is. Every row is scaled against the **chart's** range rather than its own, so two distributions can be compared cell by cell — a row sitting entirely higher is a distribution shifted upward, not merely one with a different spread. `docs/BRAILLE.md` carries the argument and the two example runs.

## Smaller decisions

- **The description reports ladder depth** (`Quantile levels: 2 to 4`, or a single number when uniform). A library adds rungs as it gains confidence in the tail, so depth is a fact about the *sample*.
- **Outliers are counted.** A boxen has far fewer than a box plot because the deep rungs absorb what a whisker would have thrown out, so a small count is itself informative.
- **Highlighting is one element per distribution**, repeated across its rungs. Chart libraries draw a boxen as one nested stack per category and do not emit a per-quantile element to pair with; every rung highlighting its own distribution is honest about what was drawn, rather than inventing elements that were not.
- **`supportsExtrema` is false.** "The largest value" on a quantile ladder is always the deepest upper rung, by construction — an extrema menu offering it would navigate to a foregone conclusion.

## One process note

The factory registration silently did nothing on the first pass: my edit anchored on a `case` that lives on another branch, and the string replace matched nothing without complaining. The unit suite caught it (`Invalid trace type: boxen`), and I now assert every anchor before replacing. All six sites verified present afterwards.

## Verification

- `npm test` — **2206 + 73 passed**, 170 suites, none failing
- `npx playwright test --project=chromium boxen` — **5 passed**

---

---


jooyoungseo:
Both taken in `faafd273`. You were right that neither blocks, and right about why — but in both cases the analysis is the argument for fixing rather than against.

## The selector guard

Your trace is correct and I confirmed it: `isUsableSelector` does `typeof query === 'string'`, so a `string[][]` entry is rejected before `querySelectorAll` ever sees it, the list comes back empty, the count check fails, and highlighting is withdrawn without a crash. It fails closed.

That is precisely what makes it worth closing. The guard **claimed** to have ruled out everything that is not `string | string[]`, and then relied on a check three call frames away to actually do it. The next person to read it sees a settled question. It now narrows on the elements:

```ts
if (!Array.isArray(selectors) || !selectors.every(one => typeof one === 'string')) {
  return null;
}
```

The count check moved into a `pairWith` helper so both branches reach it by the same path — the string branch previously shared it only by falling through, which is the arrangement that lets one branch drift.

## `p = 0.5`

Taken, and extended to `p <= 0` for the same reason: a rung is a *tail*, so its probability lies strictly between nothing and the median's own 0.5. Anything outside that is not a rung.

You called it cosmetic; I'd put it slightly higher, because of what this chart type is. Two rungs labelled `50th percentile` sitting either side of the rung already called `median` is **three names for one place**, and on a boxen the label is the whole of what tells a reader where on the distribution they are — there is no other cue. It is the same failure class as the `toPrecision(3)` decision this PR already argued for: the deep rungs are only distinguishable by their names.

Both new tests **fail on the parent commit**, so the filter is pinned rather than assumed.

## Verification

- `npx playwright test --project=chromium boxen` — **5 passed**

On the CI question at the end: the previous commit was green 14/14, and this push will re-run.

---

jooyoungseo:
Both taken in `5a4534b7`.

## The stale doc

Caught cleanly, and it was a doc *I* made wrong with the previous commit. `p = 0.5` was documented as "the median's own rung" while `ladderOf` had just started dropping it — a type doc telling a producer a value is accepted while the code silently discards it, which is the exact failure this branch has been arguing against everywhere else. It now says the range is strictly between 0 and 0.5, and *why* 0.5 is out of range rather than a way of naming the median.

## The selector-guard tests, and what they actually are

Added — and worth being exact, because the honest answer is not the flattering one: **they do not fail on the parent commit.**

I checked rather than assumed, by restoring the old guard and re-running:

```
Tests: 20 passed, 20 total
```

The old code reached the same observable result by a different route — `isUsableSelector` rejected each entry, the list came back empty, the count check failed, highlighting was withdrawn. That is exactly the "fails closed" property you identified in the first pass, and it means no test of *behaviour* can separate the two implementations.

So what these pin is the **contract** — an unusable selector shape withdraws rather than mispairs — not the narrowing. That is still worth having, since the contract is what a future refactor could break in a way that does *not* fail closed. But calling them regression tests would make them look like something they are not, and I have spent this PR arguing that a test which reads like coverage and is not is the worse kind.

## Verification

- `npm test` — **2298 + 73 passed**, 167 suites
- `npx playwright test --project=chromium boxen` — **5 passed**

CI was 14/14 on the rebased parent commit.

---

PR #842, opened 2026-08-12
feat(hexbin): keep a vertical walk over the x it started from

### LINKED ISSUE #799: feat: support hexbin (hexagonal binning) plots
## Description

Hexagonal binning is the standard answer to an overplotted scatter: bin the points into hexagons and encode count as fill. There is no trace type for it.

## Proposed mapping

Very close to `heat` — a lattice of cells, each carrying a count. The one genuine difference is that a hex lattice has **offset rows**, so:

- moving left/right within a row is unambiguous;
- moving up/down lands between two cells, and the trace has to define which one takes focus (nearest centre is the usual choice, and matches `NearestPoint` in `src/model/abstract.ts`);
- the announcement should give the bin's x/y centre and its count, not a row/column index.

Braille reuses `HeatmapBrailleEncoder`.

## Note

Rectangular 2D histograms (`hist2d`, `stat_bin2d`) need none of this — they are `heat` already, and are tracked in the binding repos.

## Difficulty

Medium, and confined to the lattice indexing.

jooyoungseo:
Closes #799. This is the last open item in tier A of #814.

## Mostly a heatmap, as #799 says

Read as a lattice of cells each carrying a count, a hexbin **is** a heatmap — so the grid navigation, the pitch, and `HeatmapBrailleEncoder` all transfer unchanged. The work is confined to the lattice indexing, exactly as the issue predicted.

## The stagger, and the part that took two attempts

A hex lattice offsets alternate rows by half a cell — that is what lets the hexagons tessellate. Two consequences:

**A column index is not a position.** Bin 3 of one row and bin 3 of the next sit at different x. So the announcement gives the bin's **centre**, per #799.

**A bin has no neighbour directly above it.** It has two, and they are *exactly* equidistant. My first implementation chose the nearer bin relative to the bin being left, which — because the two candidates always tie — means choosing the same way every time. The drift test caught it immediately:

```
from x = 2, two rows up  →  x = 1  →  x = 0
```

Half a cell per row, accumulating, which is precisely the slide the feature exists to prevent.

So the choice is made against **where the reader started**, held in an anchor and reset the moment they move sideways. That is the "desired column" a text editor keeps while arrowing past short lines — same problem, same solution. Now `2 → 1 → 2`, and stepping up and straight back down returns to where it started.

**Getting this wrong is silent**, which is what earns the mechanism. A drifting cursor announces a real bin with a real count at every step; the one thing that would give it away is a column index, and a column index is exactly what this trace does not announce.

## Smaller decisions

- **`min` count ignores empty bins.** A hexbin is mostly empty margin; reporting `0` as the minimum count would describe a density the occupied part of the chart does not have.
- **`Occupied bins` is reported alongside `Number of bins`.** A scatter spread evenly fills most of its lattice and a tight one leaves most of it empty — the shape of the cloud, not recoverable from the counts without walking every cell.
- **The densest bin is named by its centre**, for the same reason the announcement is.
- **Selectors are withdrawn on a count mismatch.** A hexbin routinely omits empty cells from the DOM, so a list off by one highlights a bin that is not even a neighbour in the direction a reader would guess.
- **Ragged rows are handled**: a move to a shorter row still lands inside it, where an index-based move would be out of bounds.

## Braille

`HeatmapBrailleEncoder`, unchanged. The stagger has no representation and needs none: a display is a line of cells with nowhere to put half of one. Empty bins render as spaces — so the e2e asserts a row of *cells and spaces* with at least one real cell, rather than pure braille, which would have been the wrong assertion rather than a stricter one.

## Verification

- `npm test` — **2203 + 73 passed**, 170 suites, none failing
- `npx playwright test --project=chromium hexbin` — **5 passed**, including a keyboard walk two rows up and two back down asserting the announcement is identical

---

---

### DIFF
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -105,6 +105,28 @@ Multiline braille displays represent grouped boxplots, such as horizontal boxplo
 
 Single-line braille displays represent grouped boxplots by allowing users to navigate vertically between groups using the up and down arrow keys. As the user navigates, the braille representation updates to show the boxplot for the current group, enabling users to explore each group's distribution one at a time.
 
+## Hexbin
+
+One row per lattice row, one cell per bin, using the heatmap encoding above —
+because read as a lattice of cells each carrying a count, that is what a
+hexbin is.
+
+**The stagger has no representation here, and needs none.** A hex lattice
+offsets alternate rows by half a cell, and a braille display is a line of
+cells with nowhere to put half of one. What survives intact is the thing the
+chart is drawn for: the density pattern, as a run of cells that rises towards
+the cloud and falls away from it, with the empty margins reading as spaces.
+
+Where the offset *does* matter is navigation rather than encoding — moving up
+or down lands between two bins, so the trace keeps the reader over the x they
+started from rather than stepping by index. A display shows the whole row at
+once and never has to make that choice.
+
+### Multiline Displays
+
+Hexbin plots use one line per lattice row on a multiline display, so the shape
+of the cloud can be felt at once rather than row by row.
+
 ## Scatter plot
 
 In the Braille representation of a scatter plot, the encoding is performed only for the line layer (layer 2). Stand alone scatterplots without a line layer are not represented in braille.
--- a/e2e_tests/specs/hexbin.spec.ts
+++ b/e2e_tests/specs/hexbin.spec.ts
@@ -1,0 +1,154 @@
+import type { HexbinPoint, Maidr } from '../../src/type/grammar';
+import { expect, test } from '@playwright/test';
+import { BasePage } from '../page-objects/base-page';
+import { TestConstants } from '../utils/constants';
+import { extractMaidrData } from '../utils/maidr-data';
+import { normalizeText } from '../utils/text';
+
+/**
+ * A braille row of a sparse lattice: cells from the Unicode braille block
+ * (U+2800 to U+28FF), and spaces where a bin held nothing.
+ *
+ * The heatmap encoding this reuses renders an empty cell as a space, and on a
+ * hexbin that is most of the margin around the cloud -- so a row of pure
+ * braille would be the wrong assertion, not a stricter one.
+ */
+const BRAILLE_ROW = /^[\u2800-\u28FF ]+$/;
+
+/** At least one real cell, so an all-space row is not mistaken for output. */
+const BRAILLE_CELL_ANYWHERE = /[\u2800-\u28FF]/;
+
+/** The example's own page, driven through the shared base helpers. */
+class HexbinPlotPage extends BasePage {
+  protected override readonly selectors = {
+    notification: `#${TestConstants.MAIDR_NOTIFICATION_CONTAINER} ${TestConstants.PARAGRAPH}`,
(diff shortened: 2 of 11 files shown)


jooyoungseo:
Both real, both fixed in `7add5cc4`. You were right to call these keyboard-accessibility gaps rather than nits — I ran each before changing anything.

## 1 — the first keypress

```
first press DOWNWARD  →  moveOnce returned false   (row -1, edge tone on the opening keystroke)
first press UPWARD    →  landed on row 1, x = 1     (row 0 never announced)
```

Exactly as you traced, including *why* the tests missed it: the e2e always moves right first and the unit helper goes through `moveToIndex`, which sets `isInitialEntry = false` directly. Neither path can reach a bare vertical `moveOnce` on a fresh trace.

The second case is the worse of the two, and it compounds: it also seeds the anchor from a bin the reader was never told about, so the *entire* vertical walk then tracks a column they did not choose. The branch now delegates to `super.moveOnce(direction)` when `isInitialEntry` — reusing the convention rather than reimplementing the clamp, so it cannot drift from `MovableGrid`'s version later.

## 2 — `moveToExtreme`

Confirmed, and worse on this trace than the general case, for the reason the type exists: **a hexbin is sparse.** An empty bin is not drawn, so rows genuinely hold different numbers of bins at different offsets — index-clamping is not "off by half a cell", it is off by however much the two rows disagree.

My first fixture didn't show it, which is worth saying: a *regular* lattice makes index-clamp and nearest-by-x agree almost everywhere, so the test passed and proved nothing. The fixture is now a diagonal cloud, where the bottom row occupies the right of the field and the top row the left:

| | |
|---|---|
| Ctrl+Down from x = 4 | nearest-by-x lands on **x = 8** |
| index 2, clamped into a 2-bin row | lands on **x = 10** — the wrong end of a row that does not reach x = 4 at all |

Also confirmed your point that `supportsExtrema` doesn't gate these — it gates the statistical extrema menu, and `MoveToTopExtremeCommand`/`MoveToBottomExtremeCommand` are bound in `TRACE` scope unconditionally.

## 3 — where I went the other way

You suggested resetting `anchorX` on the jump. **I kept it**, and I think the class's own argument decides it: the anchor is *the column the reader chose*. Where a jump lands is decided by whatever the far row happens to hold — in the new fixture, x = 8, most of the field from the x = 0 they picked. Resetting would strand them there, which is the same "cursor ends up somewhere it was not asked to be" failure the anchor exists to prevent, just arrived at by a different route.

A horizontal move still ends the walk, whether one step or a jump to the end of the row. There is a test for that specifically, since `moveToExtreme('FORWARD')` is a new way to reach it.

## Coverage

Five new tests. I checked each against the code it pins rather than assuming:

- **4 of 5 fail** with the respective fix reverted
- the 5th is the keep-vs-reset one, which passes with or without the override (the base implementation leaves `anchorX` alone) but **fails if the override resets it** — so it pins the decision above, not the fix

One incidental confirmation while doing this: reverting `moveOnce`'s anchor line breaks the two existing drift tests, so that half was already pinned.

## Verification

- `npx playwright test --project=chromium hexbin` — **5 passed**

---

PR #843, opened 2026-08-12
feat(ridgeline): compare distributions at the value the reader chose

### LINKED ISSUE #803: feat: support ridgeline (joy) plots
## Description

A ridgeline plot stacks one density curve per group along a shared value axis, offset vertically, to compare distributions. `ggridges`, `sns.violinplot`'s cousins and Plotly ridgelines all produce it; MAIDR cannot read any of them.

## Why it is cheap

`ViolinKdeTrace` (`src/model/violin.ts`) already models a KDE curve per group, with `ViolinKdePoint` carrying `{x, y, density}`. A ridgeline is the same data with a different layout — offset instead of mirrored. The trace can very likely share the KDE machinery outright.

## What is actually new

- The baseline offset is presentational, not data, so it must not leak into the announced values.
- Curves overlap on purpose, which matters for highlighting but not for navigation.
- Comparison across groups at a fixed value is the point of the chart, so a "same x, next group" move is worth more here than in a violin plot.

## Difficulty

Low-to-medium, mostly reuse.

jooyoungseo:
Closes #803. First of tier B in #814.

## Mostly reuse, as the issue says

The issue is right that `ViolinKdeTrace` already models a KDE curve per group. `ViolinKdePoint` is reused **outright** — same `{x, y, density}` — and so are the grid navigation and the line braille encoder.

Two things differ enough to make this a type rather than a violin with a layout, and both are the kind of difference that produces a chart which sounds confident and says something false.

## 1 — every group is pitched against the whole chart

`ViolinKdeTrace` pitches from a **reference curve** (row 0) and uses the current group only for volume. That suits a chart drawn to compare each violin against a first one. A ridgeline is drawn so a reader can ask *which distribution is tallest, and where* — and that needs one scale across all of them.

| | scaled per group | scaled across the chart |
|---|---|---|
| a group peaking at 0.09 | top of the register | a fifth of the way up |
| a group peaking at 0.90 | top of the register | top of the register |

Per-group, the chart reports a dozen equally tall ridges. Braille does the same and for the same reason, so a row of full cells means a dense group and not merely the densest part of a sparse one.

**The floor is zero**, not the smallest density measured anywhere. A density of zero is a real reading — nobody in that group had that value — so it belongs at the bottom of the range rather than somewhere above it.

## 2 — moving between groups holds the value

Comparing at a fixed point on the value axis is what the chart is *for*, and it is exactly the comparison a listener cannot make by holding a number in their head across a dozen groups. Groups are **not obliged to share a sample grid** — a density estimated over each group's own range does not — so stepping by *index* lands at a different value, silently: every sample is a real density at a real place, and the only thing that would give it away is the value itself, which is what changed.

So the group is stepped and the sample re-chosen as whichever sits nearest the value the reader is holding.

**Holding it, rather than re-deriving it at each ridge.** This is the part worth the mechanism, and it is subtler than the hexbin's version of the same problem: there, the two candidates always tie, so re-deriving drifts immediately. Here a single step moves by at most half a sample spacing, so re-deriving *looks* correct — and accumulates only over a walk. My first fixture could not tell the two apart and passed either way; the one in the branch uses three staggered grids where the answers diverge by the third group:

```
reader chooses 10  →  b has nothing nearer than 16
                   →  c straddles: 5 is nearest to 10, 22 is nearest to 16
```

Twelve units from where they meant to be, announcing a real density at a real place. **The test fails when the anchor is removed.**

## 3 — the offset is not data

A ridgeline's vertical stagger exists so the curves do not sit on top of one another. It says nothing about any group, so the layer carries each curve on its own terms — the value it was measured at and the density there — and never the baseline it was drawn from. Same rule the area layer follows: read what the chart is *of*, not where it was put.

The e2e pins it from the other side: the emitted groups still **overlap** on the value axis, exactly as the chart draws them overlapping. An offset baked into the data would have separated them into disjoint runs.

## Smaller decisions

- **The description reports where each group peaks** (`2019 at 34, 2021 at 41, 2023 at 52`) and the span of those peaks. Whether the modes march across the axis is the ridgeline's headline finding, and a reader cannot assemble it without walking every sample of every group. The span is **withheld** when every group peaks in the same place, rather than reporting a range of zero.
- **Highlighting is one element per group**, repeated across its samples — a ridgeline is drawn as one filled path per group and does not emit a per-sample element to pair with. Selectors are withdrawn on a count mismatch: the curves overlap, so a list off by one lights a ridge that is not obviously wrong even to a sighted reader.
- **`supportsExtrema` is false.** "The largest value" on a density curve is its mode, which the description already names for every group.
- **The first keypress enters the chart** rather than stepping through it, delegating to `MovableGrid` — the gap found on #842 and worth not repeating.

## Registration

All six sites plus `docs/BRAILLE.md` — **and a seventh** the checklist does not name: `test/util/orientation.test.ts` keeps its own list of every type and asserts totality against `TraceType`. The suite caught it, which is the checklist working even where it is incomplete.

## Verification

- `npm test` — **2312 + 73 passed**, 168 suites
- `npx playwright test --project=chromium ridgeline violin` — **24 passed** (the violin suite included deliberately, since the point type is shared)

Each mechanism was checked against the code it pins rather than assumed: reverting nearest-by-value fails 3 tests, and reverting the anchor fails the drift test.

---

---

### DIFF
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -574,6 +574,46 @@ row while falling on the other.
 Diverging charts use one line per side on a multiline display, plus the
 balance, so the two sides can be compared with one sweep.
 
+## Ridgeline (joy plot)
+
+One row per group, one cell per sample of its density curve — the line
+encoding, which is what a density curve is.
+
+**Every row is scaled against the chart's peak rather than its own.** This is
+the one place a ridgeline's braille differs from a violin's, and it is
+deliberate: the chart exists so a reader can ask which distribution is tallest
+and where. Scaled per row, a group holding a tenth of another's density would
+fill its cells just as high, and the display would show a dozen equally tall
+ridges — the exact reading the chart is drawn to prevent.
+
+The floor is zero rather than the smallest density measured anywhere. A density
+of zero is a real reading — nobody in that group had that value — so it belongs
+at the bottom of the range and not somewhere above it.
+
+**The baseline offset has no representation here, and needs none.** The stagger
+that separates the curves on screen is presentation; it is not in the data and
+it is not in the display. What the fingers get is each group's shape, aligned
+row against row, which is the comparison the offset exists to make possible
+visually.
+
+```
+Group   Cells
+2019    ⠄⠆⠖⠶⣶⣶⠶⠖⠆⠄
+2021    ⠄⠆⠖⠖⠶⠶⠖⠖⠆⠄
+2023    ⠂⠆⠖⠶⣿⣿⠶⠖⠆⠂
+```
+
+The 2021 cohort's row never reaches the top because that cohort's curve never
+reaches the chart's peak — it is more spread out, so its density is lower
+everywhere. A per-row scale would have hidden that by filling its middle cells
+as high as everyone else's.
+
+### Multiline Displays
+
+Ridgelines use one line per group on a multiline display, so a reader can sweep
+across the groups at a fixed position and feel the modes march along the axis,
+or fail to — which is the finding the chart is drawn for.
+
 ## Multiline Braille Display Support
 
 By leveraging the two-dimensional nature of multiline braille displays, MAIDR can represent multiple lines of a plot simultaneously, allowing users to perceive the distribution of values across all lines at once. This is particularly beneficial for plots with multiple groups or categories, such as grouped boxplots or line plots with multiple lines, and it also applies to scatter plot grid navigation.
--- a/e2e_tests/specs/ridgeline.spec.ts
+++ b/e2e_tests/specs/ridgeline.spec.ts
@@ -1,0 +1,150 @@
+import type { Maidr } from '../../src/type/grammar';
+import { expect, test } from '@playwright/test';
+import { BasePage } from '../page-objects/base-page';
+import { TestConstants } from '../utils/constants';
+import { extractMaidrData } from '../utils/maidr-data';
+import { normalizeText } from '../utils/text';
+
+/**
+ * A braille row is cells, and may carry spaces where a curve is flat against
+ * the chart's floor.
+ */
+const BRAILLE_ROW = /^[\u2800-\u28FF ]+$/;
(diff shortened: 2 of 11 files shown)


jooyoungseo:
Confirmed and fixed in `2e7b675a`. Reproduced before touching anything:

```
from group `a` at value 10  →  Ctrl+Up landed on group `c` at value 22
```

Twelve units from where the reader was comparing, announcing a real density at a real place. Your trace was exact, including the point that `supportsExtrema = false` does not gate these keys — that flag gates the statistical extrema menu, not the grid-edge jump.

## Worth naming, because it is not a coincidence

**This is the same gap I found and fixed on `HexbinTrace` earlier today**, in the same shape, and I did not carry it across when writing this trace — despite citing that trace as the precedent in the PR body. Two keybindings reach the same movement; I overrode one of them.

That is a cheap lesson to bank: any trace that overrides `moveOnce` for a direction has to ask what `moveToExtreme` does with the same direction, because they are two entry points to one movement and only one of them is obvious.

## The anchor across the jump

Kept, matching hexbin and for the same reason: where a jump lands is decided by whatever the far group happens to have sampled, so resuming from the landing value would strand the reader at a value they never chose. A horizontal extreme still ends the comparison.

## Coverage, checked rather than assumed

Three tests. **Two fail without the override.** The third — the keep-vs-reset one — passed either way on my first attempt, so I rebuilt its fixture: `b` now straddles the two candidates, so the chosen 10 resolves to 12 while the 5 that `c` forced resolves to 2. It **fails when the anchor is reset**.

## Enum placement

Taken. It had landed between `STACKED_AREA` and `STEP` because of the anchor I inserted it against; it now sits after `RADAR`. Cosmetic, but the six-site checklist is navigated by grepping a `TraceType`, so a scannable order earns its keep.

## Verification

- `npm test` — **2315 + 73 passed**, 168 suites
- `npx playwright test --project=chromium ridgeline violin` — **24 passed**

CI was 14/14 on the parent commit.

---

PR #844, opened 2026-08-12
feat(forest): say whether a study crosses the null, and what it weighs

### LINKED ISSUE #806: feat: support forest plots

jooyoungseo:
Closes #806. Tier B of #814.

## Free once the interval trace exists, as the issue says

`ForestTrace extends ErrorBarTrace`. The sections, the navigation, the braille and the per-section scaling all transfer unchanged — there is a test asserting that walking one study's three rows gives the same three numbers an error bar would. What this adds is only what the figure is actually *read* for.

## 1 — whether an interval crosses the null is the result

A sighted reader takes it from the drawing in one glance: the interval either touches the line or it does not. Left to a listener it means holding two bounds in mind and comparing both against a number that appears nowhere in the announcement.

**The null value is declared, and there is deliberately no default.** A ratio measure is null at 1 and a difference at 0. Guessing is not a degraded reading — assumed at 0, every odds ratio reads as *not crossing*, because odds ratios are all positive. That is a confident wrong answer handed to every row of the figure. A layer that stays quiet gets the estimate, the interval and the weight, and no claim about significance.

Two cases the tests are built around:

| study | interval | reading |
|---|---|---|
| `Silva 2018` | 0.41 – 0.94, entirely **below** 1 | does **not** cross |
| `Nguyen 2020` | 0.98 – 1.83 | crosses |

The first is the one that matters. A check that only asked whether the upper bound cleared the null would call it a crossing and **report a real protective effect as no effect at all**.

A one-sided interval is answered on the bound it has — unbounded below, an upper bound of 2.2 still contains 1; bounded at 1.1 and unbounded above, it does not.

## 2 — the weight is a magnitude the reader is otherwise never told

A forest plot encodes it as marker *area*, and nothing in an error bar's reading carries it. Two studies whose intervals sound alike can contribute wholly differently: in the example `Okafor 2022` carries **55%** of the analysis and `Nguyen 2020` carries 8%.

Announced on the estimate row only. A study has three rows, and repeating the weight and the verdict at each one buries the number the reader navigated to.

## 3 — the pooled row is not a study

It is what the studies came to, so it names itself `pooled estimate, does not cross the null`, its bounds read as `pooled lower bound`, and it is **excluded from the count** — counting it among the evidence would report five studies where the analysis had four.

## The description

- `Pooled estimate` — with its verdict, or without when no null is declared
- `Studies crossing the null` — `2 of 4`; a count a reader cannot assemble without walking every row and comparing two bounds at each
- `Heaviest study` — `Okafor 2022, 55.0%`. A meta-analysis whose weight sits in one trial is a different object from one where it is spread, and the per-row announcement never says where the mass is.

Heterogeneity statistics are **out of scope**: the issue lists them, but they are not derivable from the points and the grammar carries no field for them. Inventing one without a producer to fill it would be speculative.

## Braille

The error bar's encoder, unchanged, because a forest plot's magnitudes are an error bar's. `docs/BRAILLE.md` says what the display deliberately does *not* carry: the null line is a position rather than a magnitude and a cell height cannot express it, and the weight would have to displace the value it shares its one dimension with. Both are announced instead.

## Two failures that were mine, not the code's

Recorded because both would have shipped as green tests saying the wrong thing:

- **`2 of 4`, not `1 of 4`.** Haddad's 0.60 – 1.84 spans 1 as well. I had miscounted the fixture I wrote.
- **`forest(1, 1, undefined)` still took the default.** Passing `undefined` to a default parameter takes the default, so both "declares no null" tests were silently exercising the declared path. Now `null` with an explicit `?? undefined`, and the comment says why.

## Registration

All seven sites plus docs. `IS_ORIENTED` is `true` — a forest plot is conventionally drawn with its studies down the page — and the e2e asserts orientation **where `IS_ORIENTED` actually reaches**, which is the activation announcement rather than the per-point text. That distinction came out of #840 an hour ago.

## Verification

- `npm test` — **2351 + 73 passed**, 170 suites
- `npx playwright test --project=chromium forest errorbar` — **16 passed** (the error bar suite included deliberately, since `points` was widened to `protected` for the subclass)

The two total `Record`s caught the missing `CHART_TYPE_LABEL` and `IS_ORIENTED` entries at compile time, which is the checklist working as designed.

---

---


jooyoungseo:
Both taken in `05777086`. The second is the good one — you're right that it read as an oversight, because it was one.

## The inconsistent counts

Reproduced by printing the description:

```
STAT Number of points          = 5
STAT Narrowest interval        = 0.53
STAT Studies crossing the null = 2 of 4
```

Two counts of the same thing that disagree, in one paragraph, with nothing to say which is right. My own argument for the exclusion applies to the inherited stats and I stopped applying it one line short.

The study count is now taken over the evidence and **named for what it counts** — `Number of studies is 4` rather than a corrected number under a generic label. Interval widths go the same way, and your reasoning about why is the sharper version: a pooled interval is typically the tightest on the figure *because that is what pooling is for*, so `Narrowest interval` would routinely hand the summary to a reader asking which study was most precise. There's a test with a deliberately tight pooled row where the old code would have reported `0.1`.

## Where I drew the line, and why

`Min value`, `Max value` and the extrema navigation are **deliberately left counting every row**, and the code now says so rather than leaving it to be inferred:

> Those describe the extent of the axis the figure is drawn on, and the pooled estimate sits on that axis like anything else. Jumping to the largest value should reach whatever is largest, and the pooled row announces itself as pooled on arrival.

The distinction is between **claims about the evidence**, which must exclude the pooled row, and **positions on the chart**, which must not. A reader who presses "go to the maximum" and is silently skipped past the diamond has been lied to about the chart's shape; a reader told "4 studies" when there are four studies has not.

## The orphaned doc comment

Caught cleanly — inserting `forestOptions` between the violin block and `violinOptions` left the latter undocumented and hung the violin text over the wrong property. Restored.

## Verification

- `npm test` — **2353 + 73 passed**, 170 suites
- `npx playwright test --project=chromium forest errorbar` — **16 passed**

On the note about running the suite: understood, and worth saying that your static reads have been catching things the suite could not — the pooled-count inconsistency is not a test failure, it is two green numbers that contradict each other.

---

PR #845, opened 2026-08-12
feat(survival): give a Kaplan-Meier curve its median and its censoring

### LINKED ISSUE #805: feat: support Kaplan-Meier survival curves

jooyoungseo:
Closes #805. Tier B of #814.

## Three parts, two of which already existed — as the issue says

`SurvivalTrace extends StepTrace`. The curve is a step function, so the shape, the navigation, the sonification and the braille all transfer unchanged. There is a test asserting a time reads exactly as a step chart's would. What this adds is the two things the figure is actually read for.

## 1 — median survival is the number readers came for

It is the single figure a survival analysis is quoted by, and without it a reader has to walk the curve holding every probability in mind until one crosses a half. Taken as the **first** time at or below 0.5, since survival is non-increasing — there is a test with a curve stepping 0.55 → 0.20 where "nearest to a half" would give the wrong month.

**`not reached` is a result, not a missing number.** In the example, `Treatment` never falls to half:

```
Median survival: Control: 9, Treatment: not reached
```

More than half its subjects were still alive at the end of follow-up. Omitting the arm from the list would read as a chart that forgot it.

## 2 — censoring is not an event

A censored time is a subject who left the study without the event happening, and **the curve does not step there.** In the example `Treatment` holds at 0.79 across months 9 and 12:

| month | survival | |
|---|---|---|
| 9 | 0.79 | an ordinary time |
| 12 | 0.79 | **censored** |

Nothing else in the announcement separates them. That matters because a flat tail carried by two hundred subjects and one carried by three are the same run of numbers and mean entirely different things — censoring changes not the estimate but how much of it is still supported by data.

Announced at the time, and offered as a **rotor filter**: a survival curve routinely carries hundreds of times, most of them uneventful, and censoring is sparse and consequential. Withheld on a curve where nobody was censored, since a mode whose only possible answer is "none found" is worse than not offering it. The step chart's own `Transitions` unit survives alongside rather than being displaced — there is a test for that, since appending to an inherited list is easy to get wrong by replacing it.

## Smaller decisions

- **The confidence band travels with the estimate** as `crossRange`, not as a second navigation axis. How wide the interval is at a time is how much the curve is worth there, and it is the comparison a reader makes when two arms look separated. Absent when the chart draws no band.
- **Separation at the last shared time** is reported — what a two-arm figure is drawn to show, and not assemblable by ear without holding one curve's last value while walking the other. Withheld for a single arm, where there is nothing to separate.
- **It announces itself as a survival curve.** `StepTrace` forces `plotType` to `'step'`; a survival curve *is* a step chart, but it is not *a step chart* to a reader, and the chart type is one of the few places they learn what they are looking at. The e2e asserts the announcement contains `survival` and **not** `step`.
- **The risk table is deliberately out of scope.** It is a second set of numbers on a second axis, which the issue itself describes as a companion layer rather than something to fold into the curve. Saying so rather than silently dropping it.

## Braille

The step chart's encoder unchanged. `docs/BRAILLE.md` says what the display deliberately does *not* carry and why: censoring marks would put an annotation where the reader expects the data — the cell is the same cell either way, because the estimate is the same — and the band has nowhere to go, a cell having one height already spent on the estimate. Both are announced instead.

## Verification

- `npm test` — **2367 + 73 passed**, 171 suites
- `npx playwright test --project=chromium survival step` — **23 passed** (the step suite included deliberately, since this extends it)

Both new mechanisms checked against the code they pin: taking the *last* time at or below half fails the median test, and dropping the censored marker fails the censoring test.

---

---


jooyoungseo:
Real bug, confirmed and fixed in `19b86018`. Reproduced first:

```
Control    x = [0, 2, 5, 9, 14]
Treatment  x = [0, 3, 7]

announced: "Separation at the last shared time is 0.15"
measured:  Control at month 5 (0.70) against Treatment at month 7 (0.85)
```

Two different months, under a label saying it was one. And you put your finger on why it's worse than an edge case: **my own docstring promised "the last time they share" and promised null when there is none** — and the code never read `x` at all. It was a comment describing a function that did not exist.

## The fix goes further than exact-x matching

Your suggested fix — the largest `x` present in every arm, null otherwise — is correct, and I nearly took it before noticing it would be **correct and nearly useless**: independently fitted arms share only time zero, so it would answer `null` on almost every real forest of curves.

A survival curve is a *step function*. Its value at any time is the one carried forward from its last point at or before that time, which makes two arms comparable at a time **neither of them sampled**. So shared follow-up ends at the earliest last-observation across the arms, and each arm is read there the way the chart is drawn:

```
shared follow-up ends at month 7
  Treatment 0.85 (its own last point)
  Control   0.60 (held forward from month 6)
```

The stat now **names the time**, because it is the end of the *shorter* arm's follow-up rather than the end of the chart, and a reader given a bare number would reasonably assume otherwise. A categorical `x` gets silence rather than a guess — there is no "end of follow-up" to read at.

## My first fixture didn't discriminate

Worth recording, since it is the third time this session. I built the uneven-grid test with your exact example, and **it passed with the buggy code**: Control had no point between month 5 and month 7, so the step read and the index read both landed on 0.70. Coincidence of the fixture, not coverage.

Control now has a point at month 6, so the two answers genuinely differ — `0.25 at 7` versus the old `0.15` — and the test **fails on the parent commit**.

## The two minor notes

- **`announcesSectionBeforeLabel`** — good catch that `z` being set routes the censored point through the candlestick-style branch. I agree it reads acceptably (`…, censored Survival probability is …`) and I'd rather not change a shared text-service branch from a trace PR; it would affect every trace that sets both. Left as is.
- **The redundant `Number()` casts** — kept deliberately. `yMin`/`yMax` are typed `number` but arrive from JSON that nothing validates, so `Number.isFinite(Number(x))` is doing real work at the boundary rather than restating the type. That is the same reason `magnitudeOf` in `errorBar.ts` casts.

## Verification

- `npm test` — **2369 + 73 passed**, 171 suites
- `npx playwright test --project=chromium survival step` — **23 passed**

CI was 14/14 on the parent commit, which is the confirmation of the numbers you noted you could not run.

---

PR #846, opened 2026-08-12
feat(mosaic): read a bar's width as the second magnitude it is

### LINKED ISSUE #804: feat: support mosaic and marimekko plots
## Description

A mosaic (or marimekko) plot is a stacked bar chart in which the **bar widths also encode data** — typically the marginal total of each category. Used for two-way contingency tables; unsupported today.

## Why the width matters

`SegmentedTrace` already navigates "category, then series within category", so the segmented part is solved. The gap is that MAIDR treats bar width as layout. In a mosaic it is a second magnitude, and a reader who is told only the segment heights is getting half the table: they can hear the conditional proportions but not the group sizes those proportions are computed from.

## Proposed mapping

```
{ x: 'First class', width: 0.15, segments: [{z: 'survived', y: 0.62}, {z: 'died', y: 0.38}] }
```

Announce both — "First class, 15% of all passengers; survived 62%" — and consider offering the cell count as well, since the underlying object is a contingency table.

## Difficulty

Medium.

jooyoungseo:
Closes #804. Tier B of #814.

## The segmented part is solved, as the issue says

`MosaicTrace extends SegmentedTrace`. Category-then-series navigation, the braille encoder and the summary row all transfer unchanged — there is a test asserting a cell reads exactly as a stacked bar's would. **The gap is the width.**

## Everywhere else a bar's width is layout. Here it is data.

Without it a reader has half the table: the conditional proportions with no idea what they were computed from. The example is built so the two **disagree**:

| class | survival rate | share of the ship |
|---|---|---|
| First | **62%** — the best | **15%** — the smallest |
| Third | 25% — the worst | 32% — the largest |

A heights-only reading reports the same table for a ship of six first-class passengers and one of six hundred, which is precisely the comparison a mosaic is drawn to make. The e2e asserts the inversion holds in the shipped example, so the fixture cannot quietly become one where it does not matter.

The announcement lands close to the wording the issue suggested:

> First, 15.0% of all, Proportion is 0.62, Outcome is Survived

## Four decisions in the width handling

- **Every segment of a column announces it**, not just the first. A reader who arrives at the second segment has not stopped needing it.
- **It is read from whichever segment declares it.** The width belongs to the column, so a producer that put it on one cell is emitting something perfectly sensible.
- **A layer without widths says nothing.** A mosaic without widths is a stacked bar, and inventing a share would put a number in the announcement that the chart does not contain.
- **The summary row keeps the width but loses the count.** It is a column total, not a cell — so it has no count of its own, while the column is still the same column.

## The cell count

Carried when the producer has the contingency table, since the issue is right that a mosaic is drawn *from* counts and they are the numbers a reader would quote back. Optional, because a producer working from proportions genuinely does not have them, and multiplying out a rounded share would invent one.

## The description

- `Share of all observations` — the marginal distribution, which is the axis a mosaic adds and the one a reader cannot accumulate by walking the segments
- `Largest group` / `Smallest group` — what separates a proportion computed from six observations from one computed from six hundred
- `Total observations` — when the table carries counts

Category names are resolved by orientation. A horizontal layer carries its category on `y`, and reading a fixed field would name the columns by their proportions — there is a horizontal fixture, and it fails when the field is fixed.

## Braille, and what it deliberately cannot do

The stacked bar's encoder. `docs/BRAILLE.md` argues the width's absence is a **real loss rather than a non-issue**: a cell has one dimension, already spent on the magnitude, so the only way to widen a column would be more cells — and the number of cells is the reader's index into the table. A display where `Third` took three cells and `First` one would put them in different columns and break every comparison down a row.

The total row is full across, as it must be — a mosaic's columns each sum to one whatever their width — which is exactly why the width cannot be inferred from the dots.

## Verification

- `npm test` — **2388 + 73 passed**, 172 suites
- `npx playwright test --project=chromium mosaic stacked` — **27 passed** (the stacked-bar suite included deliberately, since this extends it)

Both mechanisms checked against the code they pin: dropping the width announcement fails four tests, and naming categories from a fixed field fails the horizontal one.

*(The commit message says 38 for the e2e count; the correct figure is 27, as above.)*

---

---


jooyoungseo:
All three taken in `f3b26d2b`. The first is a serious catch and I'd have shipped it.

## The width was fused to the label

I ran the real formatter instead of reading the field, and it is exactly what you traced:

```
verbose: Class is First, 15.0% of all Proportion is 0.62, Outcome is Survived, Count is 203
terse:   First, 15.0% of all 0.62, Survived
```

The terse reading is the one most users get, and it **says the proportion is the share** — precisely the ambiguity this trace exists to remove. And you're right that the PR body claimed a sentence that was unreachable: `announcesSectionBeforeLabel` requires `z === undefined`, and `SegmentedTrace` always sets `z`, so the branch I described could never run.

## One fix, not two

Your second finding has the same root cause, so I did not add two fields. `stack` is formatted with the cross-axis formatter, and neither the width nor the count is a value on either axis — a `percent` format on a proportion axis would have announced `203` as `20300.0%`, as you say.

So both now travel in one new `TextState.asides`: already-formatted label/value pairs the service emits as their own clauses and never runs through an axis formatter. The generalisation is the honest one — **`TextState`'s slots are all either axis-formatted or fused, and a mosaic carries two magnitudes that are neither.**

```
verbose: Class is First, Proportion is 0.62, Outcome is Survived, Share of all is 15.0%, Count is 203
terse:   First, 0.62, Survived, 15.0%, 203
```

## The tests were the actual defect

Worth stating plainly: my unit tests asserted `text.section` and `text.stack`, which is *why* they stayed green while the sentence was wrong. A test that asserts the field a bug lives upstream of is not coverage.

- `test/service/mosaicText.test.ts` is new and asserts what the reader hears, through `TextService`, in both modes — including a fixture whose cross axis declares `{ type: 'percent' }`, which is what would have exposed the count.
- The e2e now asserts the **whole sentence**. You noted `toContain('15.0% of all')` was true of the fused reading; it was, and that is now written down in the test as the reason.

## Third finding

Taken. `grandTotal()` returned a partial sum when only some columns carried counts, and a partial sum announced as `Total observations` is a number the reader will take for the size of the table. It answers `null` unless every column has one, with a test.

## Verification

- `npm test` — **2392 + 73 passed**, 173 suites
- `npx playwright test --project=chromium mosaic stacked candlestick boxplot` — **77 passed**; the last three deliberately, since `section` and `stack` are theirs and I changed how the service handles neither but sits beside both

---

PR #847, opened 2026-08-12
feat(volcano): find the points that cross the line, not the ones next to them

### LINKED ISSUE #807: feat: support volcano and manhattan plots
## Description

Volcano plots (effect size vs. significance) and Manhattan plots (genomic position vs. significance) are the standard summary figures of differential expression and GWAS analyses. Both are scatter plots read almost entirely through **thresholds**.

## Why a plain scatter is not enough

These charts routinely carry tens of thousands of points, of which a few dozen matter. Reading them point by point is not a viable path. The question is always "which points cross the significance line", and secondarily "which chromosome / region are they in".

## Proposed mapping

`ScatterTrace` plus:

- **Threshold lines** as declared, navigable references (a p-value cutoff, a fold-change cutoff).
- **Rotor filter units** for "above threshold" / "significant", so a user jumps between the points that matter. `RotorFilterUnit` already supports trace-defined units — candlestick registers bullish/bearish filters the same way.
- **Point labels** (gene, SNP), since identity is the payload here, not coordinates.
- For Manhattan plots, chromosome as a grouping the reader can move between.

An "N points, 43 above the significance threshold" summary on entry does more for these charts than any per-point navigation.

## Related

- #612 and #610 (point-by-point navigation in scatter) overlap on the scatter navigation model.

## Difficulty

Medium, and mostly additive to the existing scatter trace.

jooyoungseo:
Closes #807. This is the last open item in tier B of #814.

## Point-by-point navigation is not access to this chart

The issue puts it exactly: these carry tens of thousands of points of which a few dozen matter. A reader made to walk them has not been given the chart — they have been given a very long list. So two things carry the weight, and the issue is right that the first matters more:

**The summary on entry.** `Above the threshold is 5 of 15` — the first stat, because it is the first thing a sighted reader takes from the shape of the cloud, and the last thing a per-point walk would ever assemble.

**A rotor filter over the points that clear it**, so the few that matter are a few keystrokes away rather than twelve thousand. It follows the candlestick's bullish/bearish precedent, and is **withheld** both when no threshold is declared and when one is declared but nothing clears it — a mode whose only possible answer is "none found" is worse than no mode. That second case is a real reading, and the description states it in words: `0 of 15`.

## Identity is the payload

A reader told *"x is 3.4, y is 9.1"* has the two numbers the axes already describe, and not the one thing they came for: that this is **UP1**.

The label travels in `asides` — the field #846 added — so it survives every one of the scatter's navigation modes and is never run through an axis formatter. A gene name is not a value on an axis. It is announced **only where the cursor is on exactly one point**: the scatter's column mode sits on a whole stack, and naming any of them there would name the wrong one.

## The threshold is declared, never guessed

Same argument as the forest plot's null, and it bites harder here. These charts sit on transformed axes whose conventions differ by field and by tool — `-log10(p)` at 1.3 for p < 0.05, at 7.3 for genome-wide significance, and a raw p axis runs the other way entirely. **A guessed line sorts every point on the figure onto the wrong side of it, silently.** A layer declaring none gets the scatter's reading and no claim about significance.

## Two details the fixtures exist to pin

Both **fail when reverted**, which I checked rather than assumed:

| reverted to | tests that fail |
|---|---|
| effect measured rightwards only | 4 |
| either cutoff rather than both | 6 |

- **The effect cutoff is applied to the *magnitude* of x.** A volcano is symmetric: a fold change of −3.1 is as large an effect as +3.1. A rightwards-only reading drops half the findings, and the example has two of them (`DOWN1`, `DOWN2`).
- **Both cutoffs must be cleared.** The fixture has `LOUD` (significant, small effect) and `BIG` (large effect, no significance), so taking either alone names the wrong set.

## One class, two types

`manhattan` and `volcano` share `VolcanoTrace`, the way `POLAR_AREA` and `RADAR` share `RadarTrace`: they differ in what the x axis means and in nothing a reader navigates. A Manhattan layer's points carry a `group` — the chromosome — announced alongside the point and counted in the description.

## Braille

The scatter encoding, unchanged. `docs/BRAILLE.md` says why the threshold **cannot** be represented rather than merely isn't: it is a line on the value axis, not a magnitude, and there is no dot pattern for "this is the 5.0 mark". It is carried in the other three modalities instead — which is the right division, since the finding is about a handful of points and the display is about all twelve thousand.

## A note on the e2e

There is no assertion on the entry summary in the spec, deliberately. No e2e in this suite opens the description modal, and inventing a key sequence would have tested my guess at the binding rather than the feature. The summary is pinned at the unit level, including that it is the **first** stat.

## Verification

- `npm test` — **2431 + 73 passed**, 175 suites
- `npx playwright test --project=chromium` — **the full suite, 473 passed**, run rather than a subset because `ScatterTrace`'s `flatPoints`, `readingOrder`, `isInPointMode` and `pointModeIndex` widen from `private` to `protected` (visibility only, no behaviour change)

---


jooyoungseo:
Both blocking findings are real and both are fixed in a78e47f. A third one I want to name explicitly, because the review did not raise it and it was worse than either.

**The enum.** Correct on both counts, and the first is the one that matters: `MANHATTAN` and `VOLCANO` went in between the waterfall JSDoc block and `WATERFALL = 'waterfall'`, so that doc block ended up describing `MANHATTAN`. `MANHATTAN` now sits between `LOLLIPOP` and `MOSAIC`, `VOLCANO` between `VIOLIN_KDE` and `WATERFALL`, and the waterfall block is back above its own member. The `factory.ts` case moved to match.

**The rotor scan.** `moveToRotorFilter` called `readingOrder.indexOf(...)` to place each candidate — O(n) per candidate, O(n·k) per keystroke, on exactly the charts this mode exists for: a GWAS Manhattan is hundreds of thousands of SNPs with thousands clearing genome-wide significance. I had precomputed `significantIndices` to avoid re-filtering per keystroke and then linear-scanned to place each one, which undid the point of precomputing it. `ScatterTrace` already keeps `readingPos`, the inverse permutation of `readingOrder`; it is now `protected` with a comment saying why, and both lookups are O(1).

**The one you did not flag.** My own doc said these charts "sit on transformed axes whose conventions differ by field and by tool: -log10(p) at 1.3 for p < 0.05, at 7.3 for genome-wide significance, and a raw p axis runs the other way entirely" — and then `clearsThreshold` compared with `>=` unconditionally. On a raw-p chart that does not degrade the reading, it inverts it: the trace selects precisely the points that *failed* to reach significance, offers them under a rotor labelled Significant, and opens with "43 of 12,000 above the threshold" naming the 11,957 that are not. Silently, and with a docstring claiming otherwise.

I implemented the direction rather than softening the doc, since the doc was right about the charts. `ThresholdOptions.significanceDirection?: 'above' | 'below'` defaults to `'above'`, which is the transformed-axis convention and what every existing fixture wants. Two tests cover it, and one of them deliberately pins the wrong-half selection when a raw-p layer leaves the direction at its default — that is the trace doing what it was told, and the test says so, so nobody later reads the behaviour as a bug and "fixes" the default.

Green: `lint:fix`, `type-check`, `build`, `2433` unit + `73` component tests, and the five volcano e2e specs.

---

PR #848, opened 2026-08-12
feat(contour): read a level as a value and the spacing as a gradient

### LINKED ISSUE #802: feat: support contour and filled contour plots
## Description

Contour plots (`contour`, `contourf`, `geom_contour`, `sns.kdeplot` in 2D, Plotly `contour`) show a scalar field over a plane as iso-value curves. Unsupported today.

## Proposed mapping

The underlying grid is a `heat`. What makes a contour plot a different trace is that the **level** is a first-class object rather than a colour: the reader's questions are "how many levels are there", "how far apart are they here", "where does the 0.05 contour run".

Two navigation modes serve that:

1. **Grid mode** — walk the underlying lattice, as `Heatmap` already does.
2. **Level mode** — pick a level and walk along its curve, which behaves like a line trace.

Level mode is the one that makes the chart legible; grid mode is the cheap fallback. Contour spacing is the visual cue for gradient steepness, so announcing the distance to the neighbouring level carries the information the density of lines conveys visually.

## Difficulty

Medium-to-high. Grid mode alone is low, and shipping it first is a reasonable staging.

jooyoungseo:
Closes #802. **This completes tier B of #814.**

## Why not grid mode

The issue offers two modes and is explicit about their value: *"Level mode is the one that makes the chart legible; grid mode is the cheap fallback."*

It also notes grid mode alone would be reasonable staging — but the lattice underneath a contour **is** a heatmap, and MAIDR already reads one. Shipping grid mode alone would have been a heatmap with a different name, which is exactly the failure this roadmap keeps turning up: a chart read as its parent, sounding confident and saying nothing the parent did not already say.

A contour is structurally one line per level, so `LineTrace` gives navigation, braille and highlighting unchanged, and walking a row walks one iso-value curve. What a line layer cannot give is the two things the chart is read for.

## The level is a value, not a name

Read as a plain line layer the level is a series label at best: the curve says *where* it runs and never *what value* it runs at — the first thing anyone asks of a contour. It goes out on the field's own `z` axis, so the layer's `z` format applies to it, and is read from whichever point of the curve declares it, since the level belongs to the curve.

## Spacing is the gradient

How densely the curves are drawn is the visual cue for how steeply the field changes, and it is **the one thing a reader walking a single curve can never assemble** — a curve says nothing about its neighbours. Each point announces the distance to the nearest point on the adjacent level, taken to whichever neighbour is closer, since that is the gap an eye reads as the gradient there.

In the example fixture the same two curves sit `1` apart at one end and `6.4` apart at the other: a cliff and a plateau, indistinguishable from either curve alone.

### One thing I got wrong first

My test expected `8` at that second point — the distance to the *same-index* vertex on the neighbouring curve. The code returned `6.4`, which is the nearest vertex, and the code was right:

> Two curves of a field are not sampled at the same places, so index *n* of one is simply somewhere else than index *n* of the other.

That's the index-versus-geometry error this roadmap has now produced four times, and it caught me writing the test rather than the code. The fixture pins the geometric answer and **fails when the pairing is reverted**.

The measurement is to the nearest *sampled* point rather than to the nearest place on the segments between samples. That's an over-estimate bounded by the sampling interval, vanishing as the producer samples densely — which contour producers do, since the curve has to look smooth. The code says so rather than leaving it to be discovered.

## The description answers what a curve cannot

- how many levels, and what they are
- whether the step between them is **uniform** — which is what lets a reader treat spacing as a slope, since equal value between curves means the distance between them *is* the gradient. A varying step is named `varies` rather than averaged into a number that would license that reading falsely. **This fails when the averaging is restored.**
- where the levels run closest together, which on the page is where the lines crowd, and which no single-curve walk can find

## Braille

The line encoding. `docs/BRAILLE.md` says why the level **cannot** be encoded rather than merely isn't: it is constant along a curve, so encoding it would spend every cell of a row saying one number and the shape the row exists to convey would be gone. Spacing is a fact about the gap between two rows rather than about any cell, so it is announced — but a multiline display gives it back, since two fingers on adjacent rows feel exactly where the curves converge.

## Verification

- `npm test` — **2426 + 73 passed**, 175 suites
- `npx playwright test --project=chromium contour lineplot multiLineplot` — **57 passed**, the line suites included because this extends them

---


jooyoungseo:
All four addressed in 2eb3026. The second one is the one that mattered, and it was a better catch than "worth considering" suggests.

**1. The orphaned JSDoc — fixed.** `ContourPoint` now sits immediately after `LinePoint`, which is where this file already puts a type that extends another (`MosaicPoint` after `SegmentedPoint`, `SurvivalPoint` after `StepPoint`, `ForestPoint` after `ErrorBarPoint`), and `StepPoint` has its `@example` back.

Worth noting this is the second time in two PRs: #847 had `MANHATTAN`/`VOLCANO` inserted between the waterfall JSDoc and `WATERFALL`. Same edit, same file, same failure mode — appending near a landmark without checking what the landmark is attached to. Two occurrences is a habit, not a slip.

**2. "Number of lines / Points per line / Line names: Line 1, Line 2, Line 3" — you're right, and it undercut the PR's own argument.** The body says the level is a value rather than a name, and then the data table — the part of the description a reader actually consults — labelled every curve with an index into the order the producer happened to emit them in, in a column headed `Line`.

`seriesLabels` now returns `Number of levels / Points per level / Levels / Level`, and `groupNameAt` returns the level itself. The table reads `['X', 'Y', 'Level']` with rows `[0, 0, '0.1']`. That also removed a duplicate: my own `description` was pushing `Number of levels` and `Levels` alongside the inherited `Number of lines` and `Line names`, so the same three curves were counted twice in two vocabularies. Both pushes are gone; the inherited stats carry it, which is what `seriesLabels` is for.

Two edges the move opened, both now covered:
- A single-curve layer has no series list, so the parent says nothing about its level. `Level: 0.05` is pushed for that case, since otherwise the chart's only number would be missing from its own description.
- A curve declaring no level would have inherited `Line 2` — which in a column headed `Level` reads *as* a level. It falls back to the authored `z` if there is one and `Curve 2` otherwise.

Widening `authoredGroupNameAt` to `protected` is what makes that fallback possible without duplicating the parent's logic; `groupNameAt` calling itself would recurse.

**3. The unused `to` — put to use rather than dropped.** The gap is measured to whichever neighbour is nearer, so which side that is changes as the reader walks the curve: on the fixture, curve 0.2 is nearest the 0.1 curve at x=0 and nearest the 0.3 curve at x=10. A bare distance has the reader comparing two numbers taken from opposite sides without knowing it. It goes in the same clause — `Spacing is 1 to level 0.2` — rather than as a second aside, because terse drops aside labels and `, 1, 0.2` would arrive looking like two distances.

**4. Quadratic `steepestPoint` — memoised, and the bound written down.** Your estimate of the risk is right, and it is worth being precise about where it stops being fine: 3 levels × 500 points is ~1.5M distances and imperceptible; 20 levels × 2000 points is ~160M and would visibly hang the description dialog on the UI thread. Memoising does not change the asymptotics, but it does mean at most one such pass per trace lifetime rather than one per dialog open. A spatial index would fix it properly and is not worth the code until someone has that chart.

**One thing you did not flag, found while adding the tests you noted were missing.** `Level step` was computed over `levels.filter(Number.isFinite)`, so a layer where a middle curve declares no level — levels `[1, undefined, 3]` — reported `Level step: 2`, uniform. The step is 1, over two gaps. And a uniform step is precisely the licence a reader needs to treat spacing as a gradient, so that announcement grants it on a chart that does not support it. Now all-or-nothing.

**Tests.** Six new, in a `the description speaks of levels, not lines` block covering the vocabulary, the data table, both fallbacks, the single-curve level, and the step guard. Revert-checked: disabling the `groupNameAt` override fails four (including the pre-existing `counts and names the levels`, which now runs through the new path), and dropping the step guard fails one.

Green: `lint:fix`, `type-check`, `build`, `2433` + `73` unit and component tests, and `contour lineplot multiLineplot` e2e — 57 passed.

---

Issue #809, opened 2026-08-10
feat: support sunburst and icicle charts


jooyoungseo:
## Description

Sunburst (radial) and icicle (rectangular) charts show the same hierarchical part-of-whole data as a treemap, laid out differently. Plotly, ECharts, amCharts and Vega-Lite all produce them; MAIDR reads none.

## Relationship to the treemap issue

The data model is identical — a tree with a magnitude per node. Once hierarchical navigation exists for treemaps, these are close to free: the layout differs, the traversal does not.

They are filed separately because two things are genuinely their own:

- a sunburst is **radial**, so it inherits the angular-position work also needed by radar charts and by #780;
- an icicle reads as depth-ordered bands, which suggests defaulting the primary axis to depth rather than to sibling order.

## Suggested sequencing

Land treemap hierarchy first, then add these two as layouts over the same trace.

## Difficulty

Low-to-medium, conditional on the treemap work.

PR #849, opened 2026-08-12
feat(treemap): navigate the hierarchy as a hierarchy

### LINKED ISSUE #808: feat: support treemaps with hierarchical navigation
## Description

A treemap encodes a hierarchy as nested rectangles whose area is a magnitude. MAIDR cannot express it, because every trace today is a flat `(rows × cols)` grid — see `AbstractPlot.dimension` in `src/model/abstract.ts`.

## Why it is worth breaking that assumption

Hierarchy is one of the few chart structures where a non-visual reading can be **better** than the visual one. A sighted reader estimates areas and compares them badly; a screen-reader user given a tree can be told exact values and exact shares at every level. The tree-view metaphor is already familiar from file managers and outline views, so the interaction is not novel to the user even though it is novel to MAIDR.

## What has to be built

- A hierarchical position (path from root) alongside the existing row/column cursor.
- Down / up moves (enter a child, return to the parent) distinct from sibling moves.
- Announcements that carry depth and share: "Europe, 2.4 million, 31% of total, 5 children".
- A breadcrumb on entry so the reader knows where they are.

## Strategic value

This is the first of a family. Sunburst and icicle charts are the same model with a different layout, and the path/rotor work here is the groundwork for graph navigation (sankey, network diagrams).

## Difficulty

High. Worth staging: a depth-limited flat reading first, then true tree navigation.

jooyoungseo:
Closes #808. **First of tier C in #814**, and the groundwork the issue names for #809 sunburst.

## The assumption this breaks

Every trace in MAIDR is a flat `(rows × cols)` grid. A tree read as one becomes a list, and a list of thirty leaves says nothing about which three of them make up most of Europe — which is the whole content of the chart.

## The tree navigates as a tree, on the keys that already exist

Up returns to the parent, down enters the first child, left and right walk the siblings. That is the file-manager interaction, which is the one a reader already has — and the issue's own proposal, verbatim: *"Down / up moves (enter a child, return to the parent) distinct from sibling moves."*

**No new command, keybinding, ViewModel or movable class.** A node's address is `(depth, index-within-that-depth)`, and `MovableGraph` — which `LineTrace` already uses — navigates an arbitrary graph over such addresses. The whole hierarchy is a graph-building step in the model layer; nothing above the model changed.

Siblings link to each other and to nothing else, so the boundary holds in both places it can be crossed:

- **A single step.** Germany is depth 1 index 1; Japan is depth 1 index 2. On a flat grid, right from Germany lands in Asia and announces it as though it followed Germany. Here it is refused.
- **The jump to the extreme.** The same boundary, and the gap that shipped broken in hexbin and again in survival before being caught in review. `end` is the last *sibling*, not the last entry of the level.

A cousin is reached by going up, across and down, which is what the structure means.

## A chart a non-visual reading can do better

Comparing areas by eye is the standing criticism of the treemap. A reader walking this tree is told each node's **exact share of its parent** — the comparison the rectangles are drawn to support, and the one the eye estimates worst. The parent is named in the aside's label (`Share of Asia is 43.8%`) so terse mode, which drops aside labels, still says which whole the percentage is of.

## The hierarchy is a path, never a parent pointer

A path is acyclic by construction and cannot dangle — there is no id to point at a node that was never emitted. Every producer has one: a `d3.hierarchy` walk yields it directly and Plotly's `labels`/`parents` pair resolves to it. It is also the breadcrumb, so the reader is located out of the same field that placed them.

**Interior nodes need not be declared.** A treemap draws its leaves, so the example page declares ten countries and no continents at all; Asia, Africa and Europe exist because the leaves named them, and their totals are their children's. A *declared* total is kept even where it disagrees with that sum — a parent may carry mass no child accounts for, and correcting it would be inventing data.

## Pitch is scaled within the level

A tree's magnitudes fall by an order of magnitude per level. Scaled across the whole chart, France and Germany both sit on the floor of a range topped by Asia's 1577 and are indistinguishable — and the level a reader is walking is the one they are comparing. Per-depth scaling is what `LineTrace` already does per-row.

## Braille, and its real limit

One row per level, each scaled to itself. `docs/BRAILLE.md` says what the display **cannot** show rather than only what it does: a row is a sequence of cells, a level is a set of siblings belonging to different parents, and nothing in the dots marks where one parent's children end. A reader on the display alone would read a level as one run of magnitudes — the flat reading this trace exists to replace. The shape is the display's contribution; the structure is the announcement's.

## Verification

- `npm test` — **2485 + 73 passed**, 177 suites (28 new unit tests)
- `npx playwright test --project=chromium treemap` — **7 passed**

**Revert-checked, each mechanism separately:**

| Reverted | Fails |
|---|---|
| sibling-scoped `left`/`right` → depth neighbours | `right stops at the last sibling rather than entering a cousin` |
| sibling-scoped `start`/`end` → level extremes | both extreme tests |
| per-depth pitch → whole-tree pitch | `pitch is scaled within the level` |

One of those started out **not** discriminating, and it is the failure this roadmap has now produced five times: my first FORWARD-extreme test drove from Asia, whose last child *is* the last node of its level, so it returned the same answer with the boundary reverted. It now drives from Europe, and the test says in a comment why.

---


jooyoungseo:
This is right, and it is worse than a missing feature — it is a **keyboard trap**, which is a WCAG 2.1.2 failure in a library whose whole purpose is keyboard access. Fixed in eb2dc771.

I verified the chain rather than taking it on the description. `AbstractTrace.getExtremaTargets` returns `[]`, `GoToExtremaToggleCommand` gates only on `supportsExtremaNavigation()` → the flag, and `GoToExtrema.tsx` renders `null` on an empty target list. So `g` moved the reader into a scope with no dialog, no announcement, and arrow keys answering to a keymap that was no longer the tree's — and the exit is registered `showInHelp: false`. One detail beyond what you found: `AbstractTrace.navigateToExtrema` *throws* under exactly this condition, so the flag was arming a second failure behind the first.

The flag is now `false`. A tree has nothing to put in that dialog: every node is already one step from its parent and its siblings, and the largest leaf is in the description. Worth stating explicitly for anyone reading later, since the two are easy to conflate — **`supportsExtrema` does not gate `moveToExtreme`.** The Ctrl+arrow jump out to the branch root and in to the deepest descendant goes through the movable and is untouched; all four of its tests still pass.

**I had the same bug uncommitted in a second trace when your review arrived**, which is what decided the shape of the fix. Turning one flag off does nothing about the next trace, and this is invisible to every other kind of test: nothing crashes, nothing is announced, and the trace's own navigation tests all pass because they never enter the scope. So `test/model/extremaContract.test.ts` pins the invariant across `src/model/` — a trace that sets the flag must declare the method. Every existing trace already satisfies it; your list of nine implementers and ten opt-outs is exactly what it reproduces.

That test also caught itself being wrong, which is worth recording. The first version asked whether the file *contains* the string `getExtremaTargets`, and it passed with the flag deliberately turned back on — because the comment I had just written explaining why the flag is off names the method. It now requires a declaration (`public`/`override`/`protected` … `getExtremaTargets(`), and a second case pins that distinction using the treemap itself, which is the one file where mention and declaration disagree.

Green: `lint:fix`, `type-check`, `build`, **2488 + 73** unit and component tests across 178 suites, and the 7 treemap e2e specs.

---

PR #850, opened 2026-08-12
feat(choropleth): read a map as a map, not as a list of places

### LINKED ISSUE #812: feat: support choropleth maps
## Description

A choropleth shades geographic regions by a value. Google Charts `GeoChart`, Plotly `choropleth`, Highcharts Maps, Vega-Lite `geoshape` and `geopandas.plot()` all produce them; none are readable.

## Two very different halves

**The values are easy.** Region name plus magnitude is a bar chart. A minimal version — walk the regions in value or alphabetical order, hear each one — is cheap and already a large improvement over nothing.

**The geography is hard.** What a sighted reader gets from a map is spatial: clusters, gradients, borders where the value jumps. Reproducing that needs adjacency-aware movement — "the region north of here", "the neighbours of this region" — which requires a topology MAIDR has no way to receive today. It would have to come from the producer, since it cannot be recovered from rendered SVG paths.

## Suggested staging

1. Value-ordered navigation with region names. Low cost, immediately useful.
2. Declared adjacency in the layer schema, enabling directional movement and neighbour comparison.
3. Cluster detection ("this region and its four neighbours are all in the top decile"), which is the actual visual takeaway.

## Difficulty

Stage 1 low; stages 2 and 3 high.

jooyoungseo:
Closes #812. Tier C of #814.

## All three stages, not just the cheap one

The issue stages this deliberately, and is candid that stage 1 is the easy half: *"Region name plus magnitude is a bar chart."* Shipping only that would be the failure this roadmap keeps turning up — a chart read as its parent, sounding confident and saying nothing the parent did not already say. All three stages are here.

## Stage 1 → the arrows move across the map, out of centroids alone

Regions are banded by latitude and ordered west to east inside each band, so **up is north, down is south, left is west, right is east**. No adjacency needed for this — a centroid is enough.

Two details that are easy to get backwards, and are pinned by tests:

- **Band zero is the southern one.** `MovableGrid` moves `UPWARD` by *incrementing* the row, so north-first would make up go south. The braille service already reverses row order on a physical multi-line display for the same reason, so the map renders the right way up there too.
- **Bands hold equal counts, not equal spans of latitude.** Equal spans leave an empty band wherever the map has a gap, and an empty row is a position the cursor reaches and reads nothing from. The *direction* of every move is still exact, since the ordering is by latitude and longitude throughout; only a band's thickness varies. `docs/BRAILLE.md` says so rather than leaving it to be discovered.

A centroid is a **longitude and a latitude in degrees**, never a projected coordinate. Every producer has the pair (`d3.geoCentroid` returns exactly it), and asking for degrees removes the one thing MAIDR could not otherwise resolve: whether a rising `y` means north or south. A layer declaring none keeps its declared order in a single band — the region list, which is the reading that data supports.

## Stage 2 → borders, declared because they cannot be recovered

The issue is right that adjacency has to come from the producer, and understates it slightly: it is not recoverable from centroids either. Two regions can have near centroids and no shared edge, and a long region can border one whose centroid is far away. So `neighbors` is declared, and a layer that declares none is told nothing about borders rather than something guessed.

With it: every region announces how it sits against the ones around it — `Neighbours is 2, all lower` is a **local maximum**, which on the page is a hotspot and is the shape a reader is hunting for. A rotor unit walks the bordering regions, ordered as the grid is rather than in whatever order the producer listed them.

## Stage 3 → the cluster, which is the actual finding

The description names the **sharpest borders** — where the shading jumps hardest between two regions that touch, which an eye finds instantly and a value-ordered walk never reports, because the two sides are far apart in that ordering — and the **largest cluster of high regions**.

That last one is the issue's own stated goal, and the distinction is the whole point: not that some regions are high, but that the high ones *touch*. A ranked list gives the first fact and can never give the second. "High" is the top fifth by value, a cut of the data rather than a threshold on it, since a fixed number would mean something different on every map. A component of one is not reported — that is already the `Highest` stat, and calling one region a cluster asserts a geography that is not there.

## Verification

- `npm test` — **2476 + 73 passed**, 177 suites (19 new unit tests)
- `npx playwright test --project=chromium choropleth` — **7 passed**

**Revert-checked:** flipping the banding to north-first fails 8 tests; making the cluster walk ignore adjacency fails its test.

That second one **did not fail at first**, and the reason is worth recording. My original fixture had three regions with the outer two high — but the top-fifth cut on a three-region map admits only one, so the cluster was refused for being a single region, one step before adjacency was ever consulted. The test passed with the bug in place. It is now six regions in a chain with the two highest at opposite ends, and it asserts `Highest` as well, so the fixture proves both regions cleared the cut before claiming adjacency is what refused them.

## Two things found on the way

- **`supportsExtrema` is `false`, deliberately.** The review of #849 found that turning it on without overriding `getExtremaTargets` is a keyboard trap: the toggle gates only on the flag, the dialog is gated on a non-empty target list, and the reader lands in a scope with no dialog and arrow keys that no longer answer. I had written the same bug here before that review arrived; it is fixed in both, and #849 adds a repo-wide test pinning the invariant. Note `supportsExtrema` does **not** gate `moveToExtreme` — the Ctrl+arrow jump to the end of a band still works.
- **A drive-by doc repair, in its own commit** (605558d). The contour type went in between the diverging bar's JSDoc and `DIVERGING = 'diverging_bar'`, orphaning it — the third occurrence of this exact edit in this file after `WATERFALL` and `StepPoint`, both caught in review, and this one is on `main` today.

---


jooyoungseo:
The rotor bug is real, your trace of it is exact, and it broke the feature the PR body leads with. Fixed in d8ff109, along with both minor points.

**I reproduced it before touching the code.** A test that walks right from Nevada and collects where it lands ran past its own eight-step guard instead of terminating — the oscillation, not a slow walk. Your worked example is precisely what happens: `here` is `-1` on every call because a region is never its own neighbour, so right takes `around[0]` and left takes `around[length - 1]`, and since the list is rebuilt from the region it just moved to, the next press comes straight back. Two of Nevada's four borders, forever; Utah and Idaho unreachable.

Your diagnosis of *why* the other rotors get away with it is the part that decided the fix. `survival`, `step` and `volcano` derive their candidates from the cursor because they search a **fixed axis** and the cursor has an index on it — `index > this.col` is meaningful. A border walk has no axis. It belongs to the region it started from, so it is now anchored there and carries its own index, and it notices being abandoned by checking whether the cursor is still where its last step left it; if not, the reader moved with the arrows and it re-anchors on wherever they now are. Three tests: all four of Nevada's borders in order, walking back the way you came, and re-anchoring after an arrow move. Reverting the anchor fails the first two.

It stops at the ends rather than cycling, matching every other rotor in the model layer. Cycling is arguably nicer for a set with no natural start, but not at the cost of this mode behaving unlike the others.

**Region-name collisions** — you are right that last-wins was the wrong choice, and it is worse than "corrupting the earlier region": every border the earlier region declares would resolve to the later one, so the reader is walked to a region that does not border it and told it does. First-wins now, which is not a repair — a name two regions share cannot be an adjacency key at all — but it leaves one of the two correct and the other merely unreachable by name, rather than making the first actively wrong. Tested with `5` against `'5'`.

**Partial centroids** — worth more than a doc mention, because the failure is not "some regions are unplaced". `arrange` would lay out the placed ones in real compass directions and leave the rest wherever the declared order dropped them, so the chart reads as a map throughout and is only a map in part. It falls back for all of them, and that is now tested rather than only implied by the all-or-nothing guard.

Green: `lint:fix`, `type-check`, `build`, **2481 + 73** unit and component tests, and the 7 choropleth e2e specs.

One note on the coverage gap you identified: the e2e test asserting the word "Neighbours" appears is exactly the kind of test that confirms a feature is wired up and says nothing about whether it works. The unit tests now drive the mode rather than observe it.

---

jooyoungseo:
Taking the non-blocking one too, in 00ae4de — it is one line, and the failure mode is worse than "an edge case of an edge case" suggests.

You framed it as the walk using the wrong neighbour *list*. What that means at the announcement is that a reader standing on the second `5` presses right and is walked to a region that does not border them, and is told it does. The ambiguity is the author's; reading out a confident wrong answer is ours. A fresh walk now anchors on the region object already in hand rather than round-tripping through `placeOf`; the name is still what a *continuing* walk is keyed by, since that is what survives the cursor moving. Reverting it fails the new test.

Two things in your review I want to confirm rather than let pass, because both are cases where the right answer was to write nothing:

**The partial-centroid path needed no code.** You're right that `arrange()`'s `every(...)` guard already handled it. I checked before adding anything, and the test just closes the coverage gap — adding a second guard that duplicated a correct one would have been the easy mistake.

**`notifyRotorBounds` deliberately does not update `borderWalk`.** You traced that a bounds-hit leaves an in-progress walk intact; that is intended, so running off the end and coming back continues from where you were rather than restarting.

On the e2e gap: agreed it belongs at the model layer and not in the spec. The e2e assertion that the word "Neighbours" appears is worth keeping for what it does check — that the aside reaches the announcement at all — but it is exactly the kind of test that confirms a feature is *wired up* while saying nothing about whether it *works*, which is how the original bug got through. The unit tests now drive the mode across presses instead of observing one.

Green: `lint:fix`, `type-check`, `build`, **2514 + 73** unit and component tests, and the 7 choropleth e2e specs.

---

jooyoungseo:
Correct, and it is a regression I introduced — the code I replaced wrapped a `-1` to the end of the list, which happened to be right for exactly this case while being wrong for the one the rewrite fixed. Fixed in d15710a3, seeded from whichever end the reader is entering from, as you suggested.

**This is in two other traces as well.** The sankey and network rotors were written from the same pattern, and both have the identical `-1` sentinel. Your one comment here found three bugs:

| PR | Symptom of a first `left` press |
|---|---|
| #850 choropleth | "no bordering regions" on Nevada, which has four |
| #852 sankey | "no outgoing flows" on Coal, which has three |
| #853 network | "no links" on Ada, who has four |

All three are fixed and pushed, each with two tests that enter leftwards. I reproduced each before touching it rather than assuming the diagnosis carried across.

**The coverage-gap pattern you named is the more useful finding.** Every rotor test I have written across three traces presses `right` first — nine tests, one direction. That is not nine independent checks of the mode; it is one check run nine times. A mode with two entry directions needs both exercised at *entry*, not just mid-walk, because entry is the only moment with no position to step from and therefore the only moment the seed matters.

Two smaller notes on your review:

- The **doc comma** is fixed. It was a genuine misreading, not just punctuation: without it, "places it loses" parses as a relative clause and the sentence has no main verb.
- Your **hand-trace was the whole diagnosis** — `index = -1`, `left` → `step = -2` → `around[-2] === undefined`. You flagged that you could not run the suite; you did not need to. I ran it, and the failure is exactly what you predicted, on the exact fixture you named.

Green on this branch: `lint:fix`, `type-check`, `build`, **2529 + 73** unit and component tests across 180 suites, and the 7 choropleth e2e specs.

---

PR #851, opened 2026-08-12
feat(sunburst): read a sunburst and an icicle as the tree they are

### LINKED ISSUE #809: feat: support sunburst and icicle charts

jooyoungseo:
Closes #809. Tier C of #814, on top of #808 (merged as #849), which is the sequencing the issue asks for: *"Land treemap hierarchy first, then add these two as layouts over the same trace."*

## One trace, three layouts

The issue is right that the traversal does not differ, and that is the design: `TreemapTrace` reads all three, and **the keys are identical everywhere**. A reader who meets a treemap in one paper and a sunburst in the next should not have to learn two sets of them. An e2e test drives the same key sequence over all three pages and asserts the same landings.

Registration is the only per-layout cost — enum member, factory case, chart-type label, orientation, braille encoder — and each announces itself as the chart the author actually drew.

## What is genuinely the sunburst's own: the rings are angular

The issue names this: *"a sunburst is radial, so it inherits the angular-position work."*

A node's children divide its arc in proportion to their values, so the pan follows the node **around the dial** the way `PieTrace`'s does — `cols: 2` with `x = (sin θ + 1) / 2` lands the pan on `sin θ` exactly. Sweeping a ring goes out to one side and comes back, which is what distinguishes a circle from a row; a rectangular layout keeps panning by index, where the first node of a row is hard left.

The angle is **derived from the tree, not declared**. It is the layout, computed the way the chart is drawn: the top level divides the circle, each node's children divide its own arc. One consequence worth stating — a parent carrying mass no child accounts for leaves the remainder of its arc unclaimed, exactly as the drawn chart does, because the division is by the node's own value rather than by its children's total.

## What the icicle asks for: the band — but not on the arrows

The issue suggests *"defaulting the primary axis to depth rather than to sibling order."* **I did not do that, and the reasoning is worth stating rather than burying.**

An icicle is drawn as a full row of blocks per depth, so "everything at this level, in layout order" is a question its layout puts directly, and it deserves an answer. But making it the *arrow* axis would mean one data model with two different arrow semantics, selected by whichever layout the producer happened to pick — a reader moving between a sunburst and an icicle of the same data would find left and right meaning different things for no reason they can see. It would also re-open the cousin problem #849 was built to close: on a depth axis, stepping right from the last child of one parent lands in another subtree unannounced.

So the band goes on a **rotor unit** — an explicit mode, discoverable, and offered on every layout since a treemap reader may want it too. The arrows stay the tree. If that call is wrong I would rather it be argued than silently taken, so it is in the code comment, the example page, and here.

## Verification

- `npm test` — **2499 + 73 passed**, 178 suites (9 new unit tests)
- `npx playwright test --project=chromium hierarchyLayouts treemap` — **17 passed**

**Revert-checked:** making the sunburst pan rectangularly fails 2 tests; confining the level rotor to siblings fails its test.

The angle test pins the arithmetic rather than just the shape — Europe is 150 of 1727, so its mid-angle is `π × 150/1727` and the pan is `sin` of it — and a second test walks into France to check a child's arc opens inside its parent's rather than being laid out against the whole circle.

---


jooyoungseo:
Correct, and fixed in e22ae30 — but this is the **fourth** time this exact edit has landed in this file, after `WATERFALL` (#847), `StepPoint` (#848) and `DIVERGING` (fixed in #850). Four occurrences is not a slip, so I wrote the check instead of just the fix.

`test/type/grammarDocs.test.ts` fails on two JSDoc blocks separated by nothing but whitespace. **It immediately found two more that nobody had caught**, and between them they had stranded five members' documentation:

| Stranded doc | Was sitting above | Now above |
|---|---|---|
| `Maidr` — the root data structure, with its whole `@example` | `NavigateCallback`'s block | `export interface Maidr` |
| `DumbbellPoint` | `GanttPoint`'s block | `export interface DumbbellPoint` |
| `GAUGE` | `FOREST`'s block | `GAUGE = 'gauge'` |
| `GANTT` | `FOREST`'s block | `GANTT = 'gantt'` |
| `MOSAIC` | `MANHATTAN`'s block | `MOSAIC = 'mosaic'` |

Three of those were stacked *three deep* above `FOREST`, so `GAUGE` and `GANTT` had both been undocumented in the generated output for some time.

Your diagnosis of why this survives review is exactly right and worth stating as the general rule: **the new member always reads correctly.** Its own comment is the one immediately above it. The damage is entirely to the member that was already there, several lines further down, which nobody is looking at because it is not what changed. Nothing else catches it either — the file compiles, `npm test` passes, and the loss is to documentation that only surfaces on hover and in `npm run docs`.

I also got the repair wrong on the first attempt and had to reset the file: with three blocks stacked, I mapped two of them to the wrong members. The second pass keys each block on its own first line and prints where every one landed, and I verified all six members afterwards.

One note on scope: the `DIVERGING` fix is also in #850, so whichever of these merges second will conflict on that hunk — either side is correct.

Green: `lint:fix`, `type-check`, `build`, **2501 + 73** unit and component tests across 179 suites.

---

PR #852, opened 2026-08-12
feat(sankey): follow the ribbon, which is what the chart is drawn for

### LINKED ISSUE #810: feat: support sankey, alluvial and chord diagrams
## Description

Sankey diagrams, alluvial diagrams and chord diagrams all encode weighted flow between nodes. None can be authored, and unlike most gaps there is no partial reading available — the chart is a graph, and MAIDR's model is a grid.

## Why this is the highest-value item in the backlog

These charts exist precisely to show routing and proportion at once: where a quantity comes from, where it splits, how much is lost at each hop. That is a question about *topology*, and topology is what alt text is worst at. A sighted reader traces a ribbon in a second; there is currently no equivalent action available at all.

## What has to be built

A graph cursor, distinct from both the grid cursor and the tree cursor:

- position = a node, or an edge;
- moves = "next outgoing edge", "next incoming edge", "follow this edge to its other end", "back";
- announcements that give flow magnitude and share: "Coal → Electricity, 34 PJ, 61% of Coal's output";
- entry summary: node count, stage count, total flow.

## Sequencing

The path-tracking and breadcrumb work in the treemap issue is a prerequisite in spirit if not in code — doing hierarchy first de-risks this considerably.

## Difficulty

High. Highest cost and highest accessibility payoff in this set.

jooyoungseo:
Closes #810. Tier C of #814, and the item that issue calls *"the highest cost and highest accessibility payoff in this set."*

## The gap this closes is total, not partial

Most items on this roadmap were charts MAIDR read *badly*. This one it could not read at all: the chart is a graph and the model was a grid. A sighted reader traces a ribbon in a second; there was no equivalent action available.

## Following a ribbon is the primary move — on the keys that already exist

A sankey is drawn as columns of nodes with flow running left to right, so the arrows are given the chart's own geometry:

| key | move |
|---|---|
| **right** | follow the largest flow **out** of this node, downstream |
| **left** | follow the largest flow **in**, back upstream |
| **up / down** | walk the other nodes stacked in this column |

That is simultaneously the drawing *and* MAIDR's ordinary reading of its arrows — left and right move along, up and down move between — so nothing has to be relearned. The **largest** flow is the default because it is the ribbon an eye follows: the widest band leaving a node is what the chart draws attention to. Every other flow is one rotor away (`Outgoing`, `Incoming`).

This needed no new command, keybinding, ViewModel or movable class. `MovableGraph` navigates an arbitrary graph over `(row, col)` addresses, and a step along a ribbon simply changes both coordinates — which is exactly what a grid cannot express and what the treemap already proved out.

## The edge is announced, not just the node it lands on

A move that said only "Electricity, 46" would withhold the entire content of the step. So the trace remembers what it arrived by:

> **Coal to Electricity, 34, 60.7% of Coal**

which is the announcement the issue asks for, verbatim in shape. It is recorded *before* the move, because `moveOnce` notifies its observers during the call and the announcement is assembled from `text` at that moment. A move within the column names no ribbon; a refused move clears it rather than leaving the last one announced; and a jump to the end of a route names none, because it followed several.

## Stages are derived, and a cycle is answered honestly

A node's column is its longest distance from a source, which is how these layouts are built. A **chord diagram is cyclic** and has no such ordering, so when the graph does not sort, every node goes in one stage — the ribbons still follow, and nothing is claimed about an ordering the data does not have. Every cycle-walking routine (`end`/`start`, the dominant path) is guarded against revisiting, so a chord diagram terminates.

Nodes are derived from the flows, since a flow names both of its ends — the treemap's reasoning about paths, applied to a graph.

## Braille, and its real limit

One row per stage, scaled per stage. `docs/BRAILLE.md` states plainly that **the ribbons have no representation at all** — a cell is a node's throughput, and nothing in the dots says where that quantity came from or how it splits. The topology is carried entirely by the announcement. The state is transposed on the way out, because the cursor is addressed as `[indexWithinStage][stage]` to put the arrows on the chart's geometry, while a braille line has to be a stage.

## Verification

- `npm test` — **2538 + 73 passed**, 180 suites (34 new unit tests)
- `npx playwright test --project=chromium sankey` — **8 passed**

**Revert-checked:** taking the first authored edge instead of the largest fails 6 tests; unanchoring the rotor fails 3; removing the cycle check fails 1.

Three notes on how those were arrived at, because two of them started out not discriminating:

- The **largest-edge** revert passed at first — my fixture declared Coal's flows in descending order, so sorted and declared order coincided. The fixture now declares them **smallest first**, and says why.
- The **cycle-check** revert passed at first too. On a pure circle, "no stages assigned" happens to equal "one stage", so the fallback and its absence agree. The test is now a *partly* cyclic graph — a real source feeding a cycle — which is the case where the missing check produces a half-layering that reads as meaningful.
- The `Math.max` in the stage assignment **cannot** be distinguished by any test, because the in-degree gate makes writes arrive in non-decreasing depth order. Rather than write a test that would pass either way, the code says so and explains why the line is kept.

## One thing worth reporting

`test/type/grammarDocs.test.ts` — added in #851 after four orphaned-JSDoc incidents — **failed on this branch**, catching me putting `CHORD` between the contour doc block and `CONTOUR`. Fifth occurrence of that edit, first one caught by a machine instead of a reviewer.

---


jooyoungseo:
Real bug, correctly diagnosed, and fixed in 498c645. Both findings taken.

**The highlight mismap.** Your trace is exact, including the part I want to underline: `buildNodes` sorts every edge list by value *before* `mapToSvgElements` runs, so re-deriving a position from those lists pairs a node with whichever ribbon sits at that rank. On the example page in this very PR, arriving on Coal announces `Coal to Electricity, 34` and lights up the **Coal-to-Losses** ribbon.

There is a sharp irony in how it got there: the fixture declares Coal's flows smallest-first *because I reordered it in this PR* to make the largest-edge sort observable. That change fixed one blind test and created this bug, in the same file, in the same session.

The declared position now travels on the edge itself (`Edge.at`) rather than being recovered from a list that no longer has that order. `test/model/flowHighlight.test.ts` is new and jsdom-based; reverting the fix makes it fail with `Expected "Coal-Electricity", Received "Coal-Losses"` — literally your reported symptom.

Your point about **why this needed its own test** is the one worth recording. Audio, text and braille all read from `this.nodes`/`this.grid`, which the bug does not touch, so all three were correct and none of them could have caught it. A highlight bug is invisible to every non-visual assertion in the suite — which is a peculiar failure mode for this project, since the modality that broke is the one our own users do not use, while the ones they do use kept saying the right thing. The e2e spec asserted announcements and braille and would never have found it.

**The doc string.** Fixed — it quoted `61% of Coal's output` where the code produces `60.7% of Coal`. That is the same class of defect as the one you caught on #847, where a doc promised a behaviour the code did not have, so it is worth more than a typo fix: I have stopped writing example strings into these class docs by hand and now copy them from a passing assertion.

Two things in your review I'll confirm rather than let stand as open questions:

- **`Math.max` in `assignStages`** — you reasoned independently to the same conclusion as the code comment, that the in-degree gate makes it provably redundant. That is why no test claims it; I would rather say so than write a test that passes either way.
- **`Svg.selectAllElements` usage** — matches the existing pattern, and the count-mismatch withdrawal is deliberate: a selector list that has slipped by one highlights the wrong ribbon for the whole session, which is exactly the bug you just found, arriving by a different route.

Green: `lint:fix`, `type-check`, `build`, **2540 + 73** unit and component tests across 181 suites, and the 8 sankey e2e specs.

---

PR #853, opened 2026-08-12
feat(network): reach every node, which following links cannot

### LINKED ISSUE #811: feat: support network (node-link) diagrams
## Description

Force-directed and other node-link graphs (Highcharts `networkgraph`, ECharts `graph`, D3 force layouts, `packedbubble`) are unsupported.

## Relationship to sankey

Same requirement: a graph cursor that moves between nodes along edges. A sankey is a directed, layered, weighted graph; a general network drops the layering and may be undirected. If the graph navigation built for sankey is written against nodes and edges rather than against stages, this becomes largely a matter of relaxing constraints.

## What is specific here

- Node **degree** and neighbour count are primary information and belong in the announcement.
- Layout position is emergent and carries no meaning, so it must not be announced as if it did — a real risk given that highlighting works in SVG coordinates.
- Useful orderings (by degree, by cluster) matter more than any spatial order, and fit the rotor.
- Connected components should be discoverable; a reader must not silently get stuck in one.

## Difficulty

High, substantially lower if sankey lands first.

jooyoungseo:
Closes #811. **This completes tier C of #814** — the last of the five.

The issue frames this as relaxing the sankey's constraints, and that is what it is: undirected links, and no stages, because a force layout has no columns to lay anything out along. It addresses all four of the issue's "what is specific here" points directly.

## Layout position carries no meaning, so it is not in the schema at all

The issue flags this as *"a real risk given that highlighting works in SVG coordinates."* The strongest available answer was to make the temptation impossible: **`NetworkPoint` has no `x` or `y`.** Where a force-directed node lands is a fact about the solver's seed rather than about the data, and a field that existed would eventually be announced. Highlighting goes through selectors, as everywhere else. An e2e test asserts the declared keys are exactly `source` and `target`.

## A reader must never be stuck in one group

This is the requirement that decided the whole navigation. Following links only ever reaches the component you are in — so on the example chart, an arrow that followed links from Ada would reach five people and leave the other four unreachable *and unannounced*. So the arrows do not follow links:

| key | move |
|---|---|
| **left / right** | walk this group, most connected first |
| **up / down** | move to another group entirely |

Between them they reach every node, and a test walks the chart with arrows alone and asserts it collects all eight names. Every announcement also carries `Group 1 of 3, 5 nodes`, so a reader in one group knows the chart is larger than what they can walk.

Following a link is the **rotor** — which is also where it belongs on an undirected graph. A node's neighbours have equal status, so there is no "the" edge for an arrow to take, unlike a sankey where the widest ribbon is the one an eye follows.

## Degree is primary information, so it is in every announcement

`Links is 4, to Grace, Alan, Edsger and 1 more`. Nodes are ordered by degree within their group, so a reader arrives on the hub — which is usually the whole reason the chart was drawn. An isolated node says `Links is none` rather than staying silent: it is drawn sitting apart, and that is what the picture shows.

Links are undirected, so a link is a fact about both ends. One declared **twice**, as several libraries emit an undirected edge, counts once — counting it twice would report a degree the picture does not show. A self-link is not a connection to anyone.

## Useful orderings over spatial ones

Degree order is the arrow order; group order is the row order, largest group first. Both are real structure. The description names the hubs, the group sizes — *"structure a reader following links can never discover, because links do not cross between groups, which is what makes them groups"* — and the unconnected nodes.

## Braille

One row per group, scaled across the **whole chart** rather than per group. That is the opposite of the treemap's choice and deliberate: degree is comparable everywhere, and a per-group scale would make a hub of three feel like a hub of thirty. `docs/BRAILLE.md` states that the links have no representation at all, which is most of the chart, and that the display gives the degree distribution while the announcement gives the topology.

## Verification

- `npm test` — **2526 + 73 passed**, 180 suites (24 new unit tests)
- `npx playwright test --project=chromium network` — **7 passed**

**Revert-checked:** removing the cross-group arrows fails 5 tests including "every node is reachable with the arrows alone"; making links one-way fails 6; unanchoring the rotor fails 2.

One arithmetic error of mine on the way: I wrote the fixture believing Ada had three collaborators when the links give her four. The code was right; six expectations and a comment were wrong, and they are fixed.

---


jooyoungseo:
You are right, and this is the more embarrassing of the two options you offered: not "highlighting is out of scope here", but a stub I left behind with a docstring describing what it was going to do. `flat` was resolved, length-checked, and discarded. Fixed in 6f05d23.

**Wiring it up properly, not documenting it away.** You noted that "no visual highlight by design" is arguably defensible given that layout position carries no meaning — I don't think it is. The links are drawn, they are real structure, and a sighted collaborator or a magnifier user watching the cursor should see where it is. What position carries no meaning rules out is *announcing* a location, not lighting one up.

A node now highlights the line drawn for its **most connected neighbour**, which is also the rotor's first step from there — so the highlight and the next keystroke never disagree about which neighbour is nearest. An isolated node keeps the placeholder: it has no line to point at, and somebody else's would be worse than none.

You also spotted the missing piece exactly — `NetworkNode.links` held neighbour node indices with no way back to the declared link. It now carries `linkAt` alongside, and the reason is the same one from #852: `links` is sorted by degree while a selector list is in declared order, so pairing by rank highlights whichever line happens to sit at that position. `test/model/networkHighlight.test.ts` is deliberately built so those two orders disagree — Ada's first link is authored to Edsger but her most connected neighbour is Grace — and it fails **both** ways: reinstating the stub, and pairing by rank (`Expected "Ada-Grace", Received "Ada-Alan"`).

**Why this slipped past `npm test` and the e2e suite, which is the part worth recording.** Audio, text and braille all read from `this.nodes`/`this.grid`, which the bug never touched. All three were correct. On this project that produces an odd shape of failure: the modality that broke is the one our users do not use, while the three they do use kept saying exactly the right thing — so nothing in an accessibility-focused suite could notice. That is now the second highlight-only bug in two PRs, so both traces have a jsdom highlight test rather than relying on the other three modalities.

Both minor points taken: each node stores its own `id` instead of an `indexOf` scan per rotor step, and the factory case is back in alphabetical order.

Green: `lint:fix`, `type-check`, `build`, **2529 + 73** unit and component tests across 181 suites, and the 7 network e2e specs.

---

PR #855, opened 2026-08-12
fix(text): render a section's case as authored wherever it is announced

### LINKED ISSUE #830: fix(text): a section reads in one case in verbose mode and another in terse
## What

`TextService` renders `state.section` two different ways depending on the text mode:

- `formatVerboseTraceText` (`src/service/text.ts:511`) — `state.section.toLowerCase()`
- `formatTerseTraceText` (`src/service/text.ts:624`) — `state.section`, verbatim

So the same point announces as `minimum Value is 3.2` in verbose and `Minimum 3.2` in terse. Toggling `T` changes the capitalisation of a label that came from neither the user nor the data.

## Why it has not surfaced

The three traces that reach the lower-casing branch (`section` set, `z` unset) all carry constants:

| Trace | Source | Case |
|---|---|---|
| `BoxTrace` | `BoxplotSection` | **Capitalised** — `'Minimum'`, `'Maximum'`, `'Mean'` |
| `ErrorBarTrace` | `SECTION_LABEL` | lower |
| `WaterfallTrace` | `KIND_LABEL` | lower |

Only the box plot actually differs between modes, and both readings are intelligible, so nothing has ever failed. `text.ts:434-437` already documents the fragility from the other side — that the error bar and waterfall labels "read correctly on this branch, but only because their section labels are already lower case".

## Why it is worth fixing now

`DumbbellTrace` (#829) is the first trace to put a **producer-supplied** string in `section` — the chart's own names for the two ends of each pair, `1990`/`2020` or `Control`/`Treatment`. A producer's `Control` is announced as `control` in verbose and `Control` in terse, and neither the trace nor the producer chose that.

The trace deliberately does **not** work around it by lower-casing in the model, because doing so would destroy a distinction the data carries (`Q1 2024`) to compensate for something the formatting layer chose to drop. There is a comment at `src/model/dumbbell.ts` pointing here.

## Options

1. **Drop the `.toLowerCase()`.** One line, and the two modes agree. It changes every box plot's verbose announcement from `minimum Value is 3.2` to `Minimum Value is 3.2` — arguably an improvement, but it is a visible change to an existing trace and wants a product call plus an e2e update.
2. **Lower-case in terse too.** The two agree, and every producer-supplied label loses its case in both modes. Worse for the dumbbell, no better for the box plot.
3. **Have the trace declare it.** A `sectionCase: 'preserve' | 'lower'` on `TextState`, defaulted so nothing existing changes. Most flexible, most surface.

(1) is the smallest and probably right; the reason it is not folded into #829 is that it changes a chart type that PR does not touch.

The separate outlier branch at `text.ts:499-503` lower-cases too, and would want the same treatment for consistency.

jooyoungseo:
Closes #830. Takes **option 1** — dropping the `.toLowerCase()` — which the issue identified as the smallest and probably right choice.

## The defect

`formatVerboseTraceText` lower-cased `state.section`; `formatTerseTraceText` rendered it verbatim. The same point read `minimum Value is 3.2` in one mode and `Minimum 3.2` in the other, so toggling `T` changed the capitalisation of a label that came from neither the user nor the data.

## Why it is worth changing rather than documenting

When the issue was filed, one trace put a producer-supplied string in `section`. **There are now two**: a dumbbell's end names and a ridgeline's group names are both authored strings.

- A chart's `Control` was announced as `control` in verbose and `Control` in terse.
- `Q1 2024` is not a preference about capitalisation — lower-casing it makes a *different string*.

Neither trace works around it in the model, and shouldn't: doing so would destroy a distinction the data carries in order to compensate for something the formatting layer chose to drop. There is a comment in `src/model/dumbbell.ts` pointing at this issue.

## Scope

Nine call sites in two files, all of the trace's `section`:

| File | Site | Was | Now |
|---|---|---|---|
| `src/service/text.ts` | verbose, section-before-label | lower-cased | verbatim |
| `src/service/text.ts` | verbose, outlier branch (×2) | lower-cased | verbatim |
| `src/service/text.ts` | terse, outlier branch (×2) | lower-cased | verbatim |
| `src/command/describe.ts` | box position, verbose + terse | lower-cased | verbatim |
| `src/command/describe.ts` | candlestick position, verbose + terse | lower-cased | verbatim (no-op today) |

The outlier branch lower-cased on *both* sides, so those two agreed with each other while disagreeing with every other section. Including them is what the issue asked for under "the same treatment for consistency."

**`describe.ts` was added after review flagged it, and it is not optional.** Before this PR, the box plot's verbose per-point reading also lower-cased, so the position key agreed with it. Fixing only `text.ts` would have moved the disagreement rather than closed it — and moved it somewhere harder to notice, since it would then be two different *keys* saying different things rather than two modes of one key. The two candlestick sites read identically either way (`CANDLESTICK_SECTIONS` is already lower case) and go with the other two anyway: the trace owns the case of its own label, and one of four sites deciding differently is how these drifted apart.

`state.z.value.toLowerCase()` is left alone — that is the candlestick trend, a different field with a different argument.

## The visible change, stated plainly

A box plot's **verbose** reading changes:

- `minimum Value is 3.2` → `Minimum Value is 3.2`
- `upper outlier(s) for Value is [...]` → `Upper outlier(s) for Value is [...]`

**Terse already said both of those that way**, so this is the two modes agreeing rather than a new wording. The position key changes in both modes — `Position is 2 of 2 in minimum` → `... in Minimum` — which is it agreeing with the reading the reader just heard. Nothing else moves: error bar, waterfall and survival author their sections in lower case already, and a test pins that.

## What I found running it

**No test asserted the old behaviour in either direction.** `npm test` and the full e2e suite — including both box plot specs — passed with the change and with it reverted. Nothing in `e2e_tests/` asserts either wording, for the per-point reading or the position key. The issue predicted this "wants a product call plus an e2e update"; the e2e update turns out not to exist, because the rendering was never pinned at all. That is precisely how the two modes drifted apart, and it means the change is safe in the narrow sense and unguarded in the important one.

So two test files cover it:

- **`test/service/sectionCase.test.ts`** (new) asserts the **rendered sentence** through `TextService` in both modes: the box section, the outlier section, a producer-supplied name, a name whose case carries meaning (`Q1 2024`), and one already-lower-case section to prove nothing else moved. Reverting the fix fails four of the five.
- **`test/command/announce-position.test.ts`** gains four cases built on **real traces** rather than hand-written state — a fixture that writes `section: 'Minimum'` itself would decide the thing under assertion. Reverting fails three of the four; the candlestick case passes either way on purpose, since it is the only honest way to cover a site whose behaviour genuinely does not change.

## Verification

- `npm test` — **2609 + 73 passed**, 185 suites
- `npx playwright test --project=chromium boxplotVertical boxplotHorizontal dumbbell ridgeline errorbar waterfall survival forest` — **91 passed**, every trace that sets a `section`
- `npx playwright test --project=chromium boxplotVertical boxplotHorizontal candlestick` — **50 passed**, after the `describe.ts` change

### DIFF
--- a/src/command/describe.ts
+++ b/src/command/describe.ts
@@ -800,6 +800,11 @@ export class AnnouncePositionCommand extends AnnounceCommand {
   /**
    * Announces position for boxplots with section information.
    * Uses braille state which normalizes box position regardless of orientation.
+   *
+   * The section is rendered as it was authored, the same as
+   * {@link TextService} renders it for the per-point announcement (#830).
+   * Lower-casing it here meant a reader who pressed the position key heard
+   * "in minimum" for the section that had just announced itself as "Minimum".
    */
   private announceBoxplotPosition(state: NonEmptyTraceState): void {
     const section = state.text.section || '';
@@ -813,15 +818,20 @@ export class AnnouncePositionCommand extends AnnounceCommand {
 
     if (this.textService.isTerse() || this.textService.isOff()) {
       const percent = totalBoxes > 1 ? Math.round((boxIndex / (totalBoxes - 1)) * 100) : 0;
-      this.textViewModel.update(`${percent}%, ${section.toLowerCase()}`);
+      this.textViewModel.update(`${percent}%, ${section}`);
     } else {
-      this.textViewModel.update(`Position is ${position} of ${totalBoxes} in ${section.toLowerCase()}`);
+      this.textViewModel.update(`Position is ${position} of ${totalBoxes} in ${section}`);
     }
   }
 
   /**
    * Announces position for candlestick charts with component information.
    * Uses braille state to get correct candle position.
+   *
+   * Verbatim for the same reason as {@link announceBoxplotPosition}. A
+   * candlestick's sections are authored in lower case already, so this reads
+   * the same either way today -- but the section belongs to the trace, and
+   * the one place that decides its case should not be here.
    */
   private announceCandlestickPosition(state: NonEmptyTraceState): void {
     const section = state.text.section || '';
@@ -835,9 +845,9 @@ export class AnnouncePositionCommand extends AnnounceCommand {
 
     if (this.textService.isTerse() || this.textService.isOff()) {
       const percent = totalCandles > 1 ? Math.round((candleIndex / (totalCandles - 1)) * 100) : 0;
-      this.textViewModel.update(`${percent}%, ${section.toLowerCase()}`);
+      this.textViewModel.update(`${percent}%, ${section}`);
     } else {
-      this.textViewModel.update(`Position is ${position} of ${totalCandles}, ${section.toLowerCase()}`);
+      this.textViewModel.update(`Position is ${position} of ${totalCandles}, ${section}`);
     }
   }
 
--- a/src/service/text.ts
+++ b/src/service/text.ts
@@ -518,11 +518,11 @@ export class TextService implements Observer<PlotState>, Disposable {
       const formattedMainValue = this.formatSingleValue(state.main.value as number | string, mainAxisType);
       if (outliers.length === 0) {
         // No outliers
-        return `${state.main.label} is ${formattedMainValue}, no ${state.section.toLowerCase()} for ${label}`;
+        return `${state.main.label} is ${formattedMainValue}, no ${state.section} for ${label}`;
       } else {
         // Outlier values present
         const verb = outliers.length === 1 ? 'is' : 'are';
-        return `${state.main.label} is ${formattedMainValue}, ${state.section.toLowerCase()} for ${label} ${verb} ${outlierStr}`;
+        return `${state.main.label} is ${formattedMainValue}, ${state.section} for ${label} ${verb} ${outlierStr}`;
       }
     }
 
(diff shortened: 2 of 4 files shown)


jooyoungseo:
Taken, and it belongs in this PR rather than a follow-up — because this PR is what makes it wrong.

**What the flag found.** `src/command/describe.ts` lower-cased `state.text.section` at four sites. Before this change the box plot's verbose per-point reading also lower-cased, so the two agreed with each other and disagreed only with terse. After it, the per-point reading says `Minimum` in both modes and the position key still answered `Position is 2 of 2 in minimum`. So merging the fix without this would have moved the disagreement rather than closed it, and moved it into a place harder to notice — two different keys, not two modes of one key.

**Fixed in `839ac8e`.** All four sites, including the two candlestick ones. Those two read identically either way today — `CANDLESTICK_SECTIONS` (`src/model/candlestick.ts:49`) is `['volatility', 'open', 'high', 'low', 'close']`, already lower case, exactly as you said. They go with the other two anyway: the trace owns the case of its own label, and one of four sites deciding differently is how these drifted apart in the first place.

**Four cases in `test/command/announce-position.test.ts`,** built on real traces rather than hand-written state — a fixture that writes `section: 'Minimum'` itself would be deciding the thing under assertion:

| case | with the fix reinstated |
|---|---|
| box, verbose — `Position is 2 of 2 in Minimum` | **fails** |
| box, terse — `100%, Minimum` | **fails** |
| every box section: the announcement contains what the trace reports | **fails** |
| candlestick — `Position is 2 of 3, close` | passes either way |

The last one passes both ways on purpose: it is the case that shows nothing else moved, and it is the only honest way to cover a site whose behaviour genuinely does not change.

Your other five checks all match what I found tracing it, including that `errorBar.ts` and `waterfall.ts` author lower case and are untouched.

## Verification

| Gate | Result |
|---|---|
| `npm test` | **2609 + 73 passed**, 185 suites (was 2605 + 73) |
| `npx playwright test --project=chromium boxplotVertical boxplotHorizontal candlestick` | **50 passed** |
| `npm run type-check`, `npm run lint:fix`, `npm run build` | clean |

Nothing in `e2e_tests/` asserted either wording — I grepped for the position phrasing before running, and there was no case to update. Same gap the PR body describes for the per-point rendering, in the same place.

---

PR #856, opened 2026-08-12
ci: stop a malformed commit message from skipping every test

### LINKED ISSUE #763: fix(ci): a commitlint failure still hides every verification job
# Bug Report

## Description

Five jobs in `ci.yml` declare `needs: [commitlint, lint]`:

```
.github/workflows/ci.yml:130    E2E (chromium)
.github/workflows/ci.yml:178    Build Project
.github/workflows/ci.yml:202    Unit Tests
.github/workflows/ci.yml:226    Type Check
.github/workflows/ci.yml:253    Docs Build
```

A `needs` dependency does not just order jobs — it **skips** them when the dependency fails. So a commit message that does not parse takes down the entire verification surface, and the PR shows two red checks out of fourteen rather than "nothing was tested."

#651 already recognised this and fixed the bot case, and `commitlint.ignores.ts` states the reasoning outright:

> A failing commitlint job skips every job declaring `needs: [commitlint, lint]` in ci.yml -- build, unit-test, type-check, e2e-smoke and docs-build -- so dependency updates would merge without ever being built or tested.

But the fix was scoped to `/^(?:build|chore)\(deps(?:-dev)?\)/`. **The hazard is unchanged for every human commit.** A typo in a commit subject still hides the build, the type check, 1,664 unit tests, 330 e2e specs and the docs build behind a message-formatting error.

This is not hypothetical. It is exactly what happened on #669, where the pattern hid three independent breakages.

## Steps to Reproduce

1. Open a PR whose commit subject does not satisfy commitlint (e.g. `Fix the thing.` — capitalised, trailing period).
2. Let CI run.
3. Look at the checks list.

## Actual Behavior

`Lint Commit Messages` fails. `Build Project`, `Unit Tests`, `Type Check`, `E2E (chromium)` and `Docs Build` report **skipped**, which GitHub renders in grey rather than red. A reviewer scanning the list sees one failure about commit formatting and no evidence that anything is broken — because nothing ran to find out.

Reproduced on #669, whose CI showed:

```
Lint Commit Messages   failure
Build Project          skipped
Unit Tests             skipped
Type Check             skipped
E2E (chromium)         skipped
Docs Build             skipped
```

Under those greyed-out checks were three real defects: `npm ci` failing on npm 10, 24 test suites (341 tests) silently not running, and a published bundle that threw `process is not defined` on load before MAIDR could attach to a chart. The last one is the serious kind — the build, the type check and the full unit suite all pass on that bundle, because they import source rather than the built artefact. Only e2e catches it, and e2e was skipped.

## Expected Behavior

A malformed commit message fails its own check and nothing else. Build, tests, type check and e2e report the true state of the branch regardless.

## Additional Information

**Suggested fix.** Drop `commitlint` from the `needs` of the five jobs, keeping `lint`:

```diff
-    needs: [commitlint, lint]
+    needs: [lint]
```

Commitlint is a required check in its own right, so a bad message still blocks the merge — it just stops taking the test results down with it. This mirrors the reasoning already applied to `actionlint`, which the same file describes as gating nothing and reporting on its own:

> The check is pure Node with no dependencies, so it skips npm ci and gates nothing — it reports drift on its own rather than blocking the real jobs.

Whether `lint` should stay in `needs` is a separate judgement. It is a genuine cost saver (no point running e2e on code that will not lint), and unlike commitlint it inspects the code under test rather than the message describing it. I would keep it.

**Alternative worth weighing.** Keeping `needs: [commitlint, lint]` but adding `if: always()` to the dependent jobs makes them run regardless. That preserves the ordering but is easy to get wrong — `always()` also runs jobs after a cancellation, which is usually not wanted. Removing the dependency is simpler and has fewer edges.

**On the bot-bump ignore.** With this fixed, `isBotDependencyBump` becomes a much smaller concession: it would only exempt bot commits from the message rule itself, rather than being load-bearing for whether dependency updates get tested at all. Worth revisiting whether it is still wanted, though that is a separate question and there is a decent argument for keeping it either way.

**Why this is worth doing now.** The five PRs that replaced #669 (#758, #759, #760, #761, #762) all went green on fourteen checks, so the current state is healthy. But the mechanism that made #669 dangerous is still in place, and the next time it fires it will be on a commit nobody thought to rebase.

jooyoungseo:
# Pull Request

## Description

`needs` does not just order jobs — it **skips** them when the dependency fails, and GitHub renders a skip in grey. Five jobs declared `needs: [commitlint, lint]`, so a commit subject that did not parse took the build, the type check, the whole unit suite and every e2e spec down with it.

The checks list then reads as *one formatting problem* rather than as *nothing was tested*. Those are opposite conclusions, and the greyed-out rendering argues for the wrong one.

## Related Issues

Closes #763.

Not hypothetical. On #669 the list was:

```
Lint Commit Messages   failure
Build Project          skipped
Unit Tests             skipped
Type Check             skipped
E2E (chromium)         skipped
Docs Build             skipped
```

Under those greyed-out checks sat three real breakages: `npm ci` failing on npm 10, 24 test suites silently not running, and a published bundle that threw `process is not defined` on load. The last is the one that matters here — the build, the type check and the full unit suite all pass on that bundle, because they import source rather than the built artefact. **Only e2e catches it, and e2e was skipped.**

#651 already recognised the hazard and fixed the bot case; `commitlint.ignores.ts` says so outright. But that fix was scoped to `/^(?:build|chore)\(deps(?:-dev)?\)/`, so the hazard was unchanged for every human commit.

## Changes Made

```diff
-    needs: [commitlint, lint]
+    needs: [lint]
```

on `e2e-chromium`, `build`, `unit-test`, `type-check` and `docs-build`, plus a comment on the `commitlint` job saying why it gates nothing — the same note `copilot-sync` and `actionlint` already carry.

**A bad message still blocks the merge**, because commitlint remains a required check in its own right. It just stops taking the evidence down with it.

**`lint` stays in `needs`**, deliberately. Unlike commitlint it inspects the code under test rather than the message describing it, and there is no point running a 25-minute e2e job on code that will not lint. The issue reached the same conclusion.

I did not take the `if: always()` alternative. It preserves the ordering but also runs the jobs after a cancellation, which is not what anyone wants, and it is easy to get subtly wrong. Removing the dependency has fewer edges.

n/a — the change is to job wiring.

## Checklist

## Additional Notes

**`test/scripts/ciJobDependencies.test.ts`** pins it, because the one-line form of this mistake is easy to reintroduce and invisible until the day it fires. Four cases:

- no job lists `commitlint` in its `needs`;
- the `commitlint` job still exists — removing it from `needs` is only safe while it remains a required check, and deleting it would be the opposite mistake;
- the four verification jobs still declare exactly `needs: [lint]`, so a change that dropped every `needs` could not pass as a fix for the one that was removed;
- the parser found more than five jobs and found `commitlint` among them, so a regex that stopped matching cannot make the rest vacuous.

Parsed by hand rather than with a YAML library: neither `yaml` nor `js-yaml` is a direct dependency, and a test guarding CI should not be the thing that adds one.

**Verified it bites:** restoring `needs: [commitlint, lint]` fails 2 of the 4.

| Gate | Result |
|---|---|
| `node scripts/test.js -- test/scripts/ciJobDependencies.test.ts` | **77 passed** |
| `actionlint` v1.7.12 on `.github/workflows/ci.yml` | exit 0, no findings |

`eslint` rejected the first `needs:` regex under `regexp/no-super-linear-backtracking` — `\s*` and `.+` can exchange characters. Narrowed to `/^ {4}needs: *(\S.*)$/`, where the capture cannot start with a space and the exchange is impossible.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,16 @@ jobs:
       - name: Check generated instruction files
         run: npm run sync:copilot:check
 
+  # Gates nothing, and must not: `needs` does not just order jobs, it *skips*
+  # them when the dependency fails, so a commit subject that does not parse
+  # used to take the build, the type check, the whole unit suite and every e2e
+  # spec down with it -- reported in grey rather than red, which reads as
+  # "nothing wrong here" instead of "nothing was tested" (#763). It happened on
+  # #669, where three real breakages sat under the greyed-out checks, one of
+  # them a published bundle that threw on load and which only e2e could catch.
+  #
+  # A bad message still blocks the merge, because this is a required check in
+  # its own right. It just stops taking the evidence down with it.
   commitlint:
     name: Lint Commit Messages
     runs-on: ubuntu-latest
@@ -137,7 +147,7 @@ jobs:
   e2e-chromium:
     name: E2E (chromium)
     runs-on: ubuntu-latest
-    needs: [commitlint, lint]
+    needs: [lint]
     timeout-minutes: 25
 
     permissions:
@@ -185,7 +195,7 @@ jobs:
   build:
     name: Build Project
     runs-on: ubuntu-latest
-    needs: [commitlint, lint]
+    needs: [lint]
 
     permissions:
       contents: read
@@ -209,7 +219,7 @@ jobs:
   unit-test:
     name: Unit Tests
     runs-on: ubuntu-latest
-    needs: [commitlint, lint]
+    needs: [lint]
 
     permissions:
       contents: read
@@ -233,7 +243,7 @@ jobs:
   type-check:
     name: Type Check
     runs-on: ubuntu-latest
-    needs: [commitlint, lint]
+    needs: [lint]
 
     permissions:
       contents: read
@@ -260,7 +270,7 @@ jobs:
   docs-build:
     name: Docs Build
     runs-on: ubuntu-latest
-    needs: [commitlint, lint]
+    needs: [lint]
 
     permissions:
       contents: read
--- a/test/scripts/ciJobDependencies.test.ts
+++ b/test/scripts/ciJobDependencies.test.ts
@@ -34,14 +34,31 @@ interface Job {
 /**
  * Read the jobs of a workflow and what each one waits for.
  *
+ * Scanning starts at `jobs:` rather than at the top of the file. `on:` puts
+ * its triggers at the same two-space indent, so an unscoped scan reports
+ * `workflow_dispatch` and `pull_request` as jobs — harmless for the
+ * assertions that name an id, but it makes the "there are jobs here" sanity
+ * check count things that are not jobs, which is the one case whose whole
+ * purpose is to be trustworthy.
+ *
  * @param body The workflow file
  * @returns One entry per job, in file order
  */
 function jobsOf(body: string): Job[] {
   const lines = body.split('\n');
+  const start = lines.findIndex(line => /^jobs:\s*$/.test(line));
   const jobs: Job[] = [];
 
-  for (const line of lines) {
+  if (start < 0) {
+    return jobs;
+  }
+
+  for (const line of lines.slice(start + 1)) {
+    // A key at the top level ends the `jobs:` block.
+    if (/^\S/.test(line)) {
+      break;
+    }
+
     const job = /^ {2}([\w-]+):\s*$/.exec(line);
     if (job) {
       jobs.push({ id: job[1], needs: [] });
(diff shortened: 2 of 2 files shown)


jooyoungseo:
Taken — `67c8830f`. You were right, and I checked before fixing:

```
matched: workflow_dispatch, pull_request, copilot-sync, commitlint, lint,
         actionlint, e2e-chromium, build, unit-test, type-check, docs-build
```

Two phantoms. As you say, no assertion that names an id was affected — but the one case whose entire purpose is to be trustworthy was counting things that are not jobs, and a sanity check that is itself unsound is worse than no sanity check.

`jobsOf()` now starts at `^jobs:$` and stops at the next top-level key, and the sanity case asserts the phantoms are *absent* rather than only that the count is large. Reverting the scope makes 3 of the 4 cases fail, which is the check I wanted before calling it fixed.

`npm run lint:fix`, `npm run type-check` clean; `node scripts/test.js -- test/scripts/ciJobDependencies.test.ts` — 77 passed.

On the verification note: understood, and the same limitation showed up on the sibling PRs today. Worth saying that this PR makes that gap smaller rather than larger — the reason a reviewer has to fall back on reading the diff is that CI is the thing that runs it, and the whole point here is that CI could previously be silenced by a typo in a commit subject.

---

PR #858, opened 2026-08-13
refactor(build): delete the root vite config no build reads

### LINKED ISSUE #765: fix(build): the root vite.config.ts is a mirror that nothing builds through
# Bug Report

## Description

Nothing in this repository builds through `vite.config.ts`. Three separate mechanisms each rule it out:

- **`scripts/build.js:347` passes `configFile: false`.** This is `createViteConfig`, the factory behind every published bundle. Vite's `build()` API auto-loads `vite.config.ts` by default; that flag turns it off, so the root config is never merged into a library build.
- **Every example names its own config.** `dev:recharts`, `dev:victory`, `build:recharts-example`, `build:victory-example` and `build:react-example` all pass `--config examples/*/vite.config.ts`.
- **No script invokes `vite` bare**, which is the only remaining way the root config would get picked up implicitly.

Yet the file is maintained as though it were live, and `test/scripts/publishedPackage.test.ts:91` asserts it stays in step with the real build config:

```ts
it('should match the standalone vite config', () => {
  const source = readFileSync(resolve(ROOT, 'vite.config.ts'), 'utf8');
  ...
  expect(code).toMatch(/sourcemap:\s*'hidden'/);
```

So it is a **mirror nothing validates by building** — the test compares its text to the real config, but no build ever proves the two agree in effect.

## Steps to Reproduce

1. `grep -n "configFile" scripts/build.js` → `configFile: false` at line 347
2. `node -e "const s=require('./package.json').scripts; for (const [k,v] of Object.entries(s)) if (v.includes('vite') && !v.includes('--config')) console.log(k,v)"` → prints nothing
3. Edit `vite.config.ts` — put anything invalid in `define` — and run `npm run build`

## Actual Behavior

Step 3 builds successfully. The file is not read.

## Expected Behavior

Either the file is the build config and a change to it affects the build, or it does not exist. A config that looks authoritative, is kept in sync by a test, and is silently inert is the shape that produces wrong conclusions during debugging.

## Additional Information

**How this surfaced.** In #759 I fixed a `define` bug in both `scripts/build.js` and `vite.config.ts`, and wrote a comment saying the root config "serves the dev servers and the example builds." That was wrong, and review on #762 caught it. Checking properly turned up the above. The comments in both files were corrected in `f66fea1`, and the measured −300 kB improvement in #762 comes entirely from `scripts/build.js` — the root config contributed nothing.

Concretely, the risk it creates: a reviewer verifying a build change reads `vite.config.ts`, finds it correct, and concludes the build is correct. That is a plausible way to miss a defect in `createViteConfig`.

**Options:**

1. **Delete it,** and move the two invariants `publishedPackage.test.ts` checks (`sourcemap: 'hidden'`, and the `define` keys guarded by `buildDefineNodeEnv.test.ts`) onto `createViteConfig` alone, where they already have coverage. Simplest end state; the mirror-sync tests become unnecessary rather than merely passing.
2. **Make it authoritative** — drop `configFile: false` and let `createViteConfig` merge the root config. Removes the duplication in the other direction, but changes how every bundle is built for no functional gain, and `configFile: false` may well be deliberate (it makes the build hermetic and independent of what happens to sit in the repo root).
3. **Keep it and say so.** A header comment stating the file is a template for future builds and is not currently loaded. Cheapest, and honest, but leaves two configs to keep in step by hand.

I would take option 1, but this is a maintainer's call — whether the file is meant as a template for a future build path is not something the code records, and I did not want to guess inside a bundle-size PR.

**Current state is safe.** #762 left both configs agreeing on `define`, and `test/scripts/buildDefineNodeEnv.test.ts` now pins the root config's value as well as its key, precisely so that whoever eventually points a build at this file does not rediscover the `process is not defined` crash from #759. Nothing is broken today; the cost is the confusion.

jooyoungseo:
# Pull Request

## Description

`vite.config.ts` at the repository root was not on any build path. It was a second, unread copy of the build's settings, kept in step with the real one by tests that compared the two files' texts.

That is worse than having no copy. A reader checking whether the published bundles substitute `process.env.NODE_ENV` by its full member expression, or emit hidden sourcemaps, finds the root config, finds it correct, and concludes the build is — when the build's answer lives somewhere else entirely. #759 is the case in point: the `process.env.NODE_ENV` bug was in `scripts/build.js`, and the inert config had it right the whole time.

Nothing reads it:

| Entry point | What it uses |
| --- | --- |
| `createViteConfig` in `scripts/build.js` | its own object, with `configFile: false` |
| `npm run dev:recharts` / `dev:victory` | `--config examples/*/vite.config.ts` |
| the example builds | `--config examples/*/vite.config.ts` |
| any script invoking `vite` bare | there is none |

### Why this is deleted where `playwright.config.ts` was kept

#687 met a similar situation — a root config that was a full second copy and had drifted — and did **not** delete it. It became a thin `export { default } from './e2e_tests/config/test-config'`, guarded by `test/scripts/playwrightRootConfig.test.ts` against gaining settings of its own.

The deciding difference is not the shape of the real config, it is whether anything would break. Playwright's root config **has callers**: a bare `npx playwright test`, and the Playwright VS Code extension, both of which look for a config at the repository root and collect nothing — silently — without one. A file with no settings still earns its place when deleting it would break a real workflow quietly.

Vite has no equivalent. Every invocation passes `--config`, `createViteConfig` passes `configFile: false`, a bare `vite` at the root would serve the library itself (a workflow this repo does not have — there is no `npm run dev`, only the per-example servers), and there is no Vitest, so nothing auto-discovers a root vite config the way the Playwright extension does.

So: Playwright's root config has a caller and no settings, and stays. This one had settings and no caller, and goes.

## Related Issues

Closes #765

## Changes Made

- **Deleted `vite.config.ts`.**
- **`test/scripts/publishedPackage.test.ts`** — removed `should match the standalone vite config`, which asserted `sourcemap: 'hidden'` against the deleted file's source text.
- **`test/scripts/buildDefineNodeEnv.test.ts`** — removed the `the root vite config's define` describe, two cases asserting on the same file's text.
- **`scripts/vite-plugin-woff2-only.d.ts`, `scripts/vite-plugin-math-stylesheet.d.ts`** — the headers said `vite.config.ts` needed these types. They now say no TypeScript imports the plugins today, and why the declarations stay anyway.
- **`scripts/build.js`** — the `define` comment absorbed what the deleted file's comment explained (the whole-member-expression rule and #759), and records why the second copy is gone.

### Coverage

Both invariants the deleted cases stood for keep **functional** coverage against the real factory — which is what a text comparison was standing in for in the first place:

- `publishedPackage.test.ts` calls `createViteConfig` through `sourcemapSettings()` and asserts every bundle is `hidden`. That case is untouched.
- `buildDefineNodeEnv.test.ts` executes `scripts/build.js` in a subprocess and asserts the substituted value reaches the output.

So what was removed is exactly the cases asserting on a file's source text, and nothing that runs is lost. I checked this before deleting rather than after.

N/A — no user-facing change.

## Checklist

## Additional Notes

Local verification on this branch:

```
> eslint . --fix                            # clean
> tsc --noEmit && tsc -p tsconfig.ci.json   # clean
All builds complete in 128.5s               # npm run build
Test Suites: 187 passed, 187 total          # unit + component
Tests:       2618 passed, 2618 total
Test Suites: 7 passed, 7 total              # esm
Tests:       73 passed, 73 total
```

The build was run deliberately rather than trusted: if anything had in fact been reading the root config, deleting it would have surfaced there.

The two `.d.ts` files are kept rather than deleted with their last caller. They are hand-written declarations that describe modules which still exist and are still imported by `scripts/build.js`; the next TypeScript caller needs them, and reading `vite-plugin-woff2-only.js` is easier with them present. Deleting them would be a second, separable decision.


jooyoungseo:
The Playwright question is the right one to ask, and worth answering in more than a line — the deciding difference is not the shape of `createViteConfig`, it is whether anything would break.

`playwright.config.ts` was kept as a re-export because **there are callers that need a config at the root**: a bare `npx playwright test`, and the Playwright VS Code extension. Its own header says so:

> This exists so the paths that look for a config at the repository root find the same one: a bare `npx playwright test`, and editor integrations such as the Playwright VS Code extension. Without it those collect nothing, silently.

That is the whole reason a file with no settings still earns its place: deleting it would break a real workflow *quietly*, and quietly is the failure mode the repo keeps writing tests against.

Vite has no equivalent. Every invocation passes `--config`:

```
build:recharts-example: vite build --config examples/recharts/vite.config.ts
build:react-example:    vite build --config examples/react-app/vite.config.ts
build:victory-example:  vite build --config examples/victory/vite.config.ts
dev:recharts:           vite --config examples/recharts/vite.config.ts
dev:victory:            vite --config examples/victory/vite.config.ts
```

and `createViteConfig` passes `configFile: false`. A bare `vite` at the root would start a dev server for the library itself, which is not a workflow this repo has — `CLAUDE.md` says outright there is no `npm run dev`, only the per-example servers. There is no Vitest either; `@vitest/eslint-plugin` is an ESLint plugin, not a runner, so nothing auto-discovers a root vite config the way the Playwright extension does.

So: Playwright's root config has a caller and no settings, and stays. Vite's had settings and no caller, and goes. Your point about `createViteConfig` returning an object rather than a config module is also true and would have made a re-export awkward — but even if it were a clean module, re-exporting it would create a file for nobody.

I have added this to the PR description rather than leaving it in a thread.

On the verification note: the local run stands as reported, and CI has since gone green on all fourteen checks including E2E and the example preview build.

---

PR #870, opened 2026-08-14
feat(frappe): support the chart types added since the pie chart

### LINKED ISSUE #866: feat(frappe): support the chart types added since the pie chart
## Problem

MAIDR core has gained ~34 new trace types since the pie chart landed (radar, gauge, funnel, waterfall, gantt, treemap, sankey, sunburst, dumbbell, dot/lollipop, diverging, area family, and more). The `frappe` adapter has not kept up: of those, it currently emits **none**.

A chart Frappe Charts draws perfectly well is therefore either dropped from the figure or announced as the wrong chart — which for a screen-reader user is worse than not being supported at all.

There is no detection at all — the TraceType is chosen by a `switch` on a string the *caller* passes in. `buildLayers` (/home/user/maidr/src/adapters/frappe/converters.ts:315) switches on `options.chartType`, an adapter-level union declared at converters.ts:53 and types.ts:57 (`'axis-mixed' | 'bar' | 'donut' | 'line' | 'pie' | 'scatter'`); an unrecognised value throws (converters.ts:327-332). Each case calls a per-type builder that hardcodes one TraceType literal: `buildBarLayer` → `TraceType.BAR` (converters.ts:360), `buildLineLayer` → `TraceType.LINE` (converters.ts:392), `buildScatterLayer` → `TraceType.SCATTER` (converters.ts:412), `buildPieLayer` → `TraceType.PIE` (converters.ts:511), and `buildMixedLayers` → a second per-dataset switch on `dataset.chartType ?? 'bar'` yielding LINE (converters.ts:447) or BAR (converters.ts:461). Note the adapter string is NOT the Frappe string: Frappe v1.6.2 has no `scatter` type (chart.js only maps bar/line/percentage/heatmap/pie/donut plus the `axis-mixed` alias), so `'scatter'` is already an adapter-invented type for `type:'line' + lineOptions.hideLine` (docs/frappe.md "Scatter note"), which is the precedent any new type would follow. Only two things are inferred from the chart itself: multi-line (`data.datasets.length > 1` → per-dataset selectors + subplot legend, converters.ts:261 and 373) and pie aggregation; the only instance field read besides `chart.data` is `chart.config?.maxSlices` (converters.ts:255).

## In scope — 4 trace types Frappe Charts can render

| TraceType | Frappe Charts construct | Detection / payload |
|---|---|---|
| `AREA` | line chart with `lineOptions.regionFill: 1` (Frappe's own "area" look) | Verified in the v1.6.2 source, not assumed: `regionFill` is wired through AxisChart.js:300 and draws `path.region-fill` in draw.js:615, on top of the normal `g.dataset-units.dataset-line.dataset-{i}` group (ChartComponents.js:359) and its per-point `<circle>` dots. Detection: `lineOptions` is an instance field (`this.lineOptions = args.lineOptions \|\| {}`, AxisChart.js:17), NOT under `config`, so add `lineOptions?: { regionFill?: number; hideLine?: number; dotSize?: number }` to `FrappeChart` in types.ts:39 and read `chart.lineOptions?.regionFill` in `buildSubplot` (converters.ts:253-256, the same place `config.maxSlices` is already read) — or accept an explicit `chartType: 'area'` the way `'scatter'` already works. Payload is byte-identical to `buildLineLayer` (converters.ts:368-397): `LinePoint[][]`, one inner array per dataset, `z` = dataset name, `selectors` = `lineSelectorForDataset` per dataset (selectors.ts:86) or `lineSelector` (selectors.ts:81) for one series, `axes` from `buildAxes`, plus the existing multi-series legend at converters.ts:263. Only `type` changes to `TraceType.AREA`; AreaTrace extends LineTrace and treats a plain area as a line with an honest chart-type announcement (factory.ts:57-59, src/model/area.ts). Multi-dataset regionFill charts overlap rather than stack, which is exactly what plain AREA means per the grammar doc comment — so they stay AREA, never STACKED_AREA. |
| `DOT` | line chart with `lineOptions: { hideLine: 1, dotSize: N }` — the identical rendering the adapter already calls 'scatter' | Frappe places marks at evenly spaced label positions regardless of the label's numeric value (documented at docs/frappe.md, "Scatter note"), so this rendering is a categorical dot plot, not a true scatter. Today `buildScatterLayer` coerces the label with `x: Number(label)` (converters.ts:406), which yields `NaN` for every categorical label ('Mon', 'North', …) — so a Frappe dot plot over category names is currently unemittable. Detection: either an explicit `chartType: 'dot'` case in `buildLayers` (converters.ts:315), or the zero-new-API route — inside the existing `'scatter'` case, emit DOT when `data.labels.some(l => !Number.isFinite(Number(l)))` and keep SCATTER for numeric labels. Payload is the bar payload, not the scatter one: `BarPoint[]` of `{ x: label, y: dataset.values[i] }` (`{ x: string \| number; y: number \| string }`, grammar.ts:256), `orientation: Orientation.VERTICAL`, `selectors: scatterSelector(containerId)` (selectors.ts:91 — unchanged, the dots are the highlight targets), `axes` from `buildAxes`. `TraceType.DOT` routes to `BarTrace` (factory.ts:67-69: "A dot plot draws a point where a bar chart draws a bar"), whose single-row `mapToSvgElements` (bar.ts:230) matches the one-circle-per-point DOM exactly. |
| `BUMP` | multi-dataset line chart (`type: 'line'`, one dataset per competitor) | Structurally this is the multi-line chart the adapter already emits — one line per competitor over shared categorical labels, which is precisely a bump chart's geometry. Detection has to be an explicit `chartType: 'bump'` opt-in (a new case in converters.ts:315 plus the union at types.ts:57): nothing in Frappe's data or config distinguishes ranks from magnitudes, so inferring it would be guessing. Payload is `buildLineLayer` unchanged — `LinePoint[][]` with `y` = the true rank (1 = best), `z` = dataset name, `lineSelectorForDataset` per series (selectors.ts:86), plus the existing legend at converters.ts:263 — with `type: TraceType.BUMP` (factory.ts:74 → BumpTrace, src/model/bump.ts, which derives places-gained/lost and inverts the pitch itself). ACCURACY CAVEAT worth stating in the docs: Frappe cannot invert or reverse an axis — there is no `invert`/`reverse` anywhere in v1.6.2's chart sources, only `axisOptions: { xAxisMode, yAxisMode, xIsSeries, shortenYAxisNumbers }` (AxisChart.js:35-41) — so a rank chart drawn in Frappe shows rank 1 at the BOTTOM. The emitted `y` must still be the true rank because BumpTrace does the inversion; an author who pre-inverts the values to make the picture look right would make MAIDR announce the wrong ranks. |
| `DIVERGING` | grouped (non-stacked) bar chart with signed values — two bar datasets, one all-negative and one all-positive, drawn either side of Frappe's zero line | Frappe draws negative bars below the zero line and renders each dataset into its own `g.dataset-units.dataset-bars.dataset-{i}` group (ChartComponents.js:296), so a two-series signed bar chart genuinely renders. Detection: an explicit `chartType: 'diverging'` case, validated as exactly two datasets whose values have opposite signs (silently accepting anything else would mis-announce sides). Payload: `SegmentedPoint[][]` (grammar.ts:953 — `BarPoint` plus a required `z`), one inner array per side, `{ x: label, y: signedValue, z: dataset.name }`, values emitted AS THE CHART DRAWS THEM (left/down series negative — DivergingTrace takes the magnitude for pitch and names the side, src/model/diverging.ts). Selectors: `SegmentedTrace` takes ONE selector string matching every group's rects, not a per-series array (bar.ts:94 + segmented.ts:315), which is exactly what the currently-unused broad `barSelector(containerId)` (selectors.ts:60) returns — Frappe emits dataset-0's rects then dataset-1's, so pair it with `domMapping: { order: 'row' }` (grammar.ts:1148-1159). TWO CAVEATS to state honestly: (1) Frappe has no horizontal bars, so the classic back-to-back population pyramid orientation is not drawable — only the vertical up/down form; (2) `barOptions.stacked` must NOT be used, because `getAllYValues` cumulates raw values (AxisChart.js:165-175) and so mis-stacks a negative series onto a positive one. This is also the natural place to fix the multi-dataset bar drop at converters.ts:345-350. |

## Not renderable by Frappe Charts

<details><summary>29 trace types the library genuinely cannot draw</summary>

- ALLUVIAL — no flow/ribbon geometry; v1.6.2 ships only AxisChart, PieChart, DonutChart, PercentageChart, Heatmap (chart.js:10-17).
- CHORD — no circular layout beyond pie/donut wedges, and no ribbons.
- CHOROPLETH — no geographic projection or path/region rendering of any kind.
- CONTOUR — no field/iso-line rendering; `getPaths` draws one polyline per dataset (draw.js:598) and nothing level-aware.
- DUMBBELL — two dot series are drawable (two `hideLine` datasets) but there is no mark that joins the pair, and the connector is the whole point of the chart.
- ERROR_BAR — no per-point interval mark; `yMarkers`/`yRegions` (AxisChart.js:144-165) are chart-wide horizontal lines/bands, not per-sample whiskers.
- FOREST — needs per-study intervals plus a null line and a pooled row; the interval mark does not exist (see ERROR_BAR).
- GANTT — no floating/offset bars: bars are always drawn from the zero line, and there are no horizontal lanes.
- FUNNEL — no funnel/trapezoid/centred-bar mark. The nearest thing, `percentage`, is a single 100%-wide bar segmented by label (PercentageChart.js:52-66), which is a share-of-total breakdown, not ordered shrinking stages.
- GAUGE — no dial, arc or bullet mark. `percentage` is a share-of-total bar with no min/max/target/band semantics (GaugePoint needs value+min+max, grammar.ts:530-543); a two-slice donut is a hack, not a rendering the library offers.
- HEXBIN — no binning and no hexagon mark; Frappe's only lattice is the calendar Heatmap's day squares (`heatSquare`, draw.js:192).
- ICICLE — no hierarchy support at all; the data model is a flat `{ labels, datasets }` (types.ts:30-33).
- LOLLIPOP — dots are drawable (`hideLine: 1`) but there is no stem-to-baseline mark and no way to draw one.
- MANHATTAN — needs continuous genomic x over tens of thousands of points; Frappe spaces labels evenly regardless of value (docs/frappe.md scatter note) and renders one `<circle>` per point per dataset.
- MOSAIC — bar widths are computed from `barsWidth / datasets.length` and a fixed `spaceRatio` (AxisChart.js:251-258); per-column data-driven widths are impossible.
- NETWORK — no node-link rendering, no layout solver.
- NORMALIZED_AREA — follows STACKED_AREA: no stacking exists for line/area datasets, so there is nothing to normalise.
- PARALLEL — a multi-line chart over categories is close in shape, but every dataset shares ONE y scale and there are no per-variable axes; PARALLEL is defined precisely by pitching each value against its OWN axis, so emitting it here would describe a chart Frappe did not draw.
- POLAR_AREA — pie/donut wedges all share one radius; the wedge radius cannot encode a value (PieChart/DonutChart draw fixed-radius arcs).
- RADAR — no polar axes, no closed outline.
- RIDGELINE — no per-series vertical offset; multiple `regionFill` regions all fill down to the same zero line and overlap.
- SANKEY — no flow/ribbon geometry (see ALLUVIAL).
- STACKED_AREA — `barOptions.stacked` applies only to bar datasets (AxisChart.js:128, 167, 251-280); line/area datasets are never accumulated, so several `regionFill` regions overlap from a shared baseline instead of stacking.
- SUNBURST — no hierarchy, no multi-ring arcs.
- SURVIVAL — needs step interpolation plus censoring marks; `lineOptions` offers spline/heatline/regionFill/hideLine/hideDots/dotSize (AxisChart.js:299-321) and no step mode.
- TREEMAP — no hierarchy and no nested-rectangle layout.
- VOLCANO — needs continuous x (log2 fold change); Frappe positions points at evenly spaced label slots, which destroys the axis the chart exists to show.
- WATERFALL — bars always run from the zero line; there is no base/offset per bar, so floating running-total bars cannot be drawn.
- WORD_CLOUD — no text-glyph layout; the library draws no data-bearing text marks.

</details>

## Acceptance criteria

- [ ] Unit tests under `test/adapters/frappe/` cover every newly detected type.

## Implementation notes

Verification method: I pulled the real library (`npm pack frappe-charts@1.6.2`, extracted to /tmp/claude-0/-home-user/f931b544-743a-5579-836a-62b28330b0c9/scratchpad/package) rather than relying on memory, so every capability claim above is checked against v1.6.2 source — chart.js (type table), AxisChart.js (bar/line options, markers, stacking), ChartComponents.js (group class names), draw.js (marks), PercentageChart.js / AggregationChart.js. Delete that scratch dir when done; nothing in /home/user/maidr was modified.

Files that change for any new type: converters.ts:315 (`buildLayers` switch) + a new `buildXLayer`, types.ts:57 (`FrappeChartType` union — note it is an ADAPTER type, so 'area'/'dot'/'bump'/'diverging' belong there even though Frappe's own `type` stays 'line'/'bar'), selectors.ts if a new element class is targeted, docs/frappe.md ("Supported Chart Types" table ~line 95 and the options table ~line 300), and an example page following the `examples/frappe-*.html` pattern.

Trap 1 — highlight targets must be one SVG element per data point AND must be SVG: HighlightService throws on non-SVGElement (src/service/highlight.ts:120-121, 315-317). Line/area/dot layers therefore target the per-point `<circle>` dots, never `path.line-graph-path` or `path.region-fill`, and the host page must keep `lineOptions.dotSize > 0` / `hideDots: 0` (selectors.ts:17-19). This also rules out any mapping built on Frappe's calendar Heatmap or `percentage` chart being casually reshaped.

Trap 2 — element-count mismatch silently kills highlighting: `AbstractBarPlot.mapToSvgElements` returns null when the queried count differs from the point count (src/model/bar.ts:230-250), which is why every selector is scoped to `.dataset-{i}` (converters.ts:341-344). Any new multi-series layer must either scope per dataset (LINE-family, `string[]`) or use one combined selector plus `domMapping` (SEGMENTED-family, single string — see bar.ts:94 and segmented.ts:315).

Trap 3 — timing: Frappe's entrance animation re-creates the SVG nodes, so all examples and the docs use an `activateMaidrWhenSettled` MutationObserver helper (docs/frappe.md:47-66). Any new example must copy it verbatim or highlighting appears broken until a tab switch.

Unused building block already in place: `barSelector(containerId)` (selectors.ts:60) is exported but referenced only in a comment (converters.ts:342) — it is exactly the combined all-groups selector a DIVERGING/segmented layer needs.

Biggest real gap overall is out of the post-pie scope: multi-dataset bar charts are silently reduced to dataset 0 with a `console.warn` (converters.ts:345-350), so Frappe's grouped bars (default) and `barOptions: { stacked: 1 }` never become DODGED/STACKED. Implementing DIVERGING would build the whole segmented-selector plumbing those need.

Test conventions to follow: test/adapters/frappe/converters.test.ts is pure JSDOM against `chart.data` stubs (`makeChart`, `makeDom`, `panel` helpers) and asserts the exact selector string and the exact emitted point array per type — add one describe block per new chartType there. test/adapters/frappe/multiPanelLayout.test.ts is a separate contract test that builds fake Frappe SVG (`svg.frappe-chart` > `g.dataset-units.dataset-{bars|line}.dataset-0` > marks) and feeds the first layer selector through `resolveSubplotLayout`; a new chart type whose marks live in a different group class must be added to its `makePanel` factory or multi-panel grids of that type will lose their vertical ordering (the adapter deliberately emits no `MaidrSubplot.selector`, converters.ts:236-241).

jooyoungseo:
## Description

The Frappe Charts adapter previously converted bar, line, multi-line, scatter, mixed-axis and pie/donut charts. This adds the statistical chart types MAIDR has gained since the pie work landed and that Frappe can actually draw: **area**, **bump**, **dot** and **diverging bar**.

Frappe v1.6.2 draws several distinct statistical charts with the *same* `type: 'line'` or `type: 'bar'` — they differ only in their options or in what the numbers mean, and nothing on the chart instance records which one the author intended. So the adapter's own `chartType` union (already distinct from Frappe's `type` strings, since it carried an invented `'scatter'`) grows to name them. This is adapter-local work: no core file was touched, and the existing traces already support every payload emitted here.

Every new layer both sonifies **and** highlights — the line-family types target the per-point `<circle>` dots, and the dot layer is scoped to its own `.dataset-0` group.

## Related Issues

Closes #866

### Trace types added

| Trace | Adapter `chartType` | How it is detected |
|-------|--------------------|--------------------|
| `AREA` | `'area'` | Explicit, **and inferred** from `chart.lineOptions.regionFill` |
| `BUMP` | `'bump'` | Explicit — a multi-dataset line chart whose y values are ranks |
| `DOT` | `'dot'` | Explicit, **and** a `'scatter'` whose labels are categorical |
| `DIVERGING` | `'diverging'` | Explicit, validated as two opposite-signed datasets |

Notes on each:

- **Area** is additionally inferred from `lineOptions.regionFill`, an instance field Frappe keeps verbatim (`AxisChart.js:17`), so a filled chart passed as `chartType: 'line'` is still announced as an area chart — an author cannot mislabel one. Several filled bands *overlap* in Frappe rather than stacking, so they stay independent `area` series, never `stacked_area`.
- **Bump** shares the line payload; only the announced type differs, which is what flips MAIDR's pitch so rank 1 is the highest note. `buildLineLayer` gained a 4th `type` parameter over the `AREA | BUMP | LINE` union rather than forking the builder.
- **Dot** vs scatter is decided by the labels, because Frappe spaces its marks evenly whatever the label holds. A `chartType: 'scatter'` over categorical labels is converted as a dot plot with a `console.warn`, since `Number('Mon')` is `NaN` and a scatter layer built from those labels would have had no x values at all. A blank label counts as categorical, because `Number('')` is `0` and would otherwise plant a phantom point at the origin.
- **Diverging** throws (rather than warns) unless the chart really is exactly two datasets with opposite signs — `DivergingTrace` strips the sign from its announcement, so a mis-shaped chart would name sides that do not exist. It uses the broad all-groups bar selector plus `domMapping: { order: 'row' }`, which is the series-major order Frappe appends its dataset groups in (`AxisChart.js:340`).

### Supporting changes

- `buildAxes` now emits `axes.z`; `FrappeChartAdapterOptions.axes` and `FrappePanel.axes` gained `z?: string`. Deliberately no fallback — only the types with a third dimension read it.
- New shared `seriesName(dataset, index)` helper (`dataset.name ?? \`Series N\``), used by both the multi-series legend and `SegmentedPoint.z`, which the grammar requires.
- `LINE_FAMILY_TYPES` drives the multi-series legend that used to be gated on `chartType === 'line'` alone.
- `barSelector(containerId)` gained a doc comment explaining the one-selector-plus-`domMapping` contract; the diverging layer is its first caller.
- Docs: `docs/frappe.md` gains a row per type in the supported table, a worked code example for each, and a caveat note per type.
- Examples: `examples/frappe-{area,bump,dot,diverging}.html`.
- Tests: `test/adapters/frappe/converters.test.ts` extended, plus a new `test/adapters/frappe/selectorResolution.test.ts`.

### Trace types deliberately NOT added

- **`DODGED` / `STACKED`** — out of scope for this change. Multi-dataset bar charts are still reduced to dataset 0 with a `console.warn`, exactly as before. The diverging work does build all the segmented plumbing these need (broad selector, `domMapping: { order: 'row' }`, `SegmentedPoint` construction), so this is the obvious next increment.
- **`HEATMAP`** — Frappe's only heatmap is a GitHub-style *calendar* heatmap, structurally unlike MAIDR's matrix heatmap. Not a payload mismatch that a converter can bridge.
- **`STACKED_AREA` / `NORMALIZED_AREA`** — Frappe overlaps filled regions rather than stacking them, so the drawn picture is not a stacked area chart and announcing it as one would be wrong.
- **Frappe's `percentage` chart** — no MAIDR trace is an equivalent.
- **`GAUGE`, `GANTT`, `DUMBBELL`, `CANDLESTICK`, `BOX`, `HISTOGRAM` and the rest of the grammar** — Frappe Charts v1.6.2 has no drawing primitive for any of them, so there is no chart to convert.

One limit is Frappe's, not MAIDR's, and is documented rather than faked: Frappe has **no horizontal bars**, so a diverging chart can only be the vertical up/down form — the back-to-back population-pyramid orientation is not drawable.

## Checklist

On the third box: the automated suite is fully green (details below), but `ManualTestingProcess.md` does not exist in this repository and no manual screen-reader pass was run, so the box is left unticked rather than half-claimed. The four new example pages are the natural place for that manual pass.

## Additional Notes

**Gate results** — all green, no fixes were needed:

| Check | Result |
|-------|--------|
| `npm run lint:fix` | clean, no files modified |
| `npm run type-check` | clean (`tsc --noEmit` + `tsconfig.ci.json`) |
| `npm run build` | all 12 adapter bundles built |
| `npm test` | 188 suites / 2640 tests + 7 ESM suites / 73 tests, 0 failures |
| `npm run sync:copilot:check` | in sync (11 files) |

`src/adapters/frappe` sits at 96.96% statement coverage.

**Testing caveat worth knowing for review:** jsdom implements neither `SVGRectElement` nor `SVGPathElement`, and `SegmentedTrace.mapToSvgElements` branches on exactly those two `instanceof` checks — so a trace-level highlight test for any segmented type (diverging included) cannot run under jest here; it falls through both branches. The new `selectorResolution.test.ts` works around this by building a realistic Frappe SVG and asserting each emitted selector resolves to exactly one element per point, in the emitted order. Real end-to-end highlight verification for the segmented types belongs in Playwright.

**Known gap, outside this change's file scope:** `scripts/build-site.js` lists the frappe example pages in the docs-site nav. The four new `examples/frappe-*.html` pages work standalone but are not yet linked from the built site nav, and `scripts/` was out of scope here. Someone with that access should add the four entries.

---

### DIFF
--- a/docs/frappe.md
+++ b/docs/frappe.md
@@ -80,7 +80,7 @@ Once the page loads, click on the chart (or Tab to it) and MAIDR activates with:
 The Frappe Charts adapter:
 
 1. **Reads data** - takes the chart's `{ labels, datasets }` and converts it to MAIDR's schema
-2. **Creates selectors** - generates CSS selectors (scoped to the chart container) for the Frappe SVG elements used in visual highlighting. Line and scatter traces highlight one element per point, so they target the per-point `<circle>` dots — keep `lineOptions.dotSize > 0`.
+2. **Creates selectors** - generates CSS selectors (scoped to the chart container) for the Frappe SVG elements used in visual highlighting. MAIDR highlights one element per data point, so every line-based chart — line, area, bump, scatter, dot — targets the per-point `<circle>` dots rather than the one `<path>` drawn for the whole series. Keep `lineOptions.dotSize > 0` and leave `hideDots` off.
 3. **Returns MAIDR config** - produces a complete `Maidr` object you set as the `maidr` attribute
 
 ### Wait for the chart to settle
@@ -98,16 +98,28 @@ The `activateMaidrWhenSettled` helper in the Quick Start handles this: it waits 
 | Bar | `'bar'` | `'bar'` |
 | Line | `'line'` | `'line'` |
 | Multi-line | `'line'` (multiple datasets) | `'line'` |
+| Area | `'line'` + `lineOptions: { regionFill: 1 }` | `'area'` (also inferred) |
+| Bump (rank over time) | `'line'` (one dataset per competitor) | `'bump'` |
 | Scatter | `'line'` + `lineOptions: { hideLine: 1 }` | `'scatter'` |
+| Dot plot | `'line'` + `lineOptions: { hideLine: 1 }` | `'dot'` |
+| Diverging bar | `'bar'` (two signed datasets) | `'diverging'` |
 | Mixed axis (bar + line) | `'axis-mixed'` | `'axis-mixed'` |
 | Pie | `'pie'` | `'pie'` |
 | Donut | `'donut'` | `'donut'` |
+
+The `chartType` names above are the **adapter's**, not Frappe's. Frappe draws several distinct statistical charts with the same `type: 'line'` or `type: 'bar'`, differing only in their options or in what the numbers mean — nothing a chart instance records — so naming the chart is how you tell MAIDR which one to announce.
 
 **Not supported:** Percentage charts (no MAIDR equivalent), and Frappe's calendar-style Heatmap (structurally unlike MAIDR's matrix heatmap).
 
 > **Pie note:** Frappe aggregates before it draws — it sums every dataset at each label, drops labels whose total is negative, and collapses everything past `maxSlices` (default 20) into one "Rest" wedge. The adapter reproduces that aggregation so each announced slice is the wedge it highlights. Pass the chart instance (not a plain `{ data }` object) when you override `maxSlices`, so the adapter can read it.
 
-> **Scatter note:** Frappe Charts v1.6.2 has no native `scatter` type. Render a dot plot with a line chart whose connecting line is hidden (`lineOptions: { hideLine: 1 }`); pass `chartType: 'scatter'` to the adapter so it is treated as a scatter plot. Frappe spaces x-axis labels evenly regardless of their numeric value, so use evenly spaced numeric labels.
+> **Area note:** `lineOptions.regionFill` is an instance field the adapter reads directly, so a chart passed as `chartType: 'line'` with the region filled is announced as an area chart anyway — you cannot mislabel one as the other. Pass `chartType: 'area'` when you hand the adapter a plain `{ data }` object, which has no instance to read. Several filled bands in one chart **overlap** in Frappe rather than stacking, so they stay independent series (`area`, never `stacked_area`). Keep `dotSize > 0`: the fill is one `<path>` for the whole series and cannot highlight individual points.
+
+> **Scatter / dot note:** Frappe Charts v1.6.2 has no native `scatter` type. Render either chart with a line chart whose connecting line is hidden (`lineOptions: { hideLine: 1, dotSize: 6 }`). Frappe places its marks at **evenly spaced label positions whatever the label holds**, so which of the two you have depends on the labels: numeric, evenly spaced labels are a scatter plot (`chartType: 'scatter'`), and category names are a Cleveland dot plot (`chartType: 'dot'`). Passing `'scatter'` with categorical labels is converted as a dot plot with a console warning, because `Number('Mon')` is `NaN` and a scatter layer built from those labels would have no x values at all.
+
+> **Bump note:** A bump chart is a multi-dataset line chart whose y values are **ranks** — one dataset per competitor, `1` = best. MAIDR inverts the pitch so first place is the highest note and announces the places gained or lost at each period, which is why the type has to be declared: nothing in the data distinguishes a rank from a magnitude. Emit the **true ranks**. Frappe v1.6.2 cannot invert or reverse an axis, so a rank chart it draws shows rank 1 at the *bottom*; pre-inverting the values to make the picture look right would make MAIDR announce the wrong ranks.
+
+> **Diverging note:** A diverging bar chart is two bar datasets with opposite signs, drawn either side of Frappe's zero line — negate the values of the series that should grow downwards and emit them signed, exactly as the chart draws them. MAIDR takes the magnitude for the pitch and names the side. The adapter throws unless the chart is exactly two datasets, one all-negative and one all-positive, rather than announce sides that are not there. Two limits are Frappe's: it has no horizontal bars, so the back-to-back population pyramid orientation is not drawable — only the vertical up/down form — and `barOptions: { stacked: 1 }` must **not** be used, because Frappe cumulates raw values and would stack the negative series onto the positive one.
 
 ## Code Examples
 
@@ -181,6 +193,123 @@ The `activateMaidrWhenSettled` helper in the Quick Start handles this: it waits 
     chartType: 'line',
     title: 'Monthly Product Sales',
     axes: { x: 'Month', y: 'Sales (thousands USD)' },
+  });
+</script>
+```
+
+### Area Chart
+
+```html
+<div id="area-chart"></div>
+<script>
+  const data = {
+    labels: ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun'],
+    datasets: [{ name: 'Rainfall', values: [42, 55, 71, 63, 48, 30] }],
+  };
+  // regionFill: 1 fills the band between the line and the baseline. The
+  // adapter reads it off the instance, so `chartType: 'line'` is announced as
+  // an area chart too — pass 'area' when handing over a plain { data } object.
+  const chart = new frappe.Chart('#area-chart', {
+    data,
+    type: 'line',
+    height: 400,
+    lineOptions: { regionFill: 1, dotSize: 5 },
+  });
+
+  activateMaidrWhenSettled(document.querySelector('#area-chart'), {
+    chartType: 'area',
+    title: 'Monthly Rainfall',
+    axes: { x: 'Month', y: 'Rainfall (mm)' },
+  });
+</script>
+```
+
+### Bump Chart (Rank Over Time)
+
+```html
+<div id="bump-chart"></div>
+<script>
+  // The values are RANKS, 1 = best. Emit them as ranks: MAIDR inverts the
+  // pitch itself, and Frappe cannot reverse an axis, so the drawn chart shows
+  // rank 1 at the bottom while the announcements stay correct.
+  const data = {
+    labels: ['Week 1', 'Week 2', 'Week 3', 'Week 4'],
+    datasets: [
+      { name: 'Alpha', values: [1, 2, 2, 1] },
+      { name: 'Beta', values: [2, 1, 3, 3] },
+      { name: 'Gamma', values: [3, 3, 1, 2] },
+    ],
+  };
+  const chart = new frappe.Chart('#bump-chart', {
+    data,
+    type: 'line',
+    height: 400,
+    lineOptions: { dotSize: 6 },
+  });
+
+  activateMaidrWhenSettled(document.querySelector('#bump-chart'), {
+    chartType: 'bump',
+    title: 'Weekly League Standings',
+    axes: { x: 'Week', y: 'Rank' },
+  });
+</script>
+```
+
+### Dot Plot
+
+```html
+<div id="dot-chart"></div>
+<script>
+  const data = {
+    labels: ['North', 'South', 'East', 'West', 'Central'],
+    datasets: [{ name: 'Sales', values: [64, 41, 78, 55, 92] }],
+  };
+  // Same rendering as the scatter plot below — a line chart with the line
+  // hidden. The labels are what tells the two apart: category names place one
+  // mark per category, which is a dot plot.
+  const chart = new frappe.Chart('#dot-chart', {
+    data,
+    type: 'line',
+    height: 400,
+    lineOptions: { hideLine: 1, dotSize: 8 },
+  });
+
+  activateMaidrWhenSettled(document.querySelector('#dot-chart'), {
+    chartType: 'dot',
+    title: 'Sales by Region',
+    axes: { x: 'Region', y: 'Sales (thousands USD)' },
+  });
+</script>
+```
+
+### Diverging Bar Chart
+
+```html
+<div id="diverging-chart"></div>
+<script>
+  // Two datasets with opposite signs. The downward series is NEGATIVE: MAIDR
+  // takes the magnitude for the pitch and names the side, so stripping the
+  // sign would leave it nothing to name the sides with. Do not set
+  // barOptions.stacked — Frappe cumulates raw values and would stack the
+  // negative series onto the positive one.
+  const data = {
+    labels: ['0-14', '15-29', '30-44', '45-59', '60-74', '75+'],
+    datasets: [
+      { name: 'Men', values: [-1200, -1150, -1080, -990, -720, -310] },
+      { name: 'Women', values: [1140, 1100, 1060, 1010, 830, 520] },
+    ],
+  };
+  const chart = new frappe.Chart('#diverging-chart', {
+    data,
+    type: 'bar',
+    height: 420,
+    colors: ['#4c72b0', '#c44e52'],
+  });
+
+  activateMaidrWhenSettled(document.querySelector('#diverging-chart'), {
+    chartType: 'diverging',
+    title: 'Population by Age Band',
+    axes: { x: 'Age band', y: 'People, thousands', z: 'Sex' },
   });
 </script>
 ```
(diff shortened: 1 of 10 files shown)


jooyoungseo:
All 14 checks were already green on `9dc0d5b`, so there was nothing to repair on the CI side. The review found no blocking issues and one nit, which I took.

**Fixed**
- `buildLineLayer`'s `type` parameter no longer defaults to `TraceType.LINE` (`src/adapters/frappe/converters.ts`). All three call sites in `buildLayers` pass the type explicitly, so the default was unreachable and contradicted the parameter's own doc comment. Making it required also stops a future caller from silently falling back to a line when it meant an area or a bump. The function is module-local and not exported, so this is not a public API change.

**Declined**
- Nothing. The rest of the review was confirmation of the trace wiring, selector scoping and sign validation, with no requested changes.
- The Copilot "quota limit" review and the preview-URL bot comment need no action.

Gates re-run after the change: `lint:fix`, `type-check`, `build`, and `npm test` (188 + 7 suites, 2713 tests) all pass locally. Pushed as `942db83`.

---

jooyoungseo:
All 14 checks are green on `942db83`, and the re-review's only prior nit (the dead `TraceType.LINE` default) was already taken in that commit. Nothing new to repair, so no code changed this round.

**Declined**

- The `hasNumericLabels` empty-`labels` edge case (`src/adapters/frappe/converters.ts`). It's correctly observed — `[].every(...)` is vacuously true, so `chartType: 'scatter'` with no labels builds a `ScatterLayer` rather than falling into the dot-plot warning — but the two branches are not meaningfully different here: both produce a layer with zero points, because the points are mapped from the same empty `labels` array either way. Guarding on `labels.length > 0` would only reroute an already-empty chart to an equally empty layer while printing a warning that says the chart "has non-numeric labels" when in fact it has none, which is worse than the current silence. Empty input is degenerate for every chart type in this adapter, not just scatter; if it's worth handling it belongs in one up-front validation, not in the scatter/dot disambiguation, and that's outside this PR's scope.
- The Copilot "quota limit" review and the preview-URL bot comment need no action.

---

PR #871, opened 2026-08-14
feat(vegalite): support the chart types added since the pie chart

### LINKED ISSUE #859: feat(vegalite): support the chart types added since the pie chart
## Problem

MAIDR core has gained ~34 new trace types since the pie chart landed (radar, gauge, funnel, waterfall, gantt, treemap, sankey, sunburst, dumbbell, dot/lollipop, diverging, area family, and more). The `vegalite` adapter has not kept up: of those, it currently emits **AREA, STACKED_AREA, NORMALIZED_AREA**.

A chart Vega-Lite draws perfectly well is therefore either dropped from the figure or announced as the wrong chart — which for a screen-reader user is worse than not being supported at all.

A single `switch` on the Vega-Lite **mark type string** in `resolveTraceType` (src/adapters/vegalite/converters.ts:643-709), refined by encoding channels: `bar` splits on `bin` → HISTOGRAM (:650) and on `color`/`fill` + `stack`/`xOffset` → DODGED/NORMALIZED/STACKED (:653-666); `line` splits on `mark.interpolate` → STEP vs LINE (:670-671); `area` splits on `stack` → NORMALIZED_AREA/STACKED_AREA/AREA (:677-690); `point|circle|square|tick` → SCATTER, `rect` → HEATMAP, `boxplot` → BOX, `arc` + `theta` → PIE; **every other mark falls through `default: return null` (:706-707)**. The mark string comes from `getMarkType` (:567-571), which reads only `spec.mark` — so the adapter is a Vega-Lite-*spec* reader; the compiled Vega `view` is used solely for row data (`resolveData`, :878) and DOM ordering, never to classify. `convertLayerSpec` calls it at :1447, drops the layer on null (:1449-1450), then a second `switch` on the resolved TraceType picks the extractor and selector builder (:1481-1583). A spec whose every layer resolves to null produces an empty figure and `bindVegaLite` refuses to bind it (src/vegalite-entry.ts:1654-1660) — the chart renders but is not keyboard-navigable at all.

## In scope — 11 trace types Vega-Lite can render

| TraceType | Vega-Lite construct | Detection / payload |
|---|---|---|
| `ERROR_BAR` | `errorbar` composite mark (and `errorband`); also `rule` with `y`+`y2` | Add `case 'errorbar':` / `case 'errorband':` to resolveTraceType (converters.ts:643-709) — exactly the by-name handling the other VL composite mark `boxplot` already gets at :698. Payload ErrorBarPoint[] `{x, y, yMin, yMax}`: read the compiled aggregate columns `lower_<valField>` / `upper_<valField>` / `center_<valField>` off the view dataset, reusing the sniff-then-fallback pattern extractBoxData already implements for `lower_box_<field>` at :1293-1300 (verify exact column names against a compiled view first). Orientation from the same x-quant/y-nominal test at :1572-1578. selectors.ts markToCssClass (:25-45) needs an entry — VL renders an errorbar as `mark-rule` + `mark-tick` groups, not a single path. |
| `DOT` | Cleveland dot plot — `point`/`circle` mark with one nominal/ordinal axis | Today `point\|circle\|square\|tick` map unconditionally to SCATTER (:691-695) and extractScatterData coerces BOTH channels with `Number()` (:1232-1235), so a categorical dot plot currently emits `x: NaN` on every point — this is a live correctness bug, not just a missing type. Branch: if exactly one of x/y is non-quantitative → DOT, and reuse `extractBarData` (:1031-1071), which already emits `{x: category, y: value}` and handles the horizontal (x-quant/y-nominal) flip. BarTrace serves DOT, so no new model work. |
| `GANTT` | Ranged bar / Gantt chart — `bar` (or `rule`) with `x`+`x2` (or `y`+`y2`) | `VegaLiteEncoding` (types.ts:125-147) does not model `x2`/`y2` at all — add them first. Then in the `bar` case (:649-669): both a positional channel and its `2` counterpart carry fields, and the other axis is nominal/ordinal → GANTT. Payload is `GanttData` (grammar.ts:579), not a flat array: group rows by the lane field into `points: GanttPoint[][]` with `{x: lane, start: row[xField], end: row[x2Field]}`, and fill `lanes` from `view.scale('y').domain()` (types.ts:36) so lanes with zero intervals still exist. `unit` has no Vega-Lite source — omit it rather than guess. |
| `WATERFALL` | Waterfall chart — `bar` with `y`+`y2` over a `window` running-sum transform (the official VL "Monthly Profit and Loss" example) | Same `x2`/`y2` encoding extension as GANTT. Discriminator against a plain ranged bar: `spec.transform` contains a `window` with `op: 'sum'` and both y bounds are data-driven rather than a constant baseline. Payload WaterfallPoint[] (grammar.ts:460): `start = row[yField]`, `end = row[y2Field]`, `delta = end - start`, `kind` from the sign — with `kind: 'total'` for the rows the spec's own `calculate` marks as Begin/End (the official example emits a label field for exactly those bars). Sharing the x2/y2 plumbing with GANTT makes these two one piece of work. |
| `POLAR_AREA` | Radial plot / coxcomb — `arc` mark with a `radius` encoding | The `arc` case (:704-705) currently returns PIE when `theta` exists and null otherwise, so a radial plot is either dropped outright or announced as a pie with its radius channel silently ignored — the code comment at :700-703 already names this gap. Add `radius` to VegaLiteEncoding and branch: arc + a `radius` **field** → POLAR_AREA. RadarTrace (model/radar.ts:47, extends LineTrace) reads LinePoint[][], so emit one series of `{x: category from color/fill or theta, y: radius value}`; the wedge selectors are the same `mark-arc` paths the PIE path already targets at :1553. |
| `LOLLIPOP` | Lollipop chart — layered `rule` (baseline → value) + `point` | `rule` hits `default: return null` (:706-707), so today the stem layer is dropped silently and only the head survives, as a SCATTER (a live example of this is the mean-rule layer in test/adapters/vegalite/fixtures/layeredNonLine.ts:215). In the layered path (:130-142) recognise a two-layer run of rule(with its second bound at zero/baseline) + point sharing an x field and collapse it into one layer typed LOLLIPOP with `extractBarData`'s BarPoint[] payload — the same collapse machinery `coalesceSiblingLineLayers` (:172-220) already provides for line runs. |
| `DUMBBELL` | Ranged dot plot / dumbbell — layered `rule`/`line` connector + `point`s (VL's "Ranged Dot Plot" example) | Same layered-collapse hook as LOLLIPOP; the discriminator is that the connector carries exactly two values per category — either via `x`+`x2` on a rule, or via a `detail`/`color` channel whose field has cardinality 2. Payload is `DumbbellData` (grammar.ts:651), an object not an array: `points: [{x: category, start, end}]` plus `startLabel`/`endLabel` taken from the two values of the grouping field in `view.scale('color').domain()` order — those labels are the whole content of the comparison and there is no other source for them. |
| `BUMP` | Bump chart — `line` (+`point`) over a `window: [{op: 'rank'}]` transform | In the `line` case (:670-671), scan `spec.transform` for `{window: [{op: 'rank', as: F}]}` and require `encoding.y.field === F` (a reversed y scale or descending sort is a corroborating but not required signal). Payload is exactly the LinePoint[][] `extractLineData` (:1190-1223) already builds — BumpTrace consumes the same shape and only inverts the pitch mapping — so this is a classification change with no new extractor. |
| `PARALLEL` | Parallel coordinates — `line` mark over a `fold` transform with a `detail` channel (official VL gallery example) | `line` + `spec.transform` containing `{fold: [...], as: [key, value]}` + `encoding.x.field === key` + a `detail` channel → PARALLEL. Payload LinePoint[][] with one inner array **per observation**, which needs a new grouping key: `extractLineData` groups only by `color`/`fill` (:1196), so it must also accept the `detail`/index field. Caveat worth planning around: the canonical VL example min-max-normalises before plotting, so per-axis real ranges have to be recovered from the pre-normalisation columns or ParallelTrace's per-column axes announce normalised units. |
| `RIDGELINE` | Ridgeline / joyplot — row-`facet`ed `area` over a `density` transform (official VL "Ridgeline plot") | `describeFacet` (:107-110) currently routes every faceted spec to `buildFacetMaidr`, so a ridgeline becomes N independent AREA subplots the reader must page between. Intercept before that call: facet is row-only (or `encoding.row`), child mark is `area`, and `transform` contains `{density: field, groupby: [...]}` → emit one RIDGELINE layer instead. Payload ViolinKdePoint[][] (model/ridgeline.ts:76): one inner array per group, `{x: group, y: value position, density: <density column>}` — the density transform's output columns are already on the compiled dataset the adapter reads. |
| `DIVERGING` | Diverging bar / population pyramid / Likert — `bar` with signed quantitative values (usually via `calculate: -datum.x`) | No spec-level marker exists; the only signal is in the resolved rows. After `extractSegmentedData` in the segmented branch (:1499-1526), if the quantitative channel carries both signs across series → DIVERGING rather than STACKED. DivergingTrace extends SegmentedTrace (model/diverging.ts:35), so the SegmentedPoint[][] payload is already the right shape — this is a one-line type swap. Honest caveat: an ordinary stacked bar containing a few negatives would be reclassified, so this likely belongs behind an opt-in flag on VegaLiteToMaidrOptions (types.ts:206) rather than firing automatically. |

## Deferred to a follow-up

These need a new author-facing opt-in API or semantics Vega-Lite does not carry in its own configuration, so they are out of scope here and tracked separately:

- `CHOROPLETH` — `geoshape` hits `default: return null` (:706-707), so a map produces zero layers and bindVegaLite skips binding entirely (vegalite-entry.ts:1654-1660) — the map is completely inaccessible today. Add `case 'geoshape'` → CHOROPLETH when a `color`/`fill` field is present. Payload ChoroplethPoint[] (gra

## Not renderable by Vega-Lite

<details><summary>14 trace types the library genuinely cannot draw</summary>

- ALLUVIAL — neither Vega-Lite nor Vega has a flow layout; every alluvial in either grammar is hand-computed ribbon geometry, so there is no mark or transform to key off.
- SANKEY — same: Vega has `linkpath` for drawing one edge but no sankey layout, and Vega-Lite has neither; community VL sankeys are stacks of rects and rules indistinguishable from an ordinary chart.
- CHORD — no chord/ribbon layout in either grammar; the arcs and ribbons are hand-computed.
- NETWORK — Vega's `force` transform draws node-link diagrams, but Vega-Lite has no node-link layout at all, and this adapter classifies solely off `spec.mark` (converters.ts:567), so it is unreachable without a separate Vega-spec reader.
- TREEMAP — Vega's `treemap` transform draws it; Vega-Lite has no hierarchy layout, so no VL spec can express one.
- SUNBURST — Vega's `partition` transform draws it; Vega-Lite has no hierarchy layout.
- ICICLE — same `partition` transform in Vega (rect instead of arc); no Vega-Lite equivalent.
- CONTOUR — Vega's `kde2d` + `isocontour` transforms draw it; Vega-Lite has no contouring transform.
- WORD_CLOUD — Vega's `wordcloud` transform draws it; Vega-Lite has no text-layout transform.
- RADAR — Vega-Lite's only polar geometry is the `arc` mark (theta/radius), with no polar line or area; Vega's radar example hand-computes x/y trig, producing a spec indistinguishable from an ordinary line chart.
- HEXBIN — neither grammar has a hexbin transform; the VL "hexbin" gallery example fakes the lattice with `calculate` formulas and a hand-written hexagon `shape` path, leaving nothing detectable.
- BOXEN — no letter-value composite mark in either grammar (unlike `boxplot` and `errorbar`); a hand-stacked pile of rects is indistinguishable from any other layered rect chart.
- FUNNEL — no funnel mark; it is drawn as a centred or normalised bar chart, which the adapter already reads correctly as BAR/STACKED, and nothing in the spec says "these bars are stages".
- GAUGE — no gauge or bullet mark; drawn as layered arcs or a bar+tick stack, with no signal identifying one measure against a range, a target, and named bands.

</details>

## Acceptance criteria

- [ ] Unit tests under `test/adapters/vegalite/` cover every newly detected type.

## Implementation notes

SCOPE: the adapter reads Vega-**Lite** specs only. `getMarkType` (converters.ts:567-571) reads `spec.mark`, a field a raw Vega spec (which uses a `marks[]` array) does not have; the compiled Vega `view` is consulted only for rows, scale domains and dataset names. So "Vega can draw it" is not sufficient for feasibility here — several notRenderable entries above are Vega-renderable but unreachable from the spec shape this adapter parses.

THIRD BUCKET — renderable by Vega-Lite, but NOT detectable from the spec (deliberately kept out of both lists above, since claiming them means guessing):
- SURVIVAL — a Kaplan-Meier curve is a `line` with `interpolate: 'step-after'` plus a censoring tick layer and an `errorband`. The adapter already emits STEP for it correctly. Nothing distinguishes it from any other step chart except heuristics (y in [0,1], monotone non-increasing, starts at 1), and `censored` flags have no VL encoding.
- MANHATTAN / VOLCANO — both are plain scatters; the adapter already emits SCATTER, which is right. The extra payload (point identity, thresholds) has no Vega-Lite home, and misclassifying an ordinary scatter would cost more than it gains. Worth revisiting only if `thresholdOptions` can be passed through VegaLiteToMaidrOptions.
- FOREST — drawable as `errorbar` + `point` + a null `rule`, so it will fall out of the ERROR_BAR work as an ERROR_BAR layer. `weight` could plausibly come from a `size` encoding, but `pooled` (which row is the summary) is unmarked in any VL spec, so promoting to FOREST would need an explicit caller hint.
- MOSAIC — there is a real signal (a `rect` with both `x`+`x2` and `y`+`y2` from stacked `window` transforms), but the VL mosaic construction is a community pattern, not a native mark, and recovering `width`/`count` means reverse-engineering the transform chain. Reachable, but I would not plan it before the x2/y2 work for GANTT/WATERFALL exists.

SEQUENCING: three of the twelve gaps share one prerequisite — adding `x2`/`y2` to `VegaLiteEncoding` (types.ts:125-147) unlocks GANTT, WATERFALL and part of DUMBBELL. Two more share the layered-collapse hook (LOLLIPOP, DUMBBELL), which should extend `coalesceSiblingLineLayers` (:172-220) rather than duplicate it. ERROR_BAR and DOT are the two clear first moves: ERROR_BAR because `boxplot` (:698) is an exact precedent for a composite mark, DOT because it fixes a live NaN bug.

TRAPS:
1. Adding a mark to `resolveTraceType` is only half the job — `markToCssClass` (selectors.ts:25-45) must learn the mark's Vega CSS class or highlighting silently matches nothing. Its `default` returns `mark-<mark>`, which is right for `rule`/`arc` but wrong for `geoshape` (renders as `mark-shape`) and for composite marks that expand into several groups.
2. `extractScatterData` (:1225-1236) `Number()`s both channels; any new categorical-axis payload must not route through it.
3. Layers that resolve to null vanish without warning (:1449-1450) — only a spec whose layers ALL resolve to null produces the console warning (vegalite-entry.ts:1654-1660), so a dropped rule/errorbar layer inside a layered chart is currently invisible. A `console.warn` on the null return would be a cheap standalone improvement.
4. Vega/vega-lite are not dependencies of the repo at all (not in `dependencies`, `peerDependencies`, or `devDependencies`) — tests drive the converter with hand-written spec fixtures and a stubbed `VegaView` (test/adapters/vegalite/fixtures/testView.ts). New work follows that convention: a spec fixture plus a fake view, no live rendering. Any claim about compiled column names (`lower_<field>`, density outputs, window sums) is therefore untested against a real compiler and should be verified against a real view before coding.
5. Examples exist for bar, box (v+h), dodged, heatmap, histogram, line, normalized, scatter, stacked, pie, facet-bar and hconcat-box (examples/vegalite-*.html) — note there is **no** area example even though AREA/STACKED_AREA/NORMALIZED_AREA are supported, so the example set is not a reliable index of adapter coverage.

jooyoungseo:
## Description

The Vega-Lite adapter stopped at the pie chart, so every trace type the grammar has gained since then fell through to `null` (left unbound) or, worse, to a neighbouring type that announced the wrong numbers. This PR extends the adapter to read eleven of those types.

The work is adapter-local: it converts Vega-Lite's own model into the MAIDR JSON grammar and builds the highlight selectors for it. No file under `src/model/`, `src/service/`, `src/state/`, `src/ui/`, `src/command/`, or `src/type/grammar.ts` is touched — the core already supported every type here.

Every claim was verified against a real compiled Vega view rather than inferred from the spec: `vega`/`vega-lite` were installed in a scratch directory outside the tree, the adapter was bundled against it, and `vegaLiteToMaidr(spec, view)` was run on live views. Emitted selectors were then resolved through jsdom against the real `view.toSVG()` output to confirm they match the elements they claim to. The fixture rows in the new tests are captured from those views.

## Related Issues

Closes #859

### Trace types added

| Type | Recognised from |
|---|---|
| `ERROR_BAR` | `errorbar` / `errorband` marks, across all four `extent` values |
| `DOT` | a point-family mark with one positional channel nominal/ordinal |
| `GANTT` | a `bar` with `x` + `x2` (or `y` + `y2`) against a categorical lane axis |
| `WATERFALL` | the same ranged bar, plus a `window` sum building a running total |
| `POLAR_AREA` | an `arc` whose `radius` reads a data field |
| `LOLLIPOP` | a `rule` + `point` layer pair sharing category and value channels |
| `DUMBBELL` | a connector + `point` pair, two values per category told apart by `color` |
| `BUMP` | a `window` `rank`/`dense_rank` whose output column is the one on `y` |
| `PARALLEL` | a `fold` transform with `detail` splitting the polylines |
| `RIDGELINE` | a `row` facet over a `density` transform grouped by the facet field |
| `DIVERGING` | a stacked bar with each series wholly one side of the baseline |

Highlighting is wired in the same change as the detection for each of these. It is deliberately **withheld** — rather than pointed at the wrong element — in three cases, each documented at the point it happens: an `errorband` draws every sample into a single `<path>`; an `errorbar` with `ticks: true` moves the whip out of the layer group; and a gantt whose lanes are interleaved in the source rows cannot be sliced back apart in DOM order.

### Types deliberately NOT added

These are the remaining types the grammar has gained, and why each is absent:

- **`ALLUVIAL`, `CHORD`, `NETWORK`, `SANKEY`, `TREEMAP`, `SUNBURST`, `ICICLE`, `WORD_CLOUD`** — Vega-Lite has no graph or hierarchy layout; these are authored in Vega proper, which this adapter does not read. There is no Vega-Lite spec to detect.
- **`BOXEN`, `FOREST`, `FUNNEL`, `GAUGE`, `MANHATTAN`, `MOSAIC`, `SURVIVAL`, `VOLCANO`, `HEXBIN`, `RADAR`** — each is a domain convention drawn with ordinary marks. Separating one from the plain chart it is built out of means guessing intent from field names, and a wrong guess misannounces a chart that currently reads correctly. Detecting them needs an author-supplied hint, which is a grammar question rather than an adapter one.
- **`CHOROPLETH`** — `geoshape` is an unambiguous mark and is the clearest candidate for a follow-up, but the spatial handling it needs was out of scope for this pass.
- **The second dumbbell shape** — a `rule` with `x` + `x2` plus two separate point layers. It is three layers, and its rule-plus-`x2` shape is indistinguishable from a rule-drawn gantt. `GANTT` is therefore restricted to the `bar` mark and a bare `rule` still resolves to null outside the paired-layer pre-pass.

### Fixes to existing behaviour

- `markToCssClass('tick')` returned `mark-tick`, which matches nothing in any Vega SVG — Vega-Lite compiles `tick` to a Vega `rect`. Corrected, and `buildSelector`'s child element now comes from `markToChildElement`.
- Two facet tests expected `SCATTER` for a point layer with a nominal `x`. That layer was emitting `x: NaN` on every point; the expectations are now `DOT`, with a comment.
- `resolveData` gained an optional `requiredFields` filter. Without it the parallel-coordinates polylines were handed the axis-rule layer's `{key, __count}` tally, because the layer index and the compiled dataset index do not line up.

### Supporting work

- New helpers in `converters.ts` (`isCategorical`, `hasField`, `isHorizontalEncoding`, `readScaleDomain`, `resolveRangedFields`, `resolveCategoryChannels`, `resolveFoldedAxes`, `resolveFacetCellKeys`, `sidesDiverge`, and others), all reused across more than one type.
- `types.ts` widened for the channels and transforms these specs use (`x2`, `y2`, `radius`, `detail`, `window`, `fold`, `density`, `groupby`, `as`, mark `extent`).
- Seven runnable examples and a documentation pass on `docs/vegalite.md` covering all eleven types.

## Checklist

The two unticked boxes are unticked on purpose: `CONTRIBUTING.md` was not read as part of this work, and the `ManualTestingProcess.md` walkthrough was not performed — no browser or e2e run was done. The automated gate below was run in full.

### Verification

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm run sync:copilot:check` all exit clean. `npm test` is green: **196 suites / 2691 tests** (unit + component, up from 192 / 2647 on `main`) and **7 suites / 73 tests** (esm). No e2e run.

Note for anyone running the suite directly: `npx jest test/` reports the 7 esm suites as "failed to run". That is pre-existing and environmental — use `node scripts/test.js`, or `--selectProjects esm` with `NODE_OPTIONS=--experimental-vm-modules`.

### Known limitations and follow-ups

Found and verified during this work, left alone as out of scope:

- `sortSimpleBarsByVisualOrder` and `sortLinesByVisualOrder` live in `src/vegalite-entry.ts` and filter on `TraceType.BAR` / `TraceType.LINE` only, so `DOT`, `LOLLIPOP`, `GANTT`, `BUMP` and `PARALLEL` navigate in data order rather than visual order. Not currently wrong for any of them — Vega emits the sibling groups in data order — but it is a latent trap.
- The gallery's "Diverging Stacked Bar Chart (with Neutral Parts)" uses `x` + `x2` and is classified `GANTT`. Reading it as `DIVERGING` needs an `x2`-aware segmented extractor plus a decision about what "side" means for a bar that never crosses zero. A partial fix trades one misannouncement for another, so the `GANTT` reading — which at least reports the right span lengths — stands.
- Pre-existing facet ordering bug: `buildFacetMaidr` takes panel order from `row_domain`, so for any facet field whose first-seen order differs from its sorted order, panel titles and `stampFacetCells` pairing are off by a permutation. `resolveFacetCellKeys` (added here for ridgeline) is the fix, but wiring it in ripples through `facets.test.ts`.
- The gallery's own `line_bump` example is a plain nominal-`y` line, not a rank window. It still resolves to `LINE`, and `extractLineData` `Number()`s `y`, giving `y: NaN` on every point — the same class of bug as the `DOT` fix above.
- The paired-layer pre-pass only runs for top-level layered specs, so a lollipop nested inside a concat is not collapsed.

---


jooyoungseo:
Both bugs from the review were real — I reproduced each against the converter before touching it — and both are fixed in `4d93610`.

**Dumbbell false positive.** A `line` + `point` pair over an ordinal x with a two-value colour field and one row per category came back as `dumbbell` with `startLabel: 'A'`, `endLabel: 'B'`: the ordinary line-with-markers idiom announced as a comparison of its two series. A `line` connector now has to name the category in `detail`, the way Vega-Lite's Ranged Dot Plot does; without it the pair falls through and the layers convert as `LINE` + `DOT`. A `rule` is unchanged — it is one segment per row by construction and draws no series of its own. Regression test in `pairedLayers.test.ts`.

**Horizontal waterfall.** Confirmed: `x: previous_sum, x2: sum` against a categorical y with the `window` sum came back as `gantt` with the running totals announced as start/end times. Both branches of `resolveRangedBarType` now check `hasRunningSumTransform`. One thing the finding didn't cover: `WaterfallTrace` carries no orientation and reads the step label from `axes.x`, so the sideways layout also has to swap the two axis titles or the step is announced under the amount axis' title. The test asserts that.

**Minor items.** Took the untyped-category-axis warning (`console.warn`, matching the adapter's existing unsupported-shape warnings, plus a docs note), the linear dumbbell end scan with an early bail on a third value, and the removal of the `spec.layer!` assertions.

**Declined, with reasons.**

- *Splitting `converters.ts` into modules.* Agreed on the direction, but a ~3,600-line file split is a pure-motion change that would bury these fixes in a diff nobody can review against the original. Better as its own PR before the next batch of chart types.
- *Modelling `ticks` in `types.ts` to make the errorbar claim testable.* The comment describes what Vega renders — the extra composite parts shift the whip out of `layer_<N>_marks`, so `ErrorBarTrace`'s count check fails at runtime. The adapter has no branch on `ticks` and adding an unused field to the spec type wouldn't create one; a unit test over the spec can't observe it either way. I'd rather leave the type honest about what the adapter reads.

Everything green locally at this head: lint, type-check, build, and 2694 + 73 tests. CI for the new commit hadn't finished when I wrote this.

---

PR #872, opened 2026-08-14
feat(chartjs): support the chart types added since the pie chart

### LINKED ISSUE #860: feat(chartjs): support the chart types added since the pie chart
## Problem

MAIDR core has gained ~34 new trace types since the pie chart landed (radar, gauge, funnel, waterfall, gantt, treemap, sankey, sunburst, dumbbell, dot/lollipop, diverging, area family, and more). The `chartjs` adapter has not kept up: of those, it currently emits **AREA, STACKED_AREA**.

A chart Chart.js draws perfectly well is therefore either dropped from the figure or announced as the wrong chart — which for a screen-reader user is worse than not being supported at all.

Dispatch is a single `switch` on the Chart.js controller name. `extractChartData` reads `chart.config.type` at `src/adapters/chartjs/extractor.ts:74` and hands it to `extractLayers`, whose switch at `src/adapters/chartjs/extractor.ts:561-586` routes `bar | line | scatter | bubble | pie | doughnut | boxplot | candlestick | ohlc | matrix` to a per-type extractor and **throws** for everything else (`extractor.ts:579-585`); `plugin.tsx:302-310` catches that throw and leaves the chart un-instrumented with a `console.warn`. There is no per-dataset dispatch — `ChartJsDataset.type` is declared at `types.ts:63` but never read, so a mixed bar+line chart is extracted wholly as its `config.type`. Two secondary decisions refine the type *inside* a branch: `extractBarLayers` picks `STACKED` vs `DODGED` vs `BAR` from dataset count plus `isStacked(chart)` (scales `.stacked`) at `extractor.ts:601-606`, and `extractLineLayers` buckets datasets by the pair (`stepDirectionOf`, `isFilledLine`) at `extractor.ts:831-842` and then maps each bucket through `lineGroupType` (`extractor.ts:775-783`) to `STACKED_AREA | AREA | STEP | LINE`. No heuristics run on the data values themselves.

## In scope — 11 trace types Chart.js can render

| TraceType | Chart.js construct | Detection / payload |
|---|---|---|
| `RADAR` | Chart.js native `type: 'radar'` (RadarController, core) | Add `case 'radar':` to the switch at extractor.ts:561. It is currently the *explicit* rejection case — test/adapters/chartjs/extractor.test.ts:460-471 asserts `extractChartData` throws for it. Payload is exactly what extractLineLayers already builds: `LinePoint[][]`, one row per dataset, one column per spoke — `{x: data.labels[i], y: value, z: dataset.label}` — since RadarTrace extends LineTrace (src/model/radar.ts:47) and only re-pans by spoke angle. Two adjustments: a radar has no x/y scales, so `getAxisLabel` (extractor.ts:463) would return the 'X'/'Y' fallback — read `chart.options.scales?.r?.title?.text` and default the way `getPieAxes` does (extractor.ts:957-962). Then add RADAR to the LINE/AREA/STEP case in highlightTargets.ts:148-159 (one `finiteIndices` row per backing dataset); `resolveActiveTargets` already falls through to the bar/line branch (highlightTargets.ts:231-239) correctly, and radar draws PointElements so overlay.ts:353-362's point branch needs no change. |
| `POLAR_AREA` | Chart.js native `type: 'polarArea'` (PolarAreaController, core) | Add `case 'polarArea':` alongside `radar` in extractor.ts:561 — same `LinePoint[][]` payload (typically one row), same RadarTrace consumer (src/model/factory.ts:145-148 routes POLAR_AREA and RADAR to the same class). Same `scales.r` axis-label fix. Highlight is free in both directions: polarArea draws ArcElements, and overlay.ts:281-295 already emits a `wedge` shape from `outerRadius`/`startAngle`/`endAngle` for the pie path. |
| `GANTT` | Chart.js floating bars — `type: 'bar'`, `indexAxis: 'y'`, `data: [[start, end], …]`, usually against a time x-scale (core, v3+) | Detect a two-number tuple entry in `dataset.data`. Note this is currently a silent-data-loss bug, not just a missing type: `toFiniteNumber` (extractor.ts:495-505) sees an array as an object with no `y`/`v` key and returns `null`, and `singleDatasetToBarLayer` (extractor.ts:622-629) skips every null — so a Chart.js gantt today extracts as a BAR layer with zero points. Add an `isRangeValue` guard beside `isBoxplotValue`/`isMatrixValue` (extractor.ts:511-534) and widen `ChartJsDataValue` (types.ts:13-26) with `[number, number]`. Payload is `GanttData` (grammar.ts:579-612): `points: GanttPoint[][]` with one lane per `data.labels[i]` (or per dataset when several datasets share the label axis), each `{x: lane, start: v[0], end: v[1], label: dataset.label}`, plus `lanes` from `data.labels` so empty lanes survive, and `unit` from `plugins.maidr` or the time scale's unit. Disambiguate from WATERFALL by `indexAxis: 'y'` + a time/linear x-scale and non-chaining intervals. |
| `WATERFALL` | Chart.js floating bars — `type: 'bar'` with `data: [[start, end], …]`; this is Chart.js's own documented waterfall recipe (core, v3+) | Same tuple guard as GANTT (and same current silent-empty-layer bug via extractor.ts:495). The signal that separates a waterfall from a range bar is chaining: `end[i] === start[i+1]` across consecutive entries, on the default `indexAxis: 'x'`. Payload is `WaterfallPoint[]` (grammar.ts:460-478): `{x: labels[i], start: v[0], end: v[1], delta: v[1]-v[0], kind}` where `kind` is `'total'` when the bar sits on the baseline (start equals 0 / the value-scale minimum — the opening, closing and subtotal bars) and otherwise `'increase'`/`'decrease'` from the sign of delta. Highlight needs no new geometry: these are BarElements with `base`, covered by overlay.ts:298-315; highlightTargets would treat it like BAR (one row, `finiteIndices` replaced by an index list over the tuple entries). |
| `DIVERGING` | Chart.js `type: 'bar'` with `indexAxis: 'y'` and stacked scales, one series negated — the standard Chart.js population-pyramid / Likert recipe (core) | In `extractBarLayers` at extractor.ts:601-603, where `isStacked(chart)` currently picks STACKED unconditionally: additionally test whether the datasets partition cleanly into an all-≤0 group and an all-≥0 group (both non-empty) and, if so, emit `TraceType.DIVERGING`. The payload needs no new builder at all — `extractSegmentedBarLayers` (extractor.ts:644-686) already produces `SegmentedPoint[][]`, and DivergingTrace subclasses SegmentedTrace and reads the sign itself (src/model/diverging.ts:34, which pitches on \|value\| and announces the side). One caution: the gap→0 collapse at extractor.ts:665 is fine here, but the sign must not be normalised on the way out. |
| `NORMALIZED_AREA` | Chart.js `type: 'line'` with `fill` and `scales.y.stacked: true`, values precomputed as percentages (core — Chart.js has no `stack: 'normalize'` mode of its own) | Extend `lineGroupType` (extractor.ts:775-783), which today returns STACKED_AREA for every filled+stacked group. Before returning, sum each category column across the group's datasets; when every column sums to the same constant (100 or 1) within a small tolerance, return `TraceType.NORMALIZED_AREA` instead. The emitted payload is byte-identical to the STACKED_AREA path (`LinePoint[][]` from `linePoints`, extractor.ts:809-822) and AreaTrace already handles all three area variants from `layer.type` (src/model/factory.ts:56-59), so only the type string changes. Honest caveat: unlike Vega-Lite, which reads a declared `stack: 'normalize'` (src/adapters/vegalite/converters.ts:678-681), Chart.js states nothing — this is a value heuristic and will miss a normalized chart whose totals were rounded unevenly. |
| `BUMP` | Chart.js `type: 'line'` with `scales.y.reverse: true` and integer rank values — the standard Chart.js bump-chart recipe (core) | In `extractLineLayers`, before the group loop at extractor.ts:831: if the y scale declares `reverse: true` (a field to add to `ChartJsScale`, types.ts:113-133) and, at every x index, the values across datasets are a permutation of 1..N, emit `TraceType.BUMP` for the unstepped/unfilled group. The permutation test is what makes this safe — a reversed y-axis alone is not enough. Payload is the existing `LinePoint[][]` unchanged (BumpTrace reads it as a multi-line layer with inverted pitch, src/model/bump.ts), and highlightTargets only needs BUMP added to the LINE case at highlightTargets.ts:148. |
| `GAUGE` | Chart.js `type: 'doughnut'` with `options.circumference: 180, rotation: 270` and a two-value dataset (value + remainder) — the doughnut-gauge idiom, no plugin needed | In `extractPieLayers` (extractor.ts:973-1007), branch when `chart.options.circumference` is present and < 360 and the dataset has exactly two finite values. Payload is `GaugePoint` (grammar.ts:530-543), a single object rather than an array: `{value: data[0], min: 0, max: data[0]+data[1], label: labels[0] ?? dataset.label}`, with `target`/`bands` only from `plugins.maidr`. Highlight is free — the ArcElement wedge branch (overlay.ts:281-295) already covers it. Caveat worth stating in the plan: a 180° doughnut is also a legitimate half-pie, so the safe shape is an explicit `plugins.maidr` declaration with this geometry as the auto-detect fallback; `MaidrPluginOptions` (types.ts:192-205) has no trace-type override field today. |
| `DOT` | Chart.js `type: 'line'` with `showLine: false` on a category axis — Chart.js's own way to draw points without a connecting line, i.e. a Cleveland dot plot | Add `showLine?: boolean` to `ChartJsDataset` (types.ts:60-92) and read it with the same dataset-over-`options.showLine` precedence `stepDirectionOf` uses (extractor.ts:709-719). When every dataset in a line group has `showLine === false` and the x axis is categorical, emit `TraceType.DOT`. The payload changes shape, which is the one real cost: DOT routes to BarTrace (src/model/factory.ts:69-71) and so wants a flat `BarPoint[]` — `{x: label, y: value}`, as `singleDatasetToBarLayer` builds at extractor.ts:609-642 — not the `LinePoint[][]` the line path emits. Modest payoff: a dot plot already navigates correctly as BAR, so the gain is that the chart announces itself as what the author drew. |
| `DUMBBELL` | Chart.js horizontal floating bars — `indexAxis: 'y'`, `data: [[start, end], …]` with round point styling; also drawn as two point datasets plus a thin connecting bar | Same tuple guard as GANTT/WATERFALL (and same current empty-layer bug, extractor.ts:495). Payload is `DumbbellData` (grammar.ts:653-665): `{points: [{x: labels[i], start: v[0], end: v[1]}], startLabel, endLabel}` with the two end names from `plugins.maidr` or the two dataset labels. Detection is the weak point and should be stated plainly: a one-interval-per-category horizontal floating bar is shape-identical to a one-interval-per-lane gantt, so the disambiguation has to be either the x scale kind (time ⇒ gantt, linear ⇒ dumbbell) or an explicit `plugins.maidr` trace-type declaration. Recommend building the tuple guard once for GANTT/WATERFALL/DUMBBELL and adding the opt-in field at the same time. |
| `SURVIVAL` | Chart.js `type: 'line'` with `stepped: 'after'` — how a Kaplan-Meier curve is drawn in Chart.js (core) | Nothing in the Chart.js config distinguishes a KM curve from any other staircase, so today it correctly falls out as STEP (extractor.ts:782) and this needs an explicit `plugins.maidr` trace-type declaration. Payload is `SurvivalPoint[][]` — a `StepPoint` extended with the censoring flag (grammar.ts:1034) — and the censored marks and at-risk counts are not in Chart.js's data model either; the page author would ride them on the point objects (`{x, y, censored: true}`, which Chart.js ignores as extra props) and `isPointValue` (extractor.ts:507-509) already tolerates them. Worth doing only alongside the general opt-in mechanism, since the automatic path genuinely cannot see it. |

## Deferred to a follow-up

These need a new author-facing opt-in API or semantics Chart.js does not carry in its own configuration, so they are out of scope here and tracked separately:

- `VOLCANO` — Renderable natively but not inferable: it is indistinguishable from any scatter, so it needs a `plugins.maidr` trace-type declaration plus `thresholdOptions` (MaidrLayer, grammar.ts:1195-1196) for the fold-change and significance cutoffs. `VolcanoPoint extends ScatterPoint` (grammar.ts:895) with the
- `MANHATTAN` — Same situation and same class as VOLCANO (src/model/factory.ts:129-134 routes both to VolcanoTrace), so it rides on the same three pieces of work: a `plugins.maidr` trace-type declaration, `thresholdOptions` passthrough, and a per-point label preserved through `datasetToScatterPoints` (extractor.ts:

## Not renderable by Chart.js

<details><summary>19 trace types the library genuinely cannot draw</summary>

- ALLUVIAL — no Chart.js controller draws stage-to-stage flows; chartjs-chart-sankey (kurkle) covers sankey only and is not a repo dependency.
- CHORD — no Chart.js controller or community plugin draws a circular chord diagram.
- SANKEY — core cannot draw ribbons; chartjs-chart-sankey exists but the repo does not depend on it (only boxplot, financial, matrix are used).
- NETWORK — no force-directed / node-link controller in core or in any plugin the repo uses.
- TREEMAP — core cannot draw nested rectangles; chartjs-chart-treemap (kurkle) exists but is not a repo dependency.
- SUNBURST — no hierarchy controller. A multi-dataset doughnut draws concentric rings, but the datasets carry no parent-child relation, so the tree TreemapTrace needs is simply absent from the data.
- ICICLE — no depth-band layout in core or any plugin the repo uses.
- CHOROPLETH — core has no geographic projection; chartjs-chart-geo (sgratzl) provides `choropleth`/`bubbleMap` but is not a repo dependency.
- HEXBIN — no hexagonal binning controller anywhere in the Chart.js ecosystem the repo uses; the matrix plugin draws rectangular cells only.
- CONTOUR — no iso-line controller. Several line datasets can be drawn, but the level value each curve represents exists nowhere in the config, so nothing recoverable identifies them as contours.
- MOSAIC — Chart.js resolves bar sizing (`barThickness`, `barPercentage`, `categoryPercentage`) at the dataset/controller level, not per data index, so per-bar variable widths — the half of the table a mosaic exists to encode — cannot be drawn.
- PARALLEL — a line dataset plots against one value scale, so per-axis scales for a parallel-coordinates plot are not expressible; chartjs-chart-pcp (sgratzl) exists but is not a repo dependency.
- ERROR_BAR — core has no error-bar element; @sgratzl/chartjs-chart-error-bars adds `barWithErrorBars`/`lineWithErrorBars` but the repo does not depend on it. (See notes: this is the cheapest dependency to add.)
- FOREST — same gap as ERROR_BAR, plus per-study weight and a pooled-row flag that no Chart.js data shape carries.
- BOXEN — @sgratzl/chartjs-chart-boxplot registers `boxplot` and `violin` only; there is no letter-value / variable-depth quantile controller.
- RIDGELINE — the boxplot plugin's `violin` draws a KDE outline, not offset ridgelines. Core can stack several filled lines, but the vertical offsets would be baked into the values and are not recoverable, so each group's own curve cannot be reconstructed.
- FUNNEL — no tapering-mark controller in core; a descending horizontal bar chart is a bar chart, and chartjs-funnel-chart is not a repo dependency.
- LOLLIPOP — no lollipop mark. The thin-bar-plus-point-dataset hack is undetectable from the config, and anything drawn that way already reads correctly as BAR (src/model/factory.ts:69-71 routes BAR/DOT/LOLLIPOP to one class), so nothing is lost.
- WORD_CLOUD — chartjs-chart-wordcloud (sgratzl) exists but is not a repo dependency; core has no text-layout controller.

</details>

## Acceptance criteria

- [ ] Unit tests under `test/adapters/chartjs/` cover every newly detected type.

## Implementation notes

Plugin set actually in use (this is the whole of "already depends on"): `@sgratzl/chartjs-chart-boxplot` v4, `chartjs-chart-financial` 0.2.1, `chartjs-chart-matrix` v2, and `chartjs-adapter-luxon` (a date adapter, not a controller) — see examples/chartjs/*.html and docs/chartjs.md:78-80. `chart.js ^4.0.0` is the only peer dependency (package.json).

Three shared pieces of groundwork carry most of the missingFeasible list, and are worth landing before any individual trace type:

1. **Floating-bar support is a prerequisite for GANTT, WATERFALL and DUMBBELL, and is a live bug today.** `ChartJsDataValue` (src/adapters/chartjs/types.ts:13-26) has no `[number, number]` member and `toFiniteNumber` (extractor.ts:495-505) returns `null` for an array — so any Chart.js floating-bar chart currently extracts as a BAR layer with **zero points**, silently, with no warning. Fixing the guard is a prerequisite for all three and a correctness fix on its own.

2. **`MaidrPluginOptions` has no trace-type override.** types.ts:192-205 carries only `enabled`, `title`, `axes`, `highlightColor`. SURVIVAL, VOLCANO, MANHATTAN, and the safe form of GAUGE and DUMBBELL all depend on an author declaration, so one `traceType?: TraceType` field unlocks a whole tier. It also gives every heuristic above (normalized-area column sums, bump permutation test, diverging sign split) an explicit escape hatch.

3. **Radar and polarArea are the only native Chart.js controllers the adapter rejects outright**, and they cost the least: same `LinePoint[][]` payload the line path already builds, and the overlay needs no change in either case (polarArea reuses the pie wedge branch at overlay.ts:281-295, radar reuses the point branch at overlay.ts:353-362). extractor.ts's own header comment (lines 10-12) names them as deliberately rejected, and test/adapters/chartjs/extractor.test.ts:460-471 pins that behaviour — that test has to be updated, not just extended.

Also worth knowing for planning:

- **`violin` is a free adjacent win.** `@sgratzl/chartjs-chart-boxplot` registers both `boxplot` and `violin` (examples/chartjs/boxplot.html:8), and the adapter maps only `boxplot` (extractor.ts:572-573). VIOLIN_KDE/VIOLIN_BOX are pre-pie types so they are out of scope for this survey, but the plugin is already a documented dependency and the payload is adjacent to `extractBoxplotLayers`.
- **@sgratzl/chartjs-chart-error-bars is the highest-value new dependency**: its datasets are `{y, yMin, yMax}`, which maps 1:1 onto `ErrorBarPoint` (grammar.ts:380-390), and it unlocks FOREST cheaply on top.
- **Mixed charts are a latent trap.** Dispatch is purely on `chart.config.type` (extractor.ts:74); `ChartJsDataset.type` is declared (types.ts:63) but never read, so a `type: 'bar'` chart with a `type: 'line'` dataset extracts every dataset as bars. Any new type added by dataset shape rather than chart type should not deepen that assumption.
- **Axis labels assume x/y scales exist.** `getAxisLabel` (extractor.ts:463-477) falls back to the literal `'X'`/`'Y'`. Radar and polarArea use an `r` scale and gauges use none, so each needs the `getPieAxes` treatment (extractor.ts:948-962) — name what the positions mean rather than letting the fallback announce a nonexistent axis.
- **Test conventions to follow:** test/adapters/chartjs/areaFill.test.ts is the cleanest model for a new type — a small `lineChart(...)`/`series(...)` fixture pair plus `it.each` over the config variants, asserting `layers[0].type`. Add an example page under examples/chartjs/ (each loads `chart.js@4` from jsDelivr plus `../../dist/chartjs.js`) and a row in the docs/chartjs.md:73-83 support table, which is currently stale — it lists neither Area nor Stacked Area even though both are emitted.

jooyoungseo:
## Description

The Chart.js adapter had fallen behind MAIDR core: of the trace types added since the pie chart it emitted only `AREA` and `STACKED_AREA`. Charts Chart.js draws perfectly well were either announced as the wrong figure or dropped from the figure entirely — which for a screen-reader user is worse than no support at all.

This adds all 11 in-scope trace types from #860. It is adapter-local: nothing under `src/model/`, `src/service/`, `src/state/`, `src/ui/`, `src/command/` or `src/type/grammar.ts` is touched, because the core already supports every one of these payloads.

It also fixes a live correctness bug found along the way. `toFiniteNumber` returned `null` for an array, so **every Chart.js floating-bar chart extracted as a `BAR` layer with zero points** — silently, with no warning. The `isRangeValue` guard that gantt, waterfall and dumbbell are built on fixes that on its own.

## Related Issues

Closes #860

### Trace types added (11 of 11 in scope)

| Trace type | How it is detected |
|---|---|
| `RADAR` | native `type: 'radar'` — previously an explicit rejection |
| `POLAR_AREA` | native `type: 'polarArea'` — previously an explicit rejection |
| `GANTT` | `[start, end]` bar data that does not chain; one lane per label |
| `WATERFALL` | one series of `[start, end]` bars on a vertical index axis where each step begins where the last ended |
| `DIVERGING` | stacked bars whose datasets each sit wholly on one side of the baseline, both sides occupied |
| `NORMALIZED_AREA` | stacked filled lines whose every category totals the same whole (100 or 1) |
| `BUMP` | `scales.y.reverse` **and** values at every period a permutation of `1..N` |
| `GAUGE` | `doughnut` with `circumference` under 360 and exactly two finite values, or declared |
| `DOT` | every dataset `showLine: false` on a **category** axis |
| `DUMBBELL` | declared — shape-identical to a one-interval-per-lane gantt |
| `SURVIVAL` | declared — a `stepped` line is how every staircase is drawn |

Highlight was wired in the same change as detection for each, per the acceptance criteria: `computeTargetMaps` gained the DOT/GAUGE and SURVIVAL/BUMP/RADAR/POLAR_AREA cases, `DUMBBELL` got its own `resolveActiveTargets` branch (its rows are the two ends of one drawn element, so the row must not reach `rowDatasetIndex`), and gantt needed a dedicated `ganttTargets` map because its grid is transposed against Chart.js's — MAIDR row = lane, MAIDR column = which dataset booked it.

### Supporting groundwork

- `plugins.maidr.traceType` — the author declaration the issue called the unlock. Read through one helper and consulted in exactly three places. A declaration wins over every heuristic **in both directions**: declaring `pie` on a two-slice half-doughnut keeps it a pie.
- `plugins.maidr` also gained `unit` (gantt interval axis), `startLabel`/`endLabel` (dumbbell ends), and `target`/`bands` (gauge).
- New `ChartJsPointValue` with `censored`, `yMin`, `yMax`. This was forced, not cosmetic — TS excess-property checking made a datum carrying survival extras unrepresentable in the adapter's own types even though Chart.js passes them through.

### Trace types deliberately NOT added

- **`VOLCANO`** — deferred by #860 itself. Indistinguishable from any scatter; needs a `thresholdOptions` passthrough on top of the declaration.
- **`MANHATTAN`** — same situation and same consumer class as VOLCANO; needs the same `thresholdOptions` work plus a per-point label preserved through scatter extraction.
- **The 19 types Chart.js cannot draw** — ALLUVIAL, CHORD, SANKEY, NETWORK, TREEMAP, SUNBURST, ICICLE, CHOROPLETH, HEXBIN, CONTOUR, MOSAIC, PARALLEL, ERROR_BAR, FOREST, BOXEN, RIDGELINE, FUNNEL, LOLLIPOP, WORD_CLOUD. Either no controller exists, or the controller lives in a plugin this repo does not depend on, or the data Chart.js carries genuinely lacks the semantics (a mosaic's per-bar widths, a contour's level values, a ridgeline's offsets). LOLLIPOP costs nothing today: it already reads correctly as BAR.

### Two places the issue's plan was corrected

- The issue said the radar payload "is exactly what `extractLineLayers` already builds". Radar datasets are very commonly `fill: true`, and that path buckets by (stepDirection, filled) — it would have emitted an AREA layer. A dedicated `extractRadarLayers` was written instead, with a test pinning it.
- The issue offered "x scale kind (time ⇒ gantt, linear ⇒ dumbbell)" as an alternative to declaring a dumbbell. That would have been a regression: every real gantt on a numeric week or sprint axis is linear. Implemented as declaration-only.
- Gantt `unit` announces `'milliseconds'` on a time scale — which is what `end - start` actually is — rather than `time.unit`, which is a *display* unit and would have announced "5 days" for a five-millisecond span.

### Tests, docs and examples

- 9 new test files plus `traceConstruction.test.ts`, which builds each type through `TraceFactory` and asserts `getStateAt(row, col)` — it catches payload-shape errors extraction-only assertions miss.
- `docs/chartjs.md`: the support table was stale (it listed neither Area nor Stacked Area despite both being emitted); every new type is now in it, plus sections on the value heuristics and on the declaration-only readings.
- 9 new example pages under `examples/chartjs/`.

## Checklist

The manual-testing box is left unticked deliberately: the automated gate below passes in full, but no manual browser pass was done against the new example pages, and `ManualTestingProcess.md` is not present in the repo.

## Additional Notes

**Gate — all green.** `npm run lint:fix` (no changes produced), `npm run type-check`, `npm run build`, `npm test` (2709 unit/component + 73 ESM, 0 failures). Also verified against the rest of CI: `npm run lint`, `npm run docs`, `npm run build:preview` and `npm run sync:copilot:check` all pass. `npm run e2e` was not run locally.

**One follow-up needed outside this diff's scope.** The 9 new example pages ship to the site — the examples folder is copied wholesale — but they are not linked from the gallery index, whose Chart.js list is hardcoded in `scripts/build-site.js` (~line 421). Rows are needed for `chartjs/radar.html`, `gantt.html`, `waterfall.html`, `bump.html`, `bar-diverging.html`, `gauge.html`, `dot.html`, `dumbbell.html` and `survival.html`. That file sits outside the adapter and was left alone rather than edited opportunistically.

**Known limits, stated plainly.** The normalized-area, bump, diverging and waterfall readings are value heuristics — Chart.js declares none of these — so each is deliberately strict and each has `plugins.maidr.traceType` as an escape hatch in both directions. A normalized chart whose totals were rounded unevenly beyond half a percent will read as a plain stacked area. Mixed charts remain a pre-existing latent trap: dispatch is still on `chart.config.type`, and `ChartJsDataset.type` is still never read — this change does not deepen that assumption, but it does not fix it either.

---

### DIFF
--- a/docs/chartjs.md
+++ b/docs/chartjs.md
@@ -72,15 +72,36 @@ MAIDR's Chart.js adapter is a standard Chart.js plugin:
 | Bar | `'bar'` (one dataset) | — | [Bar chart](examples.html) |
 | Stacked Bar | `'bar'` with `scales.x.stacked` / `scales.y.stacked` | — | [Stacked bar](examples.html) |
 | Dodged Bar | `'bar'` with multiple datasets (no stacking) | — | [Dodged bar](examples.html) |
+| Diverging Bar | stacked `'bar'` with one series negated | — | [Diverging bar](examples.html) |
+| Gantt / Range Bar | `'bar'` with `[start, end]` data | — | [Gantt chart](examples.html) |
+| Waterfall | `'bar'` with chained `[start, end]` data | — | [Waterfall](examples.html) |
 | Line | `'line'` | — | [Line chart](examples.html) |
 | Step | `'line'` with `stepped` on the dataset (or `elements.line`) | — | [Line chart](examples.html) |
+| Area | `'line'` with `fill` | — | [Line chart](examples.html) |
+| Stacked Area | `'line'` with `fill` and a stacked value scale | — | [Line chart](examples.html) |
+| Normalized Area | stacked area whose categories all total 100 (or 1) | — | [Line chart](examples.html) |
+| Bump | `'line'` with `scales.y.reverse` and ranked values | — | [Bump chart](examples.html) |
 | Scatter | `'scatter'` | — | [Scatter plot](examples.html) |
+| Radar | `'radar'` | — | [Radar chart](examples.html) |
+| Polar Area | `'polarArea'` | — | [Radar chart](examples.html) |
 | Box Plot | `'boxplot'` | `@sgratzl/chartjs-chart-boxplot` | [Box plot](examples.html) |
 | Candlestick | `'candlestick'` | `chartjs-chart-financial` + a date adapter | [Candlestick](examples.html) |
 | Heatmap | `'matrix'` | `chartjs-chart-matrix` | [Heatmap](examples.html) |
 | Pie / Doughnut | `'pie'`, `'doughnut'` | — | [Pie chart](examples.html) |
 
 > **Pie note:** a pie has no Chart.js scales, so there is no axis title to read. `axes.x` and `axes.y` default to `Category` and `Value`; set `plugins.maidr.axes` to name what the slice labels and their values actually mean. Multiple datasets are concentric rings, not slices of one circle — each becomes its own MAIDR layer with its own total and percentages, and Page Up / Page Down move between them.
+
+> **Radar note:** a radar and a polar area are drawn against a single radial `r` scale, so `axes.y` reads `scales.r.title.text` and `axes.x` defaults to `Category` — set `plugins.maidr.axes` to name what the spokes and their magnitudes mean. Both are read as a multi-series layer, one row per dataset and one column per spoke, with each spoke's stereo position following its angle rather than its index.
+
+### Charts Chart.js Does Not Declare
+
+Chart.js has no configuration for a waterfall, a bump chart, a normalized area or a diverging bar — each is a recipe built out of a plain bar or line chart — so MAIDR reads them off the **values**. Each test is deliberately strict, because announcing a chart as something it is not is worse than announcing it plainly:
+
+- **Waterfall** — one series of `[start, end]` bars on the default vertical index axis where each step begins where the previous one ended. Bars that sit on the baseline (`start` of `0`) are read as the opening, closing or a subtotal. Unchained intervals are read as a gantt instead.
+- **Gantt** — any other `[start, end]` bar chart. `indexAxis: 'y'` is the ordinary schedule, one lane per label; several datasets put several intervals in the same lane, each named by its dataset label. A lane whose entry is `null` stays a navigable row with nothing booked. Bounds may be numbers or `Date`s; on a `type: 'time'` scale MAIDR announces both ends as dates and measures lengths in milliseconds, so set `plugins.maidr.unit` when the schedule reads in days or sprints.
+- **Normalized area** — two or more stacked, filled line datasets whose every category totals the same whole (100 or 1, within half a percent so rounded shares still count).
+- **Bump** — a line chart with `scales.y.reverse` whose values at every period are a permutation of `1..N` across the series. The reversed axis alone is not enough; the permutation is what makes the rank reading safe.
+- **Diverging bar** — a stacked bar chart whose datasets each sit wholly on one side of the baseline, with both sides occupied. The values stay signed: MAIDR pitches the magnitude and announces the side.
 
 ## Code Examples
 
--- a/docs/chartjs.md
+++ b/docs/chartjs.md
@@ -75,12 +75,15 @@ MAIDR's Chart.js adapter is a standard Chart.js plugin:
 | Diverging Bar | stacked `'bar'` with one series negated | — | [Diverging bar](examples.html) |
 | Gantt / Range Bar | `'bar'` with `[start, end]` data | — | [Gantt chart](examples.html) |
 | Waterfall | `'bar'` with chained `[start, end]` data | — | [Waterfall](examples.html) |
+| Dumbbell | horizontal `'bar'` with `[start, end]` data and `plugins.maidr.traceType` | — | [Dumbbell](examples.html) |
 | Line | `'line'` | — | [Line chart](examples.html) |
 | Step | `'line'` with `stepped` on the dataset (or `elements.line`) | — | [Line chart](examples.html) |
 | Area | `'line'` with `fill` | — | [Line chart](examples.html) |
 | Stacked Area | `'line'` with `fill` and a stacked value scale | — | [Line chart](examples.html) |
 | Normalized Area | stacked area whose categories all total 100 (or 1) | — | [Line chart](examples.html) |
 | Bump | `'line'` with `scales.y.reverse` and ranked values | — | [Bump chart](examples.html) |
+| Dot Plot | `'line'` with `showLine: false` on a category axis | — | [Dot plot](examples.html) |
+| Survival | `'line'` with `stepped` and `plugins.maidr.traceType` | — | [Survival curve](examples.html) |
 | Scatter | `'scatter'` | — | [Scatter plot](examples.html) |
 | Radar | `'radar'` | — | [Radar chart](examples.html) |
 | Polar Area | `'polarArea'` | — | [Radar chart](examples.html) |
@@ -88,6 +91,7 @@ MAIDR's Chart.js adapter is a standard Chart.js plugin:
 | Candlestick | `'candlestick'` | `chartjs-chart-financial` + a date adapter | [Candlestick](examples.html) |
 | Heatmap | `'matrix'` | `chartjs-chart-matrix` | [Heatmap](examples.html) |
 | Pie / Doughnut | `'pie'`, `'doughnut'` | — | [Pie chart](examples.html) |
+| Gauge | `'doughnut'` with `circumference` under 360 and two values | — | [Gauge](examples.html) |
 
 > **Pie note:** a pie has no Chart.js scales, so there is no axis title to read. `axes.x` and `axes.y` default to `Category` and `Value`; set `plugins.maidr.axes` to name what the slice labels and their values actually mean. Multiple datasets are concentric rings, not slices of one circle — each becomes its own MAIDR layer with its own total and percentages, and Page Up / Page Down move between them.
 
@@ -102,6 +106,16 @@ Chart.js has no configuration for a waterfall, a bump chart, a normalized area o
 - **Normalized area** — two or more stacked, filled line datasets whose every category totals the same whole (100 or 1, within half a percent so rounded shares still count).
 - **Bump** — a line chart with `scales.y.reverse` whose values at every period are a permutation of `1..N` across the series. The reversed axis alone is not enough; the permutation is what makes the rank reading safe.
 - **Diverging bar** — a stacked bar chart whose datasets each sit wholly on one side of the baseline, with both sides occupied. The values stay signed: MAIDR pitches the magnitude and announces the side.
+- **Dot plot** — a line chart whose every dataset sets `showLine: false` on a category axis. This one Chart.js does say outright: switching the line off is its own way of drawing a Cleveland dot plot. Unjoined points along a *linear* axis are a scatter plot drawn by the line controller, and stay one.
+- **Gauge** — a `doughnut` swept through less than the full circle (`circumference` under 360) whose single dataset holds exactly two values: the measure, and the rest of the dial drawn empty. The remainder is spent on the dial's `max` rather than announced as a second reading.
+
+### Charts Only the Page Can Declare
+
+Three readings are shape-identical to another recipe, so no test on the config or the values can reach them. Set `plugins.maidr.traceType` to say which figure the chart is; the declaration wins over every heuristic above, in both directions — declaring `pie` on a two-slice half-doughnut keeps it a pie, and declaring `gantt` on a chained bar chart keeps it a schedule.
+
+- **Dumbbell** (`traceType: 'dumbbell'`) — a horizontal floating bar chart, which is the same `[start, end]` datum a one-interval-per-lane gantt uses. `plugins.maidr.startLabel` and `endLabel` name the two ends ("1990", "2020"); without them the reader is told which dot they are on but not which year it is. Rows with no pair are skipped rather than kept as empty rows, unlike a gantt lane.
+- **Survival** (`traceType: 'survival'`) — a `stepped: 'after'` line, which is how every staircase is drawn. Chart.js ignores properties it does not know, so ride the two things a survival figure carries and a step chart does not on the points themselves: `{x, y, censored: true}` for a censoring mark and `{yMin, yMax}` for the confidence band. Each dataset is one arm, gathered into a single layer.
+- **Gauge** (`traceType: 'gauge'`) — for a dial the geometry above misses, and for the target and bands Chart.js records only as styling: `plugins.maidr.target` is the bullet marker and `plugins.maidr.bands` is a `[{ to, label }]` list in ascending order.
 
 ## Code Examples
 
--- a/docs/chartjs.md
+++ b/docs/chartjs.md
@@ -101,7 +101,7 @@ MAIDR's Chart.js adapter is a standard Chart.js plugin:
 
 Chart.js has no configuration for a waterfall, a bump chart, a normalized area or a diverging bar — each is a recipe built out of a plain bar or line chart — so MAIDR reads them off the **values**. Each test is deliberately strict, because announcing a chart as something it is not is worse than announcing it plainly:
 
-- **Waterfall** — one series of `[start, end]` bars on the default vertical index axis where each step begins where the previous one ended. Bars that sit on the baseline (`start` of `0`) are read as the opening, closing or a subtotal. Unchained intervals are read as a gantt instead.
+- **Waterfall** — one series of `[start, end]` bars on the default vertical index axis where each step begins where the previous one ended. Bars that sit on the baseline (`start` of `0`) are read as the opening, closing or a subtotal. Unchained intervals are read as a gantt instead. A waterfall is one running total, so only the first dataset is read — a chart with several series of intervals is a gantt.
 - **Gantt** — any other `[start, end]` bar chart. `indexAxis: 'y'` is the ordinary schedule, one lane per label; several datasets put several intervals in the same lane, each named by its dataset label. A lane whose entry is `null` stays a navigable row with nothing booked. Bounds may be numbers or `Date`s; on a `type: 'time'` scale MAIDR announces both ends as dates and measures lengths in milliseconds, so set `plugins.maidr.unit` when the schedule reads in days or sprints.
 - **Normalized area** — two or more stacked, filled line datasets whose every category totals the same whole (100 or 1, within half a percent so rounded shares still count).
 - **Bump** — a line chart with `scales.y.reverse` whose values at every period are a permutation of `1..N` across the series. The reversed axis alone is not enough; the permutation is what makes the rank reading safe.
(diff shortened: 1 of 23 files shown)


jooyoungseo:
All 14 checks were green at `587c96b`, so nothing here is a CI fix. Responding to the three non-blocking review findings:

**Fixed**
- **Stale type count (`docs/chartjs.md`)** — the integration-comparison row said "8 Chart.js types"; it now reads 23 and links to the supported-types table so the number has a source rather than a second place to go stale.
- **Single-dataset assumption, documented and pinned** — the dumbbell and waterfall bullets now say outright that only the first dataset is read and that several series of intervals are a gantt. Added a test that pins it *and* that highlighting resolves to the same dataset the extractor read, since extraction hardcodes `datasets[0]` while highlighting goes through `firstDatasetIndex` — a drift between those two would outline a bar the reader was never told about.

**Declined**
- **`extractBarLayers` routing on `some(isRangeValue)`** — tightening this to "every value is a range or null" would trade the flagged low-probability case for a likelier one. A mixed chart routed to the bar path loses the range semantics of *every* floating bar; routed to the floating-bar path it loses only the entries that are not ranges. Losing the ranges is the worse of the two, and Chart.js configs are shape-homogeneous per dataset in practice. Left as is.
- **Multi-dataset support for dumbbell/waterfall** — out of scope, and both figures are single-series by definition; a multi-series set of intervals already has a correct reading in the gantt path. Documented and tested instead of widened.

`lint:fix`, `type-check`, `build` and `test` all pass locally (2710 + 73 tests, 203 suites).

---

jooyoungseo:
Second review round. All 14 checks are green at `f5c6e350`, so no CI fix was needed, and I pushed nothing — the review round found nothing worth changing in the adapter. Two things worth recording rather than a fix:

**One gap neither review caught — example gallery discoverability.** This PR adds 9 example pages under `examples/chartjs/`, and the supported-types table links each new row to `examples.html`. That gallery is a hardcoded list in `scripts/build-site.js` (lines 423-431) and still names only the original 9 Chart.js pages, so the new ones deploy (the whole `examples/` tree is copied recursively, and the preview URLs resolve) but are reachable only by typing the path. Not introduced here — `line-stacked-panels.html` was already unlisted on `main` — but this widens it from 1 unlisted page to 10. I left it alone deliberately: the fix belongs in `scripts/build-site.js`, which is outside this PR's declared file scope and is edited by the other adapter branches in flight, so touching it here would just create a conflict. Worth a follow-up that adds all 10 at once.

**Correction to finding 2 of the last review pass.** It restated dumbbell/waterfall multi-dataset as "not test-covered"; that is stale for dumbbell — `dumbbell.test.ts:106` (`reads the first dataset only, and highlights the bar it read`) pins it, and pins the part that actually matters, that extraction's hardcoded `datasets[0]` and highlighting's `firstDatasetIndex` resolve to the same element. Waterfall needs no equivalent: `isWaterfallSequence` requires `datasets.length === 1`, so only an explicit `traceType: 'waterfall'` can reach that path with more than one dataset, and the docs state the first-dataset rule for it.

I re-derived the rest independently rather than taking the prior pass's word for it: `STEP_DIRECTION_BY_OPTION` maps `after` to `vh` and the survival default is `vh`, so the KM default matches its comment; gauge routing resolves `row=0, col=0` to arc index 0 through `finiteIndices`; the gap-skipping in radar/survival extraction and in `barLineIndices` both key off `toFiniteNumber`, so highlights stay aligned across nulls; and each new example's config genuinely trips the heuristic it advertises (checked the waterfall chain, the bump `scales.y.reverse`, the dot-plot category axis). Type-check clean and 173/173 chartjs tests pass locally.

I agree with the earlier decline on `extractBarLayers` routing on `some(isRangeValue)` for the reason already given — the alternative loses range semantics for every floating bar instead of losing the non-range strays.

---

PR #873, opened 2026-08-14
feat(victory): support the chart types added since the pie chart

### LINKED ISSUE #864: feat(victory): support the chart types added since the pie chart
## Problem

MAIDR core has gained ~34 new trace types since the pie chart landed (radar, gauge, funnel, waterfall, gantt, treemap, sankey, sunburst, dumbbell, dot/lollipop, diverging, area family, and more). The `victory` adapter has not kept up: of those, it currently emits **none**.

A chart Victory draws perfectly well is therefore either dropped from the figure or announced as the wrong chart — which for a screen-reader user is worse than not being supported at all.

The adapter is a React-children introspector, and it decides the TraceType in two hops: Victory `displayName` → an internal `kind` tag → `TraceType`. First, `getVictoryDisplayName` (src/adapters/victory/converters.ts:38-62) reads `displayName`/`name` off the element type (unwrapping `WrappedComponent`/`render`/`type` HOCs) and only accepts names starting with `Victory`; `isDataComponent` (converters.ts:69-79) then gates that name against a hardcoded 7-name whitelist (VictoryBar, VictoryLine, VictoryScatter, VictoryBoxPlot, VictoryCandlestick, VictoryHistogram, VictoryPie). Second, `extractLayerFromElement` switches on that name (converters.ts:471-493) to pick a per-component extractor, each of which stamps a `kind` discriminant onto `VictoryLayerData` (types.ts:80-93); `VictoryStack` is intercepted earlier in `collectDataLayers` (converters.ts:599-604) and routed to `extractSegmentedLayer` → `kind: 'segmented'`. Finally `toMaidrLayer` switches on `data.kind` (converters.ts:762) and emits the TraceType literal there. So dispatch is component-identity-driven, not data-driven — with exactly one exception: LINE vs STEP is decided from the `interpolation` prop via `STEP_DIRECTION_BY_INTERPOLATION` / `stepDirectionOf` (converters.ts:171-187), consumed at converters.ts:775. A parallel switch on `layer.victoryType` in `tagLayerElements` (selectors.ts:167-204) picks the DOM-tagging strategy, so every new type needs a branch in both switches.

## In scope — 9 trace types Victory can render

| TraceType | Victory construct | Detection / payload |
|---|---|---|
| `AREA` | VictoryArea | Add 'VictoryArea' to VictoryComponentType (types.ts:63-71) and to the isDataComponent whitelist (converters.ts:69-79), then a `case 'VictoryArea'` in the extractLayerFromElement switch (converters.ts:471-493). Extraction is byte-for-byte extractLineData (converters.ts:192-216): same `data` prop, same resolveAccessor x/y convention. Payload is LinePoint[][] — `[[{x, y}, ...]]` — which is exactly what AreaTrace consumes, since it extends LineTrace (src/model/area.ts:80, factory.ts:57-60). Emit a new `kind: 'area'` in VictoryLayerData so toMaidrLayer can return TraceType.AREA without colliding with the line/step branch. DOM tagging needs no new code: victory-area/src/area.tsx:95-96 renders a single `<path role="presentation">`, identical to VictoryLine, so routing 'VictoryArea' to the existing tagLineElements (selectors.ts:270-301) works as-is. |
| `STACKED_AREA` | VictoryStack wrapping VictoryArea children | extractSegmentedLayer (converters.ts:518-570) currently hard-filters stack children with `if (name !== 'VictoryBar') return;` (converters.ts:533). Branch on the child component name instead: VictoryBar children keep producing `kind: 'segmented'` → TraceType.STACKED; VictoryArea children produce a new `kind: 'stackedArea'` → TraceType.STACKED_AREA. Note the payload shape differs — STACKED_AREA wants LinePoint[][] (one inner array per band, `z` carrying the series name from the child's `name` prop, as converters.ts:543 already does), not SegmentedPoint[][]. Emit each band's own y value untransformed: AreaTrace deliberately computes the running total itself from the series it is given (src/model/area.ts, "The total is computed here rather than expected from the producer"), and VictoryStack children carry exactly that per-series value (victory-stack accumulates via `_y0` at render time only). |
| `NORMALIZED_AREA` | VictoryStack of VictoryArea with author-normalized data (Victory has no stack-normalize prop) | Verified: victory-stack/src has zero occurrences of `normalize`/`percent`, so unlike Vega-Lite's `stack: 'normalize'` there is no declarative flag to read. Detection must be numeric: after building the STACKED_AREA bands, sum the series values at each x and, if every column total is ~1 or ~100 within a tolerance, emit TraceType.NORMALIZED_AREA instead. Same LinePoint[][] payload; the trace's variantOf (src/model/area.ts:31-39) does the rest. Be explicit that this is a heuristic — a genuine unnormalized stack whose totals happen to be 1.0 would be mislabelled — so it is worth gating behind an explicit opt-in prop rather than inferring silently. |
| `ERROR_BAR` | VictoryErrorBar | VictoryErrorBar is a first-class Victory component (victory-errorbar package; displayName set at victory-errorbar/src/victory-errorbar.tsx:64), so detection is a pure whitelist addition, no heuristics. Data is `{x, y, errorX, errorY}` per datum, plus `errorX`/`errorY` accessor props (victory-errorbar/src/helper-methods.tsx:36 lists accessorTypes as x, y, errorX, errorY); the error value is either a scalar (symmetric) or a `[plus, minus]` array (victory-errorbar/src/helper-methods.tsx:135-137). Critical conversion: Victory's error is a DELTA, MAIDR's ErrorBarPoint wants ABSOLUTE bounds (grammar.ts:380-391), so build `{x, y, yMin: y - minus, yMax: y + plus}`; leave yMin/yMax undefined for a one-sided interval, which ErrorBarTrace already handles as NaN (src/model/errorBar.ts:38-52). Selector work is the real cost: VictoryErrorBar renders its caps/whiskers as `<line role="presentation">` via renderBorder/renderLine (victory-errorbar/src/error-bar.tsx:12-34), so the default tagDiscreteElements path (which queries `path[role="presentation"]`, selectors.ts:210-242) finds nothing and a new geometry-classifying branch is needed in tagLayerElements (selectors.ts:167-204). |
| `RADAR` | VictoryLine / VictoryArea with polar={true} inside a polar VictoryChart + VictoryPolarAxis | Polar is genuinely supported: victory-line/src/helper-methods.ts:26 and victory-area/src/helper-methods.tsx:40 both branch on props.polar, and the victory-polar-axis package ships VictoryPolarAxis. Detection: a line/area layer whose own `polar` prop is true, OR whose enclosing `<VictoryChart polar>` is true. The chart-level case needs new plumbing — extractVictorySubplots/extractVictoryLayers (converters.ts:624-710) currently reads only `chartProps.children` and `title`, and extractAxisLabels (converters.ts:106-129) only looks at VictoryAxis, so `polar` would have to be threaded into collectDataLayers alongside axisLabels. Payload is LinePoint[][], one row per series, since RadarTrace extends LineTrace and derives spoke angles itself (src/model/radar.ts:47-76). The extra work is series merging: extractLineData wraps a single series as `[points]` (converters.ts:209-211) and each sibling VictoryLine becomes its own layer, whereas a radar needs sibling polar series coalesced into one multi-row layer (the Vega-Lite adapter's coalesceSiblingLineLayers, src/adapters/vegalite/converters.ts:172, is the precedent). |
| `POLAR_AREA` | VictoryBar with polar={true} (radial/coxcomb bars) | Same polar plumbing as RADAR — victory-bar/src/helper-methods.ts:37 branches on props.polar. Detection: a VictoryBar layer that is polar (own prop or inherited from `<VictoryChart polar>`). Payload is the same LinePoint[][] as RADAR, because POLAR_AREA and RADAR share RadarTrace and differ only in the spoken label (factory.ts:150-152, src/model/radar.ts:29). Once the polar flag is threaded for RADAR this is a one-line branch, so bundle the two; standalone it costs the same plumbing. |
| `WATERFALL` | VictoryBar with per-datum y0 (floating bars) | `y0` is a public accessor prop on every Victory data component (victory-core/src/victory-util/common-props.tsx:31, honored per-datum at victory-core/src/victory-util/data.ts:81-86), so a bar drawn from `start` to `end` is native Victory, not a hack. Detection: a VictoryBar whose data carry `y0` (or which passes a `y0` prop), confirmed as a waterfall when the bars chain — `y0[i+1] === y[i]` for the running steps. Payload is WaterfallPoint[] (grammar.ts:460-479): `{x, start: y0, end: y, delta: end - start, kind}`, with `kind` from the sign of delta and a `total` kind for any bar whose y0 sits at the baseline. Reuses the existing tagDiscreteElements path unchanged, since VictoryBar renders one `<path role="presentation">` per datum (already relied on at selectors.ts:196-203). |
| `DIVERGING` | VictoryStack (usually horizontal) of two VictoryBar series, one signed negative — a population pyramid | Detect inside the existing extractSegmentedLayer result (converters.ts:518-570): exactly two series, where one series' y values are all <= 0 and the other's all >= 0. Emit TraceType.DIVERGING in place of TraceType.STACKED at converters.ts:837. The payload needs no change at all — DivergingTrace extends SegmentedTrace and consumes the same SegmentedPoint[][], expecting the values already signed as the chart draws them (src/model/diverging.ts). Selector tagging is likewise unchanged. The one caveat is that BarPoint.y is typed `number \| string` and extractSegmentedLayer passes the accessor result through uncoerced (converters.ts:547), so the sign test must coerce before comparing. |
| `DOT` | VictoryScatter over a categorical x axis — a Cleveland dot plot | Detection is a property of the extracted data rather than of the component: in extractScatterData (converters.ts:221-236), if every raw x is non-numeric (a string), the layer is a category-vs-value dot plot rather than a bivariate scatter. Emit TraceType.DOT with BarPoint[] `{x, y}`, since DOT is routed to BarTrace (factory.ts:66-69). This doubles as a real bug fix: the current extractor does `x: Number(getX(d))` (converters.ts:231), so a categorical VictoryScatter today produces NaN x on every point and sonifies as garbage. Selector tagging is unaffected (still the discrete-path branch plus stampPathCenter, selectors.ts:236-238) — though note stampPathCenter exists to satisfy ScatterTrace's cx/cy grouping, and BarTrace does not need it, so the DOT branch can skip it. |

## Not renderable by Victory

<details><summary>13 trace types the library genuinely cannot draw</summary>

- ALLUVIAL — Victory has no flow/ribbon primitive and no stage layout; every data component consumes a flat array against a cartesian or polar scale.
- CHORD — no circular arc-link or ribbon primitive; VictoryPie draws wedges only, with no chords between them.
- SANKEY — no sankey/flow layout, no link geometry, no node placement solver.
- NETWORK — no node-link rendering and no force/graph layout; Victory has no graph data model.
- TREEMAP — no hierarchical rectangle packing; Victory data components take flat arrays, never a tree.
- ICICLE — same missing hierarchy layout as TREEMAP; no depth-banded rectangle placement.
- SUNBURST — VictoryPie renders exactly one flat ring from a flat array (victory-pie); there is no nested-ring or hierarchy support.
- CHOROPLETH — no geographic projection, no geoJSON/topoJSON ingestion, no arbitrary map-path rendering.
- CONTOUR — no scalar-field or isoline generation; Victory has nothing that turns a grid of values into level curves.
- HEXBIN — no hexagonal binning primitive and no hex lattice geometry.
- WORD_CLOUD — no text-packing or glyph-layout engine; VictoryLabel places single labels at fixed coordinates only.
- BOXEN — VictoryBoxPlot is hardwired to the five-number summary (min/q1/median/q3/max, plus derived-from-raw-y mode); it has no variable-depth letter-value ladder.
- FUNNEL — no funnel or trapezoid primitive, and no stage/segment geometry that narrows across ordered steps.

</details>

## Acceptance criteria

- [ ] Unit tests under `test/adapters/victory/` cover every newly detected type.

## Implementation notes

SCOPE OF THE FINDING: the adapter emits exactly nine TraceTypes — BAR, LINE, STEP, SCATTER, BOX, CANDLESTICK, HISTOGRAM, PIE, STACKED (converters.ts:766, 775, 785, 794, 803, 812, 821, 837). Every one is pre-pie, so `alreadySupported` is genuinely empty: not a single post-pie TraceType is emitted, and `grep -rn 'TraceType\.'` over src/adapters/victory returns only those nine. There are also no comment-level or imported-but-unused mentions of any post-pie type anywhere in the directory.

WHERE THE WORK LANDS: adding any type touches five places in lockstep — (1) VictoryComponentType in types.ts:63-71, (2) the VictoryLayerData union in types.ts:80-93, (3) isDataComponent in converters.ts:69-79, (4) the name switch in extractLayerFromElement converters.ts:471-493, and (5) the `data.kind` switch in toMaidrLayer converters.ts:762-842. Anything with a new DOM shape additionally needs a branch in tagLayerElements (selectors.ts:167-204). Both switches are exhaustive over a discriminated union, so tsc will point at every site you missed — a genuinely safe extension path.

DRAWABLE BUT NOT DETECTABLE (deliberately excluded from missingFeasible — Victory renders these fine, but nothing in the React props distinguishes them from a type the adapter already emits):
- BUMP — multiple VictoryLine series with rank y values; indistinguishable from a plain multi-line chart without an author hint.
- SURVIVAL — VictoryLine interpolation="stepAfter" already becomes STEP (converters.ts:174, 775); censoring marks live in a separate sibling component and the censored flag SurvivalPoint needs cannot be recovered.
- VOLCANO / MANHATTAN — both are scatters; ThresholdOptions (grammar.ts) has no Victory-side source, and point identity is not in the x/y accessors.
- FOREST — composable as a horizontal VictoryErrorBar, but ForestPoint's `weight` and `isSummary` (grammar.ts:399-410) have no Victory equivalent. Worth revisiting as an opt-in layered on top of ERROR_BAR.
- GANTT and DUMBBELL — both reduce to the same per-datum `y0` floating-bar mechanism as WATERFALL. The adapter cannot tell the three apart from the data alone; WATERFALL is the one listed as feasible because chaining totals (y0[i+1] === y[i]) is a checkable signal that the other two lack. If GANTT/DUMBBELL are wanted, they need an explicit author prop, not inference.
- LOLLIPOP — the Victory idiom is a thin VictoryBar overlaid with a VictoryScatter on identical data. Detection means correlating two sibling layers by x/y plus a barWidth/style-width threshold; real but fragile, and it fights the adapter's one-component-one-layer model. Left out as not actionable.
- GAUGE — VictoryPie accepts startAngle/endAngle and innerRadius (victory-pie/src/victory-pie.tsx:35, 41, 56), so a half-donut gauge draws fine, but recovering GaugePoint's min/max/target/bands (grammar.ts:530-543) from two pie slices is guesswork.
- MOSAIC — VictoryBar's barWidth accepts a per-datum callback (victory-bar/src/victory-bar.tsx:46), so a marimekko is contrivable, but it is not a Victory chart type and the column shares would have to be reverse-engineered from style callbacks.
- RIDGELINE and PARALLEL — both are hand-composed in Victory (offset VictoryArea groups; multiple offset VictoryAxis components). In each case the author applies the transform the trace explicitly must not receive — the ridgeline baseline (src/model/ridgeline.ts contract, grammar.ts:1462-1471) and the per-column independent scales ParallelTrace needs — so the original values are unrecoverable from the props.

TRAPS FOR WHOEVER IMPLEMENTS THIS:
- extractLineData wraps one series as `[points]` (converters.ts:209-211) and never merges siblings, so any multi-series type (RADAR, STACKED_AREA, POLAR_AREA) needs a coalescing step the adapter does not have today. src/adapters/vegalite/converters.ts:172 (coalesceSiblingLineLayers) is the in-repo precedent.
- extractVictorySubplots reads only `children` and `title` off VictoryChart (converters.ts:701-709); chart-level props such as `polar` and `horizontal` are not threaded to layers at all. That plumbing is shared prerequisite work for RADAR and POLAR_AREA.
- subplotFingerprint (useVictoryAdapter.ts:69-82) serializes `victoryType` and `data`, so new component types and payload shapes participate in change detection automatically — no extra work there.
- Test/example conventions: tests are plain jest with hand-built React elements and jsdom fragments (test/adapters/victory/converters.test.ts for extraction/TraceType assertions, selectors.test.ts for DOM tagging); examples are one .tsx per chart type under examples/victory/examples/, registered in App.tsx. Today those cover only bar, line, scatter, histogram, candlestick, boxplot and stack — notably there is no pie example even though pie is implemented, so example coverage already lags the code.
- All Victory-side claims above were verified against the installed victory@37.3.6 in node_modules, matching the ^37.0.0 peer/dev dependency in package.json.

jooyoungseo:
# Pull Request

## Description

The Victory adapter emitted nine trace types, all of them pre-dating the pie chart. Charts that Victory draws perfectly well were therefore either dropped from the figure or announced as the wrong chart — which for a screen-reader user is worse than no support at all.

This teaches the adapter eight of the nine trace types listed as in scope, along with the detection, the payload conversion and the highlight tagging for each. Two structural changes make room for them:

- **Dispatch is now data-driven, not component-driven.** One Victory component can produce several trace types — `VictoryBar` is a bar, a waterfall or a coxcomb depending on its props and its values — so `tagLayerElements` switches on `layer.data.kind` first and only falls through to the old `victoryType` branches. The component name no longer decides the DOM shape.
- **`polar` is threaded from `VictoryChart` down to its children**, the way Victory passes it internally. A child's own `polar` prop wins, so `polar={false}` opts back out.

Everything is adapter-local: no model, service, command, state, ui or `grammar.ts` change.

## Related Issues

Closes #864

### Trace types added (8)

| TraceType | Detected from | Notes |
|---|---|---|
| `AREA` | `VictoryArea` | Read exactly as a line; Victory draws both from the same props. |
| `STACKED_AREA` | `VictoryStack` of `VictoryArea` children | One band per child, each carrying its own **untransformed** value — `AreaTrace` computes the running total itself. |
| `NORMALIZED_AREA` | such a stack whose columns sum to a common whole | Victory has no normalize prop, so the evidence is arithmetic. Deliberately narrow: ≥2 bands, ≥2 columns, every column total within 0.5% of 1 or 100. |
| `ERROR_BAR` | `VictoryErrorBar` | Converts Victory's error **deltas** to the absolute bounds the schema fixes, one side at a time so a one-sided interval survives. |
| `POLAR_AREA` | `VictoryBar` drawn polar | A coxcomb, emitted as the single-series line grid the radar trace reads. |
| `WATERFALL` | `VictoryBar` floating on a per-datum `y0` | Gated on the bars chaining (`y0[i+1] === y[i]`), baseline-resting bars exempt as totals. |
| `DOT` | `VictoryScatter` over a categorical x | Doubles as a bug fix — see below. |
| `DIVERGING` | `VictoryStack` of two oppositely-signed bar series | A population pyramid. Values stay signed as the chart draws them. |

### Trace types NOT added, with the reason

- **`RADAR`** — the only in-scope type skipped. A radar is several sibling `VictoryLine`/`VictoryArea` series coalesced into one multi-row layer, and the adapter has no sibling-coalescing step (each sibling stays its own layer; the Vega-Lite `coalesceSiblingLineLayers` is the in-repo precedent for what is missing). The polar plumbing this needs is now in place, so what remains is the coalescing. Until then a polar line or area still reads as `LINE`/`AREA` — the same behaviour as before this PR, not a regression.
- **`GANTT` / `DUMBBELL`** — the issue lists these as drawable but not detectable, and implementing `WATERFALL` confirms why: all three are the same floating-`y0` bar. Chaining is the only checkable signal, and only a waterfall has it. An unchained floating bar deliberately falls back to `BAR` rather than guessing.
- **`FOREST`** — composable on top of `ERROR_BAR`, but `ForestPoint.weight` and `isSummary` have no Victory-side source.
- **`GAUGE`, `MOSAIC`, `LOLLIPOP`, `BUMP`, `SURVIVAL`, `VOLCANO`/`MANHATTAN`, `RIDGELINE`, `PARALLEL`** — as analysed in the issue: Victory can be contrived into drawing them, but the props carry no signal that separates them from a type the adapter already emits, or the author has already applied the transform the trace must not receive.
- **The 13 types Victory cannot render** (`ALLUVIAL`, `CHORD`, `SANKEY`, `NETWORK`, `TREEMAP`, `ICICLE`, `SUNBURST`, `CHOROPLETH`, `CONTOUR`, `HEXBIN`, `WORD_CLOUD`, `BOXEN`, `FUNNEL`) — no corresponding primitive or layout in the library.

### Bug fixed along the way

`extractScatterData` coerced x with `Number()`, so a **categorical `VictoryScatter` produced `NaN` on every point and sonified as garbage**. Categorical is now decided by Victory's own test (a datum goes on the category scale when its x is a string), and such a layer becomes a `DOT`.

### Highlighting

Every type above is tagged as well as detected. The two that needed real work:

- **Stacked area** — `VictoryStack` reverses render order, so band *i* of the payload is path *(N-1-i)* in the DOM. Band selectors are inverted to match, and each band gets a distinct attribute so `LineTrace` maps them positionally instead of falling into `selectNthElement`.
- **Error bar** — `VictoryErrorBar` renders `<line>` elements, not the `<path role="presentation">` the discrete tagger looks for, so the whiskers are grouped by their parent `<g>` to reach the one-element-per-sample `ErrorBarTrace` requires exactly.

### Tests and examples

- `test/adapters/victory/converters.test.ts` and `selectors.test.ts` extended for every new type.
- New `test/adapters/victory/renderedDom.test.ts` renders the real Victory components and tags the result, so the DOM assumptions are checked against the library rather than against a hand-built fragment.
- New examples under `examples/victory/examples/`, registered in `App.tsx`: area, stacked area, error bar, polar area, waterfall, dot plot, population pyramid.

## Checklist

Note on the unticked boxes: the automated suite passes in full (`lint:fix`, `type-check`, `build`, `test` — 2677 unit + 73 ESM tests green) and the new examples build, but the `ManualTestingProcess.md` walkthrough with a screen reader was not performed, so that box is left unticked rather than claimed. No prose documentation was changed; the new behaviour is documented in code comments and in the examples.

## Additional Notes

Two things worth a reviewer's attention, both deliberate:

**1. `NORMALIZED_AREA` is a heuristic.** Victory has no declarative normalize flag (`grep normalize|percent` over `victory-stack` returns nothing), so a genuine unnormalized stack whose columns happen to sum to 1.0 would be mislabelled. The tolerance is tight on purpose. If that is not acceptable, the alternative is an author-supplied opt-in prop.

**2. A latent core bug this PR does not fix.** Plain `STACKED` is mis-highlighted today, independently of this work: `VictoryStack` reverses render order, and `SegmentedTrace.mapToSvgElements` takes one flat `string` selector, resolves it in document order and maps row-major — so a stack of two bar series binds series 0 to series 1's paths. `DIVERGING` sidesteps this because `DivergingTrace` resolves which side is which by reading values rather than by index, so this PR emits its sides in Victory's paint order. A true stack's row order is semantic, so the adapter cannot fix it by reordering; it needs either `SegmentedTrace` to consume `string[][]` (the grammar already allows it) or a `domMapping.seriesDirection`. Out of scope here — happy to file it as a separate issue.

Also unchanged: `<VictoryChart horizontal>` is still ignored. That is correct for the payload (Victory's `horizontal` flips the drawing, not the data), but it means highlight pan geometry is never orientation-aware. A `VictoryErrorBar` carrying only `errorX` is dropped rather than read as a vertical interval on the wrong axis.

---

### DIFF
--- a/examples/victory/App.tsx
+++ b/examples/victory/App.tsx
@@ -4,11 +4,13 @@ import { AreaChartExample } from './examples/AreaChartExample';
 import { BarChartExample } from './examples/BarChartExample';
 import { BoxPlotExample } from './examples/BoxPlotExample';
 import { CandlestickExample } from './examples/CandlestickExample';
+import { DotPlotExample } from './examples/DotPlotExample';
 import { ErrorBarExample } from './examples/ErrorBarExample';
 import { HistogramExample } from './examples/HistogramExample';
 import { LineChartExample } from './examples/LineChartExample';
 import { MultiPanelExample } from './examples/MultiPanelExample';
 import { PolarAreaExample } from './examples/PolarAreaExample';
+import { PopulationPyramidExample } from './examples/PopulationPyramidExample';
 import { ScatterChartExample } from './examples/ScatterChartExample';
 import { StackedAreaExample } from './examples/StackedAreaExample';
 import { StackedBarExample } from './examples/StackedBarExample';
@@ -18,8 +20,10 @@ const examples: { name: string; component: () => JSX.Element }[] = [
   { name: 'Bar Chart', component: BarChartExample },
   { name: 'Line Chart', component: LineChartExample },
   { name: 'Scatter Chart', component: ScatterChartExample },
+  { name: 'Dot Plot', component: DotPlotExample },
   { name: 'Area Chart', component: AreaChartExample },
   { name: 'Stacked Bar', component: StackedBarExample },
+  { name: 'Population Pyramid', component: PopulationPyramidExample },
   { name: 'Stacked Area', component: StackedAreaExample },
   { name: 'Histogram', component: HistogramExample },
   { name: 'Box Plot', component: BoxPlotExample },
--- a/examples/victory/App.tsx
+++ b/examples/victory/App.tsx
@@ -1,22 +1,32 @@
 import type { JSX } from 'react';
 import { useState } from 'react';
+import { AreaChartExample } from './examples/AreaChartExample';
 import { BarChartExample } from './examples/BarChartExample';
 import { BoxPlotExample } from './examples/BoxPlotExample';
 import { CandlestickExample } from './examples/CandlestickExample';
+import { ErrorBarExample } from './examples/ErrorBarExample';
 import { HistogramExample } from './examples/HistogramExample';
 import { LineChartExample } from './examples/LineChartExample';
 import { MultiPanelExample } from './examples/MultiPanelExample';
+import { PolarAreaExample } from './examples/PolarAreaExample';
 import { ScatterChartExample } from './examples/ScatterChartExample';
+import { StackedAreaExample } from './examples/StackedAreaExample';
 import { StackedBarExample } from './examples/StackedBarExample';
+import { WaterfallExample } from './examples/WaterfallExample';
 
 const examples: { name: string; component: () => JSX.Element }[] = [
   { name: 'Bar Chart', component: BarChartExample },
   { name: 'Line Chart', component: LineChartExample },
   { name: 'Scatter Chart', component: ScatterChartExample },
+  { name: 'Area Chart', component: AreaChartExample },
   { name: 'Stacked Bar', component: StackedBarExample },
+  { name: 'Stacked Area', component: StackedAreaExample },
   { name: 'Histogram', component: HistogramExample },
   { name: 'Box Plot', component: BoxPlotExample },
   { name: 'Candlestick', component: CandlestickExample },
+  { name: 'Error Bar', component: ErrorBarExample },
+  { name: 'Waterfall', component: WaterfallExample },
+  { name: 'Polar Area', component: PolarAreaExample },
   { name: 'Multi-Panel', component: MultiPanelExample },
 ];
 
--- a/examples/victory/examples/AreaChartExample.tsx
+++ b/examples/victory/examples/AreaChartExample.tsx
@@ -1,0 +1,32 @@
+import type { JSX } from 'react';
+import { VictoryArea, VictoryAxis, VictoryChart } from 'victory';
+import { MaidrVictory } from '../../../src/adapters/victory/MaidrVictory';
+
+const data = [
+  { x: 2018, y: 42 },
+  { x: 2019, y: 55 },
+  { x: 2020, y: 38 },
+  { x: 2021, y: 71 },
+  { x: 2022, y: 90 },
+  { x: 2023, y: 84 },
+];
+
+export function AreaChartExample(): JSX.Element {
+  return (
+    <div>
+      <h2>Area Chart</h2>
+      <p>Area chart showing annual rainfall.</p>
+      <MaidrVictory
+        id="victory-area"
+        title="Annual Rainfall"
+        subtitle="2018–2023"
+      >
+        <VictoryChart domainPadding={{ y: 24 }}>
+          <VictoryAxis label="Year" />
+          <VictoryAxis dependentAxis label="Rainfall (cm)" />
+          <VictoryArea data={data} />
+        </VictoryChart>
+      </MaidrVictory>
+    </div>
+  );
+}
--- a/examples/victory/examples/DotPlotExample.tsx
+++ b/examples/victory/examples/DotPlotExample.tsx
@@ -1,0 +1,38 @@
+import type { JSX } from 'react';
+import { VictoryAxis, VictoryChart, VictoryScatter } from 'victory';
+import { MaidrVictory } from '../../../src/adapters/victory/MaidrVictory';
+
+// A Cleveland dot plot is a `VictoryScatter` whose x values name categories
+// rather than measure a position. One magnitude per named category is what a
+// bar chart is, drawn as a point, and MAIDR reads it as one.
+const data = [
+  { x: 'Finland', y: 7.8 },
+  { x: 'Denmark', y: 7.6 },
+  { x: 'Iceland', y: 7.5 },
+  { x: 'Sweden', y: 7.3 },
+  { x: 'Netherlands', y: 7.4 },
+  { x: 'Norway', y: 7.3 },
+];
+
+export function DotPlotExample(): JSX.Element {
+  return (
+    <div>
+      <h2>Dot Plot</h2>
+      <p>
+        Life satisfaction by country, one dot per country. The categories are
+        announced by name rather than by position.
+      </p>
+      <MaidrVictory
+        id="victory-dot"
+        title="Life Satisfaction"
+        subtitle="Score out of 10"
+      >
+        <VictoryChart domainPadding={24}>
+          <VictoryAxis label="Country" style={{ tickLabels: { fontSize: 8 } }} />
+          <VictoryAxis dependentAxis label="Score" />
+          <VictoryScatter data={data} size={5} />
+        </VictoryChart>
+      </MaidrVictory>
+    </div>
+  );
+}
--- a/examples/victory/examples/ErrorBarExample.tsx
+++ b/examples/victory/examples/ErrorBarExample.tsx
@@ -1,0 +1,35 @@
+import type { JSX } from 'react';
+import { VictoryAxis, VictoryChart, VictoryErrorBar } from 'victory';
+import { MaidrVictory } from '../../../src/adapters/victory/MaidrVictory';
+
+// Victory's `errorY` is a distance from the estimate, either symmetric (a
+// number) or asymmetric (`[plus, minus]`). MAIDR announces absolute bounds.
+const data = [
+  { x: 'Control', y: 4.2, errorY: 0.6 },
+  { x: 'Low dose', y: 5.1, errorY: 0.9 },
+  { x: 'Mid dose', y: 6.4, errorY: [1.2, 0.7] },
+  { x: 'High dose', y: 6.9, errorY: 1.4 },
+];
+
+export function ErrorBarExample(): JSX.Element {
+  return (
+    <div>
+      <h2>Error Bar</h2>
+      <p>
+        Group means with 95% confidence intervals. Up and down move between the
+        lower bound, the estimate and the upper bound at one sample.
+      </p>
+      <MaidrVictory
+        id="victory-error-bar"
+        title="Response by Dose"
+        subtitle="Mean with 95% CI"
+      >
+        <VictoryChart domainPadding={40}>
+          <VictoryAxis label="Group" />
+          <VictoryAxis dependentAxis label="Response (mmol/L)" />
+          <VictoryErrorBar data={data} />
+        </VictoryChart>
+      </MaidrVictory>
+    </div>
+  );
+}
--- a/examples/victory/examples/PolarAreaExample.tsx
+++ b/examples/victory/examples/PolarAreaExample.tsx
@@ -1,0 +1,36 @@
+import type { JSX } from 'react';
+import { VictoryBar, VictoryChart, VictoryPolarAxis } from 'victory';
+import { MaidrVictory } from '../../../src/adapters/victory/MaidrVictory';
+
+// A `VictoryBar` inside a `<VictoryChart polar>` draws a coxcomb: each value
+// is a wedge whose radius is its magnitude. MAIDR pans by the wedge's angle
+// rather than by its column, so the sweep goes out and back around the circle.
+const data = [
+  { x: 'N', y: 18 },
+  { x: 'NE', y: 12 },
+  { x: 'E', y: 7 },
+  { x: 'SE', y: 9 },
+  { x: 'S', y: 15 },
+  { x: 'SW', y: 22 },
+  { x: 'W', y: 27 },
+  { x: 'NW', y: 21 },
+];
+
+export function PolarAreaExample(): JSX.Element {
+  return (
+    <div>
+      <h2>Polar Area</h2>
+      <p>Wind rose showing hours of wind from each compass direction.</p>
+      <MaidrVictory
+        id="victory-polar-area"
+        title="Wind by Direction"
+        subtitle="Hours per week"
+      >
+        <VictoryChart polar>
+          <VictoryPolarAxis />
+          <VictoryBar data={data} />
+        </VictoryChart>
+      </MaidrVictory>
+    </div>
+  );
+}
--- a/examples/victory/examples/PopulationPyramidExample.tsx
+++ b/examples/victory/examples/PopulationPyramidExample.tsx
@@ -1,0 +1,51 @@
+import type { JSX } from 'react';
+import { VictoryAxis, VictoryBar, VictoryChart, VictoryStack } from 'victory';
+import { MaidrVictory } from '../../../src/adapters/victory/MaidrVictory';
+
+// A population pyramid is a `VictoryStack` of two bar series with one of them
+// signed negative, drawn horizontally so the two sides grow left and right.
+// The sign is a direction rather than a magnitude, and MAIDR reads it as one:
+// the pitch carries the size and the announcement names the side.
+const men = [
+  { x: '0-14', y: -1200 },
+  { x: '15-29', y: -1150 },
+  { x: '30-44', y: -1080 },
+  { x: '45-59', y: -990 },
+  { x: '60-74', y: -720 },
+  { x: '75+', y: -310 },
+];
+
+const women = [
+  { x: '0-14', y: 1140 },
+  { x: '15-29', y: 1100 },
+  { x: '30-44', y: 1060 },
+  { x: '45-59', y: 1010 },
+  { x: '60-74', y: 830 },
+  { x: '75+', y: 520 },
+];
+
+export function PopulationPyramidExample(): JSX.Element {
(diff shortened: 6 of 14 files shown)


jooyoungseo:
Thanks for the review — both findings addressed. All 14 checks were already green at `9381682`; these are correctness/coverage changes, not CI fixes.

**Fixed**

- **#2 exact float equality in the waterfall chain check** (`7db4b66`) — the chain test now compares within `1e-9` instead of `===`, and the same tolerance applies to the two baseline checks so "rests on the baseline" means one thing throughout. Added a test that builds each bar's foot as `end - delta` from an accumulated total, which is the realistic way a caller produces a chain that misses itself by a ULP; before the change it fell back to `BAR`.
- **#1 untested `[plus, minus]` mapping** (`0441e6c`) — added a `renderedDom` case that renders an asymmetric `errorY: [0.4, 0.2]`, calibrates px-per-unit from two rendered sample centres, and checks both bounds against the pixels Victory drew. The mapping was right (`victory-errorbar/helper-methods` scales `errors[0]` above the estimate and `errors[1]` below), but nothing in the suite could have falsified it — now something can.

**Not changed**

- The nitpicks were confirmations, not requests. The two disclosed limitations (`VictoryStack` reversed paint order for plain `STACKED`, `horizontal` polar/error-bar orientation) stay out of scope as agreed.

`lint:fix`, `type-check`, `build` and `npm test` (2679 unit + 73 ESM) all pass locally.

---

PR #874, opened 2026-08-14
feat(plotly): support the chart types added since the pie chart

### LINKED ISSUE #861: feat(plotly): support the chart types added since the pie chart
## Problem

MAIDR core has gained ~34 new trace types since the pie chart landed (radar, gauge, funnel, waterfall, gantt, treemap, sankey, sunburst, dumbbell, dot/lollipop, diverging, area family, and more). The `plotly` adapter has not kept up: of those, it currently emits **none**.

A chart plotly.js draws perfectly well is therefore either dropped from the figure or announced as the wrong chart — which for a screen-reader user is worse than not being supported at all.

Two-stage dispatch, both driven by plotly's own resolved trace objects (`gd._fullData`) rather than by the SVG. Stage one is `mapTraceType(trace)` at /home/user/maidr/src/adapters/plotly/extractor.ts:263 — a plain `switch` on `trace.type` (defaulting to `'scatter'`) that returns BAR/BOX/VIOLIN_KDE/HEATMAP/HISTOGRAM/CANDLESTICK/PIE, delegates `scatter`/`scattergl` to `mapScatterMode` (extractor.ts:301, which reads `trace.mode` for markers-vs-lines and `trace.line.shape` via `isStepShape` at extractor.ts:343 to split LINE from STEP), and for anything else logs `Unsupported plotly trace type` and returns `null` (extractor.ts:296). Stage two is layout-level, in `buildSubplotLayers` (extractor.ts:473): entries are bucketed by their stage-one type (extractor.ts:505-523), then multi-trace shapes are decided — line/step traces merge into one multi-series layer (extractor.ts:527-541), violins expand into a VIOLIN_BOX + VIOLIN_KDE pair (extractor.ts:552-554 → 1469), and >1 bar traces are re-typed from `_fullLayout.barmode`/`barnorm` into DODGED / STACKED / NORMALIZED (extractor.ts:557-580). Payload construction is a third switch, `extractLayer` (extractor.ts:1017, switch at 1036), and CSS selectors are a fourth, independent switch in /home/user/maidr/src/adapters/plotly/selectors.ts:40. Panels come from `groupTracesBySubplot` (extractor.ts:384), keyed on the `xaxis`+`yaxis` pair (extractor.ts:406) with a single hand-written exception for pie, which has no axes and is keyed/positioned by `trace.domain` instead (extractor.ts:403-412).

## In scope — 21 trace types plotly.js can render

| TraceType | plotly.js construct | Detection / payload |
|---|---|---|
| `AREA` | scatter with `fill: 'tozeroy' \| 'tozerox' \| 'toself'` (no `stackgroup`) | In `mapScatterMode` (extractor.ts:301), before the LINE/STEP decision, check `trace.fill` on the resolved `_fullData` trace: a truthy fill other than `'none'` with no `trace.stackgroup` is an AREA. Payload is identical to what `extractMultiLineLayer` (extractor.ts:1218) already builds — `LinePoint[][]`, one inner array per trace, `{x, y}` per sample — since AreaTrace extends LineTrace (src/model/area.ts:81). Only the emitted `type` changes; selectors.ts:53 already handles the scatter `.point` case. |
| `STACKED_AREA` | scatter with `stackgroup` set (plotly's native area stacking) | Same hook in `mapScatterMode`: `trace.stackgroup` non-empty and `trace.groupnorm` empty/absent → STACKED_AREA. Bucket the traces by `stackgroup` value the way `stepTraces` is bucketed by direction (extractor.ts:489, 534) so two independent stacks do not merge into one layer, then reuse `extractMultiLineLayer`. Note plotly's `trace.y` for a stacked-area trace is the trace's OWN value, not the running total, which is exactly what AreaTrace wants; the running total is derived by the model. `PlotlyTrace` in types.ts needs `stackgroup`/`groupnorm`/`fill` fields added. |
| `NORMALIZED_AREA` | scatter with `stackgroup` + `groupnorm: 'percent' \| 'fraction'` | Same branch as STACKED_AREA, one extra condition on `trace.groupnorm` — the direct analogue of the existing `barnorm === 'percent' \| 'fraction'` test at extractor.ts:566. Payload identical (`LinePoint[][]`); plotly has already normalised the values it drew, and `gd.calcdata[i][j].y` carries the post-normalisation value if the authored `trace.y` is preferred to be raw. |
| `FUNNEL` | `funnel` trace (and `funnelarea` for the area variant) | Add `case 'funnel'` to `mapTraceType` (extractor.ts:266). It is a bar in disguise — horizontal by default (`orientation: 'h'`), stages on the position axis, counts on the value axis — so `extractBarLayer` (extractor.ts:1171) works unchanged apart from the emitted type: FunnelTrace extends BarTrace (src/model/funnel.ts:43) and reads `BarPoint[]`. Stage order must be preserved as authored (plotly draws top-to-bottom). Selector: `.funnellayer .trace .point > path` scoped by the existing `subplotCssPrefix`. |
| `WATERFALL` | `waterfall` trace | Add `case 'waterfall'`. `WaterfallPoint` needs `{x, start, end, delta, kind}` and plotly's own calc already produces all four: each `gd.calcdata[i][j]` carries `s0`/`s1` (the floating bar's start and end on the value axis), `rawS` (the signed contribution) and `isSum` (the flag for a total/absolute step). Map `trace.measure[j]` — `'relative' \| 'total' \| 'absolute'` — plus the sign of `rawS` onto `WaterfallKind` (`increase`/`decrease`/`total`). Falls back to accumulating `trace.y` against `trace.base` when calcdata is absent, matching the existing calcdata-first pattern of `barPoint` (extractor.ts:1158). |
| `ERROR_BAR` | `error_y` / `error_x` on a scatter or bar trace (not a trace type of its own) | This is a modifier, so it belongs in the bucketing pass of `buildSubplotLayers` rather than in `mapTraceType`: when a scatter/bar trace has `error_y.visible !== false` and a source of values (`type: 'data'` with `array`/`arrayminus`, or `'percent'`/`'constant'`/`'sqrt'` with `value`/`valueminus`), emit ERROR_BAR instead of SCATTER/LINE/BAR. `ErrorBarPoint` wants ABSOLUTE `yMin`/`yMax`, and plotly's errorbars calc has already resolved them per point onto calcdata as `yh`/`ys` (and `xh`/`xs` for x errors), so read those first and fall back to `y ± array[i]`. Horizontal errors map to the same shape with the axes swapped, following the existing `isHorizontal` convention (extractor.ts:1184). |
| `TREEMAP` | `treemap` trace (`labels` / `parents` / `values` / `ids` / `branchvalues`) | New `case 'treemap'` plus a domain-panel branch in `groupTracesBySubplot` (see notes — it needs the same `isPie` treatment at extractor.ts:403, since a treemap has no `xaxis`/`yaxis` and would otherwise be filed into the cartesian panel). `TreemapPoint` wants `{x: label, y: value, path: ancestors-root-first-excluding-self}`; plotly gives a parent POINTER (`parents[i]`, matching `ids[i]` when present else `labels[i]`), so walk the parent chain once per node to build `path`. Emit only leaves — the grammar derives interior totals (src/type/grammar.ts:793 doc) — or emit interior nodes too when `branchvalues: 'total'` declares a value the children do not sum to. Highlighting selector: `.treemaplayer .trace .slice path`. |
| `SUNBURST` | `sunburst` trace (same labels/parents/values shape) | Identical code path to TREEMAP — one shared `extractHierarchyLayer(trace, type)` serves both, because TraceFactory routes ICICLE/SUNBURST/TREEMAP to the same TreemapTrace (src/model/factory.ts) and only the emitted `type` differs (the sunburst's angular panning reads `this.type`). Selector: `.sunburstlayer .trace .slice path`. |
| `ICICLE` | `icicle` trace | Third case of the same shared hierarchy extractor; plotly's `icicle` carries exactly the same `labels`/`parents`/`values`/`ids` attributes. Selector: `.iciclelayer .trace .slice path`. Effectively free once TREEMAP lands. |
| `SANKEY` | `sankey` trace | New `case 'sankey'` plus the domain-panel branch. `FlowPoint[]` is `{source, target, value}` by NAME, while plotly stores `link.source[k]`/`link.target[k]` as integer indices into `node.label` — so the extractor resolves each index through `trace.node.label` (falling back to the index as a string when labels are absent). `link.value[k]` is the magnitude verbatim. Nodes are derived from the edges by the model, so no node list is emitted. Highlight selector: `.sankey .sankey-link path` / `.sankey-node`. |
| `GAUGE` | `indicator` trace with `mode` containing `'gauge'` (gauge and bullet chart) | New `case 'indicator'`, gated on `trace.gauge` being present. `GaugePoint` is a single OBJECT, not an array (src/type/grammar.ts:530): `value` ← `trace.value`; `min`/`max` ← `trace.gauge.axis.range`; `label` ← `trace.title.text`; `target` ← `trace.gauge.threshold.value` (or `trace.delta.reference`); `bands` ← `trace.gauge.steps[]`, whose `range[1]` gives `GaugeBand.to`. Honest caveat: plotly's steps carry only `range` + `color`, never a name, so `GaugeBand.label` has no source — emit bands only when the author supplied `name`, otherwise omit `bands` rather than invent 'poor/ok/good'. Needs the domain-panel branch (indicator is positioned by `trace.domain`). |
| `RADAR` | `scatterpolar` (mode lines/lines+markers, usually `fill: 'toself'`) | Polar traces do not carry `xaxis`/`yaxis`; they carry `trace.subplot` (e.g. `'polar'`, `'polar2'`) and are laid out by `layout.polar*.domain`, so `groupTracesBySubplot` (extractor.ts:384) needs a polar key alongside the pie key. RadarTrace extends LineTrace, so the payload is `LinePoint[][]`: one inner array per trace, `{x: theta[i] (spoke label), y: r[i]}`. Axis labels come from `layout.polar.angularaxis.title` / `radialaxis.title` via a polar-aware variant of `resolveAxisLabel` (extractor.ts:734). Selector: `.polarlayer .trace .point`. |
| `POLAR_AREA` | `barpolar` (the rose / coxcomb chart) | Same polar grouping work as RADAR; the only difference is the mark, and TraceFactory sends POLAR_AREA and RADAR to the same RadarTrace. Detection is exactly `trace.type === 'barpolar'`, payload the same `LinePoint[][]` built from `r`/`theta`. Nearly free once the polar subplot branch exists. |
| `PARALLEL` | `parcoords` (parallel coordinates) | New `case 'parcoords'` + domain-panel branch. ParallelTrace extends LineTrace and reads `LinePoint[][]` as one ROW PER OBSERVATION and one column per axis — plotly stores the transpose, `trace.dimensions[c] = {label, values[]}` — so emit `data[r][c] = {x: dimensions[c].label, y: dimensions[c].values[r]}`. The trace computes its own per-column extents (src/model/parallel.ts:50-67), so no per-axis range needs emitting, though `dimensions[c].range` is available if wanted. Highlighting is effectively unavailable (parcoords draws to canvas, not per-line SVG) — emit without selectors, which the grammar allows. |
| `CHOROPLETH` | `choropleth` (and `choroplethmapbox` / `choroplethmap`) | New case + a geo-panel branch (`trace.geo` / `trace.subplot` instead of an axis pair). `ChoroplethPoint` is `{x: region name, y: value, lon?, lat?, neighbors?}`: `x` ← `trace.locations[i]` (or `trace.text[i]` when locations are ISO codes and text carries the names), `y` ← `trace.z[i]`. The optional centroid is genuinely recoverable after draw — plotly resolves each region to a topojson/GeoJSON feature and hangs it on calcdata with a precomputed `ct` ([lon, lat]) — so read `cd[i].ct` when present and omit `lon`/`lat` otherwise. `neighbors` has no source in plotly and must stay omitted (the grammar explicitly prefers silence to a guess). |
| `RIDGELINE` | `violin` traces with `side: 'positive'` (plotly's own documented ridgeline recipe: horizontal violins, one per y category, half-shown and overlapped) | In the violin bucket of `buildSubplotLayers` (extractor.ts:552), check whether every violin trace in the subplot has `side === 'positive'` (typically with `orientation: 'h'`, `points: false`, `width > 0` overlapping); if so emit ONE ridgeline layer instead of the VIOLIN_BOX + VIOLIN_KDE pair. The payload is the same `ViolinKdePoint[][]` that `buildViolinKdeLayer` already builds from `cd.density` samples (extractor.ts:1697-1718) — `{x: group label, y: sample.t, density: sample.v}` — because RidgelineTrace holds exactly that shape. This is the cheapest of the non-native wins: the density extraction, the label resolution (extractor.ts:1568) and the selector building already exist. |
| `GANTT` | horizontal `bar` traces with a `base` array — what `plotly.express.timeline` (and the `figure_factory.create_gantt` output) emits | In the bar bucket, before the barmode branch (extractor.ts:557), detect `orientation === 'h'` + `Array.isArray(trace.base)` + an x axis of `type: 'date'`. Each bar is then an interval: `start` ← `base[i]`, `end` ← `base[i] + x[i]` (plotly encodes the duration in `x`, in ms for date axes), lane ← `y[i]`. `GanttData` is an object, not an array (src/type/grammar.ts:579): group intervals by lane into `points: GanttPoint[][]`, fill `lanes` from the y axis `_categories` so empty lanes survive, and set `unit` from the axis (e.g. 'days') after dividing the ms durations down. One layer per subplot, merging all such bar traces. |
| `ALLUVIAL` | `parcats` (parallel categories diagram) | New `case 'parcats'` + domain-panel branch. Flows are derived from consecutive dimension pairs: for each adjacent `dimensions[c]` / `dimensions[c+1]`, group the row indices by the `(values[r], values[r+1])` pair and sum `trace.counts` (or count rows when `counts` is absent) to get one `FlowPoint {source, target, value}` per pair. Node names must be stage-qualified (`'Stage: Category'` or `dimensions[c].label + '·' + value`) because the same category label recurs at every stage and FlowPoint derives its nodes from the edge names — unqualified names would fold the stages into a cyclic graph. |
| `DIVERGING` | two `bar` traces with signed values under `barmode: 'relative'` (the standard plotly population pyramid / Likert idiom) | Refine the existing barmode branch at extractor.ts:565: when the mode is `'relative'`/`'overlay'`, the bars are horizontal on a shared category axis, and one trace's values are all ≤ 0 while another's are all ≥ 0, emit DIVERGING instead of STACKED. The payload is the `SegmentedPoint[][]` that `extractSegmentedBarLayer` (extractor.ts:2020) already produces — z carries the side's legend name — with the one requirement that the values stay SIGNED, which means taking the raw `x`/`y` rather than `cd.s` (calcdata's `s` is the drawn size and would lose the sign that DivergingTrace reads as a direction). |
| `DOT` | `scatter` with `mode: 'markers'` on a categorical axis (Cleveland dot plot) | In `mapScatterMode` (extractor.ts:301), when the mode is markers-only and the resolved position axis has `type: 'category'` with at most one marker per category, emit DOT instead of SCATTER. Payload is `BarPoint[]` — `{x: category, y: value}` — since DOT, LOLLIPOP and BAR all route to BarTrace, so `extractBarLayer` (extractor.ts:1171) is reusable with the value taken from `trace.y` rather than `cd.s`. Worth flagging as a judgement call: it changes how an existing chart is announced (SCATTER today), and a genuinely scattered categorical plot would be mis-typed, so it may want to be conservative (exactly one point per category, no duplicate x). |
| `WORD_CLOUD` | `scatter` with `mode: 'text'` and a per-point `textfont.size` array (plotly has no word-cloud trace; this is its documented recipe) | Detect `mode` containing `'text'`, `trace.text` an array, and `trace.textfont.size` an ARRAY (not a scalar). `WordCloudPoint` is `{x: term, y: weight}` with x ← `text[i]`. The honest caveat is `y`: the only magnitude on the trace is the glyph size in px, which is a scaled transform of the weight rather than the weight, so announcing it as a frequency overstates what the chart carries. Emit `y` from `trace.customdata[i]` or a numeric `hovertext` when the author supplied one, and otherwise treat the size as an ordinal weight; without either, this may not be worth claiming. |

## Deferred to a follow-up

These need a new author-facing opt-in API or semantics plotly.js does not carry in its own configuration, so they are out of scope here and tracked separately:

- `CONTOUR` — New `case 'contour'`. The blocker is that `ContourPoint[][]` wants the actual iso-curve polylines, and plotly computes those (`pathinfo`, marching squares) at draw time without persisting them on the graph div — `calcdata[i][0]` keeps only the `z` grid plus `x`/`y` coordinate arrays. Two routes: (a)
- `MOSAIC` — In the barmode `'stack'`/`'relative'` branch, detect `Array.isArray(trace.width)` with non-uniform entries. `MosaicPoint` extends SegmentedPoint with `width` (the column's share of all observations, as a fraction of one) and optional `count`, so `extractSegmentedBarLayer` (extractor.ts:2020) extends

## Not renderable by plotly.js

<details><summary>3 trace types the library genuinely cannot draw</summary>

- BOXEN — plotly.js has no letter-value/boxen trace. Its `box` trace computes only the five-number summary (q1/median/q3/fences, plus outlier points); the variable-depth quantile ladder a boxen is drawn from is never computed and cannot be recovered from what plotly renders.
- CHORD — plotly.js has no circular-flow trace. It ships `sankey` (left-to-right ribbons) and `parcats` (staged ribbons) but nothing that draws ribbons around a circle; a chord would have to be hand-built from `layout.shapes`, which the adapter has no way to identify as a chord.
- HEXBIN — plotly.js has no hexagonal-binning trace. `histogram2d` / `histogram2dcontour` bin on a rectangular lattice, which is a HEATMAP and is what the adapter should emit for them. (Caveat for the record: plotly.py's `figure_factory.create_hexbin_mapbox` does produce hexagons, but it emits them as a `choroplethmapbox` of hex GeoJSON polygons, which reaches this adapter indistinguishable from any other choropleth — so it is not a signal the JS side can act on.)

</details>

## Acceptance criteria

- [ ] Unit tests under `test/adapters/plotly/` cover every newly detected type.

## Implementation notes

DRAWABLE BUT NOT IDENTIFIABLE — these belong in neither bucket honestly, so recording them here. plotly.js can draw each, but only as a composition of generic scatter/bar traces with no attribute that distinguishes it, so the adapter cannot claim them without an author opt-in (e.g. a convention on `trace.meta` or a `data-maidr-type` attribute): BUMP (a line chart with a reversed rank axis — currently emitted as LINE), LOLLIPOP (markers plus stems from `layout.shapes` or a hairline bar), DUMBBELL (two marker traces plus connecting segments), FOREST (scatter + `error_x` on a category axis — the adapter can honestly reach ERROR_BAR, but `nullValue`, per-study `weight` and the `pooled` row have no source), MANHATTAN and VOLCANO (plain scatter; the thresholds and point identities are not on the trace), SURVIVAL (an `hv` step scatter — currently emitted as STEP; censoring marks and median survival are not recoverable), NETWORK (plotly's documented node-link recipe is one scatter of edges with `None` separators plus one scatter of nodes — indistinguishable from any other scatter pair).

SHARED PREREQUISITE — panel grouping. `groupTracesBySubplot` (extractor.ts:384) keys panels on `${trace.xaxis}${trace.yaxis}` (line 406) and has exactly one hand-written exception, for pie, which is keyed by trace index and positioned from `trace.domain` via `readTraceDomain` (extractor.ts:403-412, 709). Every domain-positioned trace type — treemap, sunburst, icicle, sankey, indicator, parcoords, parcats, funnelarea — has no `xaxis`/`yaxis` and would silently collapse into the `xy` cartesian panel, inheriting that panel's axis labels. Polar (`trace.subplot === 'polar'`), geo (`trace.geo`), ternary and mapbox traces have the same problem with their own keys. Generalising the pie exception into a "panel key + domain" resolver is the single change that unblocks TREEMAP/SUNBURST/ICICLE/SANKEY/GAUGE/RADAR/POLAR_AREA/PARALLEL/ALLUVIAL/CHOROPLETH at once, and it is worth doing first rather than adding a second one-off.

SELECTORS. `generatePlotlySelectors` (selectors.ts:29) is a separate switch that returns `undefined` in its default branch (selectors.ts:70), so a new trace type will emit and navigate correctly but with no visual highlight until a case is added. Cartesian types use the `subplotCssPrefix` scoping (`.subplot.xy `, selectors.ts:85); non-cartesian types need their own layer scoping, and `pieSelector` (selectors.ts:118) is the worked example — it counts drawn siblings inside `.pielayer` because plotly gives pies no uid class. The same counting trick applies to `.treemaplayer` / `.sunburstlayer` / `.iciclelayer` / `.funnellayer` / `.polarlayer`.

CALCDATA-FIRST CONVENTION. The adapter consistently prefers what plotly computed over what the author wrote — `barPoint` reads `cd.s` for the post-`barnorm` drawn size (extractor.ts:1158-1169), `drawnPieSlices` reads calcdata because plotly reorders slices (extractor.ts:1817), histogram bins come from `cd.p`/`cd.s` and the layer is SKIPPED when calcdata is absent (extractor.ts:1901, 1951), violin KDE comes from `cd.density` (extractor.ts:1707). New types should follow it: waterfall's `s0`/`s1`/`rawS`/`isSum`, error bars' `yh`/`ys`, choropleth's resolved `ct` centroid all live on calcdata for the same reason. `PlotlyCalcData` (types.ts:163) already has an index signature, so new fields cost only a typed property.

TESTS. /home/user/maidr/test/adapters/plotly/extractor.test.ts builds fake graph divs with `createGraphDiv({id, traces, layout, bgRects, calcdata})` (test file line 24) — no real plotly needed, `_fullData`/`_fullLayout`/`calcdata` are set directly — and several tests round-trip the extracted schema through the real `Figure` model (the `core-model integration` block at line 1374), which is the pattern to copy for a new type. TRAP: there is an explicit test asserting that a `sunburst` trace warns as unsupported (extractor.test.ts:124-144); it will fail the moment SUNBURST is implemented and must be repointed at a still-unsupported type.

OTHER. plotly.js is NOT a dependency of this repo (nothing in package.json, nothing in node_modules) — examples load it from a CDN, and there are 13 `examples/plotly-*.html` files, none of them a post-pie type. Post-pie examples such as `examples/sunburst.html`, `funnel.html`, `waterfall.html` are hand-authored MAIDR JSON on static SVG, not library output, so they are useful as payload references but not as adapter fixtures. No other adapter emits most of these either — the only post-pie precedent anywhere in src/adapters is AREA / STACKED_AREA / NORMALIZED_AREA in vegalite/converters.ts (see its `case 'area'` at ~line 677) and STACKED_AREA in chartjs/extractor.ts; the vegalite stacking decision is the closest model for the plotly `stackgroup`/`groupnorm` work. Finally, docs/plotly.md carries a support matrix (lines 59-75) and a line claiming "9 Plotly types" (line 422) that needs updating alongside any addition.

jooyoungseo:
## Description

The `plotly` adapter had not kept up with the trace types MAIDR core gained
since the pie chart: of the types in scope it emitted none, so a chart
plotly.js draws perfectly well was either dropped from the figure or announced
as the wrong chart.

This adds **20 of the 21 in-scope trace types**, taking the adapter from 9 to
29 recognised Plotly constructs. The work is adapter-local — detection,
payload construction and highlight selectors only. No file under `src/model/`,
`src/service/`, `src/state/`, `src/ui/`, `src/command/` or `src/type/grammar.ts`
is touched; the core already supported every payload emitted here.

Each type is wired end to end in the same change: detection, the grammar
payload, **and** the CSS selector path. A type that sonifies but does not
highlight was not counted as done.

## Related Issues

Closes #861

### Trace types added (20)

| TraceType | Detected from |
|---|---|
| `AREA` | `scatter` with `fill` other than `none`, no `stackgroup` |
| `STACKED_AREA` | `scatter` with `stackgroup` |
| `NORMALIZED_AREA` | `stackgroup` + `groupnorm: 'percent' \| 'fraction'` |
| `FUNNEL` | `type: 'funnel'` |
| `WATERFALL` | `type: 'waterfall'` |
| `ERROR_BAR` | visible `error_y` / `error_x` on a scatter or bar trace |
| `SUNBURST` | `type: 'sunburst'` |
| `ICICLE` | `type: 'icicle'` |
| `TREEMAP` | `type: 'treemap'` |
| `SANKEY` | `type: 'sankey'` |
| `GAUGE` | `type: 'indicator'` with `gauge` in `mode` |
| `RADAR` | `type: 'scatterpolar'` |
| `POLAR_AREA` | `type: 'barpolar'` |
| `PARALLEL` | `type: 'parcoords'` |
| `RIDGELINE` | `violin` traces with `side: 'positive'` |
| `GANTT` | horizontal `bar` traces with a `base` array on a date axis |
| `DIVERGING` | `barmode: 'relative'` + bar traces with opposed signs |
| `DOT` | `scatter`, `mode: 'markers'`, one marker per category |
| `WORD_CLOUD` | `scatter`, `mode: 'text'`, array `textfont.size` |
| `CHOROPLETH` | `type: 'choropleth'` |

### Trace types NOT added

- **`ALLUVIAL`** (`parcats`) — skipped. Two blockers, both verified against
  upstream. (1) `traces/parcats/calc.js` models a path as a full
  N-dimensional `categoryInds` tuple keyed across *all* dimensions plus
  colour, not as pairwise flows; deriving `FlowPoint`s would mean inventing a
  pairwise re-aggregation rather than reading what Plotly computed, against
  the adapter's calcdata-first convention. (2) `FlowPoint` keys its nodes by
  *label* and has no id field, so the same category label recurring in several
  dimensions (`"Yes"` under both `Survived` and `Smoker`) would fold into one
  node and produce a cyclic graph with wrong totals. Stage-qualifying the
  names, as the issue suggests, makes the node labels an adapter invention
  rather than the chart's own text. Additionally `traces/parcats/parcats.js`
  draws one `path.path` per whole path across every dimension, so there is no
  DOM element per pairwise flow to attach a highlight to.

The following were declared out of scope by the issue itself and are not
attempted here: `CONTOUR` and `MOSAIC` (deferred to a follow-up), and
`BOXEN` / `CHORD` / `HEXBIN` (not renderable by plotly.js).

`choroplethmapbox`, `choroplethmap`, `scattergeo` and the 3D traces continue
to warn as unsupported — they draw to a WebGL canvas with no element per
region, so there is nothing to highlight.

### Shared machinery

- `groupTracesBySubplot` generalised from the one hand-written pie exception
  into a panel-key + domain resolver. Domain-positioned types
  (`DOMAIN_POSITIONED_TYPES`) and subplot-keyed families (`POLAR_TYPES`,
  `GEO_TYPES`) are now one set entry plus a `mapTraceType` case.
- `extractMultiLineLayer` made generic over the line family; `extractBarLayer`
  parameterised by emitted type; new `extractHierarchyLayer` shared by
  sunburst / icicle / treemap.
- `drawnBefore`, `scatterTraceScope`, `polarSeriesSelectors`,
  `barPointSelector` in `selectors.ts` for the layers Plotly hangs no uid
  class on.

### Docs and examples

- `docs/plotly.md` support matrix extended, per-type detection notes added,
  and the type count updated 9 → 29.
- 15 new `examples/plotly-*.html` files, one per new chart family.

### Tests

175 unit tests under `test/adapters/plotly/`, including `core-model
integration` cases that round-trip the emitted payloads through a real
`Figure` so the payload *shape*, not just the type string, is checked against
what `TraceFactory` can build.

## Checklist

The manual-testing box is left unticked deliberately: `ManualTestingProcess.md`
is not present in the repo, and plotly.js is not a dependency (the examples
load it from a CDN), so no rendered-chart pass was performed. The automated
gate below is green.

### Verification method

Because plotly.js is not in `node_modules`, every behavioural claim was checked
against the upstream source over the network rather than against the issue
spec. That caught several spec claims that would have shipped bugs:

- Waterfall does **not** get `s0`/`s1` from its own calc — those arrive later
  from bar's `setGroupPositions` as pixel-layout helpers. The authority is
  `traces/waterfall/hover.js`; reading `cd.s` for the contribution announces
  the running total twice.
- For stacked/normalized area, `calcdata[i][j].y` is the **cumulative** value,
  not the band — feeding it to `AreaTrace` double-counts every series.
- For diverging bars, `cd.s` keeps the signed authored size (contrary to the
  spec), so the existing segmented-bar extractor works unchanged.
- Emitting a hierarchy in authored order would have misaligned every
  highlight, since Plotly re-sorts children by value; the adapter walks the
  computed tree instead.

### Known limitations

- A **lines-only area or line** trace gets no highlight, matching existing
  LINE/STEP behaviour — `lineSelector` returns a selector only when the trace
  has markers. Fixing it properly means per-segment path parsing that breaks
  on gapped series, and would touch LINE/STEP too; worth its own change.
- `parcoords` draws to canvas, so the layer carries no selectors — audio,
  text, braille and navigation work, visual highlighting does not.
- A hierarchy authored with `ids` **and** duplicate sibling labels collapses,
  because the grammar keys the tree by label path. A grammar limitation, not
  an adapter one.
- `trace.node.groups` on a sankey generates nodes past `node.label`, which
  fall back to their index as a name.
- A gantt whose bar traces carry `ids` gets d3-keyed groups that can reorder
  on update and desynchronise the selectors (first render is fine). Its axis
  formatter uses `new Function`, which a strict-CSP page refuses —
  `FormatUtil.resolveFormat` catches that and falls back, so positions then
  read as scaled numbers rather than dates. A panel mixing a schedule with an
  ordinary bar chart stays a bar chart.
- Gauge qualitative bands are emitted only when the author named every step,
  since Plotly's steps carry a range and colour but no required name — an
  invented "band 2" would say nothing the numbers do not.

### Gate

`npm run lint:fix`, `npm run type-check` (both tsconfigs), `npm run build` and
`npm test` all pass; `npm run sync:copilot:check` reports in sync. Full suite:
187 suites / 2711 tests plus the ESM project's 7 suites / 73 tests, all green.

---

### DIFF
--- a/docs/plotly.md
+++ b/docs/plotly.md
@@ -78,6 +78,14 @@ For dynamically-created charts (SPAs, notebooks), a `MutationObserver` watches f
 | Funnel | `type: 'funnel'` | [Funnel chart](examples.html) |
 | Waterfall | `type: 'waterfall'` | [Waterfall chart](examples.html) |
 | Error Bars | `error_y` or `error_x` on a scatter or bar trace | [Error bars](examples.html) |
+| Sunburst | `type: 'sunburst'` | [Sunburst](examples.html) |
+| Icicle | `type: 'icicle'` | [Icicle](examples.html) |
+| Treemap | `type: 'treemap'` | [Treemap](examples.html) |
+| Sankey | `type: 'sankey'` | [Sankey](examples.html) |
+| Gauge / Bullet | `type: 'indicator'` with `gauge` in `mode` | [Gauge](examples.html) |
+| Radar | `type: 'scatterpolar'` | [Radar](examples.html) |
+| Polar Area / Rose | `type: 'barpolar'` | [Polar area](examples.html) |
+| Parallel Coordinates | `type: 'parcoords'` | [Parallel coordinates](examples.html) |
 | Subplots / Facets | multiple `xaxis`/`yaxis` pairs, `layout.grid`, or Plotly Express facets | [Subplots](examples.html) |
 
 **Notes on chart-type detection:**
@@ -125,6 +133,49 @@ For dynamically-created charts (SPAs, notebooks), a `MutationObserver` watches f
   bar's height, which is what the pitch follows) and the running total it
   produced (the bar's position). Steps whose `measure` is `total` or `absolute`
   are announced as totals, so a subtotal is not mistaken for a contribution.
+
+- A sunburst, an icicle and a treemap are one tree drawn three ways, so MAIDR
+  reads all three the same: the hierarchy navigates *as a hierarchy* on the
+  arrow keys that already exist — up to the parent, down to the first child,
+  left and right between siblings — and each node announces its share of its
+  parent. Plotly stratifies `labels`/`parents` into a tree and, unless the
+  trace sets `sort: false`, reorders every node's children largest first, so
+  the adapter walks the tree Plotly computed rather than the arrays as written.
+  A hierarchy Plotly has not computed yet is still read out, in the authored
+  order, but without selectors — sector *k* is then not slice *k*.
+
+- A sankey names both ends of every flow. Plotly stores them as indices into
+  `node.label`, so the adapter resolves them back to the labels: "34 from Coal
+  to Electricity" is the reading and "34 from 0 to 1" is not. The nodes are
+  derived from the flows, so nothing is emitted for them separately. Flows
+  Plotly drops before drawing — a non-positive value, an endpoint that is not a
+  node — are dropped here too, since a ribbon nothing draws would still put a
+  node in the graph.
+
+- An `indicator` trace is a gauge only when it draws one. Without `gauge` in
+  its `mode` it is a number set in text, which a screen reader already reaches,
+  so it is skipped rather than announced twice. A gauge's `gauge.threshold.value`
+  becomes the target; `delta.reference` stands in when there is no threshold,
+  except when it equals the measure, which is the value Plotly defaults it to
+  and not a target anyone set. Plotly's steps carry a range and a colour but no
+  name, so the qualitative bands are announced only when the author named every
+  step — an invented "band 2" would say nothing the numbers do not.
+
+- A polar trace has no axis pair either: `scatterpolar` and `barpolar` name a
+  `subplot` (`polar`, `polar2`, …) and are positioned by `layout.polar.domain`,
+  so each polar subplot becomes its own MAIDR panel. Both read as spokes and
+  values, the way a multi-line chart reads as samples and values, and the
+  circle is carried in the panning rather than in the payload. Plotly's schema
+  has no title for the angular axis, so the spokes are named `Spoke`; the
+  radial axis title is read when there is one.
+
+- A `parcoords` layer is transposed on the way in: Plotly stores a column of
+  values per axis, and MAIDR reads a row per observation. The pitch is scaled
+  per axis rather than for the layer, since the columns are different
+  quantities and one range for all of them would sonify the units instead of
+  the data. Plotly draws these lines to a canvas rather than to SVG, so the
+  layer carries no selectors — audio, text, braille, and navigation all work,
+  visual highlighting alone does not.
 
 - Plotly sorts pie slices by descending value unless the trace sets
   `sort: false`, so the authored order is not necessarily the drawn order. The
@@ -454,7 +505,7 @@ For the full list, see the [Keyboard Controls](docs/CONTROLS.html) reference.
 | Data source | Manual JSON schema | Manual JSON schema | Auto-extracted from Plotly |
 | SVG selectors | Manual CSS selectors | Manual CSS selectors | Auto-generated |
 | Configuration | Required | Required | Zero configuration |
-| Chart types | All MAIDR types | All MAIDR types | 15 Plotly types |
+| Chart types | All MAIDR types | All MAIDR types | 23 Plotly types |
 | Dynamic charts | Manual init | React lifecycle | Auto-detected |
 
 ## Python and R Binders
--- a/docs/plotly.md
+++ b/docs/plotly.md
@@ -72,6 +72,12 @@ For dynamically-created charts (SPAs, notebooks), a `MutationObserver` watches f
 | Pie | `type: 'pie'` | [Pie chart](examples.html) |
 | Grouped Bar | `barmode: 'group'` + multiple bar traces | [Grouped bar](examples.html) |
 | Stacked Bar | `barmode: 'stack'` + multiple bar traces | [Stacked bar](examples.html) |
+| Area | `type: 'scatter'`, `fill: 'tozeroy' \| 'tozerox' \| 'toself'` | [Area chart](examples.html) |
+| Stacked Area | `type: 'scatter'` + `stackgroup` | [Stacked area](examples.html) |
+| 100% Stacked Area | `stackgroup` + `groupnorm: 'percent' \| 'fraction'` | [Normalized area](examples.html) |
+| Funnel | `type: 'funnel'` | [Funnel chart](examples.html) |
+| Waterfall | `type: 'waterfall'` | [Waterfall chart](examples.html) |
+| Error Bars | `error_y` or `error_x` on a scatter or bar trace | [Error bars](examples.html) |
 | Subplots / Facets | multiple `xaxis`/`yaxis` pairs, `layout.grid`, or Plotly Express facets | [Subplots](examples.html) |
 
 **Notes on chart-type detection:**
@@ -90,6 +96,35 @@ For dynamically-created charts (SPAs, notebooks), a `MutationObserver` watches f
   whichever cartesian panel happens to use the first axis pair. `axes.x` and
   `axes.y` are named `Label` and `Value`, since there is no drawn axis title to
   read them from. A doughnut (`hole`) is the same trace and reads identically.
+
+- A filled scatter is an area chart. Naming a `stackgroup` makes it a stacked
+  one, and adding `groupnorm` makes that a 100% stacked one; each stack group
+  in a panel becomes its own layer, so two independent stacks are never merged
+  into a running total neither of them draws. The layer carries each band's
+  **own** value rather than the running edge Plotly draws it at — MAIDR derives
+  the totals and each band's share of them — except under `groupnorm`, where
+  the rescaled heights Plotly drew are read from its calculated data so the
+  announced number matches the percentage axis. A filled trace whose
+  `line.shape` is step-wise stays a step chart: with nothing accumulating, the
+  fill is decoration and the staircase convention is the part worth announcing.
+
+- Error bars are a modifier rather than a trace type, so a scatter or bar trace
+  whose `error_y` (or `error_x`) is visible becomes an error-bar layer instead
+  of a scatter, line, or bar one. The interval is what such a chart is drawn to
+  show, and a scatter reading announces the estimate and drops it. MAIDR reads
+  the absolute bounds Plotly resolved for each sample, whichever way they were
+  declared — `array`, `percent`, `constant`, or `sqrt` — and navigates them as
+  three rows: lower bound, value, upper bound.
+
+- A funnel is read as a bar chart whose order means something: the pitch
+  carries the **retention** between adjacent stages rather than the count,
+  since the drop-off is what a funnel is read for and a ratio is what a
+  listener cannot take by ear. The counts are announced alongside it.
+
+- A waterfall step carries both numbers the bar draws: the contribution (the
+  bar's height, which is what the pitch follows) and the running total it
+  produced (the bar's position). Steps whose `measure` is `total` or `absolute`
+  are announced as totals, so a subtotal is not mistaken for a contribution.
 
 - Plotly sorts pie slices by descending value unless the trace sets
   `sort: false`, so the authored order is not necessarily the drawn order. The
@@ -419,7 +454,7 @@ For the full list, see the [Keyboard Controls](docs/CONTROLS.html) reference.
 | Data source | Manual JSON schema | Manual JSON schema | Auto-extracted from Plotly |
 | SVG selectors | Manual CSS selectors | Manual CSS selectors | Auto-generated |
 | Configuration | Required | Required | Zero configuration |
-| Chart types | All MAIDR types | All MAIDR types | 9 Plotly types |
+| Chart types | All MAIDR types | All MAIDR types | 15 Plotly types |
 | Dynamic charts | Manual init | React lifecycle | Auto-detected |
 
 ## Python and R Binders
--- a/docs/plotly.md
+++ b/docs/plotly.md
@@ -86,6 +86,12 @@ For dynamically-created charts (SPAs, notebooks), a `MutationObserver` watches f
 | Radar | `type: 'scatterpolar'` | [Radar](examples.html) |
 | Polar Area / Rose | `type: 'barpolar'` | [Polar area](examples.html) |
 | Parallel Coordinates | `type: 'parcoords'` | [Parallel coordinates](examples.html) |
+| Ridgeline | `type: 'violin'` with `side: 'positive'` | [Ridgeline](examples.html) |
+| Gantt / Timeline | horizontal `bar` traces with a `base` array on a date axis | [Gantt chart](examples.html) |
+| Diverging Bar / Pyramid | `barmode: 'relative'` + bar traces with opposed signs | [Population pyramid](examples.html) |
+| Dot Plot | `type: 'scatter'`, `mode: 'markers'`, one marker per category | [Dot plot](examples.html) |
+| Word Cloud | `type: 'scatter'`, `mode: 'text'`, array `textfont.size` | [Word cloud](examples.html) |
+| Choropleth | `type: 'choropleth'` | [Choropleth map](examples.html) |
 | Subplots / Facets | multiple `xaxis`/`yaxis` pairs, `layout.grid`, or Plotly Express facets | [Subplots](examples.html) |
 
 **Notes on chart-type detection:**
@@ -505,7 +511,7 @@ For the full list, see the [Keyboard Controls](docs/CONTROLS.html) reference.
 | Data source | Manual JSON schema | Manual JSON schema | Auto-extracted from Plotly |
 | SVG selectors | Manual CSS selectors | Manual CSS selectors | Auto-generated |
 | Configuration | Required | Required | Zero configuration |
-| Chart types | All MAIDR types | All MAIDR types | 23 Plotly types |
+| Chart types | All MAIDR types | All MAIDR types | 29 Plotly types |
 | Dynamic charts | Manual init | React lifecycle | Auto-detected |
 
 ## Python and R Binders
--- a/examples/plotly-area.html
+++ b/examples/plotly-area.html
@@ -1,0 +1,71 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Plotly.js Area Charts - Auto MAIDR</title>
+    <script src="https://cdn.plot.ly/plotly-2.35.2.min.js" charset="utf-8"></script>
+    <script type="text/javascript" src="../dist/maidr.js"></script>
+  </head>
+  <body>
+    <h1>Plotly.js Area Charts</h1>
+    <p>
+      Three filled plotly.js charts auto-detected by MAIDR. Click on a chart and
+      use arrow keys to navigate; up and down move between series. On the
+      stacked charts each point is announced with the running total of the
+      column it sits in and its share of that total, which is the number the
+      band heights are drawn to compare.
+    </p>
+
+    <h2>Filled bands, independent of one another (<code>fill</code>)</h2>
+    <div id="area" style="width: 700px; height: 400px"></div>
+
+    <h2>Bands stacked on one another (<code>stackgroup</code>)</h2>
+    <div id="stacked-area" style="width: 700px; height: 400px"></div>
+
+    <h2>Bands as shares of a common total (<code>groupnorm</code>)</h2>
+    <div id="normalized-area" style="width: 700px; height: 400px"></div>
+
+    <script>
+      // Quarterly revenue by region. The stacked charts read the same numbers;
+      // only how plotly lays the bands out differs.
+      var quarters = ['Q1', 'Q2', 'Q3', 'Q4'];
+      var east = [120, 80, 140, 160];
+      var west = [90, 70, 60, 110];
+
+      function series(name, values, extra) {
+        var trace = {
+          x: quarters,
+          y: values,
+          mode: 'lines',
+          type: 'scatter',
+          name: name
+        };
+        for (var key in extra) {
+          trace[key] = extra[key];
+        }
+        return trace;
+      }
+
+      function areaChart(divId, title, extra) {
+        Plotly.newPlot(
+          divId,
+          [series('East', east, extra), series('West', west, extra)],
+          {
+            title: { text: title },
+            xaxis: { title: { text: 'Quarter' } },
+            yaxis: { title: { text: 'Revenue' } }
+          }
+        );
+      }
+
+      areaChart('area', 'Revenue by region', { fill: 'tozeroy' });
+      areaChart('stacked-area', 'Revenue by region (stacked)', {
+        stackgroup: 'one'
+      });
+      areaChart('normalized-area', 'Revenue by region (share)', {
+        stackgroup: 'one',
+        groupnorm: 'percent'
+      });
+    </script>
+  </body>
+</html>
--- a/examples/plotly-choropleth.html
+++ b/examples/plotly-choropleth.html
@@ -1,0 +1,44 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <title>Plotly.js Choropleth Map - Auto MAIDR</title>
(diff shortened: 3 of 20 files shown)


jooyoungseo:
Thanks — the error-bar finding was a real bug. All fourteen checks were green at `6a6254b`; this is review follow-up only. Pushed as `c358b65`.

**Fixed**

- **Error-bar selector scoped per trace** (critical). Confirmed: two traces with intervals on one panel got the identical subplot-wide selector, and `ErrorBarTrace.mapToSvgElements` returns `null` on the count mismatch, so both lost highlighting. Scatter traces are now named by their uid class (as `DOT`/`WORD_CLOUD` already do), bar traces by their position among the panel's drawn bar-layer groups (histograms counted, hidden/legend-only ones not). Two new tests: two error-bar scatters on one panel, and a bar one sitting behind a drawn and a `legendonly` bar.
- **Hierarchy `maxdepth`/`level`** now withholds selectors. The walk yields every computed node while only some have a `g.slice`, so sector k would highlight another sector. `maxdepth: -1` still keeps them.
- **Waterfall calc-path test** now discriminates — the calcdata amounts differ from the trace's own, so a fallback to `authoredWaterfallSteps` fails it. (I could not reach plotly.js source either; `cd.v` matches what waterfall's hover does, `initial = v - rawS`, and the test now actually pins it.)
- **Hierarchy BFS test** likewise — the fixture authors each level's siblings smallest first, so the asserted drawn order can only come from the computed tree.
- **Diverging `cd.s`** now has a case with real negative calcdata sizes, differing from the authored ones.
- **Gantt** groups intervals by lane in one pass via a `Map`.
- **`sankeyEndpoint`** — kept the guard, widened the parameter to include `null`. It is not dead: `typeof null === 'object'`, so a hole in an authored `link.source` would otherwise be read as a node.

**Declined**

- **`isRidgeline` requiring more than one curve.** It counts *traces*, not curves — one violin trace draws a violin per category position (`ViolinEntry` is "a single position within a plotly violin trace"), so a `length > 1` check would misclassify a genuine single-trace ridgeline. Reading a lone halved violin as a box+KDE pair would also announce quartiles for an inner box the ridgeline recipe turns off, which is the reason a ridgeline is one layer in the first place.
- **PR description wording on stacked/normalized area.** Fair point that it is avoidance rather than subtraction; leaving the shipped description alone rather than editing history over it.

Full gate re-run locally: lint, type-check, build, 187 suites / 2717 tests, plus the 7-suite ESM project — all passing.

---

PR #875, opened 2026-08-14
feat(recharts): support the chart types added since the pie chart

### LINKED ISSUE #863: feat(recharts): support the chart types added since the pie chart
## Problem

MAIDR core has gained ~34 new trace types since the pie chart landed (radar, gauge, funnel, waterfall, gantt, treemap, sankey, sunburst, dumbbell, dot/lollipop, diverging, area family, and more). The `recharts` adapter has not kept up: of those, it currently emits **none**.

A chart Recharts draws perfectly well is therefore either dropped from the figure or announced as the wrong chart — which for a screen-reader user is worse than not being supported at all.

The adapter never introspects the Recharts component tree, its props, or the rendered SVG — the chart type is *declared by the consumer* as a string literal. `RechartsChartType` is a closed 8-member union (`/home/user/maidr/src/adapters/recharts/types.ts:25-33`: bar, stacked_bar, dodged_bar, normalized_bar, histogram, line, scatter, pie), set either top-level (`config.chartType`), per composed layer (`RechartsLayerConfig.chartType`, types.ts:39-46), or per subplot panel (types.ts:85). `toTraceType()` at `/home/user/maidr/src/adapters/recharts/converters.ts:570-589` is an exhaustive 1:1 switch from that string to a `TraceType`, called from `buildSimpleLayers` (converters.ts:227) and `buildComposedLayers` (converters.ts:411). Two branches bypass the map: a segmented bar type with a single `yKey` is downgraded to `TraceType.BAR` (converters.ts:225-227), and `'histogram'` is routed to `buildHistogramLayer`, which hard-codes `TraceType.HISTOGRAM` (converters.ts:329). The same string also selects the payload converter (`convertData`, converters.ts:434-456) and the highlight selector (`baseRechartsSelector`, `/home/user/maidr/src/adapters/recharts/selectors.ts:145-159). Adding a trace type therefore means four coordinated edits: a union member, a `toTraceType` case, a `convertData` case, and a `baseRechartsSelector` case.

## In scope — 24 trace types Recharts can render

| TraceType | Recharts construct | Detection / payload |
|---|---|---|
| `AREA` | <AreaChart> + <Area> (recharts/es6/cartesian/Area.js) | Add 'area' to RechartsChartType; treat it exactly as 'line' does today — isLineType() (converters.ts:563-565) gains the case so multi-series routes to buildMultiSeriesLineLayer and selectors are emitted as string[]. Payload is LinePoint[][] unchanged (AreaTrace extends LineTrace, src/model/area.ts, factory.ts:57-60), so convertToLinePoints (converters.ts:478) is reused verbatim. Selector: '.recharts-area-dots .recharts-area-dot' (verified class names in recharts/es6/cartesian/Area.js), requiring the user to leave <Area dot> on, same caveat the line path already has. |
| `STACKED_AREA` | <Area stackId="a"> inside <AreaChart> (Area.d.ts:140) | Add 'stacked_area' to the union; reuse the AREA path entirely — one layer, LinePoint[][] with one row per yKey and z = series name. Nothing extra to compute: src/model/area.ts states the running total is summed in the model rather than expected from the producer, so the adapter just declares the type. Same multi-series selector limitation as line (selectors.ts:46-60). |
| `NORMALIZED_AREA` | <AreaChart stackOffset="expand"> with stacked <Area>s ('expand' confirmed in recharts types/util/types.d.ts:32) | Add 'normalized_area'; identical to STACKED_AREA — same LinePoint[][] payload, only the emitted TraceType string differs (factory.ts:57-60 routes all three area variants to AreaTrace, which reads the variant off layer.type in src/model/area.ts:31-39). Pass raw per-series values, not the expanded fractions. |
| `RADAR` | <RadarChart> + <Radar> (recharts exports both; classes recharts-radar-polygon, recharts-radar-dots/recharts-radar-dot) | Add 'radar'; config maps xKey -> PolarAngleAxis dataKey (the spoke) and each yKey -> one <Radar>. Payload is LinePoint[][] — one row per series, one column per spoke — exactly the multi-line shape buildMultiSeriesLineLayer already produces (RadarTrace extends LineTrace, src/model/radar.ts:46). Selector '.recharts-radar-dots .recharts-radar-dot'. |
| `BUMP` | <LineChart> with <YAxis reversed /> plotting ranks (reversed confirmed in types/cartesian/YAxis.d.ts:181) | Add 'bump'; identical config surface to 'line' (xKey = period, one yKey per competitor holding the RANK not the value). Payload LinePoint[][] of ranks — BumpTrace extends LineTrace and derives best/worst rank and per-period moves itself (src/model/bump.ts:76-95), so the adapter adds no arithmetic. Only guard worth adding: reject non-integer/zero ranks in dev, since a value chart mis-declared as bump inverts the pitch. |
| `DOT` | Cleveland dot plot: <ScatterChart> with a category XAxis (or <Line dot strokeOpacity={0}>) | Add 'dot'; config is the bar config (xKey category + one yKey). Payload is BarPoint[] — factory.ts:66-69 routes BAR/DOT/LOLLIPOP to the same BarTrace — so convertToBarPoints (converters.ts:461) is reused with no change, and isBarType() (converters.ts:526) should include it so orientation is emitted. Selector '.recharts-scatter-symbol .recharts-symbols' (already in baseRechartsSelector for scatter). |
| `LOLLIPOP` | No primitive; the standard Recharts composition is <ComposedChart> with a thin <Bar barSize={2}> as the stem plus a <Scatter> for the head | Add 'lollipop'; same config and same BarPoint[] payload as DOT/BAR (factory.ts:66-69). Selector should target the head, '.recharts-scatter-symbol .recharts-symbols'; when the user draws the stem with a custom Bar shape instead they fall back to selectorOverride (types.ts:312-328). Adapter cost is one union member plus three switch cases. |
| `ALLUVIAL` | <Sankey data={{nodes, links}}> with the same node set repeated at each stage | Once SANKEY exists this is one extra union member and one toTraceType case: factory.ts:88-93 routes ALLUVIAL/CHORD/SANKEY to the same FlowTrace over the same FlowPoint[] {source,target,value} (grammar.ts:824-846). Same nodes-array-index -> node-name resolution, same '.recharts-sankey-links .recharts-sankey-link' selector. |
| `FUNNEL` | <FunnelChart> + <Funnel dataKey nameKey> (native; renders path.recharts-funnel-trapezoid inside g.recharts-trapezoids) | Add 'funnel'; config is xKey = stage label, single yKey = count, in draw order. Payload is BarPoint[] — FunnelTrace extends BarTrace and computes retention/share from the counts itself (src/model/funnel.ts:66-76) — so convertToBarPoints is reused. Selector '.recharts-trapezoids .recharts-funnel-trapezoid'. Must NOT set orientation (a funnel is not oriented), so keep it out of isBarType(). |
| `VOLCANO` | <ScatterChart> + <Scatter> (native; the chart is literally a scatter) | Add 'volcano' plus config keys for the point label and the two cutoffs. Payload VolcanoPoint[] extends ScatterPoint with the gene/feature label (grammar.ts:895), and MaidrLayer.thresholdOptions must be filled from config (fold-change and significance cutoffs) — the labels and thresholds are the whole point, and neither is inferable from a Recharts <Scatter>, so both become new RechartsAdapterConfig fields. Reuses convertToScatterPoints plus a label pick; selector '.recharts-scatter-symbol .recharts-symbols'. |
| `MANHATTAN` | <ScatterChart> + one <Scatter> per chromosome (or one series with a precomputed cumulative position) | Add 'manhattan'; same VolcanoPoint[] payload and thresholdOptions as VOLCANO (factory.ts:128-136 routes both to VolcanoTrace). The adapter needs a chromosome/label key and the significance cutoff from config. Extra care over VOLCANO: the drawn x is usually a cumulative genomic offset, so the payload should carry the per-chromosome position that gets announced, not the plotting offset. |
| `SURVIVAL` | <LineChart> + <Line type="stepAfter"> ('stepAfter' confirmed in types/shape/Curve.d.ts:11) | Add 'survival'; config gains a censoredKey and optional lower/upper band keys. Payload SurvivalPoint[][] — StepPoint (= LinePoint) plus censored?, yMin?, yMax? (grammar.ts:1034-1052) — built by extending convertToLinePoints with those three fields. Also set stepDirection: 'hv' on the layer (grammar.ts:1205-1212) since Recharts' stepAfter is exactly that convention. Selector '.recharts-line-dots .recharts-line-dot'. |
| `ERROR_BAR` | <ErrorBar dataKey direction> nested inside <Bar>/<Line>/<Scatter> (native component, types/cartesian/ErrorBar.d.ts:41,55) | Add 'error_bar' plus config keys for the interval. The conversion that matters: grammar.ts:370-390 fixes yMin/yMax as ABSOLUTE positions on the value axis, while Recharts' <ErrorBar dataKey> commonly points at an offset or a [lo,hi] pair — so the adapter must accept either (errorKey for symmetric offset, or yMinKey/yMaxKey) and normalise to absolutes. Payload ErrorBarPoint[] {x,y,yMin?,yMax?}, bounds independently optional. Selector: the whiskers are '.recharts-errorBars .recharts-errorBar', but the navigable unit is the estimate, so highlight the host mark instead. |
| `FOREST` | <ScatterChart layout="vertical"> + <ErrorBar direction="x"> + <ReferenceLine x={1}> (all native) | Builds on ERROR_BAR: same absolute-bounds conversion, plus ForestPoint's extra fields (grammar.ts:399+ — study weight and whether the row is the pooled summary) and MaidrLayer.forestOptions (grammar.ts:1194) carrying the null-line position and effect-measure name. All of that is author knowledge, not Recharts state, so it arrives as new config fields: weightKey, summaryKey/summaryIndex, nullValue. Row order is the drawn order with the pooled row last. |
| `DIVERGING` | Population pyramid / Likert: <BarChart layout="vertical" stackOffset="sign"> with one series negated ('sign' confirmed in types/util/types.d.ts:32) | Add 'diverging_bar'; config is the existing segmented-bar config (xKey + exactly two yKeys + fillKeys). Payload is SegmentedPoint[][] straight out of buildSegmentedBarLayer (converters.ts:258-296) with one change that decides correctness: values arrive SIGNED (left side negative), and DivergingTrace reads sides in DECLARATION order, not the reversed stacking order SegmentedTrace defaults to (src/model/diverging.ts:59-70) — so yKeys must be [left, right] and must not be reversed. Selector is the existing bar selector. |
| `WATERFALL` | No primitive; the standard Recharts recipe is a <BarChart> with two stacked <Bar>s sharing a stackId — a transparent base bar and a visible delta bar | Add 'waterfall' with a deltaKey and an optional kindKey/totalIndices. The adapter accumulates: walk the rows carrying a running total, emitting WaterfallPoint {x, start, end, delta, kind} (grammar.ts:460-478) with start/end as ABSOLUTE axis positions — grammar.ts:456-459 explicitly requires the producer to accumulate offsets. kind is 'increase'/'decrease' from sign, 'total' for rows the config marks as restatements. Selector is the existing bar-rectangle one, but it will match both the transparent base and the delta bar, so it must be narrowed (e.g. a className on the visible <Bar> via selectorOverride). |
| `DUMBBELL` | No primitive; drawn as <ScatterChart> with two <Scatter>s plus a connecting segment via Scatter's custom shape prop (types/cartesian/Scatter.d.ts:160-163), or as a transparent-base stacked bar | Add 'dumbbell' with startKey/endKey and the two end labels. Payload is DumbbellData — an OBJECT not an array: {points: [{x,start,end}], startLabel, endLabel} (grammar.ts:629-661). The labels are chart-level by design, so they become config fields (e.g. startLabel: '1990', endLabel: '2020'); omitting them makes the trace announce 'start'/'end', which is the one thing the legend gives a sighted reader. First adapter payload that is not an array — buildLayers' return typing needs to accommodate it. |
| `GANTT` | No primitive; the standard recipe is <BarChart layout="vertical"> with a transparent offset <Bar> and a visible duration <Bar> sharing a stackId | Add 'gantt' with laneKey/startKey/endKey and an optional interval label key. Payload is GanttData {points: GanttPoint[][], labels?} (grammar.ts:554-600) — nested BY LANE, not flat, precisely so an empty lane survives; so the adapter groups rows by lane and must take the lane list from config (not from the data) or empty lanes vanish. start/end are numeric axis positions, so dates need converting to epoch ms plus an axis format. |
| `GAUGE` | <RadialBarChart startAngle={180} endAngle={0}> with a single <RadialBar> — the canonical Recharts gauge/bullet recipe (classes recharts-radial-bar-sectors/recharts-radial-bar-sector) | Add 'gauge'; the config stops being data-array-shaped — GaugePoint is a SINGLE OBJECT {value, min, max, label?, target?, bands?} (grammar.ts:530-552). value comes from the one data row's yKey; min/max mirror the RadialBarChart domain; target and bands (GaugeBand {to,label}, upper edges only, ascending) are author knowledge and must be new config fields. Needs a bypass of the yKeys.map(...) layer loop in buildSimpleLayers (converters.ts:230-248), which assumes an array payload. |
| `TREEMAP` | <Treemap data dataKey nameKey> (native chart component; nodes rendered under g.recharts-treemap-depth-N) | Add 'treemap'; the config gains a hierarchical data mode, since Recharts' Treemap takes nested {name, children[]} rather than the flat rows the adapter assumes today. Payload TreemapPoint[] {x, y?, path?} (grammar.ts:793-812) — flatten the nested Recharts data depth-first, x = node name, y omitted for interior nodes whose value is the child sum, path = ancestor names root-first EXCLUDING the node. Highlighting is the weak spot: nodes are class-grouped by depth, not index-aligned in one flat list, so ship it with selectors omitted (the adapter already does this gracefully for multi-series) or via selectorOverride. |
| `SUNBURST` | <SunburstChart> — a real native component in recharts 3.x (exported from ./chart/SunburstChart) | Same TreemapPoint[] payload and same flattening as TREEMAP (factory.ts:166-172 routes ICICLE/SUNBURST/TREEMAP to one TreemapTrace; the sunburst only differs in angular panning, read off layer.type). Selector is better here than for treemap: SunburstChart emits one <Sector> per node in depth-first draw order under g.recharts-sunburst, so '.recharts-sunburst .recharts-sector' is index-aligned with a DFS flatten — provided the flatten skips the root, which SunburstChart does not draw (it renders data.children). |
| `ICICLE` | No native icicle; drawn as a horizontal stacked <BarChart layout="vertical">, one row per depth level, one stacked segment per node | Zero new payload work — identical TreemapPoint[] to TREEMAP/SUNBURST (factory.ts:166-172). Cost is entirely in the union member, the toTraceType case, and a selector: the bar-rectangle selector matches all depths in row-major order, which will not align with a depth-first node list, so either emit no selector or require selectorOverride. The drawing itself is a user-side construction, so this is worth doing only alongside TREEMAP. |
| `SANKEY` | <Sankey data={{nodes, links}} nodePadding nodeWidth> (native chart component) | Add 'sankey' with a distinct config shape: Recharts' links use integer indices into the nodes array ({source: 0, target: 2, value: 10}), while FlowPoint (grammar.ts:824-846) names both ends as strings and derives the node set from the edges — so the converter resolves each index through the nodes array to its name. Node order in MAIDR is first-appearance among the edges, so emit links in the order Recharts draws them. Selector '.recharts-sankey-links .recharts-sankey-link' is index-aligned with the links array. |
| `POLAR_AREA` | Coxcomb/rose via <Pie> with equal-angle slices and a per-datum radius — Pie.outerRadius accepts a function of the data point (types/polar/Pie.d.ts:44) and Pie takes a custom shape (Pie.d.ts:242). <RadialBarChart> is NOT this: it puts categories on rings and encodes value as angle. | Add 'polar_area'; payload is identical to RADAR — LinePoint[][], one row per series, one column per category (factory.ts:148-152 routes POLAR_AREA and RADAR to the same RadarTrace, which only swaps the spoken label, src/model/radar.ts:29). So the adapter work is one union member plus three switch cases once RADAR lands. Selector: '.recharts-pie-sector .recharts-sector' when drawn with <Pie>, i.e. the existing pie selector, subject to the same zero-value-slice caveat documented at selectors.ts:36-40. |

## Deferred to a follow-up

These need a new author-facing opt-in API or semantics Recharts does not carry in its own configuration, so they are out of scope here and tracked separately:

- `PARALLEL` — The obstacle is real but sits on the drawing side, not the payload side: a Recharts <Line> binds to a single yAxisId, so a polyline crossing axes with different units requires pre-normalising the plotted values. MAIDR is unaffected — ParallelTrace derives each column's own extent from the raw LinePo
- `RIDGELINE` — Payload is ViolinKdePoint[][] (src/model/ridgeline.ts:76) — one density curve per group — so the KDE must be computed outside Recharts and outside the adapter, and the config supplies precomputed {x, density} arrays per group. Critically, the drawn y includes the ridgeline offset while the payload m
- `HEXBIN` — Recharts only places the marks — the binning happens outside it, so the config supplies precomputed bins. Payload is HexbinPoint[][] {x, y, count} (grammar.ts:677-687) arranged as a LATTICE, not a flat list: rows of bins with the alternate-row stagger preserved, because vertical navigation picks the
- `BOXEN` — Payload BoxenPoint carries a variable-depth LetterValueLevel ladder (grammar.ts:288-331) which must be computed from the raw sample outside Recharts and passed in via config. Worth flagging as low priority: the adapter does not support plain BOX today either, so BOXEN would be the first distribution

## Not renderable by Recharts

<details><summary>6 trace types the library genuinely cannot draw</summary>

- CHOROPLETH — Recharts has no geographic projection, no map layer, and no component that accepts geometry/topojson; nothing in the 3.8.1 export list draws a region shape.
- CHORD — no circular arc/ribbon component. <Sankey> is a strictly left-to-right staged DAG layout and cannot close a cycle around a circle.
- NETWORK — no node-link or force-directed layout, and no data-bound edge primitive. Nodes could be Scatter points, but edges would have to be hand-placed <ReferenceLine segment> annotations, which are not bound to a data series.
- CONTOUR — cartesian series are single-valued in x (one y per x per series), whereas a closed contour ring has two; Recharts has no per-datum path/polygon mark to draw a ring with.
- MOSAIC — <Bar> derives its width from the band scale and Recharts has no variable-width bar. A custom Bar shape can repaint a rectangle but not re-place it, so tiles cannot tile contiguously — which drops the width dimension, i.e. half the contingency table.
- WORD_CLOUD — no text-layout or collision-packing mark. <Text> and <LabelList> attach to a series' computed point positions; there is no way to bind free-placed, weight-sized glyphs to data.

</details>

## Acceptance criteria

- [ ] Unit tests under `test/adapters/recharts/` cover every newly detected type.

## Implementation notes

Baseline: this adapter emits ZERO post-pie trace types. Its whole vocabulary is the 8-member `RechartsChartType` union, and `toTraceType` (converters.ts:570-589) is the only map. Tests confirm the closed set — every `expect(layer.type)` in /home/user/maidr/test/adapters/recharts/converters.test.ts asserts BAR, STACKED, DODGED, NORMALIZED, HISTOGRAM, LINE, SCATTER or PIE (lines 32, 94, 139, 157, 170, 185, 198, 219, 259, 301, 364, 391, 451-453) — and the 9 examples under /home/user/maidr/examples/recharts/examples/ cover only bar, stacked, dodged, normalized, histogram, line, scatter, composed and facets (there is not even a pie example, though pie is implemented).

Structural consequence for planning: because the type is DECLARED rather than DETECTED, "how the adapter would recognise it" is always the same answer — a new string literal plus new config fields. There is no Recharts spec to sniff, which cuts both ways: adding a type is cheap and mechanical, but anything the payload needs that Recharts does not itself hold (gauge bands and target, forest null line and study weights, dumbbell end labels, volcano/manhattan thresholds, waterfall step kinds, gantt lane list) has to become a new `RechartsAdapterConfig` field with docs, since it can never be inferred.

Four places to touch per new type, all in two files: `RechartsChartType` (types.ts:25-33), `toTraceType` (converters.ts:570-589), `convertData` (converters.ts:434-456), `baseRechartsSelector` (selectors.ts:145-159). Watch the three predicate helpers too — `isBarType` (converters.ts:526), `isSegmentedBarType` (554), `isLineType` (563) — they drive orientation emission and the `selectors: string[]` vs `selectors: string` split (converters.ts:240, 420); a line-family addition that misses `isLineType` will hand LineTrace a bare string and break highlighting silently.

Two payload shapes break the current architecture and should be scheduled together: GaugePoint (a single object) and DumbbellData/GanttData (objects wrapping arrays) do not fit `yKeys.map(...)` in buildSimpleLayers (converters.ts:230-248), which assumes one array payload per yKey. Everything else reuses an existing converter untouched: LinePoint[][] for AREA/STACKED_AREA/NORMALIZED_AREA/RADAR/POLAR_AREA/BUMP/PARALLEL/SURVIVAL, BarPoint[] for DOT/LOLLIPOP/FUNNEL, SegmentedPoint[][] for DIVERGING, ScatterPoint[]-derived for VOLCANO/MANHATTAN, TreemapPoint[] for TREEMAP/SUNBURST/ICICLE, FlowPoint[] for SANKEY/ALLUVIAL.

Highlighting is the recurring trap, and the adapter already has an honest convention for it: `getRechartsSelector` returns `undefined` for multi-series (selectors.ts:116-120) and the traces degrade to audio/text/braille rather than mis-align. Several new types (treemap, icicle, ridgeline, parallel, waterfall's transparent base bar) cannot produce an index-aligned element list from CSS alone, so they should ship with selectors omitted plus a documented `selectorOverride` recipe (types.ts:312-328) rather than a guess. Note the existing zero-value-slice caveat at selectors.ts:36-40 applies to any new <Pie>-drawn type (POLAR_AREA).

Subplot mode is free for all of the above: `buildPanelSubplot` (converters.ts:132-172) merges panel config over top-level defaults and re-scopes selectors, so any new chartType works inside a facet grid the moment it works standalone — but panel-scoping assumes the panel's chart renders `svg.recharts-surface` (converters.ts:170), which SunburstChart/Treemap/Sankey should be spot-checked against since they build their own <Surface>.

Recharts version note: peerDependency is `^2.0.0 || ^3.0.0` (package.json:121) while devDependency is `^3.8.1` (package.json:193). `SunburstChart` and `Pie.shape` are v3-era APIs, so anything relying on them needs either a version guard or a documented v3 minimum. All library API claims above were verified against /home/user/maidr/node_modules/recharts/types/ and es6/ at 3.8.1, not from memory.

jooyoungseo:
## Description

The Recharts adapter has recognised eight chart types since it was written — bar, stacked bar, dodged bar, normalized bar, histogram, line, scatter and pie. MAIDR's grammar has grown a long way past that, and every trace type this pull request adds was already fully implemented in the core: the models, the sonification, the text and braille output all exist and are unchanged here. What was missing was the declaration that lets a Recharts author reach them.

This extends `RechartsChartType` with **24 further types** and wires each one end to end: payload conversion into the exact shape `src/type/grammar.ts` specifies, a highlight selector built from Recharts' own class names, a runnable example, and documentation. No file outside `src/adapters/recharts/`, `test/adapters/recharts/`, `examples/` and `docs/` is touched.

Accessibility is the acceptance bar, not sonification alone. `test/adapters/recharts/markDom.test.ts` is new for this reason: it mounts the real Recharts components in jsdom and asserts that each type's selector resolves to exactly one mark per data point — the funnel trapezoids, the whisker groups, the scatter symbols, the line dots, the sankey ribbons, the floating bar rectangles, the radial sectors. A type that sonifies but does not highlight would fail that test rather than ship.

## Related Issues

Closes #863

### Trace types added (24)

Grouped by the converter they reuse or introduce.

**Reuse the bar converter (`BarPoint[]`)**

- `'dot'` — Cleveland dot plot, a `<Scatter>` against a category axis
- `'lollipop'` — a `<ComposedChart>` of a thin `<Bar>` stem plus a `<Scatter>` head; the stem is the mark, not extra data
- `'funnel'` — `<FunnelChart>` + `<Funnel>`. The counts are passed through untouched and `FunnelTrace` derives the retention it announces, because retention is the number a funnel is read for and the one a listener cannot take by ear from two heights heard separately

**Join the line family (`LinePoint[][]`)**

- `'area'` — the fill is what the mark looks like, not a second magnitude, so the data is a line's
- `'stacked_area'` — `<Area stackId="...">`. Declared over a single `yKey` it falls back to `AREA`, mirroring how the bar family downgrades to `BAR`: one band is not stacked against anything, and announcing a total equal to the point's own value on every sample is noise
- `'normalized_area'` — `<AreaChart stackOffset="expand">`
- `'radar'` — `<RadarChart>` + `<Radar>`
- `'polar_area'` — a coxcomb/rose: the radar's spokes drawn as `<Pie>` wedges with equal angles and a per-datum `outerRadius`, so it carries the radar payload and the pie's sectors
- `'bump'` — rank over time, a `<LineChart>` with `<YAxis reversed>`

**Carry payload a coordinate pair cannot, through a new sub-config**

- `'survival'` — Kaplan-Meier, a `<Line type="stepAfter">` plus censoring marks and confidence band (`survivalConfig`)
- `'volcano'` and `'manhattan'` — effect size or genomic position against significance; point identity, region and cutoffs (`volcanoConfig`)
- `'error_bar'` — `<ErrorBar>` inside a `<Bar>`/`<Line>`/`<Scatter>` (`errorConfig`)
- `'forest'` — one interval per study against a shared null line, pooled summary last (`forestConfig`)
- `'alluvial'` and `'sankey'` — weighted flow through a `<Sankey>` (`flowConfig`). The two share payload, config and selector; they differ in whether the node set repeats at each stage, which the data says rather than the adapter
- `'diverging_bar'` — two `yKeys` back to back with `<BarChart stackOffset="sign">`
- `'waterfall'` — signed contributions carried to a running total (`waterfallConfig`)
- `'gantt'` — intervals per lane (`ganttConfig`)
- `'gauge'` — one measure against a range as a half-dial `<RadialBarChart>` (`gaugeConfig`)
- `'dumbbell'` — two values per category joined by a segment
- `'treemap'` and `'icicle'` — a hierarchy as nested area or as spans
- `'sunburst'` — the same hierarchy as rings

### Conversions that decide correctness

- **Error bar bounds.** Recharts' `<ErrorBar dataKey>` points at an *offset* from the estimate while `ErrorBarPoint` fixes *absolute* positions on the value axis. `errorKey` is resolved against `y` — a number or a `[lower, upper]` pair, the same arithmetic `ErrorBar.js` does — while `yMinKey`/`yMaxKey` pass through untouched. The two bounds stay independently optional: a one-sided interval is a real chart.
- **Waterfall totals.** The data holds contributions; `WaterfallPoint` holds absolute start and end positions. The running totals are accumulated in the converter, with a restating row (opening balance, subtotal, closing balance) sitting on the baseline and falling back to the accumulated total when it carries no number of its own.
- **Hierarchy flattening.** Treemap, sunburst and icicle data is flattened depth-first in pre-order so row *i* is node *i*, dropping an interior node's declared value because Recharts computes one from the children and ignores it. Order and magnitudes then match what the chart actually drew.
- **Non-uniform payloads.** `GaugePoint` is a single object, `GanttData` nests its intervals by lane, and `DumbbellData` wraps its rows with the names of the two ends — none of them one array per `yKey`. Those three get dedicated builders alongside histogram and survival rather than passing through the `yKeys.map()` loop.

### Highlighting

Every added type gets a selector derived from Recharts' own class names, and two need more than a class:

- the treemap selector skips the synthetic root rectangle Recharts wraps the data array in, which would otherwise be one element more than there are nodes and turn highlighting off entirely;
- the gantt withdraws its own selector when the rows interleave lanes, because Recharts draws in row order while `GanttTrace` walks lane by lane, and a wrong highlight is worse than none.

Sunburst excludes the root because `SunburstChart` renders `data.children`.

### Reuse over new configuration

Where an existing idiom already said the thing, it is reused rather than duplicated: a dumbbell's and a gantt's two positions are its two `yKeys`, and a dumbbell's end labels are its `fillKeys` — the names a legend gives a sighted reader for free. Only what Recharts genuinely cannot know becomes a new sub-config: `waterfallConfig` marks the rows that restate the running total, `ganttConfig` declares the lanes so an empty one survives, and `gaugeConfig` carries the range, target and bands a `<RadialBar>` holds nowhere.

### Examples and docs

24 new example components under `examples/recharts/examples/`, one per added type, registered in `examples/recharts/App.tsx`. `docs/recharts.md` grows by roughly 850 lines: a section per type with the data shape, the config, the Recharts markup that renders the marks the default selector expects, and the recipes that do not (the transparent-offset waterfall needs a `selectorOverride`, and a chart of values declared as `'bump'` sonifies upside down — the adapter cannot tell rank from value).

**No Recharts primitive draws them.** The library has no component for these marks, so there is nothing for a selector to target and no data an adapter could read off the chart:

- `box`, `boxen`, `violin_box`, `violin_kde` — no box or violin primitive
- `heat`, `hexbin`, `mosaic` — no cell-grid primitive
- `contour`, `choropleth` — no scalar-field or geographic primitive
- `chord`, `network` — no graph-layout primitive
- `parallel_coordinates`, `ridgeline` — no multi-axis or offset-density primitive
- `word_cloud` — no text-layout primitive
- `candlestick` — no OHLC primitive

**Not an adapter concern.** `candlestick_delta` is never declared in MAIDR JSON; the core creates it at runtime from a candlestick layer.

**Drawable in Recharts, not declared here.** `step` (`<Line type="stepAfter">`) and `smooth` (`<Line type="monotone">`) are `<Line>` curve variants Recharts renders today. This change does not add declarations for them, so an author still reaches them as `'line'` and loses the announcement that says which curve it is. They are the obvious next batch; no core change is needed for either.

No core change was required for anything in this pull request, and none was made.

## Checklist

Two boxes are left unticked deliberately. `CONTRIBUTING.md` was not read as part of this work, and the manual pass in `ManualTestingProcess.md` — a screen reader driven through each new chart in a browser — was not performed. The automated gate below did run in full and is green; it is not a substitute for that manual pass, and the examples are there so a reviewer can do it.

## Additional Notes

Verification run on this branch, in order, from a clean tree:

| Step | Result |
| --- | --- |
| `npm run lint:fix` | clean, no files rewritten |
| `npm run type-check` | clean (`tsc --noEmit` and `tsc -p tsconfig.ci.json`) |
| `npm run build` | all 12 adapter bundles built |
| `npm test` | 188/188 suites, 2731/2731 tests (unit + component), 7/7 suites and 73/73 tests (ESM) |

Nothing failed, so no fix commits were needed on top of the four feature commits. `npm run e2e` was not run — it is outside the gate for this change and needs browsers this environment does not have.

Review suggestion: the two conversions worth checking against the Recharts source are the `<ErrorBar>` offset-to-absolute arithmetic in `src/adapters/recharts/converters.ts` and the waterfall total accumulation, since both restate numbers the chart holds only implicitly. `test/adapters/recharts/markDom.test.ts` is the file to read for the highlighting claims — it is the proof, not a description of one.

---

### DIFF
--- a/docs/recharts.md
+++ b/docs/recharts.md
@@ -137,6 +137,11 @@ Use `subplots` when your figure is a grid of small multiples (faceted charts). S
 | `'stacked_bar'` | `<Bar stackId="...">` | Stacked bar chart (multiple `yKeys`) |
 | `'dodged_bar'` | Multiple `<Bar>` | Grouped/side-by-side bar chart (multiple `yKeys`) |
 | `'normalized_bar'` | `<Bar stackId="...">` | 100% stacked bar chart (multiple `yKeys`) |
+| `'diverging_bar'` | `<BarChart stackOffset="sign">` | Population pyramid / Likert: two sides of a shared baseline (exactly two `yKeys`) |
+| `'waterfall'` | `<Bar>` with a `[start, end]` `dataKey` | A total carried through signed contributions (`waterfallConfig`) |
+| `'dumbbell'` | `<Bar>` with a `[start, end]` `dataKey` | Two values compared per category (exactly two `yKeys`) |
+| `'gantt'` | `<BarChart layout="vertical">` + floating `<Bar>` | Intervals laid out in lanes (`ganttConfig`) |
+| `'gauge'` | `<RadialBarChart>` + `<RadialBar>` | One measure against a range (`gaugeConfig`) |
 | `'dot'` | `<Scatter>` | Cleveland dot plot: one point per category |
 | `'lollipop'` | `<Bar barSize={2}>` + `<Scatter>` | Lollipop chart: a stem and a head |
 | `'funnel'` | `<FunnelChart>` + `<Funnel>` | Funnel chart: a population shrinking across ordered stages |
@@ -155,6 +160,8 @@ Use `subplots` when your figure is a grid of small multiples (faceted charts). S
 | `'forest'` | `<ScatterChart layout="vertical">` + `<ErrorBar direction="x">` | Forest plot (`errorConfig` + `forestConfig`) |
 | `'pie'` | `<Pie>` | Pie chart (a doughnut is a `<Pie>` with an `innerRadius`) |
 | `'alluvial'` | `<Sankey>` | Weighted flow between nodes (`flowConfig`) |
+| `'treemap'` | `<Treemap>` | A hierarchy laid out as nested area (nested `data`) |
+| `'sunburst'` | `<SunburstChart>` | The same hierarchy drawn as rings (nested `data`) |
 
 ## Data Examples by Chart Type
 
@@ -777,6 +784,234 @@ const links = [
 
 MAIDR reads this as a graph rather than a grid: following a ribbon is the primary move, and arrow keys step between a node and the flows that leave or arrive at it. Highlighting pairs each flow with the `<path class="recharts-sankey-link">` at the same position, so the `links` array order is the order the ribbons are announced in.
 
+### Diverging Bar Chart (Population Pyramid)
+
+Two series drawn back to back across a shared category axis. Declare **exactly two `yKeys`, the left-hand side first**, and give the left side **negative** values — the same numbers `<BarChart stackOffset="sign">` needs:
+
+```tsx
+const data = [
+  { band: '0-9', men: -2_100_000, women: 2_000_000 },
+  { band: '10-19', men: -1_900_000, women: 1_850_000 },
+];
+
+<MaidrRecharts
+  id="pyramid-example"
+  title="Population by Age Band"
+  data={data}
+  chartType="diverging_bar"
+  xKey="band"
+  yKeys={['men', 'women']}
+  fillKeys={['Men', 'Women']}
+  orientation={Orientation.HORIZONTAL}
+  xLabel="Age band"
+  yLabel="People"
+>
+  <BarChart width={600} height={400} layout="vertical" stackOffset="sign" data={data}>
+    <XAxis type="number" />
+    <YAxis type="category" dataKey="band" />
+    <Bar dataKey="men" stackId="sides" fill="#8884d8" />
+    <Bar dataKey="women" stackId="sides" fill="#82ca9d" />
+  </BarChart>
+</MaidrRecharts>
+```
+
+The sign is read as the **side**, not the magnitude: MAIDR pitches the size of the bar and announces which way it points, so the biggest bar on the left is not heard as the lowest note on the chart. It also announces the **balance** — what one side has over the other — at every band, which is the subtraction a pyramid is drawn back to back to support.
+
+Order matters twice. The sides are read in **declaration order**, so the `yKeys` must run left then right, and the `<Bar>` elements must be declared in that same order for highlighting to land on the right one. Declared with a single `yKey` the layer falls back to a plain bar chart, since a pyramid with one side is not one.
+
+### Waterfall Chart
+
+The single `yKeys` entry names each step's **contribution**; the adapter accumulates the running totals, because a waterfall bar floats between the total before the step and the total after it and neither number is in the data.
+
+```tsx
+const data = [
+  { step: 'Opening', change: 1200, restates: true },
+  { step: 'New sales', change: 450 },
+  { step: 'Churn', change: -180 },
+  { step: 'Upsell', change: 90 },
+  { step: 'Closing', restates: true },
+];
+
+<MaidrRecharts
+  id="waterfall-example"
+  title="Revenue Bridge"
+  data={data}
+  chartType="waterfall"
+  xKey="step"
+  yKeys={['change']}
+  waterfallConfig={{ totalKey: 'restates' }}
+  xLabel="Step"
+  yLabel="Revenue ($k)"
+>
+  <BarChart width={600} height={350} data={steps}>
+    <XAxis dataKey="x" />
+    <YAxis />
+    <Bar dataKey={(step) => [step.start, step.end]} fill="#8884d8" />
+  </BarChart>
+</MaidrRecharts>
+```
+
+A row flagged by `totalKey` (or listed in `totalIndices`, for data with no flag column) **restates** the running total rather than changing it — an opening balance, a subtotal, a closing balance. It sits on the baseline, and it need carry no number of its own: the `Closing` row above restates whatever the steps came to. `kindKey` names the kind outright when the data already holds it.
+
+Draw the chart from a **floating bar** — a `<Bar>` whose `dataKey` returns a `[start, end]` pair — and it renders exactly one rectangle per step, which is what the default selector targets. Those are the same totals the adapter computes, so read them back off the layer instead of accumulating twice:
+
+```tsx
+const maidrData = useRechartsAdapter(config);
+const steps = maidrData.subplots[0][0].layers[0].data as WaterfallPoint[];
+```
+
+The other recipe — a transparent offset `<Bar>` stacked under a visible one — draws **two** rectangles per step, and the default selector matches both. Give the visible bar a `className` and pass the narrowed selector as `selectorOverride`:
+
+```tsx
+<Bar dataKey="base" stackId="w" fill="transparent" />
+<Bar dataKey="change" stackId="w" className="wf-delta" />
+// selectorOverride: '.wf-delta .recharts-bar-rectangle .recharts-rectangle'
+```
+
+### Dumbbell Chart
+
+Two values compared at each category, joined by a segment. Two `yKeys`, the starting end first, and `fillKeys` names them:
+
+```tsx
+const data = [
+  { country: 'Japan', then: 78.9, now: 84.6 },
+  { country: 'Russia', then: 69.2, now: 68.9 },
+];
+
+<MaidrRecharts
+  id="dumbbell-example"
+  title="Life Expectancy, 1990 against 2020"
+  data={data}
+  chartType="dumbbell"
+  xKey="country"
+  yKeys={['then', 'now']}
+  fillKeys={['1990', '2020']}
+  xLabel="Country"
+  yLabel="Years"
+>
+  <BarChart width={600} height={350} data={data}>
+    <XAxis dataKey="country" />
+    <YAxis />
+    <Bar dataKey={(row) => [row.then, row.now]} fill="#8884d8" />
+  </BarChart>
+</MaidrRecharts>
+```
+
+The two names are the content of the comparison. Announced as "start" and "end", a chart of life expectancy in 1990 against 2020 tells a reader which dot they are on and not which year it is — which is the one thing the legend gives a sighted reader for free. MAIDR derives the change rather than taking it from the data, so a row that declined (Russia, above) is announced as a decrease without anything extra declared.
+
+Highlighting pairs one element with each **row**, not with each dot: a chart draws one connector per row, so both ends highlight the same element. A dumbbell drawn as a `<ScatterChart>` with two `<Scatter>`s draws two symbols per row instead and needs a `selectorOverride` naming the connector shape.
+
+### Gantt Chart
+
+One data row is one interval: `xKey` names its lane and the two `yKeys` its start and end, both as positions on the same numeric axis.
+
+```tsx
+const data = [
+  { task: 'Design', from: 0, to: 5, phase: 'Wireframes' },
+  { task: 'Build', from: 3, to: 12, phase: 'Alpha' },
+  { task: 'Build', from: 14, to: 18, phase: 'Beta' },
+];
+
+<MaidrRecharts
+  id="gantt-example"
+  title="Release Plan"
+  data={data}
+  chartType="gantt"
+  xKey="task"
+  yKeys={['from', 'to']}
+  ganttConfig={{ lanes: ['Design', 'Build', 'Launch'], labelKey: 'phase', unit: 'days' }}
+  xLabel="Task"
+  yLabel="Day"
+>
+  <BarChart width={600} height={350} layout="vertical" data={data}>
+    <XAxis type="number" />
+    <YAxis type="category" dataKey="task" />
+    <Bar dataKey={(row) => [row.from, row.to]} fill="#8884d8" />
+  </BarChart>
+</MaidrRecharts>
+```
+
+Declare `lanes` so that a lane with **nothing booked still exists** — `Launch` above holds no row, and nothing booked is a real statement about a schedule that a list grouped by task cannot make. A lane a row names but the list omits is appended rather than dropped. `unit` names what a unit of the axis is; without it the length of an interval is announced as a bare number.
+
+Dates have to arrive as **epoch milliseconds**: a `Date` is not a position on a numeric axis.
+
+MAIDR puts the interval's length on pitch and its start on stereo position, so two lanes whose work lines up sound like they line up. That is the overlap question answered by ear, and it is the reason a gantt gets a trace type rather than being read as a table.
+
+Highlighting needs the rows **grouped by lane, in the lane order** — Recharts draws one rectangle per row in row order, while the payload is walked lane by lane. The adapter checks this and turns highlighting off for the layer when the rows interleave lanes, rather than highlighting somebody else's task.
+
+### Gauge Chart
+
+One measure read against a range. The data holds a single row; everything else the reading needs is author knowledge and arrives through `gaugeConfig`, which is therefore required:
+
+```tsx
+const data = [{ measure: 'NPS', score: 73 }];
+
+<MaidrRecharts
+  id="gauge-example"
+  title="Net Promoter Score"
+  data={data}
+  chartType="gauge"
+  xKey="measure"
+  yKeys={['score']}
+  gaugeConfig={{
+    min: 0,
+    max: 100,
+    target: 80,
+    bands: [{ to: 40, label: 'poor' }, { to: 70, label: 'ok' }, { to: 100, label: 'good' }],
+  }}
+>
+  <RadialBarChart
+    width={400} height={250}
+    innerRadius={80} outerRadius={140}
+    startAngle={180} endAngle={0}
+    data={data}
+  >
+    <PolarAngleAxis type="number" domain={[0, 100]} tick={false} />
+    <RadialBar dataKey="score" fill="#8884d8" background />
+  </RadialBarChart>
+</MaidrRecharts>
+```
+
+"73" is not the reading. "73 out of 100, 7 below target, in the 'ok' band" is, and a sighted reader gets all of it from the dial's geometry — the needle's position, the marker beside it, the coloured arc it lands on. None of that is written anywhere a screen reader can reach, and none of it is inferable from a `<RadialBar>`, so the range mirrors the chart's `domain` and the target and bands are declared.
+
+Bands are **ascending and bounded above only**: a band starts where the previous one ended, and the first starts at `min`. A value above every band belongs to none rather than to the last one. There is deliberately no default range — a guessed maximum misreports the one number the chart draws.
+
+### Treemap and Sunburst
+
+Both draw the same hierarchy, and both take the **nested** `{ name, children }` data Recharts itself is given rather than the adapter's usual flat rows. `xKey` is the `<Treemap nameKey>` and the single `yKeys` entry its `dataKey`:
+
+```tsx
+const nodes = [
+  { name: 'Europe', children: [{ name: 'France', people: 67.4 }, { name: 'Spain', people: 47.4 }] },
+  { name: 'Asia', children: [{ name: 'Japan', people: 125.1 }, { name: 'Nepal', people: 30.0 }] },
+];
+
+<MaidrRecharts
+  id="treemap-example"
+  title="Population by Region"
+  data={nodes}
+  chartType="treemap"
+  xKey="name"
+  yKeys={['people']}
+>
+  <Treemap width={600} height={350} data={nodes} dataKey="people" nameKey="name" />
+</MaidrRecharts>
+```
+
+A sunburst is the same config with `chartType="sunburst"`. Its Recharts component takes a single root object, and it draws that root's **children** — so pass the children here, and the declared nodes are then exactly the drawn ones:
+
+```tsx
+<MaidrRecharts id="sunburst-example" data={nodes} chartType="sunburst" xKey="name" yKeys={['people']}>
+  <SunburstChart width={400} height={400} data={{ name: 'World', children: nodes }} dataKey="people" />
+</MaidrRecharts>
+```
+
+MAIDR navigates the tree **as a tree**, on arrow keys that already exist: up returns to the parent, down enters the first child, left and right walk the siblings. Each node is announced with its exact share of its parent — the comparison the rectangles are drawn to support and the one the eye estimates worst.
+
+A node with children gets no value of its own, because Recharts computes an interior node's value as the sum of its children and ignores any value declared on it. MAIDR derives interior totals the same way.
+
+Highlighting relies on both components drawing their nodes in depth-first pre-order, which they do. The treemap selector deliberately skips the **synthetic root** rectangle Recharts wraps the data array in — it is one element more than there are nodes, and a count mismatch turns highlighting off entirely. A `<Treemap type="nest">`, or one given a custom `content`, draws something else and needs `selectorOverride`.
+
 ### Composed Chart (Bar + Line)
 
 Use `layers` mode to mix different chart types in a single chart:
@@ -905,6 +1140,11 @@ type RechartsChartType =
   | 'stacked_bar'
   | 'dodged_bar'
   | 'normalized_bar'
+  | 'diverging_bar'
+  | 'waterfall'
+  | 'dumbbell'
+  | 'gantt'
+  | 'gauge'
   | 'dot'
   | 'lollipop'
   | 'funnel'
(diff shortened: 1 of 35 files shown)


jooyoungseo:
Review round 2: all 14 checks are green at `b910fe9` and there are no unresolved review threads. Nothing to fix, so no new commit — but the one substantive point from the automated review deserves an answer rather than silence.

**Declined: "add a Playwright spec for each of the 24 new types" (`.claude/rules/testing.md`).**

The rule says "add a spec for each new **trace type**", and this PR adds no trace type. `TraceType.GANTT`, `WATERFALL`, `DUMBBELL`, `TREEMAP`, `SANKEY`, `FOREST` and the rest already exist in the model, and each already has its spec — `e2e_tests/specs/` carries `gantt.spec.ts`, `waterfall.spec.ts`, `dumbbell.spec.ts`, `treemap.spec.ts`, `sankey.spec.ts`, `forest.spec.ts`, `survival.spec.ts`, `volcano.spec.ts`, `errorbar.spec.ts`, `gauge.spec.ts`, `funnel.spec.ts`, `area.spec.ts`, `radar.spec.ts`, `bump.spec.ts`, `dot.spec.ts`, `lollipop.spec.ts` and `hierarchyLayouts.spec.ts`. What this PR adds is the Recharts *mapping* onto those existing types.

The suite also could not host such a spec as it stands. Every page object navigates to a static `examples/*.html` page built from the core library — `grep -ri recharts e2e_tests/` returns nothing, and no adapter, Recharts or Victory, has ever had E2E coverage. Adding one here would mean standing up a React example harness for the whole E2E suite: a larger change than this PR, and one that belongs to the suite rather than to the adapter.

What the adapter can be held to is the DOM contract, and that is where `markDom.test.ts` sits: it mounts real Recharts components in jsdom and asserts exact per-type mark counts and ordering — one trapezoid per funnel stage, one rectangle per gantt interval in lane order, treemap nodes excluding the synthetic root, sankey ribbons in declared-link order. That is what would actually catch a broken selector, and it is checked in.

Worth filing the React E2E harness as a follow-up issue if the team wants adapter-level keyboard coverage; it is not this PR's to carry.

---

PR #876, opened 2026-08-14
feat(google-charts): support the chart types added since the pie chart

### LINKED ISSUE #867: feat(google-charts): support the chart types added since the pie chart
## Problem

MAIDR core has gained ~34 new trace types since the pie chart landed (radar, gauge, funnel, waterfall, gantt, treemap, sankey, sunburst, dumbbell, dot/lollipop, diverging, area family, and more). The `google-charts` adapter has not kept up: of those, it currently emits **none**.

A chart Google Charts draws perfectly well is therefore either dropped from the figure or announced as the wrong chart — which for a screen-reader user is worse than not being supported at all.

Dispatch is a single explicit `switch` on a caller-supplied string, not on anything read off the chart instance: `buildLayer` at /home/user/maidr/src/adapters/google-charts/converters.ts:447 switches over `chartType` at converters.ts:453-484, with `default` throwing an "Unsupported Google Charts type" error (converters.ts:476-483). The string comes from the required `GoogleChartAdapterOptions.chartType` field (converters.ts:88, documented converters.ts:80-87 as necessary "because the chart instance does not expose its own type"), constrained by the `GoogleChartType` union in /home/user/maidr/src/adapters/google-charts/types.ts:128-139. That union mixes real Google class names ('BarChart', 'PieChart') with adapter-invented pseudo-types that Google has no class for ('StackedColumnChart', 'DodgedBarChart') — the established convention for "same Google class, different draw options, different MAIDR reading". Exactly one branch is data-driven: `buildBarOrSegmentedLayer` (converters.ts:495-506) calls `countDataColumns` (converters.ts:912) and promotes a ColumnChart/BarChart from `TraceType.BAR` to `TraceType.DODGED` when more than one non-role data column exists. Everything else — stacking, normalization, orientation, doughnut-vs-pie — is decided by the string alone, because the adapter is handed only `(chart, dataTable, container, {id, title, chartType})` (converters.ts:130-135) and never sees the chart's draw options.

## In scope — 20 trace types Google Charts can render

| TraceType | Google Charts construct | Detection / payload |
|---|---|---|
| `AREA` | google.visualization.AreaChart (corechart package); SteppedAreaChart for the stepped variant | New `chartType: 'AreaChart'` case in the converters.ts:453 switch. The DataTable is identical to a LineChart's (domain col 0, one non-role column per series), so `buildLineLayer` (converters.ts:605) can be reused verbatim with `type: TraceType.AREA` — AreaTrace reads `LinePoint[][]` exactly as LineTrace does (src/model/factory.ts:57-62, src/model/area.ts). Highlighting: Google draws an area as a filled `<path>` plus an optional line path; `markLinePointElements` (converters.ts:1243) currently filters on `path[fill="none"]`, which will select the outline but skip the fill, so the filter needs widening or the fill path skipping. |
| `STACKED_AREA` | AreaChart drawn with `isStacked: true` | Pseudo-type `'StackedAreaChart'` added to GoogleChartType (types.ts:128) — same convention as the existing 'StackedColumnChart'. Same `LinePoint[][]` payload as AREA; AreaTrace derives the running total itself from `layer.type` (src/model/area.ts variantOf). The adapter cannot infer `isStacked` because it never receives draw options, so the pseudo-type is the only honest detection. |
| `NORMALIZED_AREA` | AreaChart drawn with `isStacked: 'percent'` | Pseudo-type `'NormalizedAreaChart'`, identical build to STACKED_AREA with `type: TraceType.NORMALIZED_AREA`. Payload `LinePoint[][]` of the raw per-series values (not the percentages) — AreaTrace scales to the common whole itself. |
| `ERROR_BAR` | Intervals — `role: 'interval'` columns on LineChart / ScatterChart / ColumnChart / ComboChart, with `intervals: {style: 'bars'\|'sticks'\|'boxes'\|'points'\|'area'}` | The detection plumbing already exists and is currently used only to *discard* these columns: `isRoleColumn` (converters.ts:884) reads `dt.getColumnRole(c)`, declared optional at types.ts:21. Detect `getColumnRole(c) === 'interval'` and pair the interval columns that follow a data column. Payload: `ErrorBarPoint[]` — `x` from col 0 via `formatCellValue`, `y` from the data column, `yMin`/`yMax` from the interval pair. Google's interval values are absolute positions on the value axis, which matches the schema's fixed contract (grammar.ts:380-397), so no offset conversion is needed. Traps: a chart may declare several interval pairs (two CI levels) — take the outermost or the first; and `getColumnRole` is optional on the fake/DataView path, where `isRoleColumn` returns false and the intervals would be misread as extra data series (and so as DODGED via converters.ts:502). |
| `SANKEY` | google.visualization.Sankey (`sankey` package) | New `chartType: 'Sankey'` case. DataTable is fixed: col 0 = From (string), col 1 = To (string), col 2 = Weight (number). Payload is a flat `FlowPoint[]` of `{source, target, value}` (grammar.ts:824-846) — nodes are derived from the edges by FlowTrace (src/model/flow.ts:156), so no node list is built. Traps: Sankey exposes no `getChartLayoutInterface()`, so the existing bbox-matching helpers do not apply; links are `<path>`s and nodes `<rect>`s in the SVG with no data attributes, so either mark links by DOM order (Google emits them in DataTable row order) or follow the pie precedent and drop `selectors` when the count does not match (converters.ts:762-769). |
| `TREEMAP` | google.visualization.TreeMap (`treemap` package) | New `chartType: 'TreeMap'` case. DataTable is fixed: col 0 = node id/label, col 1 = parent id (null on the single root row), col 2 = size, optional col 3 = color. Payload `TreemapPoint[]` (grammar.ts:793-810) needs a *path* rather than a parent pointer, so build a `Map<id, parentId>` from the table and walk it upward per row to produce `path` (root first, excluding the node), with `x` = id and `y` = the size column. Interior nodes may be emitted or left to be derived. Traps: TreeMap is a drill-down chart that renders only `maxDepth`(+`maxPostDepth`) levels at a time and redraws on click, so DOM marking is unstable — dropping `selectors` is the safe first cut; also confirm whether the rendered cells are SVG rects or HTML elements before writing a selector. |
| `GANTT` | google.charts.Gantt (`gantt` package) and google.visualization.Timeline (`timeline` package) | Two separate cases ('Gantt', 'Timeline') feeding one builder. Gantt columns: Task ID, Task Name, Resource, Start Date, End Date, Duration, Percent Complete, Dependencies; Timeline columns: Row label, [bar label], [tooltip], Start, End. Payload is the nested `GanttData` object (grammar.ts:554-610): group rows by lane (Task Name / Resource for Gantt, Row label for Timeline) into `points: GanttPoint[][]`, set `lanes` from the lane names, and `label` from the bar/task name when the lane holds several intervals. Traps: `GanttPoint.start`/`end` are plain numbers, so Dates must be converted — converting to ms epoch and leaving `unit` unset makes every announced length an unreadable eight-digit number, so convert to days (or hours) and set `unit` to match; and Gantt draws dependency arrows as extra paths that must not be mistaken for bars. |
| `GAUGE` | google.visualization.Gauge (`gauge` package) | New `chartType: 'Gauge'` case. The DataTable is [Label, Value] rows. Payload `GaugePoint` is a single OBJECT, not an array (grammar.ts:530-552, consumed at src/model/gauge.ts:70), so a Gauge DataTable with N rows must become N layers (or N subplots) — one per dial. The blocker is that `min`/`max` (Google defaults 0/100) and the qualitative bands live in the draw options (`greenFrom/greenTo`, `yellowFrom/yellowTo`, `redFrom/redTo`) which the adapter never receives, so `GoogleChartAdapterOptions` needs a new field carrying them, mapped to ascending `GaugeBand[]` (`{to, label}`). Without it the reading loses the bands, which is most of what a gauge means. |
| `CHOROPLETH` | google.visualization.GeoChart (`geochart` package) in its default `displayMode: 'regions'` | New `chartType: 'GeoChart'` case. Regions mode: col 0 = region name or ISO code, col 1 = the shaded value → `ChoroplethPoint[]` with `x`/`y` (grammar.ts:754-771). `lon`/`lat` are unavailable in regions mode (Google does not expose centroids) so omit them — ChoroplethTrace then falls back to reading the regions in declared order, which the schema explicitly supports. A markers-mode GeoChart whose DataTable carries Lat/Long columns gives the centroids for free and should fill them. `neighbors` is not recoverable and must be omitted. Traps: GeoChart paints every region of the selected resolution, not only the data rows, and the paths carry no identifying attribute, so highlighting almost certainly has to be dropped (pie precedent, converters.ts:762-769); GeoChart also needs network access to Google's geo data at draw time, which matters for jsdom tests. |
| `WATERFALL` | CandlestickChart used as floating bars (set low=open=start, high=close=end); the other common recipe is a stacked ColumnChart with a transparent base series | Pseudo-type `'WaterfallChart'`. Google has no native waterfall class, but a candlestick with the wick collapsed onto the body draws exactly the floating bar a waterfall needs, and the rise/fall colouring falls out of open-vs-close for free. Detection is the explicit chartType over either the candlestick 5-column table or a plain [label, start, end] table. Payload `WaterfallPoint[]` (grammar.ts:460-476): `x` from col 0, absolute `start`/`end`, `delta = end - start`, `kind` from the sign — with `'total'` rows (opening/closing/subtotal bars) impossible to infer from values alone, so they need an adapter option listing the total row indices. Existing `buildCandlestickLayer` (converters.ts:793) and `markCandlestickElements` (converters.ts:1297) give a head start on both the extraction and the body/wick rect classification. |
| `DIVERGING` | BarChart / ColumnChart with two series, one negated — a population pyramid or Likert split | Pseudo-type `'DivergingBarChart'`, alongside the existing 'StackedBarChart'/'DodgedBarChart' entries in types.ts:128-139. Reuse `buildSegmentedLayer` (converters.ts:544) with `traceType: TraceType.DIVERGING` and — critically — WITHOUT stripping the sign: DivergingTrace (src/model/diverging.ts:35, extends SegmentedTrace) reads `SegmentedPoint[][]` where the sign is the side and the magnitude is the pitch. Note src/model/diverging.ts:60-70: a diverging layer's sides are read in DECLARATION order, unlike a stacked bar's reversed order, so `domMapping.groupDirection` must not be left at the segmented default. |
| `DOT` | LineChart drawn with `lineWidth: 0, pointSize: N` over a string domain column (Google's standard Cleveland dot-plot recipe); ScatterChart when x is numeric | Pseudo-type `'DotChart'`. TraceType.DOT routes to BarTrace unchanged (src/model/factory.ts:66-69), so `buildBarLayer` (converters.ts:508) can be reused verbatim with `type: TraceType.DOT` — payload is `BarPoint[]`. The type exists purely so the chart announces itself as the chart the author drew. Highlighting targets `circle` rather than `rect`, so `markScatterElements`/`buildDataSelector(container, 'circle')` (converters.ts:1103, selectors.ts:78) is the right helper, not `markBarElements`. |
| `LOLLIPOP` | ComboChart — a thin-bar column series plus a `{type: 'line', lineWidth: 0, pointSize: 12}` series over the same values | Pseudo-type `'LollipopChart'`. Like DOT it routes to BarTrace (src/model/factory.ts:66-69) and needs only `BarPoint[]`, so `buildBarLayer` is reused with `type: TraceType.LOLLIPOP`. The trap is the ComboChart recipe duplicates the value column so the stems and the dots can be styled separately: `countDataColumns` (converters.ts:912) would see two data columns, and the generic path would silently emit DODGED (converters.ts:502). The explicit pseudo-type must bypass `buildBarOrSegmentedLayer` and read a single named/first data column. |
| `FUNNEL` | No native funnel class — drawn as an ordered horizontal BarChart, or as a centred stacked BarChart with a transparent padding series for the trapezoid look | Pseudo-type `'FunnelChart'`. Honest caveat: this is a recipe, not a Google chart type — but the payload FunnelTrace needs is just `BarPoint[]` of stage counts (src/model/funnel.ts:43 extends BarTrace, reading `barValues[0]`), which is exactly what an ordered [stage, count] DataTable already gives, so `buildBarLayer` is reused with `type: TraceType.FUNNEL` and the retention ratios are derived in the model. If the padding-series recipe is used, the transparent spacer column must be excluded rather than treated as a second series. |
| `BUMP` | LineChart of ranks with `vAxis: {direction: -1}` so rank 1 sits at the top | Pseudo-type `'BumpChart'`. BumpTrace extends LineTrace (src/model/bump.ts:50) and consumes the same `LinePoint[][]`, so `buildLineLayer` (converters.ts:605) is reused with `type: TraceType.BUMP`; the rank inversion of the pitch and the places-gained announcement are handled entirely in the model. Detection has to be the explicit pseudo-type: a bump chart's DataTable is indistinguishable from any other multi-series LineChart, and the `vAxis.direction` that gives it away lives in draw options the adapter never sees. |
| `DUMBBELL` | Intervals recipe — a series drawn with `lineWidth: 0` plus two `role: 'interval'` columns rendered with `intervals: {style: 'sticks'}` (or 'points'), giving a segment with a mark at each end | Pseudo-type `'DumbbellChart'` over either that intervals table or a plain [category, start, end] table. Payload is the `DumbbellData` OBJECT, not a bare array (grammar.ts:629-673, read at src/model/dumbbell.ts:130): `points: DumbbellPoint[]` of `{x, start, end}` plus `startLabel`/`endLabel` taken from the two value columns' `getColumnLabel` — those labels are the content of the comparison ("1990" vs "2020") and the trace announces them. The delta is deliberately derived, so do not emit one. Caveat: no native dumbbell chart type exists in Google Charts; this is a documented intervals styling. |
| `SURVIVAL` | SteppedAreaChart with `areaOpacity: 0` (Google's way of drawing a step line), optionally with `role: 'interval'` columns rendered as the confidence band | Pseudo-type `'SurvivalChart'`. Payload `SurvivalPoint[][]` — a `StepPoint`/`LinePoint` with `censored`, `yMin`, `yMax` (grammar.ts:1034-1050) — so `buildLineLayer` supplies the backbone, the interval role columns supply `yMin`/`yMax` (same detection as ERROR_BAR), and `censored` has to come from a boolean/annotation role column, since a censoring tick is drawn as a separate zero-width series that the DataTable does not otherwise distinguish. Also set `stepDirection` (MaidrLayer, grammar.ts:1204+) only if the caller declares it — MAIDR substitutes no default. |
| `VOLCANO` | ScatterChart (a volcano plot IS a scatter of effect size against significance) | Pseudo-type `'VolcanoChart'` over a ScatterChart. Payload `VolcanoPoint[]` extends ScatterPoint with `label` — the gene/probe identity, which is the whole payload of the chart (grammar.ts:895-912). Google can carry that in a `role: 'annotation'` or `role: 'tooltip'` column, so `buildScatterLayer` (converters.ts:650) is extended to read the identity out of the role column it currently skips. `thresholdOptions` (`significance`, `significanceDirection`, `effect`, grammar.ts:917-950) has no representation in a Google DataTable at all and must be passed through a new `GoogleChartAdapterOptions` field; the enum doc is explicit that MAIDR guesses no default cutoff. |
| `MANHATTAN` | ScatterChart with one series per chromosome (the standard Google recipe for the alternating-colour banding) | Pseudo-type `'ManhattanChart'`. Same VolcanoPoint payload and same VolcanoTrace (src/model/factory.ts:131-136), with `group` filled from each series' `getColumnLabel` (the chromosome) and `label` from an annotation/tooltip role column. The build is a per-series flattening of the multi-column ScatterChart the current `buildScatterLayer` (converters.ts:650) does not handle — it hard-codes columns 0 and 1 (converters.ts:659-660) — so multi-series scatter extraction is the actual work here, and it is shared with VOLCANO. |
| `NETWORK` | google.visualization.OrgChart (`orgchart` package) — nodes joined by links via parent pointers | New `chartType: 'OrgChart'` case. The DataTable is [node id (+optional formatted name), parent id, tooltip], which maps one-to-one onto `NetworkPoint[]` — emit `{source: parentId, target: nodeId}` for every row with a parent, and let NetworkTrace derive the nodes and their degrees from the links (grammar.ts:848-868). Caveats worth weighing before building: an org chart is a tree, so a TREEMAP/hierarchy reading may serve a reader better than a degree-based network reading; and OrgChart renders an HTML `<table>` rather than SVG, so `selectors` (typed for SVG elements throughout the model) most likely has to be dropped. |

## Not renderable by Google Charts

<details><summary>14 trace types the library genuinely cannot draw</summary>

- ALLUVIAL — Google Charts has no alluvial layout. Its Sankey is the left-to-right budgeted diagram, and a Google Sankey should be emitted as SANKEY (both route to FlowTrace anyway, src/model/factory.ts:88-93), so claiming ALLUVIAL would only mislabel the chart.
- BOXEN — Google Charts has no box plot of any kind (box plots are faked with candlesticks), let alone the variable-depth letter-value ladder a boxen needs.
- CHORD — no circular/radial flow diagram exists in any Google Charts package.
- CONTOUR — no contour, isoline, or density-field chart; Google Charts has no 2-D scalar field renderer at all.
- FOREST — the three things that make a forest plot (weighted marker areas, the pooled diamond, the null line as a first-class element) cannot be drawn. The closest Google can do is a horizontal interval chart, which is honestly ERROR_BAR rather than FOREST.
- HEXBIN — no hexagonal binning; Google Charts offers no binning primitive beyond Histogram's 1-D bins.
- ICICLE — no icicle/partition layout. Google's hierarchy chart is TreeMap, which should be emitted as TREEMAP (same TreemapTrace, src/model/factory.ts:166-172).
- MOSAIC — bar widths in a Google BarChart/ColumnChart are not data-drivable, and the width is exactly half of what a mosaic encodes.
- PARALLEL — no parallel-coordinates chart, and no way to give each column its own axis on a corechart line chart.
- POLAR_AREA — no polar/rose/coxcomb chart; Google Charts has no polar coordinate system.
- RADAR — no radar/spider chart, for the same reason.
- RIDGELINE — no density/KDE estimation and no offset-baseline layout.
- SUNBURST — no radial partition layout; TreeMap is the only hierarchy renderer.
- WORD_CLOUD — no word cloud. WordTree is a different chart: a suffix tree of phrases whose DataTable is a list of sentences, not term/weight pairs.

</details>

## Acceptance criteria

- [ ] Unit tests under `test/adapters/google-charts/` cover every newly detected type.

## Implementation notes

Structural constraint that shapes every item above: the adapter only ever receives `(chart, dataTable, container, {id, title, chartType})` (converters.ts:130-135). It never sees the chart's draw options, so `isStacked`, `vAxis.direction`, `lineWidth: 0`, gauge `min`/`max`/band colours, and `intervals.style` are all invisible. The codebase's established answer is an invented pseudo-type in the `GoogleChartType` union (types.ts:128-139) — 'StackedColumnChart', 'DodgedBarChart' are not Google class names — so new reading models should follow that, and anything genuinely numeric that lives only in options (gauge bands, volcano thresholds, waterfall total rows) needs a new `GoogleChartAdapterOptions` field.

Reusable helpers already in place: `formatCellValue` (converters.ts:854, prefers the locale-formatted value), `numericValue` (converters.ts:869, returns NaN rather than 0 for missing data — the model treats NaN as a gap), `isRoleColumn` (converters.ts:884), `firstDataColumn` (converters.ts:900), `countDataColumns` (converters.ts:912), `buildDataSelector` / `ensureContainerId` / `nextId` (selectors.ts:78, 57, 44), plus the bbox-matching markers `markBarElements` (converters.ts:939), `markSegmentedBarElements` (converters.ts:1001), `markScatterElements` (converters.ts:1103), `markLinePointElements` (converters.ts:1243), `markPieSliceElements` (converters.ts:743), `markCandlestickElements` (converters.ts:1297).

Precedent worth reusing: when the DOM cannot be mapped to the data with confidence, the pie builder warns and returns no selector rather than highlighting the wrong element (converters.ts:762-774). Every non-corechart package below (Sankey, TreeMap, GeoChart, Gantt/Timeline, Gauge, OrgChart) lacks `getChartLayoutInterface()`, so shipping them with audio/text/braille and no `selectors` is a legitimate first cut.

Traps found while reading:
- `GoogleChartAdapterOptions.chartType`'s JSDoc advertises `'NormalizedBarChart'` (converters.ts:86), but that string is in neither the `GoogleChartType` union (types.ts:128) nor the switch (converters.ts:453), so a caller following the doc hits the `default` throw. TraceType.NORMALIZED (percent-stacked bars) is unimplemented despite the doc — it is pre-pie so out of scope here, but it is a real gap and a cheap one.
- `markSegmentedBarElements`' doc comment (converters.ts:986-995) claims SegmentedTrace wants elements ordered category-first, then the code deliberately marks series-first and compensates with `domMapping: {order: 'row'}` (converters.ts:592-595). The comment is stale; trust the code.
- `GoogleDataTable.getColumnRole` is optional (types.ts:21) and `isRoleColumn` returns false when it is absent (converters.ts:884-890). ERROR_BAR, SURVIVAL, VOLCANO and MANHATTAN all depend on role columns, and a caller passing a DataView-like object without `getColumnRole` would have its interval/annotation columns silently read as extra data series (and so as DODGED, converters.ts:502).
- `buildScatterLayer` hard-codes columns 0 and 1 (converters.ts:659-660), so multi-series scatter extraction has to be written before MANHATTAN or a hue-split VOLCANO is possible.
- Payloads that are objects rather than arrays: GaugePoint (grammar.ts:530), GanttData (grammar.ts:554), DumbbellData (grammar.ts:653). Every existing builder returns an array, so those three break the local pattern.

Test convention: /home/user/maidr/test/adapters/google-charts/pie.test.ts builds a hand-written `GoogleDataTable` literal and a `GoogleChart` fake whose `getChartLayoutInterface()` throws when it must not be called (pie.test.ts:12-36), plus a JSDOM container holding a synthetic SVG (pie.test.ts:41-65); multiPanel.test.ts does the same for the grid API. New chart types should follow that — no real Google Charts loader is involved. Example convention: one HTML page per family, /home/user/maidr/examples/google-charts.html (six charts, one per chartType) and google-charts-pie.html. The supported-types table in /home/user/maidr/docs/google-charts.md:80-94 (and its "Not supported" line at :94) is hand-maintained and must be updated alongside.

Out-of-scope escape hatch worth knowing: modern Google Charts ships a `vegachart` package (`google.visualization.VegaChart`) that renders arbitrary Vega/Vega-Lite specs, which technically makes RADAR, PARALLEL, RIDGELINE and others "drawable by Google Charts". That belongs to the existing vegalite adapter (src/adapters/vegalite/), not here — a VegaChart's payload is a Vega-Lite spec, so the right move if it ever comes up is to hand the spec to that adapter rather than to grow this one.

jooyoungseo:
## Description

The `google-charts` adapter had fallen behind MAIDR core: of the trace types added since the pie chart it emitted none, so a chart Google Charts draws perfectly well was either dropped from the figure or announced as the wrong chart. This adds all **20 in-scope trace types** from #867 to the adapter, each with its detection, its payload and — wherever the drawn chart makes it possible — its highlighting.

The structural constraint that shapes the whole change is unchanged: the adapter is handed `(chart, dataTable, container, options)` and **never sees the chart's draw options**. So `isStacked`, `vAxis: {direction: -1}`, `lineWidth: 0`, `areaOpacity: 0`, gauge bands and volcano cutoffs are all invisible to it. Where the distinction lives only in those options, the reading is named by the caller through a pseudo-type in the `GoogleChartType` union — the convention `'StackedColumnChart'` and `'DodgedBarChart'` already follow. Where it lives in the DataTable, it is detected: `role: 'interval'` columns are the one variant the adapter can find on its own.

Three numbers that exist nowhere in a DataTable get new `GoogleChartAdapterOptions` fields: `gaugeOptions` (a dial's range and bands), `thresholdOptions` (the significance and effect cutoffs a volcano and a Manhattan are read through) and `waterfallTotals` (which rows are opening, closing or subtotal bars).

No file outside `src/adapters/google-charts/`, `test/adapters/google-charts/`, `examples/` and `docs/` is touched. The model, service, state, ui, command layers and `src/type/grammar.ts` are untouched — the core already supported every payload here.

## Related Issues

Closes #867

### Trace types added (20)

Grouped by the builder they share.

**Area family** — the DataTable is a `LineChart`'s and MAIDR reads both as `LinePoint[][]`, so all three reuse the line builder. Values go out raw; `AreaTrace` derives the running total and the percentage scaling itself.

| TraceType | `chartType` | Google construct |
|---|---|---|
| `AREA` | `'AreaChart'` | `AreaChart` |
| `STACKED_AREA` | `'StackedAreaChart'` | `AreaChart` + `isStacked: true` |
| `NORMALIZED_AREA` | `'NormalizedAreaChart'` | `AreaChart` + `isStacked: 'percent'` |

**Detected from the DataTable, not named by the caller**

- `ERROR_BAR` — a single-series line, scatter, column or bar chart declaring `role: 'interval'` columns now becomes an error bar layer instead of dropping the bounds, as it did before. Two interval pairs (a 95% band drawn inside a 99% one) read as the outermost; a lone interval column reads as the one bound it is, chosen by which side of the estimate it falls on. A **multi-series** chart keeps its previous reading — `ErrorBarPoint[]` is flat, so a second estimate column has nowhere to go, and losing a series is worse than losing its intervals.

**Marks Google has no class for** — each reuses the bar or segmented builder with the trace type passed in.

- `DOT` (`'DotChart'`) — `LineChart` + `lineWidth: 0, pointSize: N`.
- `LOLLIPOP` (`'LollipopChart'`) — a `ComboChart` whose recipe **repeats the value column** so stems and dots can be styled apart; only the first is read, so it does not silently become a dodged bar.
- `FUNNEL` (`'FunnelChart'`) — an ordered `BarChart` of stage counts. The centred recipe stacks a transparent padding series, so the column read is the one whose values fall.
- `DIVERGING` (`'DivergingBarChart'` / `'DivergingColumnChart'`) — the segmented builder with **the sign left on**, because the sign is which side of the baseline the bar grows towards.
- `WATERFALL` (`'WaterfallChart'`) — reads both the candlestick-as-floating-bars recipe (low = open = start, high = close = end) and a plain `[label, start, end]` table. Totals cannot be read off the numbers, so they are named by row index in `waterfallTotals`.

**Non-corechart packages**

- `SANKEY` (`'Sankey'`) — `[from, to, weight]` to a flat `FlowPoint[]`; no node list, since a flow names both its ends.
- `TREEMAP` (`'TreeMap'`) — parent pointers walked upward into the `path` MAIDR addresses nodes by. An interior node's placeholder size is omitted so the trace sums its children rather than announcing zero.
- `GANTT` (`'Gantt'` and `'Timeline'`) — one nested `GanttData` object, lanes grouped from the task/row label. Dates convert to **days or hours** rather than epoch milliseconds, with an axis format that renders the same numbers back as dates, so an announced length is a number a listener can hold.
- `GAUGE` (`'Gauge'`) — `GaugePoint` is a single measure, so an N-row table becomes **N layers**, which Page Up and Page Down move between. Google's free-standing `greenFrom`/`redTo` spans are mapped onto bands that partition the dial; any stretch the options leave uncoloured becomes a band named `unbanded`.
- `CHOROPLETH` (`'GeoChart'`) — both modes. A markers table drawn from coordinates carries centroids, which turn the reading into a walk across the map; a regions table cannot supply them and is read in declared order, which the schema explicitly supports.
- `NETWORK` (`'OrgChart'`) — parent pointers as links, one per row that names a parent. Identity comes from the **raw** cell, since an org chart routinely puts markup in the value it draws.

**Rank, comparison, survival and threshold readings**

- `BUMP` (`'BumpChart'`) — the line builder with the trace type, which is what inverts the pitch so rank 1 is the highest note. Read as a `'LineChart'`, the same table sonifies a team climbing the table as a team falling.
- `DUMBBELL` (`'DumbbellChart'`) — both recipes (the intervals styling and a plain `[category, start, end]` table), emitting the `DumbbellData` object. `startLabel`/`endLabel` come from the two value columns, so a reader is told which end they are on. The delta is deliberately not emitted — the trace derives it.
- `SURVIVAL` (`'SurvivalChart'`) — a `SteppedAreaChart`'s arms with their interval band, plus a **boolean** censoring column which is never read as an arm. Times go out numeric so median survival can be read off them.
- `VOLCANO` (`'VolcanoChart'`) and `MANHATTAN` (`'ManhattanChart'`) — every series of a scatter is flattened and the null cells the banding recipe leaves are dropped; each point's identity comes from a `role: 'annotation'` or `role: 'tooltip'` column, which is the payload these charts exist for. The cutoffs arrive as `thresholdOptions`; MAIDR guesses none.

**Two of the twenty ship without visual highlighting** — deliberately, following the pie precedent that a highlight on the wrong element is worse than none. Audio, text and braille carry them fully:

- `CHOROPLETH` — a `GeoChart` paints every region of the chosen resolution rather than only the rows it was given, in Google's own geographic order, and its paths carry no class or id. A highlight would sit on a different country from the one being announced.
- `NETWORK` — an `OrgChart` renders an HTML `<table>` rather than SVG and draws no element per link at all; the connectors are cell borders, so there is nothing for a selector to point at.

Highlighting for `SANKEY`, `TREEMAP`, `GANTT` and `TIMELINE` is present but conditional: those packages expose no `getChartLayoutInterface()`, so drawn elements are matched by DOM order and only when the counts agree exactly. When they do not, the selector is dropped with a warning rather than pointing at the wrong element.

**Fourteen trace types are not added because Google Charts genuinely cannot draw them:**

- `ALLUVIAL` — no alluvial layout. Google's Sankey is the left-to-right budgeted diagram and is emitted as `SANKEY`; claiming `ALLUVIAL` would only mislabel the chart.
- `BOXEN` — Google Charts has no box plot of any kind, let alone a letter-value ladder.
- `CHORD` — no circular/radial flow diagram in any package.
- `CONTOUR` — no contour, isoline or density-field renderer.
- `FOREST` — weighted marker areas, the pooled diamond and the null line as a first-class element cannot be drawn. The closest Google manages is a horizontal interval chart, which is honestly `ERROR_BAR`.
- `HEXBIN` — no hexagonal binning.
- `ICICLE` — no partition layout. Google's hierarchy chart is TreeMap, emitted as `TREEMAP`.
- `MOSAIC` — bar widths are not data-drivable, and width is half of what a mosaic encodes.
- `PARALLEL` — no parallel-coordinates chart and no per-column axis.
- `POLAR_AREA` — no polar coordinate system.
- `RADAR` — same reason.
- `RIDGELINE` — no KDE and no offset-baseline layout.
- `SUNBURST` — no radial partition layout.
- `WORD_CLOUD` — no word cloud. WordTree is a different chart entirely.

Also out of scope and untouched: `NORMALIZED` (percent-stacked **bars**), which predates the pie chart and remains a real gap — `GoogleChartAdapterOptions.chartType`'s JSDoc advertises `'NormalizedBarChart'`, but the string is in neither the union nor the switch.

### Tests, examples and docs

- 18 new unit test files under `test/adapters/google-charts/`, one per family, following the `pie.test.ts` convention: hand-written `GoogleDataTable` literals, a `GoogleChart` fake whose `getChartLayoutInterface()` throws when it must not be called, and a JSDOM container holding a synthetic SVG. No real Google Charts loader is involved.
- 5 new example pages: `google-charts-area.html`, `google-charts-statistical.html`, `google-charts-relational.html`, `google-charts-marks.html`, `google-charts-gauge-map.html`.
- `docs/google-charts.md`: the supported-types table and the "Not supported" line are extended, with a note per family covering the DataTable recipe each reading expects and what is lost if it is named wrong.

### Notable internals

`markPointMarkerElements` is generalised to mark a series/row pair per data point, and now verifies that the chart drew its markers in the order the data declares them — a single attribute selector resolves in document order, so a disagreement would have highlighted another row's point silently.

## Checklist

## Additional Notes

**Gate status — all four pass on this branch, with no fixes required:**

| Step | Result |
|---|---|
| `npm run lint:fix` | clean, no changes written |
| `npm run type-check` | clean (`tsc --noEmit` and `tsconfig.ci.json`) |
| `npm run build` | all 12 bundles built |
| `npm test` | 202 suites / 2702 tests passed, plus 7 suites / 73 tests in the ESM project |

The manual-testing checklist box is left unticked honestly: the automated suites and the example pages are in place, but `ManualTestingProcess.md` was not walked through against a live Google Charts loader. `GeoChart` in particular fetches Google's geo data at draw time, which is why its tests are JSDOM-only.

The examples are the intended place to exercise these by hand — each page draws its family with the DataTable recipe the docs describe, so a reviewer can open one and navigate with a screen reader without assembling a chart first.

---


jooyoungseo:
Round-two pass. All 14 checks were green at `81acf20`, and there were no review findings to act on — worth flagging that the `claude-review` job exited `success` without producing a review (its comment is still stuck at "Deep review of `converters.ts` … (in progress)"), and Copilot declined on a quota limit. So the green check on this PR is **not** an approval; nothing has actually reviewed the diff yet.

I reviewed it myself instead. One real gap, now fixed in `f537762`:

- **Reading options undocumented in the options table.** `gaugeOptions`, `stepDirection`, `thresholdOptions` and `waterfallTotals` were explained in the per-chart prose notes but absent from "Configuration Options", which listed only `chartType` / `title` / `id`. A reader scanning that table would conclude the adapter takes nothing else — for a PR whose whole point is those four options. Added a reading-options table and noted that `createMaidrFromGoogleCharts` accepts the same four per panel, since `GoogleChartPanel` extends `GoogleChartReadingOptions`.

Two things I looked at and deliberately left alone:

- **`buildLineLayer` names `axes.y` from column 1**, which is wrong when column 1 is a role column and debatable on a multi-series chart. It is unchanged from `main` and not something this PR touches; fixing it would widen the diff and change existing line/area output. Better as its own commit.
- **Degenerate DataTables** (a waterfall with one data column, a gauge whose colour spans overlap so the derived bands come out non-monotonic). Both are malformed inputs, and the adapter's house rule throughout is to read what it can rather than validate; adding guards for them here would be inconsistent with the rest of the file.

One thing I could not fix from this branch: `scripts/build-site.js` is outside the paths this PR is scoped to, so the five new example pages (`google-charts-area`, `-gauge-map`, `-marks`, `-relational`, `-statistical`) are not linked from the example gallery — it still lists only the combined page and the pie example — and the Google Charts blurb on line 57 still reads "bar, line, scatter, candlestick, stacked, dodged, and pie charts". The pages build and are reachable by direct URL in the preview; they just have no entry point. Needs a maintainer or a follow-up PR.

CI for `f537762` is still running, so this is not merge-ready until it comes back green.

---

PR #877, opened 2026-08-14
feat(anychart): support the chart types added since the pie chart

### LINKED ISSUE #868: feat(anychart): support the chart types added since the pie chart
## Problem

MAIDR core has gained ~34 new trace types since the pie chart landed (radar, gauge, funnel, waterfall, gantt, treemap, sankey, sunburst, dumbbell, dot/lollipop, diverging, area family, and more). The `anychart` adapter has not kept up: of those, it currently emits **none**.

A chart AnyChart draws perfectly well is therefore either dropped from the figure or announced as the wrong chart — which for a screen-reader user is worse than not being supported at all.

Two dispatch paths, both in /home/user/maidr/src/adapters/anychart/converters.ts. (1) Series-backed charts: `buildSubplot` loops `chart.getSeriesAt(i)` and hands `series.seriesType()` to `mapSeriesType` (converters.ts:100), a flat lowercase-string lookup table (converters.ts:102-125) that returns one of eight values or `null` — a `null` logs a warning and skips the series (converters.ts:2712-2717). The result is narrowed by the `AnyChartTraceType` union (converters.ts:66-74) and dispatched by `buildLayer`'s exhaustive switch (converters.ts:2534-2559), with `resolveSelector`'s equally exhaustive switch (converters.ts:478-498) supplying the highlight selector. (2) Single-dataset charts that expose no series API are routed at chart level, BEFORE the series loop, by substring-testing `chart.getType()`: heatmap at converters.ts:2652 and pie via `isPieChart` at converters.ts:2668/1931. A `try/catch` around `getSeriesCount()` (converters.ts:2683-2699) is the catch-all fallback and sends anything that throws to the heatmap builder.

## In scope — 19 trace types AnyChart can render

| TraceType | AnyChart construct | Detection / payload |
|---|---|---|
| `AREA` | `area` / `spline-area` cartesian series (anychart.area()) | Already recognised and deliberately downgraded: `mapSeriesType` maps `area`→LINE and `spline-area`→LINE (converters.ts:113,115) and emits a console.warn that the fill is lost (converters.ts:131-136). Flip those two table rows to TraceType.AREA and delete the warning. Payload is unchanged `LinePoint[][]` from `buildLineLayer` (converters.ts:2175) because `AreaTrace extends LineTrace` (src/model/area.ts:81) and factory.ts:57 routes AREA there. Also add AREA to the `AnyChartTraceType` union (converters.ts:66) and to both exhaustive switches (resolveSelector converters.ts:478 — reuse the LINE marker selector; buildLayer converters.ts:2541). |
| `STACKED_AREA` | area / spline-area / step-area series with chart.yScale().stackMode('value') | Chart-level, not series-level. Read `chart.yScale().stackMode()` once in `buildSubplot` before the series loop (converters.ts:2705) — the literals 'value' and 'percent' are both present in the AnyChart 8.13.0 bundle — and promote every area-family series from AREA to STACKED_AREA. Payload stays `LinePoint[][]`: `AreaTrace` reads the stacking off `layer.type` and computes the column totals itself (src/model/area.ts:105-131), so the adapter must emit each series' OWN value, never the running total. Requires adding `yScale()` to `AnyChartInstance` (types.ts:109-141). |
| `NORMALIZED_AREA` | area-family series with chart.yScale().stackMode('percent') | Identical to STACKED_AREA above, branching on `stackMode() === 'percent'`. Same `LinePoint[][]` payload, same AreaTrace (factory.ts:58). Rides entirely on the STACKED_AREA work — one extra string comparison. |
| `FUNNEL` | anychart.funnel() and anychart.pyramid() — getType() returns 'funnel' / 'pyramid' | Both are single-dataset charts with no series API, structurally identical to the pie the adapter already special-cases. Add a branch in `buildSubplot` beside `isPieChart` (converters.ts:2668) testing `chart.getType()` for 'funnel'/'pyramid', then reuse `readPieRows` (converters.ts:1945) verbatim — the rows carry the same `x`/`name`/`value` fields `readRows` already collects (converters.ts:260-282). Payload is `BarPoint[]` of `{x: stage, y: count}` in draw order: FUNNEL routes to `FunnelTrace extends BarTrace` (src/model/funnel.ts:43) which derives share and retention itself. Highlight needs a `stampFunnelAttributes` modelled on `stampPieAttributes` (segments are `<path>`s in an `ac_layer_*` group, but trapezoids rather than arcs, so the `hasArcCommand` test is replaced by a count/geometry filter). |
| `WORD_CLOUD` | anychart.tagCloud() — getType() returns 'tag-cloud' (anychart-tag-cloud module) | Chart-level route in `buildSubplot` on `getType().includes('tag-cloud')`; data lives on `chart.data()` exactly like the pie's, rows carrying `x` and `value`, so `readPieRows` is reusable. Payload `WordCloudPoint[]` = `{x: term, y: weight}`; emission order is irrelevant because `WordCloudTrace` re-sorts by descending weight on construction (src/model/wordCloud.ts:74). Highlighting is easier than every existing family: each term is a single `<text>` element, so the stamper is a document-order walk with no geometric filtering. |
| `SUNBURST` | anychart.sunburst() — getType() returns 'sunburst' | Chart-level route. `chart.data()` is an `anychart.data.Tree`, not a flat view, so the walk uses `getChildren()` / `numChildren()` / `getChildAt(i)` (all three confirmed present in the 8.13.0 bundle) instead of an iterator. Payload `TreemapPoint[]`: one entry per leaf, `{x: node.get('name'), y: node.get('value'), path: [ancestor names, root first, excluding self]}` — `TreemapTrace` derives the interior nodes and their totals from the paths (src/model/treemap.ts:141-147). Shares one class with ICICLE and TREEMAP (factory.ts:166-168), which only reads `layer.type` to pick angular panning. |
| `TREEMAP` | anychart.treeMap() — getType() returns 'tree-map' | The same `anychart.data.Tree` walk and the same `TreemapPoint[]` payload as SUNBURST; only the emitted `layer.type` and the getType() branch differ. Whichever of the two is built first makes the other a two-line addition. Highlight: nodes are `<rect>`s in nested layer groups, so the existing `collectShapeCandidatesFromRoot` bar filter (converters.ts:507) is a usable starting point, but depth-ordering the stamps matters because the trace navigates parent/child rather than rows and columns. |
| `SANKEY` | anychart.sankey() — getType() returns 'sankey' (anychart-sankey module) | Chart-level route in `buildSubplot`; `chart.data()` rows carry `from`, `to` and `weight` (all three string literals confirmed in the 8.13.0 bundle). Add those names to `readRows`' field list (converters.ts:260) or do a dedicated iterator walk. Payload is a flat `FlowPoint[]` of `{source: from, target: to, value: weight}` — `FlowTrace` builds the node list and assigns stages from the edges alone (src/model/flow.ts:156-158), so no node payload is needed. Highlight: ribbons are `<path>`s in an `ac_layer_*` group; stamp in flow-emission order. |
| `WATERFALL` | anychart.waterfall() — getType() 'waterfall', series type 'waterfall' | Series-backed, so it is one row in the `mapSeriesType` table (converters.ts:102) plus a builder. Rows carry `x`, `value` and the boolean `isTotal` (confirmed in the bundle) — add `isTotal` to `readRows` (converters.ts:260). `WaterfallPoint` demands ABSOLUTE `start`/`end` (grammar.ts:460-485), so the builder accumulates a running total across rows and must honour `chart.dataMode()` (the literals 'diff' and 'absolute' are both in the bundle): in diff mode `value` IS the delta, in absolute mode `delta = value - runningTotal`. `kind` is 'total' when `isTotal`, else 'increase'/'decrease' from the sign of delta. |
| `RADAR` | anychart.radar() — getType() returns 'radar'; its series are ordinary line / area / marker series | MUST be detected at chart level. A radar's series report `seriesType()` as plain 'line'/'area'/'marker', so `mapSeriesType` (converters.ts:100) already emits a silent, wrong LINE layer for a radar chart today — this is a mis-description, not just a gap. Test `chart.getType() === 'radar'` in `buildSubplot` before the series loop (converters.ts:2705) and override the mapped type. Payload is unchanged `LinePoint[][]` from `buildLineLayer`: `RadarTrace extends LineTrace` (src/model/radar.ts:47) and computes the spoke angles itself from the longest series (src/model/radar.ts:87-98). |
| `POLAR_AREA` | anychart.polar() with column (or area) series — getType() returns 'polar' | The same chart-level override as RADAR: `getType() === 'polar'` plus a `column`/`area` series maps to POLAR_AREA, while a polar `line`/`marker` series reads better as RADAR. POLAR_AREA and RADAR share one `RadarTrace` (factory.ts:150-151), so the payload is the identical `LinePoint[][]` and this is a second branch on the radar work rather than independent effort. |
| `MOSAIC` | anychart.mosaic() / anychart.mekko() / anychart.barmekko() — getType() 'mosaic' / 'mekko' / 'barmekko' | Chart-level `getType()` test, but a mekko chart DOES expose the series API, so the existing `getSeriesCount()` loop still drives it — each series is one stack level. Payload is `MosaicPoint[][]` nested by column (like `SegmentedPoint[][]`), each cell `{x: category, y: value, z: series.name(), width}`. `width` is the one field with no source in the row data: compute it as columnTotal / grandTotal by summing every series at each x, and carry it on EVERY cell of the column as the grammar requires (grammar.ts:966-991). MOSAIC routes to `MosaicTrace extends SegmentedTrace` (src/model/mosaic.ts:36). |
| `GAUGE` | anychart.gauges.circular() / anychart.gauges.linear() — getType() 'circular-gauge' / 'linear-gauge'; and anychart.bullet() — 'bullet' | Chart-level route. `GaugePoint` is a single OBJECT rather than an array (grammar.ts:530-546), which no existing builder emits: `value` from the pointer's data row (`chart.data()` first row, or the `needle()`/`bar()`/`marker()`/`knob()` pointer accessors — all four names confirmed in the bundle), `min`/`max` from `chart.axis(0).scale().minimum()/maximum()`, `bands` from the chart's `range(i)` colour ranges mapped to `{to, label}` in ascending order, and `target` from a bullet chart's 'line'-type data row or its `rangeMarker`. Needs `axis()`, `range()` and the pointer accessors added to `AnyChartInstance` (types.ts:109-141). |
| `ERROR_BAR` | cartesian series error bars — series.error() with valueError / valueLowerError / valueUpperError (all confirmed in the bundle) | Series-level but conditional on configuration rather than on `seriesType()`: a line/marker/column series whose `series.error().enabled()` is true, or whose rows carry `valueLowerError`/`valueUpperError`, becomes ERROR_BAR instead of its base type. Add those field names to `readRows` (converters.ts:260). `ErrorBarPoint` wants ABSOLUTE bounds (grammar.ts:380-389) while AnyChart supplies offsets that may be a percent string such as '10%' on the series-level `error()` config — so the builder computes `yMin = y - lower`, `yMax = y + upper` with any percent resolved against `y`. `ErrorBarTrace extends AbstractTrace` (src/model/errorBar.ts:108), so no line-family reuse. |
| `DOT` | `marker` series on a cartesian chart with an ordinal x scale (Cleveland dot plot) | `mapSeriesType` currently sends `marker` to SCATTER (converters.ts:117), and `buildScatterLayer` then forces both coordinates numeric via `asNumber(r.x)` (converters.ts:2221) — so a dot plot's category labels are already being coerced to 0 today. Detect by asking whether `chart.xScale()` is ordinal, or more cheaply by testing whether the rows' `x` values are non-numeric strings; then emit `BarPoint[]` through `buildBarLayer` with TraceType.DOT. No model work: DOT shares `BarTrace` with BAR and LOLLIPOP (factory.ts:66-69). The one risk is regressing genuine numeric scatters, so the ordinal-scale test is the safer discriminator. |
| `LOLLIPOP` | `stick` series (confirmed in the bundle), usually drawn with markers enabled | Add `'stick': TraceType.LOLLIPOP` to the `mapSeriesType` table (converters.ts:102). Rows carry the same `x`/`value` pair a bar series does, so `buildBarLayer` (converters.ts:2157) produces the `BarPoint[]` payload with no change and LOLLIPOP routes to `BarTrace` (factory.ts:68). Highlight is the only real work: a stick is a thin vertical stroke, which `collectLineMarkerCandidates`' aspect-ratio filter rejects (converters.ts:780), so either add 'stick' to `LINE_LIKE_SERIES_TYPES` (converters.ts:314) so the markers get stamped instead, or write a thin dedicated stamper. |
| `DUMBBELL` | `range-column` / `range-bar` / `hilo` series (all three confirmed in the bundle) | Add them to the `mapSeriesType` table; their rows carry `low` and `high`, so add those two names to `readRows`' field list (converters.ts:260). Payload is `DumbbellData` — an OBJECT, not an array (grammar.ts:651-662): `{points: [{x, start: low, end: high}], startLabel, endLabel}`. The two end labels have no AnyChart source and should come from `AnyChartBinderOptions` or be omitted rather than invented. Honest caveat: AnyChart draws these as floating bars or hi-lo sticks, not as two dots joined by a segment, so DUMBBELL is the closest available reading of the data rather than the chart the author drew — worth a doc note the way the area→line warning already is. |
| `NETWORK` | anychart.pert() — getType() returns 'pert'; AnyChart's only node-link diagram | Chart-level route. PERT data is a flat list of activities, each with `id`, `name`, `duration` and a `dependsOn` array — literally an edge list. Payload `NetworkPoint[]`: one `{source: dependency, target: activity}` per entry of every `dependsOn`; `NetworkTrace` derives the node set from the links alone (grammar.ts:848-869), so nothing else is required and there is deliberately no position field to fill. Honest caveat: a PERT chart is a DIRECTED schedule while `NetworkPoint` is explicitly undirected, so the direction is lost. A real but imperfect fit — AnyChart ships no force-directed graph chart. |
| `DIVERGING` | a stacked anychart.bar() whose two series straddle zero — AnyChart's own tornado / population-pyramid idiom. There is no `diverging` chart type. | There is no library-level flag, so detection is necessarily heuristic: a bar/column chart with `yScale().stackMode() === 'value'`, exactly two series, and one series whose values are all <= 0. Payload is `BarPoint[]` per series with SIGNED y, which `buildBarLayer` already emits unchanged — `DivergingTrace` takes the magnitude for the pitch and the sign for the side (grammar.ts:1332). Listed here because AnyChart genuinely draws the chart, NOT because the adapter can reliably recognise it: given the misfire risk on any ordinary stacked bar chart with negative values, this is better exposed as an explicit opt-in flag on `AnyChartBinderOptions` (types.ts:146) than inferred. |

## Deferred to a follow-up

These need a new author-facing opt-in API or semantics AnyChart does not carry in its own configuration, so they are out of scope here and tracked separately:

- `CHOROPLETH` — Chart-level `getType().includes('map')` plus a series whose type is 'choropleth'. Region rows carry `id` and `value`; the human-readable region name comes from the bound geodata feature's `properties`, reachable via `chart.geoData()`. Payload `ChoroplethPoint[]` = `{x: region name, y: value}`. `lon`
- `GANTT` — Chart-level route, but the data model shares nothing with the rest of the adapter: `chart.data()` is an `anychart.data.Tree` of tasks carrying `actualStart` / `actualEnd` (plus a `periods[]` array on a resource chart), not an iterator. Payload `GanttData` (grammar.ts:579-628): `points` NESTED one ar

## Not renderable by AnyChart

<details><summary>13 trace types the library genuinely cannot draw</summary>

- ALLUVIAL — AnyChart ships no alluvial chart; anychart.sankey() is its only flow diagram and an alluvial is indistinguishable from one at the API level, so such a chart should be emitted as SANKEY.
- BUMP — no bump chart type. A line chart on an inverted rank scale draws one, but it is indistinguishable from any other line chart, so the adapter could only guess at the rank semantics that make the trace read correctly.
- BOXEN — the `box` series exposes exactly the fixed five-number summary (lowest / q1 / median / q3 / highest, already read at converters.ts:265-269). AnyChart has no letter-value ladder at any depth.
- CHORD — no chord or circular-flow diagram exists in any AnyChart module.
- CONTOUR — no contour or iso-line chart; AnyChart has no scalar-field renderer at all.
- FOREST — no forest plot. Error bars plus a reference line could compose one visually, but nothing in AnyChart's data model marks the pooled row or carries a study weight, so both facts ForestPoint exists to add over ERROR_BAR would be invented.
- HEXBIN — no hexagonal binning; the heatMap module is a rectangular lattice only, and a hex lattice's staggered centres cannot be recovered from it.
- ICICLE — treeMap and sunburst are AnyChart's only hierarchy layouts; there is no icicle / partition layout.
- MANHATTAN — no Manhattan plot. A scatter grouped by chromosome is drawable, but no AnyChart field carries the chromosome or the point label, and thresholdOptions.significance has no library source (grammar.ts:1200-1225 is explicit that guessing it is worse than omitting it).
- PARALLEL — no parallel-coordinates chart in any AnyChart module.
- RIDGELINE — no ridgeline / joyplot, and no KDE anywhere in AnyChart, so there is no density curve to read off.
- SURVIVAL — no Kaplan-Meier chart. A `step-line` draws the curve, but censoring marks and the confidence band have no representation in any AnyChart series, so emitting SURVIVAL would mean fabricating exactly the two fields that distinguish it from the STEP layer the adapter already produces.
- VOLCANO — no volcano plot. The per-point identity label and the significance/effect cutoffs the trace is read almost entirely through have no AnyChart source, leaving only the SCATTER the adapter already emits.

</details>

## Acceptance criteria

- [ ] Unit tests under `test/adapters/anychart/` cover every newly detected type.

## Implementation notes

1. alreadySupported is genuinely EMPTY — every one of the 34 post-pie TraceTypes is a gap. The adapter's entire vocabulary is the eight pre-pie types frozen in the `AnyChartTraceType` union (converters.ts:66-74): BAR, LINE, SCATTER, STEP, BOX, HEATMAP, CANDLESTICK, PIE. AREA is the near-miss and does not count: converters.ts:113,115 map AnyChart's area series to TraceType.LINE on purpose and warn about it (converters.ts:131-136).

2. Two exhaustive switches gate every addition, by design. `resolveSelector` (converters.ts:478-498) and `buildLayer` (converters.ts:2541-2558) both switch on the union with NO `default`, so adding a member is a compile error until both are updated — the comments at converters.ts:474-478 and 2528-2533 say this is deliberate (a missing selector is a chart that announces correctly but never highlights).

3. There are two distinct addition paths and they do not overlap. Series-backed types (AREA, WATERFALL, LOLLIPOP, DUMBBELL, DOT, ERROR_BAR) are one row in the `mapSeriesType` table (converters.ts:102-125) plus a builder. Chart-level single-dataset types (FUNNEL, SANKEY, TREEMAP, SUNBURST, WORD_CLOUD, GAUGE, GANTT, CHOROPLETH) expose no series API and must be routed inside `buildSubplot` beside the existing heatmap (converters.ts:2652) and pie (converters.ts:2668) branches, BEFORE the `getSeriesCount()` call at converters.ts:2685.

4. LIVE TRAP worth fixing alongside any chart-level work: the catch at converters.ts:2686-2699 sends ANY chart whose `getSeriesCount()` throws to `buildHeatmapLayerFromChart`. Today an anychart.funnel(), treeMap(), sunburst() or tagCloud() chart therefore does not fail loudly — those rows have an `x` and a `value` but no `y`, so `buildHeatmapLayerFromChart` (converters.ts:2357) collapses them into a single-row HEATMAP layer with one empty y label and binds it. Silent mis-description, not a warning.

5. Mirror-image trap: radar and polar charts DO expose the series API, and their series report `seriesType()` as plain 'line' / 'area' / 'marker' / 'column'. `mapSeriesType` therefore already emits a plain LINE layer for a radar chart. RADAR and POLAR_AREA can only be discriminated by `chart.getType()`, never by series type.

6. Highlighting is the real per-type cost everywhere. AnyChart's GraphicsJS SVG carries no stable classes — only randomised `ac_layer_*` / `ac_path_*` ids and `data-ac-wrapper-id` (converters.ts:430-448) — so each mark family needs its own geometric/heuristic stamper registered in `stampChartAttributes` (converters.ts:2063-2072). Audio, text and braille all work without one; only highlight is lost. The established house style is to warn and skip rather than stamp a guess (see `findPieWedgeLayer`, converters.ts:2004-2020, and its rationale about landing a highlight on a legend marker).

7. `AnyChartInstance` (types.ts:109-141) is a hand-written minimal interface — the repo has NO npm dependency on anychart at all, and the examples load the library from CDN. Every proposal above requires new accessors there: `yScale()`, `xScale()`, `dataMode()`, `axis()`, `range()`, `geoData()`, and the tree-node methods.

8. Verified against the actual AnyChart 8.13.0 bundle (the version examples/anychart/*.html load). Present as chart/series types: sankey, waterfall, tag-cloud, tree-map, sunburst, mekko, mosaic, barmekko, pert, timeline, funnel, pyramid, radar, polar, bullet, circular-gauge, linear-gauge, choropleth, map, gantt-project, gantt-resource, stick, hilo, range-column, range-bar, range-area, jump-line, marker. Present as data/config fields: from / to / weight, isTotal, diff / absolute, valueError / valueLowerError / valueUpperError / xError, stackMode with 'value' and 'percent', and the tree API getChildren / numChildren / getChildAt. Several of these ship as separate modules (sankey, treemap, tag-cloud, mekko, sunburst, pert, map, gantt) — which is already established practice in this repo: examples/anychart/heatmap.html:13 loads anychart-heatmap.min.js and examples/anychart/candlestick.html:13 loads anychart-stock.min.js.

9. Test conventions (/home/user/maidr/test/adapters/anychart/converters.test.ts, the adapter's only test file): AnyChart is mocked entirely by hand over a JSDOM document — `createIterator` / `createSeries` (lines 41-66) for series-backed charts and `createPieChart` (lines 76-85) for chart-level ones. Adding a chart-level type means copying `createPieChart` with a new `getType()` string. Highlight stamping is tested by hand-building the SVG the renderer would emit, including its decoys — see `appendPieWedges` (lines 92-120), which deliberately adds a straight label connector and a fill:none stroke twin per wedge.

10. Examples convention: one file per chart type under /home/user/maidr/examples/anychart/ (bar, box, candlestick, heatmap, line, multipanel, pie, scatter), each loading the CDN bundle plus ../../dist/anychart.js and calling bindAnyChart(chart). A new trace type should ship with a matching file.

11. Cheapest-first ordering if this is being planned as work: AREA (a two-row table edit), then STACKED_AREA/NORMALIZED_AREA (one stackMode read), then FUNNEL and WORD_CLOUD (both clone the existing pie chart-level path almost verbatim), then LOLLIPOP/DOT, then RADAR+POLAR_AREA together, then WATERFALL/SANKEY/MOSAIC/TREEMAP+SUNBURST/GAUGE/ERROR_BAR/DUMBBELL/NETWORK, and CHOROPLETH/GANTT last — both need a module, a new data model, and a highlight strategy with no precedent in the file.

jooyoungseo:
# Pull Request

## Description

The AnyChart adapter read six chart families (bar, line/step, scatter, box, heatmap, candlestick) plus the pie. Every trace type MAIDR has gained since then fell through to one of two bad endings: a chart with no series API (sankey, funnel, tag cloud) reached the `getSeriesCount()` fallback and was bound silently as a one-row heatmap, and a chart whose series lie about what they are (radar, dot plot) was announced as something else entirely — a radar read as a line chart, with its highlight sorted left to right across a circle.

This adds fourteen trace types to the adapter, and each one ships with its highlight path in the same change: a trace that sonifies but does not highlight is not done. Three of the additions are corrections rather than gaps — sankey, radar and the area family were all previously mis-described rather than skipped.

Two detections are chart-level rather than series-level, because the series genuinely cannot answer: stacking is a property of `chart.yScale().stackMode()`, and what separates a Cleveland dot plot from a scatter is `chart.xScale().getType()` returning `'ordinal'`. One is opt-in: see the diverging note below.

No file outside `src/adapters/anychart/`, `test/adapters/anychart/`, `examples/` and `docs/` is touched. No core change was needed — the model, services and grammar already supported every trace type here.

## Related Issues

Closes #868

### Trace types added

| MAIDR trace | Read from | Notes |
|---|---|---|
| `AREA` | `area`, `spline-area` series | Previously downgraded to `LINE` with a fill warning; now read as the filled band it is drawn as |
| `STACKED_AREA` | the same, with `yScale().stackMode() === 'value'` | Bands merged into one layer — the running total is only computable across the whole set |
| `NORMALIZED_AREA` | the same, with `stackMode() === 'percent'` | |
| `FUNNEL` | `anychart.funnel()`, `anychart.pyramid()` | Chart-level route, like the pie. Sonifies retention from the previous stage, not the raw count |
| `WORD_CLOUD` | `anychart.tagCloud()` | Terms paired to glyphs by matching rendered text — a cloud's DOM order is packing order, not data order |
| `SANKEY` | `anychart.sankey()` | Edges only; nodes and columns are derived from them. A row with no `to` is AnyChart's dropoff and names no target, so it is not a flow |
| `WATERFALL` | `anychart.waterfall()` | Running total accumulated here, honouring `chart.dataMode()` (`'diff'` / `'absolute'`), because `WaterfallPoint` fixes `start`/`end` as absolute positions. Several series are summed into one bridge |
| `MOSAIC` | `anychart.mekko()`, `mosaic()`, `barmekko()` | Column width computed as the column total over the grand total. `count` deliberately not emitted — AnyChart accepts any measure, and declaring one would announce "Count" for a chart of revenue |
| `RADAR` | `anychart.radar()`, and polar `line` / `marker` series | Routed by chart type before the series are asked anything; their `seriesType()` says `'line'` |
| `POLAR_AREA` | `anychart.polar()` with a wedge-drawn series (`column`, `area`, `polygon`) | |
| `DOT` | `marker` series on an ordinal x scale | Promoted from `SCATTER`. Read as a scatter, every category label was coerced to `0` |
| `LOLLIPOP` | `stick` series | The series' markers are enabled and stamped — the stick itself is a tall thin stroke the marker filter rejects |
| `DUMBBELL` | `range-column`, `range-bar` series | Ends named `Low` / `High`, AnyChart's own field names; a caller with better names sets `startLabel` / `endLabel` |
| `DIVERGING` | two or more `bar` / `column` series, **opt-in** via `bindAnyChart(chart, { diverging: true })` | See below |

Detection support added to `src/adapters/anychart/types.ts`: `xScale()` / `yScale()` (with `stackMode()` and `getType()`), `dataMode()`, `drawTotalsAsAbsolute()`, and the `diverging` binder option.

### Highlighting

Each new family gets its own stamper and `data-maidr-anychart-*` attribute (`funnel-stage`, `word`, `flow`, `waterfall-step`, `spoke`, `tile`, plus the previously undocumented `pie-slice`), documented in the selector table in `docs/anychart.md`. Every one of them stamps **nothing at all** when the shape count disagrees with the data rather than stamping partially — a stamp past an unexpected mark names the wrong datum for everything after it.

### Deliberately NOT added, and why

- **`hilo` series → dumbbell.** Carries the same `low`/`high` pair `range-column` does, but AnyChart draws it as a bare `fill="none"` stroke. The filled-mark lookup every stamper here is built on cannot find it, and the only shapes left to distinguish it from grid lines and axis strokes are stroked paths too. It would announce correctly and never highlight, so it stays unread.
- **Diverging bars, inferred.** AnyChart has no diverging chart type. The idiom is a stacked `anychart.bar()` whose two series straddle zero, which is indistinguishable from a stacked bar chart that merely contains negative values. A diverging trace replaces the sign in every announcement with the name of a side — the one clue a reader would have that the guess was wrong — so it is declared, never inferred. A chart with fewer than two bar series keeps its ordinary bar layers and says why.
- **`bubble` series → dot plot.** Left as `SCATTER` even on an ordinal scale: its rows carry a size, and a dot plot has nowhere to put one.
- **Polar `rangeColumn` series.** Rows carry `low`/`high` and no `value`, so reading it as a radar would announce a magnitude of zero for every spoke. Skipped.
- **`step-area` → stepped area.** MAIDR has no stepped-area trace, so the staircase-with-a-fill has to give up one of the two. It keeps the staircase (that is what its values do) and maps to `STEP`, and is now the only series still emitting the fill-loss warning.
- **Marimekko `count`.** Not emitted; see the table note above.

### Examples and docs

Eleven runnable pages under `examples/anychart/`: `area`, `diverging`, `dotplot`, `dumbbell`, `funnel`, `lollipop`, `mosaic`, `radar`, `sankey`, `waterfall`, `wordcloud`. `docs/anychart.md` gains the new rows in the supported-types table, a detection note per family explaining what is read and from where, and the new selector attributes.

## Checklist

Two boxes are left unticked honestly: `CONTRIBUTING.md` was not read as part of this work, and no manual browser pass following `ManualTestingProcess.md` was performed. The automated gate below did all pass.

## Additional Notes

Full gate, run in order on this branch:

| Step | Result |
|---|---|
| `npm run lint:fix` | clean — no files changed |
| `npm run type-check` | clean (`tsc --noEmit` and `tsc -p tsconfig.ci.json`) |
| `npm run build` | all 12 example builds succeeded |
| `npm test` | unit/component: 187 suites, 2679 tests passed. ESM: 7 suites, 73 tests passed. No failures, pre-existing or otherwise |

`e2e` was not run.

The new coverage lives in `test/adapters/anychart/converters.test.ts` (+1993 lines), which exercises each new builder against a stubbed AnyChart instance and each stamper against a stubbed SVG, including the count-disagreement paths that must stamp nothing.

The three commits on the branch are grouped by how the detection works rather than by chart family — chart-level routes first, then collected multi-series layers, then the scale-derived promotions — so each one is reviewable on its own.

---


jooyoungseo:
Second review round. All 14 checks were green at `9922ea9`, and the automated review job exited `success` without producing a review — its comment is still stuck mid-checklist, so nothing on this PR has actually been reviewed by it. I went through the diff myself instead.

**Fixed** (`7f12588`): the stamped-attribute table in `docs/anychart.md` omitted `data-maidr-anychart-pair` altogether, and did not say which attribute a lollipop, a dot plot or a diverging bar reuses. That table is the reference a consumer uses to write their own `selectors`, so a stamped attribute missing from it is undiscoverable without reading the adapter source.

**One gap I did not fix, because it is outside this PR's file scope:** the AnyChart examples list in `scripts/build-site.js` (~L492) is a hardcoded `<ul>` naming only the original seven examples. The eleven new ones — area, diverging, dotplot, dumbbell, funnel, lollipop, mosaic, radar, sankey, waterfall, wordcloud — ship but are not linked from the gallery, so the `[Dot plot](examples.html)` style links in the docs table land on a page that does not list them. They are still reachable by direct URL. The one-line chart-type blurb at L64 of the same file is stale for the same reason. That file is shared with the other in-flight adapter branches, so it seemed better to flag it than to edit it here and create a conflict.

I checked the rest and found nothing to change: the waterfall `dataMode` / `drawTotalsAsAbsolute` accumulation, the scale-derived stacked-area and dot-plot promotions, and the opt-in diverging path are all correct and covered by tests, and every stamper degrades to a warning rather than a mispaired highlight when its DOM assumptions do not hold.

---

PR #878, opened 2026-08-14
feat(highcharts): support the chart types added since the pie chart

### LINKED ISSUE #862: feat(highcharts): support the chart types added since the pie chart
## Problem

MAIDR core has gained ~34 new trace types since the pie chart landed (radar, gauge, funnel, waterfall, gantt, treemap, sankey, sunburst, dumbbell, dot/lollipop, diverging, area family, and more). The `highcharts` adapter has not kept up: of those, it currently emits **none**.

A chart Highcharts draws perfectly well is therefore either dropped from the figure or announced as the wrong chart — which for a screen-reader user is worse than not being supported at all.

Dispatch is keyed entirely off the Highcharts *series type string* resolved by `resolveSeriesType` — `series.type || series.options.type || chart.options.chart?.type || 'line'` (adapter.ts:286-288) — and happens in two stages. First `buildSubplot` (adapter.ts:150) partitions the chart's usable series against two hardcoded string sets, `lineTypes = {line, spline, area, areaspline}` and `barTypes = {bar, column}` (adapter.ts:156-163): the bar bucket goes to `convertBarGroup` (adapter.ts:474), which picks BAR / STACKED / NORMALIZED / DODGED from series count plus `getStackingMode` (adapter.ts:303) at adapter.ts:503-516, and the line bucket is merged into ONE multi-series layer that becomes STEP when `stepDirectionOf` finds a `step` option and LINE otherwise (adapter.ts:743-750, 787). Everything left over goes one-series-per-layer through the `switch (seriesType)` in `convertSeries` (adapter.ts:699-724), whose cases are scatter, boxplot, heatmap, histogram, candlestick/ohlc and pie; the `default` branch warns `Unsupported series type` and returns null (adapter.ts:720-722). There is no inspection of the data itself — the only non-type inputs to the decision are `stacking`, `series.options.step`, and `chart.options.chart.inverted` (adapter.ts:495-500), so orientation and stacking are the sole modifiers of an otherwise 1:1 series-type → TraceType table.

## In scope — 25 trace types Highcharts can render

| TraceType | Highcharts construct | Detection / payload |
|---|---|---|
| `AREA` | `area` / `areaspline` series (Highcharts core) | `resolveSeriesType(s)` is `area`/`areaspline` AND `getStackingMode(s, chart)` is falsy. Today these strings sit in `lineTypes` (adapter.ts:156) so they are silently merged into the LINE layer — the fix is to pull them out of that set and give them their own bucket, not to add a `convertSeries` case. Payload is byte-identical to what `convertLineSeries` already builds (LinePoint[][] `{x: pointLabel(p), y: p.y, z: series.name}`, adapter.ts:767-775) with `type: TraceType.AREA`; keep `lineSelectors` (selectors.ts:86-90) since an area series still renders `path.highcharts-graph` (caveat: not when `lineWidth: 0`, where `path.highcharts-area` is the only path). Precedent in the Vega-Lite adapter at src/adapters/vegalite/converters.ts:677-689. |
| `STACKED_AREA` | `area`/`areaspline` with `stacking: 'normal'` | Same series types with `getStackingMode(...) === 'normal'`. `getStackingMode` (adapter.ts:303-319) only consults `plotOptions.column` / `plotOptions.bar` / `plotOptions.series`, so an `plotOptions.area.stacking` needs adding there (and `area`/`areaspline` keys to `HighchartsChart.options.plotOptions`, types.ts:63-67). Payload: LinePoint[][], one row per series, `y` = the series' OWN value `p.y` and NOT the accumulated edge — `AreaTrace` sums the rows itself (src/model/area.ts, class docstring "The total is computed here rather than expected from the producer"). Rows must be x-aligned across series the way `buildSegmentedRows` aligns bars (adapter.ts:568). |
| `NORMALIZED_AREA` | `area`/`areaspline` with `stacking: 'percent'` | Same detection with `getStackingMode(...) === 'percent'`. Payload: LinePoint[][] with `y = p.percentage ?? p.y` — `percentage` is already declared on `HighchartsPoint` (types.ts:125) and is exactly the field the NORMALIZED bar path reads at adapter.ts:623. |
| `DOT` | `scatter` on a category axis (Cleveland dot plot); also `lollipop` with `connectorWidth: 0` | `seriesType === 'scatter'` AND `series.xAxis?.categories` is non-empty. This is also a latent bug fix: `convertScatterSeries` (adapter.ts:799-821) emits `x: p.x`, i.e. the bare category INDEX, throwing away the label, because `ScatterPoint.x` is numeric (grammar.ts:880). Payload: BarPoint[] `{x: pointLabel(p), y: p.y}` — identical to `convertSingleBar` (adapter.ts:519-544), since `TraceType.DOT` routes to `BarTrace` (src/model/factory.ts, BAR/DOT/LOLLIPOP share a class). Reuse `scatterSelector` (selectors.ts:112) for the `:not([visibility="hidden"])` tracker filtering. |
| `LOLLIPOP` | `lollipop` series (modules/dumbbell.js) | New `case 'lollipop'` in the `convertSeries` switch (adapter.ts:706). Payload: BarPoint[] `{x: pointLabel(p), y: p.y}`, same as `convertSingleBar`. Selector: `barSelector` matches the marker `.highcharts-point`; the stem renders separately as `.highcharts-lollipop-stem` and is decoration MAIDR does not need. |
| `FUNNEL` | `funnel` / `pyramid` series (modules/funnel.js) | New switch cases. Payload: BarPoint[] `{x: p.name ?? p.category, y: p.y}` in declared stage order — `FunnelTrace extends BarTrace` (src/model/funnel.ts:43), and the retention arithmetic between adjacent stages is done in the model, so the adapter just supplies stage/count pairs. Axis naming should follow the pie precedent (`Label`/`Value`, adapter.ts:828-829) because a funnel series is bound to no axis. Selector: `pieSelector` shape — wedges are `.highcharts-point` paths in `series.data` order. |
| `WORD_CLOUD` | `wordcloud` series (modules/wordcloud.js) | New switch case. Point carries `name` and `weight`; add `weight?: number` to `HighchartsPoint` (types.ts:105-132). Payload: WordCloudPoint[] `{x: p.name, y: p.weight ?? p.y}` (grammar.ts:487-501). `WordCloudTrace` re-sorts by descending weight itself (src/model/wordCloud.ts:74), so declared order is fine. Selector: each word is a `<text class="highcharts-point">` in data order, so `barSelector` works unchanged. |
| `SANKEY` | `sankey` series (modules/sankey.js) | New switch case. Sankey link points carry `from` / `to` / `weight`; add those to `HighchartsPoint`. Payload: FlowPoint[] `{source: p.from, target: p.to, value: p.weight}` (grammar.ts:824-831). Do NOT read `series.nodes` — MAIDR derives nodes from the edges by design (grammar.ts:814-820). Note the pane-detection guard at adapter.ts:350 skips axis-less series, so a sankey never joins a pane grid; that is correct, not a bug. |
| `CHORD` | `dependencywheel` series (modules/dependency-wheel.js, needs sankey); `arcdiagram` is the same data shape | New switch cases sharing the sankey converter — the point shape is identical (`from`/`to`/`weight`) and `TraceType.CHORD`, `SANKEY` and `ALLUVIAL` all route to the same `FlowTrace` (src/model/factory.ts). Only the emitted `type` differs, so the chart announces itself as the form the author drew. |
| `NETWORK` | `networkgraph` series (modules/networkgraph.js) | New switch case. Link points carry `from`/`to`. Payload: NetworkPoint[] `{source: p.from, target: p.to}` (grammar.ts:848-853) — deliberately no positions, so the force-solver layout is ignored, which is what the grammar asks for. |
| `TREEMAP` | `treemap` series (modules/treemap.js) | New switch case. Highcharts declares the hierarchy with `id` / `parent` pointers on each point; MAIDR declares it as a PATH (`TreemapPoint.path` = ancestors root-first, excluding self, grammar.ts:793-811). So the converter builds an `id → point` map, then walks `parent` upward per point to materialise the path. Payload: TreemapPoint[] `{x: p.name ?? p.id, y: p.value, path: [...ancestorNames]}`; interior nodes may be dropped since `TreemapTrace` derives them and their totals from the paths. Needs `id`/`parent`/`value` on `HighchartsPoint`. Guard against a cyclic/dangling `parent` (Highcharts tolerates a parent id that was never emitted; the path walk must not loop). |
| `SUNBURST` | `sunburst` series (modules/sunburst.js, needs treemap) | Same `id`/`parent`/`value` point shape as treemap, so it reuses the treemap converter verbatim and only changes the emitted `type` to `TraceType.SUNBURST` (both route to `TreemapTrace`, which reads `this.type` for the sunburst's angular panning). Effort is small *given* the treemap work; on its own it is the same medium job. |
| `GAUGE` | `gauge` / `solidgauge` (highcharts-more, modules/solid-gauge.js) and `bullet` (modules/bullet.js) | New switch cases. Payload is a SINGLE object, not an array (grammar.ts:518-543): `{value: series.data[0].y, min, max}` from `series.yAxis.getExtremes()` (already on `HighchartsAxis`, types.ts:139), `label: series.name`, `target: point.target` for a bullet, and `bands` mapped from `yAxis.options.plotBands` → `{to: band.to, label: band.label?.text ?? band.className}` in ascending order. Requires adding `plotBands` to `HighchartsAxis['options']` (types.ts:149-152) and `target` to `HighchartsPoint`. Trap: a gauge/solid-gauge lives in a `pane`, and a chart may hold several gauges — each is its own layer, and the pane-grid detector (adapter.ts:346) will not group them since it keys on axis geometry. |
| `WATERFALL` | `waterfall` series (highcharts-more) | New switch case (must ALSO be kept out of `barTypes`, adapter.ts:157, if a user sets `chart.type: 'waterfall'` — `resolveSeriesType` would then return `waterfall`, so no conflict, but a mixed column+waterfall chart must not fuse). Highcharts gives each point a signed delta in `p.y` plus `isSum` / `isIntermediateSum` flags on `p.options`. The adapter accumulates a running total to fill MAIDR's absolute positions: for an ordinary step `start = running`, `end = running + p.y`, `delta = p.y`, `kind = p.y >= 0 ? 'increase' : 'decrease'`; for a sum/intermediate-sum step `kind = 'total'`, `end = running`, `delta = 0` and the bar sits on the baseline (grammar.ts:445-477). Needs `isSum`/`isIntermediateSum` on the point options type. |
| `ERROR_BAR` | `errorbar` series (highcharts-more) | `seriesType === 'errorbar'`. The interval is on the errorbar points (`p.low` / `p.high`, already declared at types.ts:110-113) but the ESTIMATE lives in the linked parent series — `series.options.linkedTo` / `series.linkedParent` — normally a `column` or `scatter`. The converter resolves the parent, zips by x, and emits a FLAT ErrorBarPoint[] `{x: category, y: parentPoint.y, yMin: p.low, yMax: p.high}` (absolute bounds, not offsets — grammar.ts:370-378). Two structural edits beyond the switch: `linkedTo`/`linkedParent` on `HighchartsSeries['options']`, and suppressing the parent column series from also being emitted as its own BAR layer inside `convertBarGroup` (adapter.ts:474). An unlinked errorbar series (no parent) can still emit with `y` taken as the midpoint of low/high, or be skipped. |
| `DUMBBELL` | `dumbbell` series (modules/dumbbell.js) | New switch case; points carry `low` and `high` (already typed). Payload is an OBJECT, not an array (grammar.ts:651-658): `{points: [{x: pointLabel(p), start: p.low, end: p.high}], startLabel, endLabel}`. Highcharts has no field naming the two ends — `lowColor`/`marker` say nothing — so `startLabel`/`endLabel` should either be omitted (the trace then says "start"/"end") or exposed through a new field on `HighchartsAdapterOptions` (types.ts:12-19). Do not synthesise a delta; `DumbbellTrace` derives it (grammar.ts:624-628). |
| `RADAR` | polar chart: `chart.polar: true` with `line` / `spline` / `area` series (highcharts-more) | `chart.options.chart?.polar === true` — needs `polar?: boolean` added to `HighchartsChart['options']['chart']` (types.ts:62, which today carries only `{type, inverted}`). Because the series type is still `line`, the branch must be taken in `buildSubplot` BEFORE the `lineTypes` partition (adapter.ts:156-159), otherwise the spokes are merged into an ordinary LINE layer. Payload: LinePoint[][], one row per series, `x` = spoke name from `series.xAxis.categories`, `y` = value (`RadarTrace extends LineTrace`, src/model/radar.ts:47) — so `convertLineSeries` can be reused with the type swapped. |
| `POLAR_AREA` | polar `column` series — Highcharts' wind-rose / coxcomb (highcharts-more) | Same `chart.polar === true` flag with `resolveSeriesType` of `column`/`bar`; must branch before `convertBarGroup` is reached (adapter.ts:168) or the wedges become a DODGED layer. Payload identical to RADAR (LinePoint[][], both route to `RadarTrace`), only `type` differs. |
| `GANTT` | `gantt` series (Highcharts Gantt) or `xrange` series (modules/xrange.js) | New switch cases. Each point carries `x` (start), `x2` (end), `y` (lane index) and `name`; add `x2` and `partialFill` to `HighchartsPoint` (note `x2` is already read opportunistically for histograms at adapter.ts:1236). Payload is an OBJECT (grammar.ts:579-613): `{points: GanttPoint[][] grouped by lane index so an EMPTY lane still exists as an empty row, lanes: series.yAxis.categories, unit: derived from the axis — 'ms' for a datetime axis, otherwise omitted}`. Each GanttPoint is `{x: laneCategory, start: p.x, end: p.x2, label: p.name}`. Trap: nesting by lane is required, and lanes with no tasks must be preserved as empty arrays — that is the whole reason the shape is nested. |
| `PARALLEL` | parallel coordinates: `chart.parallelCoordinates: true` with one `line` series per observation (modules/parallel-coordinates.js) | `chart.options.chart?.parallelCoordinates === true` (new field on types.ts:62); like RADAR it must branch before the `lineTypes` merge (adapter.ts:156). In this mode `chart.yAxis` is an ARRAY with one axis per dimension and each series is one observation. Payload: LinePoint[][], one row per series (observation), one column per dimension, with `x` = that column's axis name (`chart.yAxis[i].options.title.text`) and `y` = the raw value. `ParallelTrace` computes per-column extents itself (src/model/parallel.ts:60-64), so no normalisation in the adapter. |
| `DIVERGING` | population pyramid / Likert: two `bar` (or inverted `column`) series over shared categories, one series' data negated, usually `stacking: 'normal'` | Highcharts has no diverging series type, so detection is a data heuristic inside `convertBarGroup` (adapter.ts:474): every value in one series ≤ 0 and every value in another ≥ 0 over the same category axis. Payload: SegmentedPoint[][] exactly as `buildSegmentedRows` already builds (adapter.ts:568-631) but with the SIGN preserved — `DivergingTrace extends SegmentedTrace` and reads the sign as a direction (grammar.ts:1323-1331). Because the heuristic can misfire on a legitimate stacked chart containing negatives, this is safer as an opt-in on `HighchartsAdapterOptions` than as automatic detection. |
| `BUMP` | `line`/`spline` series over a rank axis, typically `yAxis.reversed: true` (Highcharts' own bump-chart demo pattern) | No dedicated series type. Heuristic: `series.yAxis.options.reversed === true` AND at every x the values across series are a permutation of 1..n (integers, no duplicates). Payload: LinePoint[][] with `y` = the rank (`BumpTrace extends LineTrace`, src/model/bump.ts:50); the pitch inversion and places-gained announcement are all model-side. Given the false-positive risk on any reversed-axis line chart, prefer an explicit opt-in field on `HighchartsAdapterOptions` (types.ts:12-19) with the heuristic as a fallback. |
| `HEXBIN` | `tilemap` series with `tileShape: 'hexagon'` (modules/tilemap.js, needs modules/heatmap.js) | `seriesType === 'tilemap'` and `series.options.tileShape` is `hexagon` (the default) rather than `square`/`circle`/`diamond`. Payload: HexbinPoint[][] rows of `{x, y, count: p.options.value}` (grammar.ts:677-684) where the x centre of odd rows is shifted by half a cell, reproducing tilemap's own stagger, so the declared centres are real positions rather than lattice indices — that stagger is precisely what `HexbinTrace` navigates by. Honest caveat: Highcharts never bins anything; the user supplies pre-binned tiles, so this reads a hex-lattice display, not a computed hexbin. |
| `VOLCANO` | ordinary `scatter` series with `yAxis.plotLines` for the significance cutoff | Highcharts ships no volcano series, so nothing in the chart object distinguishes it from any scatter — this needs author intent through a new `HighchartsAdapterOptions` field (e.g. `traceTypes: { [seriesIndex]: TraceType.VOLCANO }`, types.ts:12-19). Given that, the payload is cheap: VolcanoPoint[] `{x, y, label: p.name, group: p.options.group ?? series.name}` (grammar.ts:895-912) plus `thresholdOptions.significance` read from `yAxis.options.plotLines[].value` and `effect` from an `xAxis` plotLine magnitude. `significanceDirection` must be declared, never guessed (grammar.ts:928-940). |
| `MANHATTAN` | ordinary `scatter` series, one series per chromosome, with a genome-wide `plotLine` | Same opt-in mechanism as VOLCANO — the two share `VolcanoTrace` (src/model/factory.ts) and the same `VolcanoPoint[]` payload; the difference is that `group` is filled from the per-chromosome series name and the several scatter series must be concatenated into ONE layer rather than becoming one layer each (today `convertSeries` emits one layer per non-bar/non-line series, adapter.ts:173-178). |

## Deferred to a follow-up

These need a new author-facing opt-in API or semantics Highcharts does not carry in its own configuration, so they are out of scope here and tracked separately:

- `CHOROPLETH` — `seriesType === 'map'` (or `mapbubble`/`mappoint` siblings, which are not this). Payload: ChoroplethPoint[] `{x: p.name, y: p.value}` plus `lon`/`lat` — and the grammar is strict that these are DEGREES, never projected units (grammar.ts:741-750). Highcharts mapdata features expose `properties['hc-mi
- `SURVIVAL` — Indistinguishable from any step chart by inspection — the adapter already turns it into a STEP layer (adapter.ts:787) and would need an opt-in to do better. Payload: SurvivalPoint[][] (nested, one row per curve, src/model/survival.ts:62) `{x: time, y: survival, censored: true where the companion sca
- `FOREST` — Builds on the ERROR_BAR plumbing (linked-series pairing) plus a category y-axis and `chart.inverted`. Two of the three things that make it a forest plot are not recoverable: `weight` is drawn as marker AREA, so it would have to be inverted out of `point.marker.radius` (fragile, and absent when the a

## Not renderable by Highcharts

<details><summary>6 trace types the library genuinely cannot draw</summary>

- ALLUVIAL — Highcharts ships no alluvial series. Its `sankey` and `dependencywheel` cover the weighted-flow forms, and nothing in a Highcharts chart distinguishes an alluvial layout from an ordinary sankey, so the adapter would be inventing the distinction.
- BOXEN — no letter-value / boxen series. The `boxplot` series (highcharts-more) is fixed at a five-number summary, which is exactly the depth a boxen exists to exceed (grammar.ts:304-312), and there is no per-point quantile ladder to read.
- CONTOUR — no contour/isoline series in Highcharts or any of its official modules. Nothing draws curves of constant value over a scalar field, so there is no `level` for the adapter to recover.
- ICICLE — no icicle series. `treemap` and `sunburst` cover the hierarchy; `treegraph`/`organization` draw node-link trees, not depth-ordered bands, so nothing on the page is an icicle.
- MOSAIC — no marimekko/mosaic series. `variwide` (modules/variwide.js) encodes a second variable as column WIDTH but does not stack, and stacked columns have no per-column width control, so the width-plus-stack combination a mosaic is defined by cannot be drawn.
- RIDGELINE — no ridgeline/joyplot series. The only way to draw one is to bake a per-group vertical offset into the data, and `RidgelineTrace` requires each group's curve on its own terms with the offset removed (grammar.ts:1462-1470); the adapter cannot subtract an offset it was never told about.

</details>

## Acceptance criteria

- [ ] Unit tests under `test/adapters/highcharts/` cover every newly detected type.

## Implementation notes

Nothing post-pie is emitted today — the adapter's full output set is BAR, DODGED, STACKED, NORMALIZED, LINE, STEP, SCATTER, PIE, BOX, HEATMAP, HISTOGRAM, CANDLESTICK (every `TraceType.*` use in the directory is at adapter.ts:509, 512, 537, 574, 644, 677, 686, 787, 812, 853, 908, 1185, 1249, 1301). The docs table at docs/highcharts.md:102-117 states the same, and it explicitly lists `area`/`areaspline` under **Line** (docs/highcharts.md:107) — so an area chart is not merely unsupported, it is actively read as a line today, and a stacked one announces one magnitude where the chart draws two.

TWO DISPATCH SITES, NOT ONE. The `convertSeries` switch (adapter.ts:706) is only reached by series that survived the `lineTypes`/`barTypes` partition in `buildSubplot` (adapter.ts:156-163). Any new type whose Highcharts series type is `line`/`spline`/`area`/`areaspline`/`bar`/`column` — that is AREA, STACKED_AREA, NORMALIZED_AREA, RADAR, POLAR_AREA, DIVERGING, BUMP, SURVIVAL — must be split out at the partition, before the merge, or it will be silently fused into the shared LINE or DODGED layer. Only genuinely new series-type strings (funnel, sankey, treemap, wordcloud, gauge, waterfall, errorbar, dumbbell, lollipop, tilemap, xrange, map, networkgraph, sunburst, dependencywheel) are pure `convertSeries` additions.

`types.ts` IS THE OTHER BOTTLENECK. It is a hand-written minimal structural subset (types.ts:50-153), not the Highcharts typings: `chart.options.chart` carries only `{type, inverted}` (types.ts:62) so `polar` and `parallelCoordinates` do not exist; `plotOptions` carries only `series`/`column`/`bar` stacking (types.ts:63-67) so area stacking is invisible to `getStackingMode` (adapter.ts:303-319); `HighchartsPoint` (types.ts:105-132) has no `from`/`to`/`weight`/`id`/`parent`/`value`/`x2`/`target`/`isSum`; `HighchartsAxis['options']` (types.ts:149-152) has no `plotBands`/`plotLines`/`reversed`; `HighchartsSeries['options']` has no `linkedTo`. Almost every item above needs a few fields added here first.

SELECTOR CONVENTION AND THE EXPENSIVE PATTERN TO AVOID. Every selector is scoped `#{containerId} .highcharts-series-group .highcharts-series-N …` specifically to exclude the legend swatch, which also carries `highcharts-series-N` and contains a `.highcharts-point` (selectors.ts:1-21, 60-77). Node-per-point types (funnel wedges, sankey nodes/links, treemap tiles, sunburst arcs, wordcloud words, lollipop markers) render one `.highcharts-point` per datum in `series.data` order, so `barSelector`/`pieSelector` (selectors.ts:75, 127) work unmodified — those are the cheap ones. The expensive precedent is a compound single-`<path>` mark: boxplot and candlestick required DOM stamping plus splitting a multi-subpath `d` into sibling paths with a `MutationObserver` to resurvive Highcharts redraws (adapter.ts:928-1090 and 1321-1477, ~350 lines). Prefer types whose marks are already one element per datum; a gauge dial or a waterfall connector will need some of that treatment.

PANE DETECTION. `detectPaneGrid` bails to `null` whenever any series lacks `xAxis`/`yAxis` (adapter.ts:350-352). Axis-less new types (funnel, sankey, treemap, sunburst, wordcloud, networkgraph, and gauges on a `pane`) therefore force the whole chart onto the single-subplot path — correct behaviour, but worth knowing before someone files it as a regression. The subplot's own `selector` (adapter.ts:222-224, selectors.ts:60) uses the first series' group and is type-agnostic, so it needs no change.

TESTING AND EXAMPLES. Jest fakes live in test/adapters/highcharts/helpers.ts — `fakeChart` only sets `options.chart.type` (helpers.ts, `fakeChart`), so any test for polar / parallelCoordinates / plotBands needs the helper extended first. Per-type test files follow test/adapters/highcharts/pie.test.ts as the model (assert `layer.type`, `layer.data`, `layer.selectors`, `layer.axes` labels). There are 13 examples/highcharts-*.html demos, one per supported type, and they load Highcharts from CDN — core, plus highcharts-more.js (box), modules/heatmap.js, modules/histogram-bellcurve.js and highstock.js (candlestick). The adapter never imports Highcharts (index.ts:11-14), so any official module the *user* loads is fair game; the module column in docs/highcharts.md:102-117 is how the repo already communicates that requirement per type. There is no Playwright coverage for Highcharts at all (e2e_tests/specs has no highcharts spec).

PRIOR ART WORTH COPYING. The area family is already solved once, in the Vega-Lite adapter's mark→type mapping (src/adapters/vegalite/converters.ts:677-689) and its line-family merge (converters.ts:180-217); Chart.js also emits AREA/STACKED_AREA (src/adapters/chartjs/extractor.ts, highlightTargets.ts). Those are the only two adapters in the repo emitting any post-pie type, so for everything else this adapter would be first.

ONE UNRELATED CORRECTNESS NOTE FOUND WHILE SURVEYING: `convertScatterSeries` emits `x: p.x` (adapter.ts:806), which on a category axis is the bare index — the category label is dropped. That is the same bug the DOT item above fixes, and it affects existing scatter output regardless of whether DOT is added.

jooyoungseo:
# Pull Request

## Description

The Highcharts adapter stopped growing at the pie chart: it read bar, line, step, scatter, box, heatmap, histogram, candlestick, pie and the stacked/dodged/normalized bar variants, while MAIDR's grammar has kept adding trace types since. This adds 25 of them.

Most are a new `switch` case over a Highcharts series type. The interesting ones are not:

- Six readings are **decided from context**, because Highcharts draws them with a series type it shares with something else — a `scatter` on a category axis is a Cleveland dot plot, `chart.polar` turns `line` series into radar spokes and `column` series into a wind rose, `chart.parallelCoordinates` turns everything into parallel coordinates, two stacked bar series growing opposite ways are a diverging bar, and a `line` series carrying a rank permutation on a reversed axis is a bump chart.
- Two are **declared by the caller** — a volcano and a Manhattan plot are both an ordinary `scatter` with the cutoff drawn as a `plotLine`, and nothing in the chart object separates either from a scatter of two variables, so the reading is opted into with `significancePlot` rather than guessed.

Every trace type is wired end to end in the same change: detection, payload, **and** the selector/highlight path. Where highlighting cannot be made safe — a word cloud term Highcharts could not fit, a treemap node it did not draw, a hexbin tile missing from a staggered lattice — the layer withdraws highlighting rather than pairing announcements with the wrong elements.

No file outside `src/adapters/highcharts/`, `test/adapters/highcharts/`, `examples/` and `docs/` is touched. The core already supported every trace type here; `src/type/grammar.ts` and the trace models are unchanged.

## Related Issues

Closes #862

### Trace types added (25)

Series types that map straight through:

| MAIDR trace type | Read from |
|---|---|
| `AREA` | `area`, `areaspline` |
| `STACKED_AREA` | `area`/`areaspline` + `stacking: 'normal'` |
| `NORMALIZED_AREA` | `area`/`areaspline` + `stacking: 'percent'` |
| `LOLLIPOP` | `lollipop` |
| `FUNNEL` | `funnel`, `pyramid` |
| `WORD_CLOUD` | `wordcloud` |
| `SANKEY` | `sankey`, `arcdiagram` |
| `CHORD` | `dependencywheel` |
| `NETWORK` | `networkgraph` |
| `TREEMAP` | `treemap` |
| `SUNBURST` | `sunburst` |
| `GAUGE` | `gauge`, `solidgauge`, `bullet` |
| `WATERFALL` | `waterfall` |
| `ERROR_BAR` | `errorbar`, zipped with the series it is `linkedTo` |
| `DUMBBELL` | `dumbbell` |
| `GANTT` | `gantt`, `xrange` |
| `HEXBIN` | `tilemap` with a tessellating `tileShape` |

Readings decided from chart context rather than the series type:

| MAIDR trace type | Recognised from |
|---|---|
| `DOT` | `scatter` on a category x axis |
| `RADAR` | `chart.polar` + `line`/`spline`/`area` series |
| `POLAR_AREA` | `chart.polar` + `column`/`bar` series |
| `PARALLEL` | `chart.parallelCoordinates` |
| `DIVERGING` | exactly two stacked `column`/`bar` series, one entirely at or below zero and one entirely at or above it |
| `BUMP` | `line`/`spline` on a `reversed` y axis whose values are a rank permutation at every period; forced or suppressed with `options.bump` |

Readings the caller opts into:

| MAIDR trace type | Declared with |
|---|---|
| `VOLCANO` | `significancePlot: { type: 'volcano' }` |
| `MANHATTAN` | `significancePlot: { type: 'manhattan' }` — the per-chromosome scatter series become one layer under one threshold |

### Corrections to existing readings

- `area`/`areaspline` sat in the line-type set, so a filled band announced itself as a line and a stacked one announced one magnitude where the chart draws two.
- A `scatter` on a category axis was read as a scatter, and `ScatterPoint.x` is strictly numeric, so the label the chart prints was dropped in favour of the bare tick index.

### New adapter options

- `bump?: boolean` — force or suppress the bump reading.
- `significancePlot?: { type, seriesIndices?, significance?, significanceDirection?, effect? }` — read the panel's scatter series as one volcano or Manhattan layer. Cutoffs default to the `plotLines` the chart already draws.
- (`dumbbellLabels?: { start, end }` also landed here, naming the two ends of a dumbbell pair, which Highcharts leaves anonymous.)

None of these were skipped for a missing core feature — the core supports them all. Each is absent because Highcharts has nothing to read:

- **`ALLUVIAL`** — Highcharts ships no alluvial series. Its closest series, `sankey`, is already read as `SANKEY`; emitting the same points twice under a second name would be a labelling choice, not a reading.
- **`BOXEN`** — no letter-value series. A `boxplot` carries five numbers, not the nested quantile fan a boxen needs.
- **`CHOROPLETH`** — needs Highcharts Maps (`map`/`mappoint` series and a GeoJSON projection), a separate product from the Highcharts core this adapter reads.
- **`CONTOUR`** — no contour series.
- **`FOREST`** — no forest-plot series. It is drawable as an `errorbar` pattern, but distinguishing one from the error bars now read as `ERROR_BAR` would need a caller-declared opt-in on the `significancePlot` model; deferred rather than guessed.
- **`ICICLE`** — no icicle series; Highcharts covers the hierarchy family with `treemap` and `sunburst`, both read here.
- **`MOSAIC`** — no mosaic/marimekko series.
- **`RIDGELINE`** — no ridgeline series.
- **`SMOOTH`** — no LOESS/regression smoother series. A smoothed curve is authored as a plain `line`/`spline` and is indistinguishable from one, so this would need an opt-in too.
- **`SURVIVAL`** — no Kaplan-Meier series. A survival curve is drawn as a `line` with `step`, which is already read as `STEP`; promoting every step chart to a survival curve would be wrong far more often than right.
- **`VIOLIN_KDE`, `VIOLIN_BOX`** — no violin series.
- **`CANDLESTICK_DELTA`** — not an adapter-emitted type. It is a reading mode the model applies to a `CANDLESTICK` layer (`src/service/candlestickDelta.ts`), and the adapter already emits `CANDLESTICK`.

### Tests, examples and docs

- 19 new Jest suites under `test/adapters/highcharts/` covering payload shape, ordering and the selector/highlight path for every trace type above.
- 22 new runnable pages under `examples/`, one per new chart form.
- `docs/highcharts.md` gains a row per trace type, the two new options, and a note per family recording the reading decisions that are not obvious from the code — the stagger that never reaches a hexbin point, why word cloud terms are emitted heaviest first, why sankey nodes are derived from links rather than read from `series.nodes`, and why `significanceDirection` must be stated for a raw-p axis.

## Checklist

## Additional Notes

**On the unticked checklist box.** The automated gate is fully green on this branch: `npm run lint:fix` (no changes), `npm run type-check`, `npm run build` and `npm test` (213 suites, 2787 tests, all passing). The box is left unticked only because `ManualTestingProcess.md` does not exist in the repository and no manual screen-reader pass was performed — the new example pages under `examples/` are the intended vehicle for that, and a reviewer with a screen reader should drive them before merge.

**Known gaps, deliberately left for follow-up.**

- There is no Playwright coverage for Highcharts at all — `e2e_tests/specs/` has no Highcharts spec, and this PR does not add one. Every trace type here is covered by unit tests against a mocked chart object, not against a real rendered Highcharts instance.
- Two readings are heuristics over data rather than declarations, and are documented as such: `DIVERGING` (two stacked series on opposite sides of the baseline) and `BUMP` (a rank permutation on a reversed axis). The bump test is deliberately narrow because the failure is not a degraded reading — MAIDR inverts the pitch for a bump chart, so an ordinary line chart read as one sonifies upside down. `options.bump` exists to override the guess in both directions.

**A spec deviation worth a reviewer's attention.** For hexbin, the adapter emits the *authored* tile centres rather than shifting alternate rows by half a cell. Highcharts applies that stagger in pixel space inside `translate`, so it never reaches the point — the tooltip, the axis and every other surface still report the authored integer coordinate. Announcing a shifted `1.5` would put the bin somewhere nothing else on the page agrees with. `HexbinTrace` still navigates correctly, since its vertical move resolves by nearest x and integer lattice coordinates land on the same column.

---


jooyoungseo:
Round 2 review pass.

**Heads-up on the review status:** the `claude-review` job exited `success`, but its comment is stuck mid-checklist — `adapter.ts`, `selectors.ts` and test coverage are all still marked in-progress and no findings were ever posted. The green check on that job is not an approval; this PR has not actually been machine-reviewed. The only other review on the PR is the Copilot quota notice. Worth a human pass on `adapter.ts` (+2188) before merge.

**Fixed:** `dumbbellLabels` was described in the dumbbell prose note but missing from the `highchartsToMaidr` options table, which otherwise lists every option the adapter reads — including the other two this branch adds (`bump`, `significancePlot`). Added it (`95f13a8`, docs-only).

**Verified rather than changed:** all 14 checks were green at `f076721`; locally, 23 highcharts suites / 139 tests pass, and type-check, build and the docs site build are clean. Every `highcharts-*.html` link in the supported-types table resolves to a file that exists.

**Noted, not fixed:** the Highcharts list in the site examples gallery (`scripts/build-site.js`) is hardcoded and still names only the original 11 chart types, so the 22 example pages added here ship but aren't linked from the gallery. That's a pre-existing pattern — `highcharts-grid.html` and `highcharts-panes.html` are already unlisted on `main` — and `scripts/` is outside this branch's scope, so I left it alone rather than widening the diff. Happy to do it as a follow-up if you'd rather the gallery stay complete.

---

jooyoungseo:
Round 3. The review that landed on `f076721` was a real one this time, and every finding in it is now either fixed or answered.

**Fixed**

- `1d4ef6d` — the Manhattan threshold bug. `significanceThresholds` inferred `effect` from `xAxis.plotLines` whatever the declared type, so a Manhattan drawing chromosome dividers got an effect cutoff in the hundreds of millions, `VolcanoTrace.clearsThreshold` rejected essentially every point, and the significance summary and the significant-points rotor filter both went silent. The guess is now a volcano's alone; a Manhattan gets an `effect` only by stating one. Two tests: dividers on the x axis must not become a cutoff, and a stated cutoff must still be honoured.
- `fb2f2cb` — the bump false positive. Every period of a one-competitor table is a permutation of `1..1`, so two series over disjoint x categories passed the rank test with no rank ever contested. At least one period must now rank two competitors. Regression test included — this one matters because MAIDR inverts the pitch for a bump, so a line chart admitted here sonifies upside down.
- `7730a76` — Gantt lane count via `reduce` rather than spreading the series into `Math.max`.
- `6722ae5` — the docs links. `docs/highcharts.md` renders to `_site/highcharts.html` at the site root while `examples/` is copied to `_site/examples/`, so all 39 bare `highcharts-*.html` links 404'd, old rows included. All now go through `examples/`, and I checked every target resolves in a real `npm run docs` build.
- `66569ce` — the undrawn-mark coverage gap you flagged. `stampPointIndices` skips rather than compacts, but only all-points-drawn fixtures exercised it; a partial-`graphic` hexbin case now pins that, since compacting would slide every later bin's highlight onto its neighbour.

**Declined**

- **`scripts/build-site.js` gallery.** `scripts/` is outside this branch's scope and the omission is pre-existing (`highcharts-grid.html` and `highcharts-panes.html` are already unlisted on `main`), so widening the diff to a shared build script on a +6k adapter PR isn't the right trade. The underlying concern — that the new examples are unreachable — is now largely met by the docs fix above: every one of the 35 example pages is linked, correctly, from the Highcharts guide. Happy to do the gallery as a follow-up.
- **`domMapping: { order: 'row' }` on `convertStackedBar`/`convertDodgedBar`.** Pre-existing and untouched by this diff. It's a behaviour change to two long-shipped readings and belongs in its own change with its own tests, not bundled into this one.
- **Collapsing the five `LinePoint[][]` converters.** They differ in per-type detail, and a structural refactor of the payload path is the wrong risk to take on a diff this size for no behaviour gain.

**On the review status.** The `claude-review` job on this head again exited `success` with its comment stuck mid-checklist — "Check test coverage" and "Post final review feedback" are still unticked and no fresh findings were posted. It did confirm the four fix commits are correct and tested, but that is not a completed review and the green check is not an approval. The substantive review remains the one on `f076721`; a human pass over `adapter.ts` is still worth having before merge.

All 14 checks are green at `66569ce`. `main` has moved 7 commits since this branch forked; `git merge-tree` against it is clean, so I've left the history alone rather than merging for tidiness. Locally: 213 suites / 2791 tests pass, and lint, type-check, build and the docs build are clean.

---

PR #880, opened 2026-08-14
feat(d3): support the chart types added since the pie chart

### LINKED ISSUE #865: feat(d3): support the chart types added since the pie chart
## Problem

MAIDR core has gained ~34 new trace types since the pie chart landed (radar, gauge, funnel, waterfall, gantt, treemap, sankey, sunburst, dumbbell, dot/lollipop, diverging, area family, and more). The `d3` adapter has not kept up: of those, it currently emits **none**.

A chart D3 draws perfectly well is therefore either dropped from the figure or announced as the wrong chart — which for a screen-reader user is worse than not being supported at all.

There is no detection: the D3 adapter never inspects the chart to decide a TraceType, because D3 has no chart-type concept to inspect. The developer picks the type by calling a specific binder, and each binder's pure extraction core hardcodes a TraceType constant onto the layer it builds — `type: TraceType.BAR` at src/adapters/d3/binders/bar.ts:117, `TraceType.LINE` at binders/line.ts:290, `TraceType.PIE` at binders/pie.ts:211, `TraceType.SCATTER` at binders/scatter.ts:105, `TraceType.BOX` at binders/box.ts:246, `TraceType.HEATMAP` at binders/heatmap.ts:169, `TraceType.HISTOGRAM` at binders/histogram.ts:129, `TraceType.CANDLESTICK` at binders/candlestick.ts:126, `TraceType.SMOOTH` at binders/smooth.ts:127. The single exception is the segmented binder, which lets the caller choose among exactly three types via `config.type` (default `TraceType.STACKED` at binders/segmented.ts:85, emitted at binders/segmented.ts:290), constrained by the `SegmentedTraceType` union at src/adapters/d3/types.ts:232-235. For multi-panel figures a string discriminator is switched to the matching builder — `buildPanelLayer`'s `switch (spec.chartType)` at src/adapters/d3/binders/subplots.ts:231-253 — and the React path does the same in `runBinder` at src/adapters/d3/useD3Adapter.ts:116-141 (mirrored in `withFacetsAutoApplyOff` at useD3Adapter.ts:150-171 and `MaidrD3.tsx:148-172`). All actual data comes from D3's `__data__` expando read by `getD3Datum` (src/adapters/d3/util.ts:29-31) over user-supplied CSS selectors, resolved through user-supplied or key-inferred accessors (`resolveAccessor` util.ts:95, `inferAccessor` util.ts:142); the adapter imports no d3 code at all, so the "library" it supports is whatever the user drew.

## In scope — 33 trace types D3 can render

| TraceType | D3 construct | Detection / payload |
|---|---|---|
| `DOT` | Cleveland dot plot — `selectAll('circle').data(rows).join('circle')` with a band scale | No detection is possible or needed in this adapter: add `bindD3Dot` (or a `mark: 'dot'` option on D3BarConfig) plus a `chartType: 'dot'` arm in the three switches. Payload is BarPoint[] `{x, y}` — byte-identical to what buildBarLayer already produces at binders/bar.ts:105-113; only the TraceType constant differs, and BarTrace already serves DOT (src/model/factory.ts:66-69). |
| `LOLLIPOP` | Lollipop chart — one `<line>` stem plus one `<circle>` head per category | Same as DOT: reuse buildBarLayer verbatim, emit TraceType.LOLLIPOP, and point `selector` at the circles (or a `<g>` per lollipop) for highlighting. Payload BarPoint[] `{x, y}`; BarTrace already handles it (factory.ts:66-69). |
| `FUNNEL` | Funnel / stage chart — trapezoid `<path>` or centred `<rect>` per stage | New `bindD3Funnel` reading `{stage, count}` off each stage element's `__data__` via the existing bar accessors. Payload BarPoint[] in draw order (stage order is load-bearing — FunnelTrace pitches retention between adjacent points). FunnelTrace extends BarTrace, so no model work. |
| `AREA` | `d3.area()` filled band, one `<path>` per series | New `bindD3Area` = buildLineLayer with `TraceType.AREA`: the datum bound to an area `<path>` is the same point array a line path carries, so the existing dual extraction (per-path `__data__` array, or `pointSelector` + fill grouping, binders/line.ts:123-226) works unchanged. Payload LinePoint[][] plus `legend`. AreaTrace extends LineTrace (src/model/area.ts). |
| `STACKED_AREA` | Stacked area / streamgraph — `d3.stack()` + `d3.area().y0(d=>d[0]).y1(d=>d[1])` | Same binder as AREA with `type: 'stacked_area'`, plus the d3.stack() unwrapping segmented.ts already models: each series `<path>` datum is an array of `[y0,y1]` tuples with a `.data` back-reference and the parent group's datum carries `.key`. Emit `y = d[1]-d[0]` (the series' own value — AreaTrace re-derives the running total itself) and `z = key`. Reuse the `groupSelector` idea and the friendly d3.stack() error at binders/segmented.ts:168-209. |
| `NORMALIZED_AREA` | 100% stacked area — `d3.stack().offset(d3.stackOffsetExpand)` | Identical to STACKED_AREA with `type: 'stacked_normalized_area'`; the only difference is the type constant, since AreaTrace reads the variant off `layer.type` (src/model/area.ts variantOf). Same `SegmentedTraceType`-style narrowed union would fit here. |
| `BUMP` | Bump / rank chart — `d3.line()` over rank values, one path per competitor | buildLineLayer with `TraceType.BUMP`; y must be the rank (1 = best), which is what the user already binds. Payload LinePoint[][] with `z` = competitor name. BumpTrace extends LineTrace and inverts the pitch itself, so nothing extra to compute. |
| `DIVERGING` | Population pyramid / diverging (Likert) bar — signed values on a symmetric scale, usually two `<g>` series of rects | Widen `SegmentedTraceType` (types.ts:232-235) to admit `TraceType.DIVERGING` and let bindD3Segmented emit it. Payload SegmentedPoint[][] `{x, y, z}` with y arriving signed (do NOT abs it — DivergingTrace reads the sign as a side and the magnitude as the pitch). The existing subject-major/series-major auto-detect (segmented.ts:255-265) already covers both D3 idioms. |
| `ERROR_BAR` | Point-range / confidence-interval chart — a `<g>` per estimate holding a `<line>` and a marker | New `bindD3ErrorBar` with `selector` on the per-estimate `<g>` and accessors `x`, `y`, `yMin`, `yMax` (alias fallbacks `lower`/`upper`, `ci_low`/`ci_high` via inferAccessor). Payload ErrorBarPoint[]; absolute bounds required, so a binder given half-widths needs a function accessor. |
| `DUMBBELL` | Dumbbell / connected-dot plot — a `<line>` connector plus two `<circle>` ends per row | New `bindD3Dumbbell`, `selector` on the connector lines (which is exactly what examples/dumbbell.html highlights). Accessors `x`, `start`, `end`; `startLabel`/`endLabel` come from config, not from data. Payload is the DumbbellData OBJECT `{points, startLabel, endLabel}` — the first non-array payload the D3 adapter would emit, so `D3BuiltLayer` handling is unchanged but the layer's `data` is not an array. |
| `WATERFALL` | Waterfall / bridge chart — floating `<rect>` per step, y from `start` to `end` | New `bindD3Waterfall` reading `start`/`end` per rect (the two numbers a floating bar is drawn from, already in the datum for any D3 waterfall). Derive `delta = end - start` and `kind` from the sign, with a `kind` accessor override so opening/closing/subtotal bars can be marked `'total'` — that flag is the only thing the binder cannot infer. Payload WaterfallPoint[]. |
| `WORD_CLOUD` | Word cloud — `<text>` per term, laid out by the d3-cloud plugin (`cloud().words(...)`) | New `bindD3WordCloud` over `text` elements; d3-cloud binds `{text, size, x, y, rotate}`, so default accessors `x:'text'`, `y:'size'` (alias `value`/`weight`/`count`) hit it directly. Payload WordCloudPoint[] `{x, y}`; layout coordinates are deliberately dropped (the grammar says the layout carries no information). NOTE: d3-cloud is not a repo dependency — the binder needs none, but a demo page would load it from CDN like examples/d3-*.html load d3 v7. |
| `GAUGE` | Gauge / bullet chart — `d3.arc()` dial, or a bullet built from nested rects | New `bindD3Gauge`: almost entirely config (`min`, `max`, `target`, `bands`) since a drawn gauge binds only the value. Payload is the single GaugePoint OBJECT (not an array) — `{value, min, max, label?, target?, bands?}`. Selector points at the needle/value arc. |
| `MANHATTAN` | Manhattan plot — scatter of `-log10(p)` against genomic position, coloured by chromosome | buildScatterLayer plus a `label` accessor (SNP id) and a `group` accessor (chromosome), emitting VolcanoPoint[] and `layer.thresholdOptions.significance`. VolcanoTrace serves both MANHATTAN and VOLCANO (factory.ts:131-136). The scatter binder already exists at binders/scatter.ts:63-112; the delta is two optional accessors plus the options passthrough. |
| `VOLCANO` | Volcano plot — scatter of significance against effect size | Same binder as MANHATTAN with `type: 'volcano'`; payload VolcanoPoint[] `{x, y, label?, group?}` plus `thresholdOptions.{significance, effect, significanceDirection}` from config. The `label` is the payload the grammar says a plain scatter withholds, so it should be required-ish (warn when absent). |
| `RADAR` | Radar / spider chart — `d3.lineRadial()` closed outline per series | buildLineLayer with `TraceType.RADAR`: each `<path>` datum is the array of per-spoke values, `x` = spoke/variable name, `z` = series. Payload LinePoint[][]. RadarTrace extends LineTrace and does the angular panning itself (factory.ts:150-153). One trap: a closed radar polygon repeats the first vertex — the binder should drop a trailing duplicate x so the spoke count is right. |
| `POLAR_AREA` | Polar area / coxcomb / rose chart — `d3.arc()` wedge per category, radius encodes the value | Wedge-per-mark rather than path-per-series, so it is closer to the pie binder's element walk than the line binder's: read `{category, value}` per `<path>` and emit a single-row LinePoint[][] with `TraceType.POLAR_AREA` (RadarTrace handles both, factory.ts:150-153). Reuse binders/pie.ts's arc unwrapping (isPieArc, pie.ts:42-47) if the wedges were joined to d3.pie() output. |
| `NETWORK` | Force-directed node-link graph — `d3-force` simulation, one `<line>` per link | New `bindD3Network` over the link elements. Trap worth encoding as the default accessor: `d3.forceLink` MUTATES each link's `source`/`target` from an id into the node object, so the accessor must be `d => d.source.id ?? d.source.name ?? d.source`. Payload NetworkPoint[] `{source, target}` — nodes are derived from the links, and positions are deliberately not carried. |
| `MOSAIC` | Mosaic / Marimekko plot — rects whose width also encodes the group size | Extend the segmented binder: same `{x, y, z}` extraction, plus a `width` accessor (the marginal proportion) and optional `count`. Where the datum lacks it, the width can be read off the rendered `rect`'s `width` attribute divided by the plot width — but that is an inference the binder should only make behind an explicit opt-in. Payload MosaicPoint[][]; MosaicTrace extends SegmentedTrace. |
| `TREEMAP` | `d3.treemap()` over a `d3.hierarchy()` — one `<rect>` per leaf | New `bindD3Treemap` over the leaf rects. The key move is deriving `path`: a d3-hierarchy node bound to a rect carries `.parent`, `.depth` and `.data`, so the default accessor walks `d.ancestors().reverse().slice(0,-1).map(n => n.data.name)` (or the parent chain manually, since `ancestors()` exists on the node). Payload TreemapPoint[] `{x, y, path}` — leaves only is fine, TreemapTrace derives interior nodes and totals from the paths. |
| `SUNBURST` | `d3.partition()` in polar coordinates — one `<path>` arc per node | Same binder and same `path` derivation as TREEMAP with `type: 'sunburst'`; TreemapTrace serves all three layouts and reads `this.type` only to pan angularly (factory.ts:166-172). Difference from treemap: a partition draws interior nodes too, so the binder must either emit them (with their declared `y`) or filter to `d.children === undefined`. |
| `ICICLE` | `d3.partition()` in cartesian coordinates — depth-ordered `<rect>` bands | Identical to SUNBURST with `type: 'icicle'`; same hierarchy walk, same TreemapPoint[] payload. Realistically one binder with a `layout` discriminator covers TREEMAP/SUNBURST/ICICLE. |
| `SANKEY` | Sankey diagram — the d3-sankey plugin, one `<path>` ribbon per link | New `bindD3Flow` over the ribbon paths. d3-sankey, like d3-force, replaces link `source`/`target` with node objects, so defaults are `d => d.source.name ?? d.source` and `value: 'value'`. Payload FlowPoint[] `{source, target, value}` in declared order — FlowTrace keys its selector list to declared order (src/model/flow.ts Edge.at), so the emitted selector array must not be reordered. d3-sankey is not a repo dependency; the binder needs none, only a demo would. |
| `ALLUVIAL` | Alluvial diagram — d3-sankey with repeated node columns, or a bespoke ribbon layout | Same `bindD3Flow` with `type: 'alluvial'`; FlowTrace handles SANKEY/ALLUVIAL/CHORD with one class (factory.ts:88-93). Payload FlowPoint[] identical. Only the announced chart name changes. |
| `CHORD` | `d3.chord()` + `d3.ribbon()` — circular flow, one `<path>` per ribbon | Same `bindD3Flow` with `type: 'chord'`, but chord ribbons bind `{source:{index,value,startAngle,…}, target:{index,value,…}}` — the ends are numeric INDICES, not names. So the binder needs a `names: string[]` config (the matrix's row labels) and maps `d.source.index -> names[i]`; value is `d.source.value`. Slightly more config than sankey but the same FlowPoint[] payload. |
| `GANTT` | Gantt / timeline / swimlane — one `<rect>` per interval on a band scale of lanes | New `bindD3Gantt` with accessors `x` (lane), `start`, `end`, optional `label`; dates need a numeric coercion (`+new Date(...)`) plus an axis `format: {type:'date'}`. Payload is the GanttData OBJECT `{points: GanttPoint[][], lanes?, unit?}` — points nested BY LANE, which the binder must group itself (the flat DOM order is not the payload). `lanes` matters only for empty lanes, which have no DOM element at all and therefore must come from config. |
| `BOXEN` | Letter-value plot — a stack of nested `<rect>` rungs per distribution | New `bindD3Boxen`, structurally like binders/box.ts: `selector` on the per-distribution `<g>`, read `median` plus a `levels` array of `{p, lo, hi}` from the group datum (a D3 boxen inevitably computes that ladder before drawing it, so it is in `__data__`). Payload BoxenPoint[]. Unlike box.ts this needs no geometric part-classification if the rungs are highlighted as one group, but per-rung highlighting would need the box.ts-style attribute stamping (box.ts:144-242). |
| `FOREST` | Forest plot — a meta-analysis figure: point-range per study plus a pooled diamond | Extend the ERROR_BAR binder with `weight` and `pooled` accessors and pass `forestOptions.nullValue` from config (ForestTrace reads `layer.forestOptions?.nullValue`, src/model/forest.ts:63, and deliberately has no default). Payload ForestPoint[]. The pooled row is usually a differently-shaped mark (a diamond `<path>`, not a line), so the binder likely needs a second selector or a `pooled` predicate rather than one flat selection. |
| `SURVIVAL` | Kaplan-Meier curve — `d3.line().curve(d3.curveStepAfter)`, plus tick marks for censored times | New `bindD3Survival` = the line binder plus a `censored` accessor and optional `yMin`/`yMax` for the confidence band. Payload SurvivalPoint[][] (2D, one row per group). Censoring marks are usually drawn as separate `<line>`/`<path>` elements from a separate join, so the binder must merge them into the curve's points by x rather than reading one selection. |
| `PARALLEL` | Parallel coordinates — one `<path>` per observation across several per-variable scales | New `bindD3Parallel`: each path's datum is one observation OBJECT, and the payload is LinePoint[][] with one row per observation and one point per dimension — i.e. the binder transposes `{mpg: 21, hp: 110, wt: 2.6}` into `[{x:'mpg',y:21},{x:'hp',y:110},…]` given a `dimensions: string[]` config (which the chart already has, since it built one scale per dimension). ParallelTrace extends LineTrace and derives each column's range from the data itself, so no per-axis min/max is needed in the payload. |
| `RIDGELINE` | Ridgeline / joyplot — one `d3.area()` density curve per group, vertically offset | New `bindD3Ridgeline` over the per-group area paths. Payload ViolinKdePoint[][] `{x: group, y: value-axis position, density}` — and the critical rule from the grammar is that `density` must be the curve's OWN half-width, never the drawn y which includes the group's offset baseline. So the binder needs either a `density` accessor or the pre-offset kernel-density array the user computed. RidgelineTrace extends AbstractTrace directly, so its selector contract must be checked against `mapToSvgElements`. |
| `HEXBIN` | Hexbin density plot — the d3-hexbin plugin, one `<path>` hexagon per bin | New `bindD3Hexbin`. d3-hexbin bins are arrays with `.x`/`.y` (the SCREEN-space centre) and `.length` (the count), so the binder needs inverse-scale accessors to recover data-space centres — `x: d => xScale.invert(d.x)`. Payload is HexbinPoint[][] — 2D, grouped into lattice ROWS by y (src/model/hexbin.ts:78), which the binder must cluster itself since the DOM is a flat list. HexbinTrace withdraws highlighting unless the selector count matches exactly, and empty bins are simply absent from the DOM, so the row grouping must come from the emitted data, not from an assumed rectangular grid. d3-hexbin is not a repo dependency. |
| `CONTOUR` | `d3.contours()` / `d3.contourDensity()` — one `<path>` MultiPolygon per threshold | New `bindD3Contour` over the level paths: each datum is a GeoJSON MultiPolygon carrying `.value` (the threshold) and `.coordinates`. Payload ContourPoint[][] — one row per level, each point `{x, y, level}` with `level` repeated on every point of its curve (ContourTrace scans the curve for the first finite level, src/model/contour.ts:72-76). The real work is coordinate space: `d3.contours` emits GRID INDICES, so the binder needs user-supplied `x`/`y` transforms back to data space, and a multi-ring level has to be flattened to one polyline (or the largest ring taken) since the payload row is a single curve. |

## Deferred to a follow-up

These need a new author-facing opt-in API or semantics D3 does not carry in its own configuration, so they are out of scope here and tracked separately:

- `CHOROPLETH` — New `bindD3Choropleth` over the region paths: `x` = region name (`d.properties.name`), `y` = the joined value. The two optional fields are what make it worth more than a bar chart and are also the cost: `lon`/`lat` must be geographic centroids — `d3.geoPath().centroid(d)` returns PROJECTED pixels, s

## Not renderable by D3

<details><summary>1 trace types the library genuinely cannot draw</summary>

- Nothing. D3 is a low-level DOM/SVG data-binding toolkit, not a chart library with a fixed catalogue: every one of the 34 post-pie types has a canonical D3 recipe, and the core d3 v7 bundle already ships the layout modules for the hard ones (d3-hierarchy for treemap/sunburst/icicle, d3-chord, d3-contour, d3-geo, d3-force, d3-shape's area/stack/lineRadial/arc). Four types normally use a plugin outside the core bundle and outside this repo's dependencies — d3-sankey (SANKEY, ALLUVIAL), d3-hexbin (HEXBIN), d3-cloud (WORD_CLOUD) — but this is not a rendering limit for the adapter: it imports no d3 code at all and reads whatever the page drew off `__data__`, so a plugin is only needed to author a demo page (which existing examples already load from a CDN).

</details>

## Acceptance criteria

- [ ] Unit tests under `test/adapters/d3/` cover every newly detected type.

## Implementation notes

Adding one trace type to this adapter touches five places, all mechanical: (1) a new `binders/<type>.ts` exporting `bindD3X` (thin, via `finalizeSingleChart`, util.ts:407) and `buildXLayer(root, config, panel?)`; (2) a config interface in types.ts plus a new arm in the `D3PanelChartSpec` union (types.ts:387-397) — that union is what makes the type available to facets AND subplots for free; (3) an arm in `buildPanelLayer` (binders/subplots.ts:231-253); (4) arms in `runBinder` and `withFacetsAutoApplyOff` (useD3Adapter.ts:116-141, 150-171) and in `MaidrD3.tsx:148-172` — these three switches are exhaustive-checked by TS, so a missing arm is a compile error, not a silent gap; (5) exports in index.ts:96-130 and a row in docs/d3.md:191-204.

Shared helpers to reuse rather than reinvent: `queryD3Elements` (util.ts:215) already filters out MAIDR's own hidden highlight clones via `isMaidrOwned` (util.ts:54) — skipping that filter is the classic rebind bug (doubled element counts, "no D3-bound data" throws). `inferAccessor` (util.ts:142) gives the alias-fallback behaviour every binder has; `buildAxes` (util.ts:349) maps the adapter's `fill` input to the canonical `z` axis. `scopeSelector`/`selectorPrefix` (selectors.ts:73, 94) produce page-unique, panel-scoped selectors — MAIDR resolves selectors through global `document.querySelector`, so no new binder may emit a bare selector.

The biggest trap for the new types is the SELECTOR SHAPE, not the data. `AbstractTrace.mapToSvgElements` silently withdraws highlighting when the selector list length does not match the row/point count, and a bare string's `.length` is its character count. binders/line.ts:271-287 (per-series `data-maidr-line-index` stamping, degrading to `undefined` rather than mis-highlighting), binders/box.ts:144-242 (a `BoxSelector` object per box, with `data-maidr-box-part` stamping) and binders/smooth.ts (a one-element array, not a string) are the three worked examples. Every 2D payload proposed above — HEXBIN (HexbinPoint[][]), CONTOUR (ContourPoint[][]), SURVIVAL, PARALLEL, RIDGELINE, the area family — will hit the same check, and HexbinTrace is explicitly strict about it (src/model/hexbin.ts, "withdrawn unless the count matches exactly").

Model-side work is already done for all 34: the factory dispatches every one of them (src/model/factory.ts:52-195) and the grammar carries every payload type. Many post-pie types subclass an existing trace — AreaTrace/BumpTrace/ParallelTrace/RadarTrace extend LineTrace, FunnelTrace extends BarTrace, DivergingTrace/MosaicTrace extend SegmentedTrace, ForestTrace extends ErrorBarTrace, SurvivalTrace extends StepTrace — which is why the "small" ratings above are honest: those binders are the existing bar/line/segmented cores with a different constant.

Precedent across the codebase is thin: grepping `TraceType.<post-pie>` over src/adapters shows only chartjs and vegalite emit AREA/STACKED_AREA (and vegalite NORMALIZED_AREA). Every other post-pie type currently reaches MAIDR only through hand-authored `maidr=` JSON in examples/ (treemap.html, sankey.html, waterfall.html, dumbbell.html, gantt.html, forest.html, hexbin.html, contour.html, choropleth.html, radar.html, parallel.html, ridgeline.html, wordcloud.html, volcano.html, survival.html, boxen.html, funnel.html, gauge.html, mosaic.html, network.html, sunburst.html, icicle.html, lollipop.html, dot.html, bump.html, errorbar.html) — those files are the best reference for the exact payload each trace expects, and are worth reading before writing any new binder.

Conventions to follow: tests are jsdom-based and stamp `__data__` by hand (test/adapters/d3/pie.test.ts:19-34 is the cleanest template); there are currently only five test files (util, pie, box, segmented, subplots), so several existing binders are untested and a new one is expected to bring its own. Examples follow `examples/d3-bind<type>.html`, loading d3 v7 from https://d3js.org plus ../dist/maidr.js and ../dist/d3.js. Note also `finalizeSingleChart` only emits a subplot `legend` when the builder returns one (util.ts:419) — line and segmented do; any new multi-series type (area family, radar, parallel, ridgeline, survival) should too.

jooyoungseo:
# Pull Request

## Description

The D3 adapter shipped ten binders (bar, box, candlestick, heatmap, histogram, line, pie, scatter, segmented, smooth) and had not kept up with the ~34 trace types MAIDR core has gained since the pie chart. A chart D3 drew perfectly well was therefore dropped from the figure or announced as the wrong chart.

This PR adds binders for **all 33 in-scope trace types** listed in #865. The work is adapter-local: no file under `src/model/`, `src/service/`, `src/state/`, `src/ui/`, `src/command/` or `src/type/grammar.ts` is touched. The core already dispatches every one of these types; what was missing was the extraction, the payload shape, and — the part that is easy to get silently wrong — the selector/highlight contract.

Each new type follows the five-place pattern already established in the adapter: a `binders/*.ts` exporting `bindD3X` plus `buildXLayer`, a config interface and a `D3PanelChartSpec` arm in `types.ts` (which makes the type available to `bindD3Facets` and `bindD3Subplots` for free), arms in the four TS-exhaustive switches (`buildPanelLayer`, `runBinder`, `withFacetsAutoApplyOff`, `MaidrD3.buildSpec`), exports in `index.ts`, and a section in `docs/d3.md`.

## Related Issues

Closes #865

### Trace types added (33)

Grouped by the extraction core they reuse.

**Bar core** (`binders/bar.ts`) — same `BarPoint[]` payload, different `TraceType`:

- `DOT` — `bindD3Dot`
- `LOLLIPOP` — `bindD3Lollipop`
- `FUNNEL` — `bindD3Funnel` (stage order is load-bearing; emitted in draw order)

**Line core** (`binders/line.ts`):

- `AREA`, `STACKED_AREA`, `NORMALIZED_AREA` — `bindD3Area` (`binders/area.ts`), selected via `config.type`. Unwraps the `d3.stack()` series datum (`[y0, y1]` tuples with a `.data` back-reference, `.key` on the parent group) and emits the series' own value, since `AreaTrace` re-derives the running total itself.
- `BUMP` — `bindD3Bump`
- `RADAR` — `bindD3Radar`, dropping the closing duplicate vertex a closed radial polygon repeats so the spoke count is right
- `POLAR_AREA` — `bindD3PolarArea`, an arc-per-category walk emitting a single-row `LinePoint[][]`
- `SURVIVAL` — `bindD3Survival` (`binders/survival.ts`), via a new internal point-decorator hook reading `censored`, `yMin` and `yMax` off the same datum. Censoring ticks drawn by their own join are merged into the curve by time: an exact vertex is flagged, otherwise a vertex is inserted carrying the probability held across that interval. Censoring is read strictly (`true`/`1`/`'1'`/`'true'`), never by truthiness.
- `PARALLEL` — `bindD3Parallel` (`binders/parallel.ts`), transposing each observation datum across a required `dimensions: string[]`

**Segmented core** (`binders/segmented.ts`):

- `DIVERGING` — `bindD3Diverging`; values arrive signed and are not `abs`'d, because `DivergingTrace` reads the sign as a side
- `MOSAIC` — `bindD3Mosaic`, adding `width` (the marginal proportion) and optional `count`

**Scatter core** (`binders/scatter.ts`):

- `VOLCANO` — `bindD3Volcano`, plus `thresholdOptions.{significance, effect, significanceDirection}`
- `MANHATTAN` — `bindD3Manhattan`, adding the SNP `label` and chromosome `group` accessors and the significance cutoff

**New shapes** (each with its own file):

- `ERROR_BAR` — `bindD3ErrorBar` (`binders/errorBar.ts`), absolute `yMin`/`yMax` bounds with alias fallbacks
- `FOREST` — `bindD3Forest`, sharing the error-bar core; adds `weight`, `pooled`, a separate `pooledSelector` for the diamond, and `forestOptions.nullValue`
- `DUMBBELL` — `bindD3Dumbbell` (`binders/dumbbell.ts`); emits the `DumbbellData` **object** `{points, startLabel, endLabel}`, not an array
- `WATERFALL` — `bindD3Waterfall` (`binders/waterfall.ts`), reading `start`/`end` and deriving the contribution, with a `kind` override so total bars can be marked
- `WORD_CLOUD` — `bindD3WordCloud` (`binders/wordCloud.ts`); d3-cloud's `{text, size}` are the default accessors, layout coordinates deliberately dropped
- `GAUGE` — `bindD3Gauge` (`binders/gauge.ts`); emits the `GaugePoint` **object**, not an array
- `NETWORK` — `bindD3Network` (`binders/network.ts`); the default endpoint accessor handles `d3.forceLink` mutating `source`/`target` from an id into the node object
- `TREEMAP`, `SUNBURST`, `ICICLE` — `bindD3Treemap` / `bindD3Sunburst` / `bindD3Icicle` (`binders/treemap.ts`), one core with a layout discriminator, deriving `path` by walking the `d3.hierarchy` ancestor chain
- `SANKEY`, `ALLUVIAL`, `CHORD` — `bindD3Sankey` / `bindD3Alluvial` / `bindD3Chord` (`binders/flow.ts`), one core. Endpoint normalisation handles the node objects d3-sankey substitutes for `source`/`target`; chord adds a `names` config mapping `d3.chord()`'s matrix indices to labels.
- `GANTT` — `bindD3Gantt` (`binders/gantt.ts`); emits the `GanttData` **object** with points nested **by lane** (grouped by the binder — flat DOM order is not the payload), plus optional `lanes` (so empty lanes, which have no DOM element at all, survive) and `unit`. Dates coerced to epoch ms.
- `BOXEN` — `bindD3Boxen` (`binders/boxen.ts`), reading `median` plus the `{p, lo, hi}` rung ladder off the per-distribution group datum, with key-alias normalisation
- `RIDGELINE` — `bindD3Ridgeline` (`binders/ridgeline.ts`); `density` comes from a dedicated accessor (aliases `kde`/`width`/`p`/`estimate` — deliberately never `y`), because the drawn y includes the group's offset baseline
- `HEXBIN` — `bindD3Hexbin` (`binders/hexbin.ts`); defaults read a d3-hexbin bin's `.x`/`.y`/`.length` and the caller supplies inverse scales. The binder assembles the lattice itself, rows keyed on the y centre and ordered **bottom-first**, because `HexbinTrace.moveOnce` steps the row index **up** for `UPWARD`.
- `CONTOUR` — `bindD3Contour` (`binders/contour.ts`); reads the GeoJSON `.value` as the level and flattens MultiPolygon/Polygon/bare-ring nesting into one curve per level, dropping each ring's closing duplicate vertex. A new `D3GridTransform` maps grid indices back to data space.

### Not added, and why

- **`CHOROPLETH`** — explicitly deferred by #865 itself ("Deferred to a follow-up"): it needs a new author-facing opt-in API for geographic centroids, since `d3.geoPath().centroid(d)` returns projected pixels rather than lon/lat. No binder, no `D3PanelChartSpec` arm, no docs row.
- **`STEP`, `VIOLIN_KDE`, `VIOLIN_BOX`, `CANDLESTICK_DELTA`** — not in the issue's in-scope table, so out of scope here. Nothing in the adapter blocks them; each would be another instance of the same five-place pattern.
- **Nothing was skipped for a missing core capability.** No trace type in scope required a change under `src/model/`, `src/service/`, `src/state/`, `src/ui/`, `src/command/` or `src/type/grammar.ts`.

### Highlighting

Selector shape was treated as part of each type, not as a follow-up, because `AbstractTrace.mapToSvgElements` **silently withdraws** highlighting when the element count does not match the row/point count — and in two cases here it does something worse than withdraw. Each binder emits page-unique, panel-scoped selectors through the existing `scopeSelector`/`selectorPrefix` helpers; none emits a bare selector.

Three distinct shapes were needed, chosen per trace against what its `mapToSvgElements`/`pairWith` actually requires:

- **One scoped selector, one element per row** — the flat types (dot, lollipop, funnel, waterfall, word cloud, gauge, network, error bar, dumbbell connectors, treemap/sunburst/icicle nodes, flow ribbons).
- **One selector per series/group**, stamped with `data-maidr-line-index` — the area family, radar, parallel, survival, contour, reusing the line binder's helper.
- **One stamped selector per element in payload order** (`data-maidr-gantt-index`, `data-maidr-hexbin-index`) — gantt and hexbin, where the payload order is *not* DOM order. Both traces slice a **flat** element list and check only the total count, so a bare selector would mis-highlight with the counts still matching, which is silent and worse than no highlighting. Generalised into a shared `stampOrderedSelectors` in `selectors.ts`.

Two traces required the opposite of stamping, and per-element selectors would have broken them: `BoxenTrace.pairWith` and `RidgelineTrace.pairWith` both require exactly **one element per group** and repeat it across the group's rungs/samples, so both emit a single scoped selector per group. Forest emits a two-entry array (studies, then pooled) in payload order and **throws** when `pooledSelector` overlaps the studies selector, since a duplicated row would keep the counts consistent and go unnoticed.

### Tests

26 new/extended files under `test/adapters/d3/`, jsdom-based, stamping `__data__` by hand in the style of the existing `pie.test.ts`. `npx jest test/adapters/d3` is now 28 suites / 246 tests (was 5 files on `main`). Coverage per type: payload shape against `src/type/grammar.ts`, accessor inference and alias fallbacks, error messages, a core-model `Figure` construction, and the selector/highlight contract specifically — element count vs payload count, lane-order and lattice-order stamping verified by resolving each selector back to its datum, and one-element-per-group for boxen/ridgeline.

### Docs and examples

- `docs/d3.md` — +552 lines: a table row and a worked section per new type, plus the exported TypeScript config types.
- `examples/d3-bindarea.html`, `examples/d3-bindfunnel.html` — two new live binder demos. See Additional Notes.

## Checklist

- [ ] I have tested the changes locally following `ManualTestingProcess.md`, and all tests related to this pull request pass. — the automated suite passes in full (see below), but the manual screen-reader pass in `ManualTestingProcess.md` has **not** been run.

## Additional Notes

**Verification.** Run in order on this branch, all green, no fixes required:

| Command | Result |
|---|---|
| `npm run lint:fix` | clean, no changes |
| `npm run type-check` | clean (`tsc --noEmit` + `tsconfig.ci.json`) |
| `npm run build` | all 12 targets built |
| `npm test` | 210 suites / 2816 tests + 7 ESM suites / 73 tests, all passing |

**Scope discipline.** The diff is 56 files across `src/adapters/d3/` (27), `test/adapters/d3/` (26), `docs/d3.md` and `examples/`. Nothing else is touched.

**No new dependencies.** The adapter imports no d3 code at all — it reads whatever the page drew off `__data__` — so the binders for the plugin-drawn types (d3-sankey, d3-hexbin, d3-cloud, d3-contour) need no package added, and none was.

**Known gap — demo pages.** Only `area` and `funnel` got new `examples/d3-bind*.html` demos; the other 31 types have no live binder demo, and the hand-authored `examples/*.html` files (which carry `maidr=` JSON directly) remain their reference payloads. These are worth adding as a follow-up, but every type is covered by unit tests and documented in `docs/d3.md`.

**Known gap — diverging in panels.** `bindD3Diverging` works standalone, but `D3PanelChartSpec` exposes `chartType: 'segmented'` and `'mosaic'` without a `'diverging'` alias, so inside a facet or subplot a diverging bar chart must be declared as `{ chartType: 'segmented', config: { type: 'diverging_bar', … } }`. Functional, but less discoverable than the other types; a one-line alias arm would fix it.

**Interpretation notes** where the issue left a decision open:

- *Chord without `names`* — the binder degrades to the bare matrix index rather than throwing. Honest but useless as an announcement, so the docs say to declare `names`.
- *Survival censoring merge* — a Kaplan-Meier curve steps only at events, so a censoring tick's time is usually not a vertex. Matching by x alone would silently drop most ticks, hence the insert-a-vertex behaviour described above. Row lengths therefore change relative to the drawn vertex count; harmless, because `LineTrace` falls back to path parsing via `reconcilePathCoordinates`.
- *Contour multi-ring levels* — all rings of a level are concatenated (each ring's closing duplicate dropped) rather than taking the largest, so no real geometry is dropped. The jump between rings is inaudible; documented.
- *Contour row order* — rows keep DOM order rather than being re-sorted by level, since `d3.contours` emits ascending thresholds and `ContourTrace` measures the gap to the **adjacent** row.

---

### DIFF
--- a/docs/d3.md
+++ b/docs/d3.md
@@ -228,6 +228,11 @@ When you omit an accessor, the binder tries a sensible default (`x`, `y`, `fill`
 | `bindD3Sankey` | Sankey | `<path>` ribbon per link | [Sankey](examples.html) |
 | `bindD3Alluvial` | Alluvial | `<path>` ribbon per link | [Alluvial](examples.html) |
 | `bindD3Chord` | Chord | `<path>` ribbon per chord | [Chord](examples.html) |
+| `bindD3Survival` | Kaplan-Meier survival curve | `<path>` per arm, plus `censoredSelector` for the ticks | [Survival](examples.html) |
+| `bindD3Parallel` | Parallel coordinates | `<path>` or `<polyline>` per observation | [Parallel](examples.html) |
+| `bindD3Ridgeline` | Ridgeline / joy plot | `<path>` density curve per group | [Ridgeline](examples.html) |
+| `bindD3Hexbin` | Hexbin density | `<path>` hexagon per bin | [Hexbin](examples.html) |
+| `bindD3Contour` | Contour / density field | `<path>` per level | [Contour](examples.html) |
 
 ## Multi-Panel Charts
 
@@ -892,6 +897,108 @@ maidrD3.bindD3Chord(svgElement, {
 
 The ribbons are emitted in the order they were declared, and the trace keys its selectors to that order: from a node, arrow keys follow the largest flow out of it and the rotor steps through the rest.
 
+### Kaplan-Meier Survival Curve
+
+A survival curve is a step line, so `selector`, `x`, `y` and `fill` are the line binder's. What a survival figure adds is the two things it is read for — which times were **censored**, and how wide the confidence band is:
+
+```js
+maidrD3.bindD3Survival(svgElement, {
+  selector: 'path.km',
+  censoredSelector: 'line.censor',     // the ticks, drawn by their own join
+  title: 'Overall Survival',
+  axes: { x: 'Months', y: 'Survival probability', fill: 'Arm' },
+  x: 'time',
+  y: 'surv',
+  fill: 'arm',
+  yMin: 'lower',
+  yMax: 'upper',
+});
+```
+
+Censoring marks are drawn as ticks precisely because censoring does not change the estimate — so they come from a separate join whose times are not, in general, vertices of the curve. The binder merges each tick into the arm its `fill` names: flagging the vertex already at that time, or inserting one carrying the probability the curve holds across the interval it falls in. On a chart whose own samples already carry a `censored` column, leave `censoredSelector` out and the `censored` accessor reads it (`true`, `1`, `'1'` and `'true'` count; nothing else does, because `'0'` is truthy and censors nobody).
+
+The trace derives median survival itself and offers a rotor over the censored times, so nothing here computes either.
+
+### Parallel Coordinates
+
+One `<path>` per **observation** crossing several per-variable scales, with the whole observation bound to it. The binder transposes that into one point per axis, and `dimensions` is the axis order — the same list you built one scale per:
+
+```js
+maidrD3.bindD3Parallel(svgElement, {
+  selector: 'path.observation',
+  title: 'Car Characteristics',
+  axes: { x: 'Variable', y: 'Value', fill: 'Car' },
+  dimensions: ['mpg', 'hp', 'weight'],
+  label: 'name',                       // what this observation is called
+  // value: (d, dimension) => d.values[dimension],   // when the numbers nest
+});
+```
+
+> **`dimensions` is required.** An object's key order is not an axis order, and the order the axes are drawn in is the order a reader arrows through them. An observation short of a declared dimension is an error rather than a gap: dropping it would shift every axis after it, announcing each value under the next variable's name.
+
+The trace derives each axis' range from the data and sonifies every value against its own axis, so no per-axis minimum is needed in the payload — a car with the best economy and the worst power sounds like exactly that, rather than like the units the two variables happen to be measured in.
+
+### Ridgeline / Joy Plot
+
+One `d3.area()` density curve per group, offset down the page. The three per-sample fields are named for what they mean rather than for the payload keys they land on, because a ridgeline's value axis is usually the drawn `x`:
+
+```js
+maidrD3.bindD3Ridgeline(svgElement, {
+  selector: 'path.ridge',
+  title: 'Delivery Time by Cohort',
+  axes: { x: 'Days', y: 'Cohort', fill: 'Cohort' },
+  group: 'cohort',                     // names the ridge
+  value: 'days',                       // position along the value axis
+  density: 'density',                  // the curve's OWN half-width
+});
+```
+
+> **`density` is never the drawn y.** A ridgeline is drawn by adding each group's baseline to its density, and that baseline is layout: fed to MAIDR it would make every group's loudness a function of where it happened to be stacked, and the lowest ridge the loudest chart-wide. Pass the kernel-density array you computed before offsetting it — the number inside the offset arithmetic in your `d3.area().y1(...)`, not its result. The binder cannot detect the mistake; it can only refuse when there is no density at all.
+
+Each path's datum supplies its group's samples: the array itself, a `d3.groups()` tuple, or a `values` / `samples` / `points` / `curve` property holding one. One element per group is exactly what the trace wants — it lights that ridge's whole curve from any of its samples — so a single scoped selector is emitted.
+
+### Hexbin Density
+
+The `d3-hexbin` plugin returns one bin per occupied hexagon, each an array of the points that fell in it carrying `.x`/`.y` (the centre) and `.length` (the count) — which is what the defaults read:
+
+```js
+const bins = hexbin(points);
+svg.selectAll('path.hexagon').data(bins).join('path').attr('d', hexbin.hexagon());
+
+maidrD3.bindD3Hexbin(svgElement, {
+  selector: 'path.hexagon',
+  title: 'Point Density',
+  axes: { x: 'Carat', y: 'Price', fill: 'Count' },
+  x: d => xScale.invert(d.x),          // the centres come out in PIXELS
+  y: d => yScale.invert(d.y),
+});
+```
+
+> **Pass the inverse scales.** `d3-hexbin` bins the projected points, so `bin.x` is a screen coordinate. Left as it is, every bin announces its position in pixels — which sounds entirely plausible and is entirely wrong.
+
+The lattice rows are assembled by the binder: bins grouped by their `y`, ordered from the lowest upward, each row ordered left to right. An empty bin is not drawn at all, so the rows do not hold the same number of bins and no rectangular grid could be laid over them. Supply `row` only if your `y` centres do not come out identical within a row. The highlight selectors are stamped per bin in that lattice order, because a bare selector would resolve in the order the plugin generated the bins in — with the counts still matching, so the trace could not tell.
+
+### Contour / Density Field
+
+`d3.contours()` and `d3.contourDensity()` emit one GeoJSON `MultiPolygon` per threshold, carrying the threshold as `.value`. Bind those objects to the paths — not the `d` strings you drew them with:
+
+```js
+const levels = d3.contours().size([n, m])(values);
+svg.selectAll('path.contour').data(levels).join('path').attr('d', d3.geoPath());
+
+maidrD3.bindD3Contour(svgElement, {
+  selector: 'path.contour',
+  title: 'Density Field',
+  axes: { x: 'X', y: 'Y', fill: 'Density' },
+  x: column => x0 + column * cellWidth,   // contours() emits GRID INDICES
+  y: row => y0 + row * cellHeight,
+});
+```
+
+> **The coordinates are not in data space.** `d3.contours()` walks a grid and emits indices; `d3.contourDensity()` bins projected points and emits pixels (use `x: px => xScale.invert(px)` for that one). Neither is a position on an axis.
+
+The level is carried on every point of its curve, which is where the grammar has room for it, and the trace announces it on the `z` axis — the field's own — rather than as a series name. A level drawn as several disjoint rings is flattened into one curve in ring order, since a payload row is a single polyline: every point announced is real, but the jump from one ring to the next is inaudible. Rows keep the order the paths were drawn in, which for `d3.contours()` is ascending threshold — the order the trace measures the gap to the adjacent level in.
+
 ### Gauge / Bullet Chart
 
 A drawn gauge binds one number, so the rest of the reading is config — and it *is* the reading: "73" means nothing without the range it sits in, the target it was aiming at, and the band it lands in:
@@ -973,6 +1080,11 @@ import type {
   D3HistogramConfig,
   D3HeatmapConfig,
   D3CandlestickConfig,
+  D3ContourConfig,
+  D3HexbinConfig,
+  D3ParallelConfig,
+  D3RidgelineConfig,
+  D3SurvivalConfig,   // extends D3LineConfig
   D3SegmentedConfig,
   D3MosaicConfig,
   D3NetworkConfig,
@@ -995,6 +1107,7 @@ import type {
   BarMarkTraceType,
   LineMarkTraceType,
   FlowTraceType,
+  D3GridTransform,    // (gridCoordinate: number) => number, for bindD3Contour
   ScatterMarkTraceType,
   SegmentedTraceType,
   TreemapTraceType,
--- a/docs/d3.md
+++ b/docs/d3.md
@@ -193,7 +193,12 @@ When you omit an accessor, the binder tries a sensible default (`x`, `y`, `fill`
 | Binder | Chart type | Selector targets | Example |
 |--------|-----------|------------------|---------|
 | `bindD3Bar` | Bar | `<rect>` per bar | [Bar chart](examples.html) |
+| `bindD3Dot` | Cleveland dot plot | `<circle>` per category | [Dot plot](examples.html) |
+| `bindD3Lollipop` | Lollipop | `<circle>` head per category | [Lollipop](examples.html) |
+| `bindD3Funnel` | Funnel / stage chart | `<path>` or `<rect>` per stage | [Funnel](examples.html) |
 | `bindD3Line` | Line / multi-line | `<path>` per series, optional `<circle>` points | [Line chart](examples.html) |
+| `bindD3Area` | Area / stacked area / 100% stacked area | `<path>` per series | [Area chart](examples.html) |
+| `bindD3Bump` | Bump / rank chart | `<path>` per competitor | [Bump chart](examples.html) |
 | `bindD3Scatter` | Scatter | `<circle>` per point | [Scatter plot](examples.html) |
 | `bindD3Box` | Box plot | `<g>` per box (containing `<rect>` + `<line>` + `<circle>`) | [Box plot](examples.html) |
 | `bindD3Histogram` | Histogram | `<rect>` per bin | [Histogram](examples.html) |
@@ -222,7 +227,7 @@ svg.selectAll('g.panel')
 
 maidrD3.bindD3Facets(svgElement, {
   panelSelector: 'g.panel',
-  chartType: 'bar',                       // any of the ten chart types
+  chartType: 'bar',                       // any of the single-chart types
   config: {                               // the usual per-type config;
     selector: 'rect.bar',                 // resolved inside each panel
     title: 'Sales by Day, faceted by Region',  // figure title
@@ -306,6 +311,42 @@ maidrD3.bindD3Bar(svgElement, {
   title: 'Tips by Day',
   axes: { x: 'Day', y: 'Count' },
   x: 'day',
+  y: 'count',
+});
+```
+
+### Dot Plot, Lollipop, and Funnel
+
+Three marks that carry the same `{ category, value }` datum a bar does, so they take the same config — only the binder differs, and with it the chart MAIDR announces:
+
+```js
+// Cleveland dot plot: categories down the page, so x is the value.
+svg.selectAll('circle.dot')
+  .data(endpoints).join('circle')
+  .attr('class', 'dot')
+  .attr('cx', d => x(d.ms))
+  .attr('cy', d => y(d.endpoint));
+
+maidrD3.bindD3Dot(svgElement, {
+  selector: 'circle.dot',
+  title: 'Median Response Time',
+  orientation: 'horz',
+  axes: { x: 'Milliseconds', y: 'Endpoint' },
+  x: 'ms',
+  y: 'endpoint',
+});
+```
+
+For a **lollipop**, point `selector` at the heads (one `<circle>` per category) or at a `<g>` wrapping each head-and-stem pair — never at a selector that also matches the `<line>` stems, which would double every category.
+
+A **funnel** is read for its drop-offs: the trace pitches the retention between adjacent stages rather than the raw counts, so stage order is load-bearing. The binder keeps the elements in the order D3 joined them, so draw them in funnel order:
+
+```js
+maidrD3.bindD3Funnel(svgElement, {
+  selector: 'path.stage',
+  title: 'Checkout Funnel',
+  axes: { x: 'Stage', y: 'People' },
+  x: 'stage',
   y: 'count',
 });
 ```
@@ -462,6 +503,57 @@ maidrD3.bindD3Candlestick(svgElement, {
 
 `trend` is auto-computed from `open` vs `close` when not supplied.
 
+### Area / Stacked Area / 100% Stacked Area
+
+`bindD3Area` handles all three variants through `type: 'area' | 'stacked_area' | 'stacked_normalized_area'` (`'area'` is the default). It reads the two shapes D3 produces, telling them apart from the datum bound to the first `<path>`:
+
+- **Plain point arrays** — one `{x, y}` array per `<path>`, exactly as `bindD3Line` reads them (`pointSelector` works here too).
+- **`d3.stack()` output** — the datum bound to each `<path>` is the *series* array itself, carrying `.key`, whose items are `[y0, y1]` tuples with a `.data` back-reference to the row. The binder unwraps that shape for you: `x` is read from the row, `y` is the band's own height (`y1 - y0`), and `.key` names the series.
+
+```js
+const series = d3.stack().keys(['Subscriptions', 'Services'])(rows);
+
+svg.selectAll('path.area')
+  .data(series).join('path')
+  .attr('class', 'area')
+  .attr('fill', d => color(d.key))
+  .attr('d', d3.area()
+    .x(d => x(d.data.year))
+    .y0(d => y(d[0]))
+    .y1(d => y(d[1])));
+
+maidrD3.bindD3Area(svgElement, {
+  selector: 'path.area',
+  type: 'stacked_area',
+  title: 'Revenue by Product',
+  axes: { x: 'Year', y: 'Revenue ($K)', fill: 'Product' },
+  x: 'year',   // a key on the stacked ROW, not on the [y0, y1] tuple
+});
+```
+
+In that shape the two accessors address different objects, because that is where the two values live: `x` is resolved against the **row** — function accessors included, so write `d => d.year`, not `d => d.data.year` — while an explicit `y` is resolved against the **tuple**, keeping `d => d[1] - d[0]` expressible.
+
+> **Give the band's own value, not the top edge.** The trace re-derives the running total by summing the series it is given, so it announces the stack height at each x and the point's share of it. The binder already emits `y1 - y0` for `d3.stack()` output; only supply your own `y` accessor if you stacked the data by hand.
+
+A streamgraph (`d3.stackOffsetWiggle`) is `'stacked_area'`; a 100% stacked area (`d3.stackOffsetExpand`) is `'stacked_normalized_area'`.
+
+### Bump / Rank Chart
+
+A bump chart is a multi-line layer whose y values are **ranks** — 1 is the best position and the smallest number. Bind the ranks you already plotted; `BumpTrace` inverts the pitch so first place is the highest note, and announces the places gained or lost at each period:
+
+```js
+maidrD3.bindD3Bump(svgElement, {
+  selector: 'path.rank-line',
+  title: 'League Table by Round',
+  axes: { x: 'Round', y: 'Rank', fill: 'Team' },
+  x: 'round',
+  y: 'rank',
+  fill: 'team',
+});
+```
+
+A slope graph of *values* is a line chart with two samples, not this.
+
 ### Stacked / Dodged / Normalized Bars
 
 `bindD3Segmented` handles the three common multi-series bar variants. Pass `type: 'stacked_bar' | 'dodged_bar' | 'normalized_bar'`:
@@ -543,8 +635,9 @@ All public types are exported from the `maidr/d3` entry (and re-exported from `m
 ```ts
 import type {
   // Per-binder configs
-  D3BarConfig,
-  D3LineConfig,
+  D3BarConfig,        // also bindD3Dot / bindD3Lollipop / bindD3Funnel
+  D3LineConfig,       // also bindD3Bump
+  D3AreaConfig,
   D3ScatterConfig,
   D3BoxConfig,
   D3HistogramConfig,
@@ -564,6 +657,9 @@ import type {
   D3BinderConfig,
   D3BinderResult,
   DataAccessor,
+  AreaTraceType,
+  BarMarkTraceType,
+  LineMarkTraceType,
   SegmentedTraceType,
   // MAIDR schema re-exports
   MaidrData,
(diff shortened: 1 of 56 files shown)


jooyoungseo:
Round 2. CI was fully green on `3b2798c` (14/14). The automated review job exited `success` but never produced a review — its comment is still stuck mid-checklist at "Deep-dive review of remaining binders (in progress)" — so there were no findings to act on. I self-reviewed the diff instead and pushed two fixes.

**Fixed**

- `feat(d3): accept diverging bars as a react and panel chart type` — `bindD3Diverging` was exported and documented, but `diverging` was the only binder missing from `D3PanelChartSpec`. A population pyramid was therefore unreachable through `<MaidrD3>`, `useD3Adapter`, `bindD3Facets` and `bindD3Subplots`, while every sibling variant (`mosaic`, `polarArea`, `forest`, `volcano`, …) was reachable. Adding the arm to the union made the three exhaustive switches require the case; the panel path re-types the shared segmented core with `TraceType.DIVERGING` exactly as `mosaic` does, so a value's sign keeps reading as a side rather than a magnitude.

- `fix(d3): clear stale mark stamps before restamping ordered selectors` — `stampOrderedSelectors` cleared the index attribute only on the elements it was about to re-stamp, which clears nothing, since the following `setAttribute` overwrites it regardless. Any element still under the root holding a stamp the new payload does not re-issue kept its old index, and the emitted `<selector>[data-maidr-*-index="N"]` then resolved to two marks for one datum. MAIDR's own highlight clones are the reachable case: `queryD3Elements` skips `data-maidr-owned` elements so a clone never joins the payload, but the emitted selector does not exclude it — and because the selector count still matched the datum count, nothing downstream could detect the mispairing. Now every stamp under the extraction root is cleared first, scoped so a panel never clears a sibling's. The regression test was confirmed to fail against the old code.

**Declined**

- The two new example pages (`d3-bindarea.html`, `d3-bindfunnel.html`) are not linked from the example gallery. The gallery list lives in `scripts/build-site.js`, outside this PR's scope, and the pre-existing `d3-bindfacets.html` / `d3-bindsubplots.html` are unlinked the same way — so this is an existing gap rather than a regression here, and worth a separate change that adds all four.

Local `lint:fix`, `type-check`, `build` and the full suite (2818 + 73 tests) pass at `050d060`. CI for that commit had not finished when this was written.

---

PR #882, opened 2026-08-14
fix(plotly): scope a candlestick's selector to its own panel and trace

### LINKED ISSUE #881: fix(plotly): the candlestick selector also matches the rangeslider's copy of the chart
## The selector

```ts
// src/adapters/plotly/selectors.ts
case TraceType.CANDLESTICK:
  // Candlestick reuses box plot rendering: boxlayer > trace.boxes > path.box
  return `${prefix}.trace.boxes .box`;
```

`prefix` here is built from the trace's axis references. For a single-subplot figure that is empty, so the selector is global.

## What it matches

Plotly gives a candlestick chart a **rangeslider by default**, and the rangeslider holds a complete second copy of the plot. Measured in Chromium, 4 candles:

```
.trace.boxes .box                                       8     ** two per candle
.subplot.xy .boxlayer > .trace.boxes:nth-child(1) …     4
```

The duplicate lives at:

```
svg.main-svg > g.infolayer > g.rangeslider-container
  > g.rangeslider-rangeplot.xy > g.overplot > g.xy > g.boxlayer.mlayer.rangeplot
```

Note the ancestor carries `xy` but **not** `subplot`, which is why a `.subplot.<id>` prefix excludes it and the current prefix does not.

Setting `xaxis_rangeslider_visible=False` drops the count to 4, confirming the extra four are the slider's.

## Consequences

With 8 elements matched for 4 data points, the index-to-element mapping is off for every candle: candles 0–3 address elements 0–3, which — depending on DOM order — are split across the main plot and the thumbnail. The audio, braille and text are all correct, so nothing else reports it.

## Second issue in the same selector

`g.boxlayer` holds one group per trace, and a `go.Box` shares it with a candlestick, drawing its own `path.box`. Measured with one box and one candlestick on a subplot:

```
.subplot.xy .trace.boxes path.box                       5     (1 box + 4 candles)
.subplot.xy .boxlayer > .trace.boxes:nth-child(1) …     1     the box
.subplot.xy .boxlayer > .trace.boxes:nth-child(2) …     4     the candlestick
```

`nth-child` follows declaration order among boxlayer traces; a `go.Violin` (which uses `g.violinlayer`) and a scatter do not shift it.

## Suggested fix

```
${prefix}.boxlayer > .trace.boxes:nth-child(N) path.box
```

with `prefix` scoped to `.subplot.<id>` and `N` the trace's 1-based position among the subplot's `{box, candlestick}` traces.

`go.Ohlc` is worth checking at the same time — it is **not** in `boxlayer`. Measured, it draws into `g.ohlclayer > g.trace.ohlc > path`, so a shared selector would match nothing for it.

## Cross-reference

py-maidr builds its plotly schema in Python rather than through this adapter, and its candlestick support (xability/py-maidr#343) uses the scoped form above, verified against the same DOM.

jooyoungseo:
Closes #881.

## The defect

```ts
case TraceType.CANDLESTICK:
  // Candlestick reuses box plot rendering: boxlayer > trace.boxes > path.box
  return `${prefix}.trace.boxes .box`;
```

For a single-subplot figure `prefix` is empty, so this was effectively global. It over-matched in two independent ways, both measured in Chromium against real plotly output.

**1. The rangeslider.** Plotly gives a candlestick chart one **by default**, and it holds a complete second copy of the plot. For a four-candle chart:

```
.trace.boxes .box                                      8    ** two per candle
.subplot.xy .boxlayer > .trace.boxes:nth-child(1) …    4
```

The duplicate lives at `g.infolayer > g.rangeslider-container > g.rangeslider-rangeplot.xy > … > g.boxlayer.mlayer.rangeplot`. That ancestor carries the panel's `xy` class but **not** `subplot`, so the prefix is what excludes it. Setting `xaxis_rangeslider_visible=False` drops the unscoped count to 4, confirming the extra four are the slider's.

With 8 elements matched for 4 points the index-to-element mapping is wrong for every candle, and half the highlights land in the thumbnail.

**2. `go.Box` shares the layer**, and draws its own `path.box`:

```
one box + one four-candle trace:
  .subplot.xy .trace.boxes path.box                       5    (1 + 4)
  .subplot.xy .boxlayer > .trace.boxes:nth-child(1) …     1    the box
  .subplot.xy .boxlayer > .trace.boxes:nth-child(2) …     4    the candlestick
```

Throughout, the audio, braille and text were all correct — only a sighted reader could see the highlight was on the wrong bar.

## The fix

`candlestickSelector` scopes three ways: the subplot prefix, the trace's position among the panel's box-family traces, and a different layer for `ohlc`.

Two things I checked rather than assumed:

**Sibling counting is safe for `boxlayer`.** It is not for `.scatterlayer` — `scatterTraceScope` in this same file notes its groups are in fill z-order rather than `_fullData` order, which is why it uses the uid class instead. Measured with three interleaved traces declared out of position order:

```
_fullData: candlestick:C3, box:B1, candlestick:C2
boxlayer:  #1 boxes=2      #2 boxes=1  #3 boxes=2
```

Declaration order, not position order.

**Only same-panel traces take a slot.** Each panel has a `boxlayer` of its own. My first version counted the figure's traces — the shape `pieSelector` correctly uses, since `pielayer` is figure-level — and the two-panel test I had written caught it: the second panel's candlestick came out at `nth-child(2)` of a layer that only has one child. A selector matching nothing loses the highlight silently, so this was worth having a test for rather than discovering later.

## Second commit: `ohlc`

`go.Ohlc` carries the same four numbers and differs only in how plotly draws a bar. It mapped to no MAIDR type at all, so the trace was dropped and a chart of nothing but OHLC bars announced nothing. It draws into `g.ohlclayer > g.trace.ohlc > path` — measured, `boxlayer` selectors match zero for it — so it takes its own layer's selector and counts among OHLC traces.

Split into its own commit since it is a feature rather than the fix.

## Falsification

Each mechanism reverted in turn, against the 8 new tests.

| reverted to | failures |
|---|---|
| the original `.trace.boxes .box` | 7 |
| `ohlc` not mapped to a trace type | 3 |
| box-family set narrowed to `{candlestick}` | 2 |
| `ohlc` sharing the box layer's selector | 2 |
| no same-panel filter on the count | 1 |

## Checks

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass: **3612 tests in 266 suites**, plus the 73 ESM tests, and all 12 example builds.

## Related

The Python binding builds its plotly schema itself rather than through this adapter, and its candlestick support (xability/py-maidr#396) uses the same scoping, verified against the same DOM. Its `layer_position` was already panel-correct — it is passed only the subplot's traces — which is why the bug this PR's second-panel test caught did not exist there.

---

### DIFF
--- a/src/adapters/plotly/extractor.ts
+++ b/src/adapters/plotly/extractor.ts
@@ -326,7 +326,12 @@ function mapTraceType(trace: PlotlyTrace): TraceType | null {
     case 'histogram':
       return TraceType.HISTOGRAM;
 
+    // `ohlc` carries the same four numbers and differs only in how plotly
+    // draws a bar — a tick either side of a vertical range rather than a
+    // filled body — so a reader is told open, high, low and close either way.
+    // It draws into a layer of its own, which the selector handles.
     case 'candlestick':
+    case 'ohlc':
       return TraceType.CANDLESTICK;
 
     case 'pie':
--- a/src/adapters/plotly/selectors.ts
+++ b/src/adapters/plotly/selectors.ts
@@ -107,8 +107,7 @@ export function generatePlotlySelectors(
       return `${prefix}.heatmaplayer image`;
 
     case TraceType.CANDLESTICK:
-      // Candlestick reuses box plot rendering: boxlayer > trace.boxes > path.box
-      return `${prefix}.trace.boxes .box`;
+      return candlestickSelector(plotlyGd, traceIndex, prefix);
 
     case TraceType.PIE:
       return pieSelector(plotlyGd, traceIndex);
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Good catch — that's a real bug, and I verified it in a browser before fixing rather than taking the static argument on trust. Fixed in `d8d9637`.

**The visibility gap is confirmed.** A hidden box trace draws no group whatsoever:

```
visible: false        boxlayer children: ["trace boxes:4"]   <- the candlestick's, alone
visible: 'legendonly' boxlayer children: ["trace boxes:4"]
  nth-of-type(1) -> 4     nth-of-type(2) -> 0
```

So counting it produced `nth-child(2)` against a one-child layer and matched nothing — exactly the silent-highlight-loss this PR set out to remove, reintroduced one case over. And as you say it is reachable by ordinary use, not an exotic figure: a legend click sets `legendonly` and re-renders.

**Reusing `drawnBefore` rather than patching the loop.** You're right that my hand-rolled loop was the problem, not just its missing guard — the helper already solves this and carries the visibility check that three other selectors rely on. It takes one plotly type at a time, so the box-family layer takes one call per type and the results are summed; a trace has exactly one type, so the sum is the count:

```ts
const types = isOhlc ? ['ohlc'] : [...BOXLAYER_TRACE_TYPES];
const position = types.reduce(
  (total, type) => total + drawnBefore(gd, traceIndex, type, onThisPanel),
  0,
);
```

That also removes the shadowing you flagged, since there is no longer a local of that name.

**`nth-of-type`** — switched, for the reason you give and one more: the two are equivalent only while the layer holds nothing but trace groups, and `nth-of-type` stays correct if plotly ever puts something else among them. So it is the more robust choice as well as the file's convention.

Added `does not count a hidden trace, which draws no group`, parametrized over both `false` and `'legendonly'`. Falsified: reverting to the hand-rolled loop fails exactly that one test, 8 others still passing.

**This applied to the Python binding too**, which builds its plotly schema independently — thanks, this review reached further than this repo. `layer_position` there has the same gap and I've filed it as xability/py-maidr#399 with the reproduction. The violin support currently in review at xability/py-maidr#397 had it as well, and is worse there: the hidden violin was also *announced*, describing a shape with its name, quartiles and full curve for something the reader cannot see. Fixed in that PR before it merges.

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass: **3613 tests in 266 suites** (up one), 73 ESM tests, all 12 example builds.

---

PR #887, opened 2026-08-14
feat(observable): make Quarto's Observable Plot charts accessible

jooyoungseo:
# Pull Request

## Description

Adds an adapter for [Observable Plot](https://observablehq.com/plot), and with it MAIDR support for the `{ojs}` cells of a [Quarto](https://quarto.org) document.

A Quarto cell cannot call a binder — the cell's value *is* the chart, and the Observable runtime inserts it — so the adapter watches the page and binds each chart as it appears, including the replacements a reactive cell draws when a reader moves a `viewof` input. Two script tags in the document header, or `quarto add xability/maidr`, are the whole integration; no cell changes.

```yaml
format:
  html:
    include-in-header:
      - text: |
          <script src="https://cdn.jsdelivr.net/npm/maidr/dist/maidr.js"></script>
          <script src="https://cdn.jsdelivr.net/npm/maidr/dist/observable.js"></script>
```

### How the data is recovered

Plot labels every mark it renders (`<g aria-label="bar">`) and hangs its scales off the node it returns, so running an element's geometry back through the matching scale recovers the datum it was drawn for. A bar drawn for `3.14159` is announced as `3.14159`, not as the `3.141589999999809` the inversion actually produces.

### What it reads

| Plot mark | MAIDR trace |
|-----------|-------------|
| `barY` / `barX` | Bar, orientation from whichever axis is a band |
| `barY` / `barX` with `fill` | Stacked bar, one series per fill colour, with a legend |
| `rectY` / `rectX` with `binX` / `binY` | Histogram, either orientation, bin edges reconstructed |
| the same with a `fill` | Stacked bar over the bins — a stacked histogram |
| `dot` | Scatter, or a Cleveland dot plot on a categorical axis |
| `line`, `area` | Line / area, one series per drawn path |
| any of the above with `fx` / `fy` | One MAIDR subplot per facet, named after it |

Date axes work: values travel as epoch milliseconds with `format: { type: 'date' }` on the axis, since a trace's point types are numeric because the value drives sonification and the min/max range.

### What it deliberately does not read

Each of these could be given a plausible-looking answer, and a reader who cannot check it against the picture would have no way to know:

- **`cell` marks (heatmaps).** The magnitude lives in an 8-bit fill colour, so distinct values render identically and no inversion separates them.
- **Box plots.** `Plot.boxY` is four marks and nothing in the DOM says they belong together; read individually the interquartile box is an ordinary bar whose value is `q3 − q1`. Recognised by geometry — median inside the box, whiskers running past both its ends — so an error-bar chart that happens to use the same three mark types still reads.
- **Lines whose path misses the data** — `curveBasis`, `curveBundle`, and the step curves. Plot binds the datum indices to the path, so the mismatch is detectable.
- **A chart the tick-derived fallback cannot fit.** When Plot's `scale` function is gone (a chart revived from saved HTML), the scales are fitted from the rendered ticks; a log axis, or a quantitative axis whose labels cannot be read back, is refused rather than approximated.

## Related Issues

None filed; this adds the Quarto/Observable integration as new capability.

## Changes Made

**Adapter** — `src/adapters/observable/`: `types.ts`, `scales.ts` (inversion, float-noise removal, tick fallback), `introspect.ts` (mark groups, axis labels, facets, box-composite detection), `converters.ts` (mark → `MaidrLayer`), `selectors.ts` (stamping and element ordering), `quarto.ts` (binding, the watcher, teardown). Entry at `src/observable-entry.ts` exposing `window.maidrObservable` and the auto-start.

**Core** — `maidr:unbindchart` alongside the existing `maidr:bindchart`, answered by `disposeMaidrOnElement`. Without it a reactive document leaks a whole chart per re-render: nothing tears down an instance whose chart the host page discarded.

**Quarto extension** — `_extensions/maidr/` (`quarto add xability/maidr`), a Lua filter with `maidr-version` and `maidr-base-url` options.

**Registration** — `scripts/build.js`, `package.json` exports, `scripts/build-site.js` (six points), `docs/template.html`.

**Docs and examples** — `docs/observable.md`, `examples/observable-quarto.html` (Quarto's real `div#ojs-cell-N > div.observablehq` structure, including a reactive cell), `examples/quarto/penguins.qmd`.

n/a — the change is non-visual by construction; the chart renders identically and gains keyboard, audio, text and braille access.

## Checklist

## Additional Notes

**Verification.** `npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` pass (3669 unit + 73 esm, 50 of them new). Beyond the suite the adapter was run against charts rendered by the real `@observablehq/plot` on **0.6.11 — the version Quarto bundles — and 0.6.17**, and the built bundles were driven in Chromium against a document produced by a real `quarto render` with the extension filter: every chart cell binds with exactly correct data, and moving a `viewof` slider redraws and rebinds.

**Adversarial review found real defects in four of the five rounds run over this diff.** They are worth naming, because they share a shape — a chart that reads as perfectly fine while telling a blind reader something untrue:

- A date axis announced every point as `2024`: an ISO timestamp parsed from the front is its year.
- Line values were reported to full precision, but a path's coordinates are rounded on the way into the `d` attribute — `4321.1234` came back as `4321.1357339`. A value a hair below a power of ten had the opposite problem: `0.001` announced as `0.00099999999996`.
- A stacked histogram was read as one bin per rect: twice the bins, half of every count. A histogram binned down the *other* axis was read from the frequency axis, so every bar reported the same invented interval.
- **The adapter chose its data order independently of the order the DOM answers in.** MAIDR pairs a selector with its data by position and a selector resolves in document order, so sorting the data moved the values and left the highlight behind — every announcement correct, the wrong bar lit up. Sorting the elements did not help either, since stamping writes an attribute and not an order. The elements are now *moved* to match, and only when nothing overlaps, so reordering can never change what is painted over what.
- A stack with both a zero segment and a category one series never reaches came back one element out of step past the gap — the two are written the same way and cancel out, so the counts agreed and nothing looked wrong.
- A box plot's interquartile box was announced as a bar whose value is `q3 − q1`, while the median and whiskers went unannounced — and three places in the docs already claimed box plots were skipped.
- The tick-derived fallback, given an axis labelled `10, 100, 1k, 10k`, took it for a set of categories and returned the chart **transposed**.
- A reactive cell leaked one whole chart per frame — and the first two attempts at fixing that tore down *live* charts instead, first a cell holding two plots side by side, then a chart the runtime was still mounting.

That last one is worth a note, since it is the only piece of this that is a judgement rather than a measurement. Binding a chart moves it — the runtime replaces it with a wrapper and adopts it into React's tree on React's schedule — so "not in the document" means either *discarded by the page* or *still being mounted*, and getting it wrong in one direction leaks while getting it wrong in the other silently strips a visible chart of every way in. The DOM does answer it exactly: mid-mount a chart has **no parent at all**, while a discarded one went down inside whatever held it. And React commits into its container whether or not the page still holds that container, so a chart discarded during its own mount acquires a parent anyway and is released on the next sweep — the rule leaks nothing and tears down nothing early. Checked in Chromium on a cell redrawn 6 times at one chart per two frames and then 8 times inside a single task: 14 redraws, 3 live instances, every visible chart still mounted.

Two behaviours worth a reviewer's attention:

1. *A plain bar or dot layer is put in axis order, and its elements are moved to match* — except when the marks overlap, where draw order is kept because moving a sibling would change the picture. So on an overlapping `sort`ed chart the arrow keys will not sweep strictly left to right. The highlight always matches the announcement.
2. *Auto-init is on for any page that loads the bundle*, not only Quarto pages, since watch-and-bind is what a plain Plot page wants too. It is idempotent, skips anything already bound, and is opt-out via `window.maidrObservableAutoInit = false` — with `stopQuartoObservable()` to stop it afterwards.

https://claude.ai/code/session_019Ki1N1MqVzd3Y8uY5nRrgE


jooyoungseo:
The `MISSING_LIMIT = 3` question was a real one, and the answer turned out to be that the constant should not exist. It is gone in `61221ee`.

The counter was a guess at how long a mount takes, and it guessed against the reader: sweeps are driven by mutations, so a document whose other cells are still rendering runs the counter down in a few milliseconds while one chart is still waiting on React's commit — torn down on its way onto the screen, visible and with no way into it. That is a worse failure than the leak it was there to prevent.

The DOM answers the question exactly, so it is asked directly now. Mounting replaces the chart with a wrapper and adopts it a commit later, so **mid-mount a chart has no parent at all**, while a discarded chart went down inside whatever held it. The sweep releases a chart that is disconnected and either settled or *parented*, and leaves the parentless case alone.

Waiting there costs nothing, which is what makes it a complete rule rather than a safer guess: React commits into its container whether or not the page still holds that container, so a chart discarded during its own mount acquires a parent on commit anyway and is released on the next sweep. No constant, no premature teardown, no leak.

Verified in Chromium against the rapid-redraw case that motivated the concern — a cell redrawn 6 times at one chart per two frames, then 8 times inside a single task so several charts are inserted before any scan runs:

```
after 6 slow redraws:  3 articles — sub1, sub2, "frame 5"   (all visible, all mounted)
after 8 rapid redraws: 3 articles — sub1, sub2, "frame 7"   (all visible, all mounted)
tab order: ["sub1","sub2","frame 7","sub1","sub2"]   console: []
```

Fourteen redraws, three instances: nothing accumulated, and every chart on screen is still mounted. The `quarto render` document (7 charts) and the side-by-side cell case were re-run against the new bundles and are unchanged.

Two other things in this push:

- A regression the 0.6.11 sweep caught: a value a hair below a power of ten kept a decade too many digits, so `0.001` was announced as `0.00099999999996`. The exponent is now taken after the noise is rounded off rather than by flooring the log.
- On CI — it had not run against `1c44029` when you looked; it is running now, and `61221ee` is the commit to judge. Locally `npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass (3669 unit + 73 esm).

---

jooyoungseo:
Two review rounds landed on this thread with opposite verdicts on `sweepDiscarded` — the first called it a leak, the second traced the same code and approved it. Both were hand-traces. I ran it, and **the first round was right about the defect and wrong about the case**, which is why the second round could not reproduce it. Fixed in `75a9086`.

The trace in the leak report walks through `FIXTURES.bar`, and for that fixture it does not hold: Plot wraps a titled chart in a `<figure>`, so `cell.innerHTML = …` removes the *figure* and the `<svg>` inside it keeps a non-null `parentNode`. That is why the test passes and why the re-review found the logic sound.

But Plot only emits a `<figure>` when the chart has a title, subtitle, caption, or legend. For a bare `<svg>` — the common case — the removal lands directly on the chart:

```
parent before discard: DIV
parent after discard:  null
released:              0
```

So the defect is real, and the mechanism is the one the first report described. It is not the MAIDR instance that leaks in every configuration, but the adapter retains the whole detached SVG tree indefinitely, one per redraw.

**The fix is not a restored countdown.** The ambiguity is not in the DOM state, it is in *how the chart got there*, and the adapter is the one that knows: the runtime mounts synchronously from the `maidr:bindchart` dispatch, and mounting is what removes the chart from the page. So a chart that still has a parent when the dispatch returns has no mount in flight, and any later parentless state is a removal rather than a commit that has not run yet. Recorded at bind time, required before the sweep waits.

| state | parent | mount in flight | sweep |
|---|---|---|---|
| mid-mount | none | yes | waits |
| commit landed, page discarded it | container | — | releases |
| bare `<svg>`, `innerHTML` discard | none | no | releases |
| `<figure>`-wrapped, `innerHTML` discard | figure | no | releases |

The regression test uses the `scatter` fixture deliberately, and the commit says why: `bar` hides this case. I checked it is not vacuous by reverting the guard — the new test goes red and the mid-mount test stays green, which is the pair that pins both directions.

Re-verified after the change: lint, type-check, build, and `npm test` (3669 unit + 73 esm) pass; in Chromium the `quarto render` document still binds 7/7, side-by-side teardown still fires exactly twice, the leak check is 8 rendered → 1 retained, and the rapid-redraw stress is still 3 live instances after 14 redraws with an empty console.

One process note, since it produced two confident and contradictory reviews: both rounds said they could not run `npm test` or `npx tsc` in the review sandbox. A hand-trace against a fixture is worth what the fixture is worth, and here the choice of fixture was the whole difference.

---

jooyoungseo:
Correcting my previous comment. I said the bare-`<svg>` leak was "the dominant real-world case". I then measured it in Chromium instead of reasoning about it, and that claim is wrong.

Driving a cell at one redraw per two frames — a slider drag, and tight enough that no sweep ever sees a chart connected between its bind and its discard:

| configuration | before `75a9086` | after |
|---|---|---|
| `observable.js` alone | 12 bound, **0 released, 11 retained** | 12 bound, 11 released, **0 retained** |
| `observable.js` + `maidr.js` (the documented setup) | 12 bound, 11 released, 0 retained | 12 bound, 11 released, 0 retained |

So the defect is real and browser-reproducible, but **not in the supported configuration**. When the runtime is loaded, mounting interposes its container between the chart and the cell, so the OJS `replaceChild` removes the *container* and the discarded chart keeps a parent — which the old rule released on. My reading of the OJS Inspector source was right about what it does and wrong about what is sitting in `firstChild` by the time it runs.

That means the second review round's approval was correct for the setup this PR actually ships, and my framing of it as a missed defect was unfair. The first round's mechanism was right; my claim about its reach was not.

The fix stays, for reasons that do not depend on the leak being reachable today:

- It removes an unbounded wait that was only safe by accident — the old rule was correct solely because the runtime happens to interpose a wrapper, which is not a property the adapter can see or rely on.
- `bindObservablePlot` is public API. A caller who binds charts without the auto-init/runtime pairing gets the retention above, and the numbers show it is not hypothetical.
- It replaces "wait on anything parentless" with the thing actually being asked — *is a mount in flight* — which the adapter can answer exactly, because it is the one that dispatched the event that starts one.

The regression test is jsdom, where there is no runtime and the case is therefore direct. I have added nothing to the branch for this; `75a9086` already contains it.

---

jooyoungseo:
**Do not merge this yet.** A fourth adversarial round found defects that every review so far — mine included — was structurally unable to see, and I have reproduced the two worst against MAIDR's real trace models.

Every test in this PR asserts the *schema the adapter emits*. None of them construct the trace that consumes it. So the adapter has been checked against my belief about the grammar rather than against the grammar, and on horizontal orientation that belief was backwards.

`AbstractBarPlot` reads the bar's **value from `point.x` when `HORIZONTAL`** and from `point.y` otherwise (`src/model/bar.ts:86`), a convention the plotly adapter documents explicitly at `src/adapters/plotly/extractor.ts:1525-1527`. This adapter emits `{x: category, y: value}` for both orientations. Feeding its own output into the real models:

```
bar  VERTICAL   → barValues=[[20,14,23]]        min=14  max=23     ok
bar  HORIZONTAL → barValues=[[null,null,null]]  min=null           ✗
hist VERTICAL   → barValues=[[4,4,3,2,2]]                          ok
hist HORIZONTAL → barValues=[[1,3,5,7,9]]                          ✗
```

So for a `barX` chart: `Number('Mon')` is `NaN`, there is **no sonification, no braille, and the description reports min/max as "missing"** — the chart looks bound and is not readable. For a horizontal histogram the bin midpoints are sonified as if they were the frequencies, so the pitch contour is a straight ramp unrelated to the data. Both are worse than the confidently-wrong-number class this PR was written to avoid, because the announcement is not merely wrong, it is absent.

The same round raised five more, which I have not yet verified myself:

| | finding |
|---|---|
| high | stacked `areaY` announces each series' cumulative edge, not its value (Services `[180,205,215]` for data `[60,70,65]`) |
| high | the tick fallback *approximates* a log axis when the axis drew only two labelled ticks, announcing `-728.74` for a point at `12` — the PR claims it refuses |
| med | a reversed or descending value axis strips the minus sign from every negative bar: `-4.2` announced as `4.2` |
| med | ranged/floating bars (`y1`/`y2`) announce the bar's height, a number in neither endpoint |
| med | `resolveSvg` picks a legend swatch, so `legend: true` — the shape of the documented example — never binds |

The lesson is about the tests, not just the code: a fixture-based adapter test can only ever check that the JSON matches what the author expected the JSON to be. The fix is to run the adapter's layers through the real `BarTrace`/`Histogram`/`SegmentedTrace` and assert on `barValues`, min/max and the rendered description — which is what caught this, and what should have been here from the start.

I am working through these. The branch stays open.

---

jooyoungseo:
Four of the six now fixed. I reproduced all five open findings against real Plot renders before touching anything — every one of them was real, and your `points.slice(1, -1)` reading of the two-tick case was exactly right.

**Fixed in `f602f89`:**

| finding | before | after |
|---|---|---|
| `legend: true` never binds | `NULL` | binds, `[1, 3]` |
| two-tick log axis fitted as linear | `-728.74` for a point at `12` | refused |
| reversed axis strips the sign | `Ops: 4.2` | `Ops: -4.2` |

On the legend one — you flagged it as plausible-but-unconfirmed and guessed the continuous `ramp` case. It is worse than that: a *discrete* legend does it too. `Plot.plot({color: {legend: true}, marks: [barY(…, {fill: 's'})]})` returns a `<figure>` containing **three** `<svg>`s, the first two being 15px swatches. So the documented example was the failing case.

Each ships with a fixture captured from a real render, and the sign fix is asserted through the emitted values because `BarTrace` reads them directly.

**Two still open, and I am not merging until they are resolved:**

- **Stacked `areaY` announces the cumulative edge.** Confirmed: series values `[60, 70, 65]` are announced as `[180, 205, 215]`. Your suggestion is to de-stack by subtracting the series below, but I think there is a better fix available: an area path is a closed band, so its `d` holds the top edge *and* the bottom edge reversed. Taking the difference between them is the series' own value by construction — correct for a stacked area and for a baseline-anchored one alike, with no need to detect which it is. That is worth doing properly rather than heuristically, so it is next.
- **Ranged `y1`/`y2` bars announce the height.** Confirmed: a waterfall segment drawn 140 → 90 is announced as `+50`, a 50-unit loss spoken as a gain. I agree with your read that MAIDR's bar grammar has no two-value point, so refusing is the right answer — but the detection has to distinguish a floating bar from a *stacked segment*, which is also not baseline-anchored and where the height is the correct value. Getting that wrong would break stacked bars, so it needs care.

Gate on `f602f89`: lint, type-check, build, 3678 unit + 73 esm. Browser re-verified — the real `quarto render` document binds all four chart cells (cells 1 and 4 are the data-load and `viewof` cells, no charts) and the slider still redraws and rebinds; alignment 6/6, box detection 9/9, teardown unchanged.

---

jooyoungseo:
All six findings from the fourth round are now fixed. The last two are in `3f44f3b`.

**Stacked area.** The band was already in the path and was being thrown away: an area's `d` is a loop whose second half is the lower edge walked back, and `parsePathVertices` kept only the first half. Taking the difference between the two edges gives the series' own value **by construction**, so nothing has to detect whether the mark is stacked — a single-series area subtracts a baseline of zero and is unchanged. That is why I did not take the de-stacking approach you suggested: subtracting the series below works, but it needs a stacking heuristic to decide when to apply it, and this needs none.

```
Hardware [120,135,150]   Services [60,70,65]   Support [30,25,40]     (data)
Hardware [120,135,150]   Services [60,70,65]   Support [30,25,40]     (announced)
single-series area [2,5,1] — unchanged
```

**Ranged bars** are refused, as you and I both concluded. One correction to the detector worth recording: my first attempt refused a mark when *no* rect touched the baseline, and that missed the waterfall entirely — its opening step is drawn from zero like any bar, so one rect did touch. It refuses when *every* rect does not touch. A stacked segment is off-baseline too and must not be caught, which it is not: a stack is built from a colour channel, so its segments carry a series, and every column still has one segment standing on zero.

Round-four scoreboard, all verified by running real Plot charts rather than by reading the code:

| # | finding | resolution |
|---|---|---|
| 1 | horizontal layers transposed → silent charts | fixed, `63ea305` |
| 2 | `legend: true` never binds | fixed, `f602f89` |
| 3 | two-tick log axis fitted as linear | refused, `f602f89` |
| 4 | reversed axis strips the sign | fixed, `f602f89` |
| 5 | stacked area announces cumulative edges | fixed, `3f44f3b` |
| 6 | ranged `y1`/`y2` bars announce their height | refused, `3f44f3b` |

Gate on `3f44f3b`: lint, type-check, build, 3681 unit + 73 esm. Browser: the `quarto render` document binds all 4 chart cells and the slider still redraws and rebinds, 7/7 charts instrumented, alignment 6/6, box detection 9/9, side-by-side teardown exactly 2, leak 8 → 1. Every chart the adapter read before still reads, and the only marks now returning `null` are the ones meant to.

---

jooyoungseo:
The coloured-waterfall finding is right, and it is a good catch — the exemption I wrote to spare stacked segments let the *common* form of the very chart it was meant to refuse straight through. Fixed in `1f80af7`.

Colour was never the right question. A stack is a mark with **more rects than categories**, because its segments share a category and pile up; a coloured waterfall still has one rect per category. That is the same test `stackedGrid` already applies, so the two now agree by construction rather than by a separate heuristic:

```
coloured waterfall : null (refused)
stacked bar        : stacked_bar  [[A 1, B 3], [A 2, B 4]]
coloured plain bar : bar          [a 5, b 3]
```

A fixture with `y1`/`y2` + `fill` pins it, alongside the uncoloured one.

**Also in this push, two things you did not ask for but that belong with it:**

- A stacked area is now typed `STACKED_AREA` rather than `AREA`. `AreaTrace` announces the running total and each band's share of it only when the layer says it is stacked, and it sums the series itself — which is exactly what reading the band gives it. Typed as a plain area the values were right and half the chart's meaning was missing. Verified in Chromium, since jsdom has no `SVGPathElement` and the area trace resolves its highlight through one: `type: stacked_area`, three series, each selector resolving to exactly one path, instrumented, no console errors.
- **CI caught a test of mine that was green locally and red there.** `settle()` in the watcher tests slept once for 5ms, which is enough only on an idle machine — the scan is scheduled through a `requestAnimationFrame` stub and a binding pass can schedule another. It now waits several turns.

That last one had a second cause worth naming: my local run was 3681 tests and CI's was 4047, because **CI tests the merge with `main` and I was testing the branch alone**. `main` had moved nine commits. I have merged it in, and the local suite now reports 4048 — CI's number plus the test added here.

Gate on `1f80af7`: lint, type-check, build, 4048 unit + 73 esm. Browser: `quarto render` document binds all 4 chart cells with the slider still rebinding, 7/7 instrumented, alignment 6/6, box detection 9/9, side-by-side teardown exactly 2, leak 8 → 1, stacked area correct.

---

jooyoungseo:
Your "worth a fixture, not a blocker" edge case is real and reachable, so I fixed it rather than filing it. `fe43d2d`.

`Plot.areaY(data, {x, y1, y2, fill})` draws independent overlapping bands through the ordinary API — no stack transform — and `series.length > 1` called them stacked. Confirmed against a real render before changing anything.

The fix does not need to know about Plot's transform at all, because **the drawn edges say it outright**: bands stack exactly when each one's floor is the ceiling of the one below. Both edges are already parsed — that is what reading the band's height needs — so the test is a comparison of coordinates I have in hand, in pixels, which is the unit the tolerance belongs in.

```
stacked      stacked_area  [[120,135,150],[60,70,65],[30,25,40]]
independent  area          [[10,15,12],[35,40,33]]
single       area          [[2,5,1]]
```

The values were right in every case already — each band's height comes from its own drawn edges — so this only changes what the reader is *told the chart is*. One number did improve, and it is the kind of thing this branch has been chasing throughout: a band's height is the difference of two coordinates the path serializer had already rounded, so it carries both errors and has to be rounded against twice the quantum. A band of exactly 33 was coming back as `33.0001`. Fixed in the same commit, with the fixture asserting `33`.

Gate on `fe43d2d`: lint, type-check, build, 4049 unit + 73 esm. Browser: `quarto render` document binds all 4 chart cells with the slider rebinding, 7/7 instrumented, alignment 6/6, box detection 9/9, teardown exactly 2, leak 8 → 1, stacked area still `stacked_area` and instrumented.

CI on `1f80af7` had not finished when you reviewed and is now superseded; I will confirm green on `fe43d2d` before merging.

---

Issue #885, opened 2026-08-14
feat(adapters): the 21 chart types deferred out of the per-adapter coverage PRs


jooyoungseo:
## Why this exists

The eleven adapter issues #859-#869 each carried a **Deferred to a follow-up** section listing the trace types their PR deliberately did not implement, with the reason for each. Every one of those issues was auto-closed by its PR merge, so those 21 items are now closed along with the research behind them and are effectively untracked. This re-opens them as one list.

## The common blocker

These are not gaps in effort. In each case the library **can** draw the chart, but nothing in its own configuration distinguishes that chart from a plainer one it already renders. An error bar in amCharts is a range column; a survival curve in Highcharts is a stepped line; an alluvial in amCharts is a sankey. Detecting them by heuristic would misannounce the plainer chart far more often than it caught the specialised one -- and a chart that sounds confident while saying something false is worse than one MAIDR declines.

So each needs an **author-facing opt-in**: an adapter option naming the layer, plus in some cases data fields carrying semantics the chart never holds (a study weight, a censoring flag, a significance threshold). That API does not exist yet, and designing it once across the adapters is the real unit of work -- which is why it did not belong inside eleven separate PRs.

## Chart.js (`chartjs`)

<sub>from #860, shipped in #872</sub>

- [ ] **`VOLCANO`** -- Chart.js `type: 'scatter'` (core) — a volcano plot is a plain scatter of log2FC against -log10(p)
  <details><summary>why it was deferred</summary>

  Renderable natively but not inferable: it is indistinguishable from any scatter, so it needs a `plugins.maidr` trace-type declaration plus `thresholdOptions` (MaidrLayer, grammar.ts:1195-1196) for the fold-change and significance cutoffs. `VolcanoPoint extends ScatterPoint` (grammar.ts:895) with the point's identity, which is the actual payload of the chart and which Chart.js does not model — the author would carry it as an extra prop on each `{x, y}` datum, which `datasetToScatterPoints` (extractor.ts:930-942) currently drops (it copies only x/y/r). So the work is: opt-in field, threshold options passthrough, and preserving a per-point label through the scatter path.

  </details>
- [ ] **`MANHATTAN`** -- Chart.js `type: 'scatter'` with one dataset per chromosome (core), typically plus a threshold line dataset
  <details><summary>why it was deferred</summary>

[...]

PR #888, opened 2026-08-14
feat(d3): read the chart types deferred out of the coverage PR

jooyoungseo:
## Description

Adds the choropleth map to the D3 adapter — the one trace type the D3 coverage
PR deferred that is implementable without a new dependency.

A choropleth in D3 is `d3.geoPath()` over a projection: one `<path>` per region,
each bound to the GeoJSON feature it was drawn from. `bindD3Choropleth` reads
those bound features and emits a `ChoroplethPoint[]` layer, and — because a
trace that sonifies but does not highlight is not done — stamps the highlight
selectors in the same pass.

Per the adapter convention's Rejected list, D3 declares by binder: calling
`bindD3Choropleth` **is** the declaration. No co-located `maidr` block was added
to D3, the public `DataAccessor<T>` type is unchanged, and this adapter makes no
`validateDeclaration` call.

## Related Issues

Part of #885

## Changes Made

**Trace type added**

- **`choropleth`** — `bindD3Choropleth` in `src/adapters/d3/binders/choropleth.ts`,
  configured by a new `D3ChoroplethConfig` (`selector`, `region`, `value`, `lon`,
  `lat`) in `src/adapters/d3/types.ts`, exported from `src/adapters/d3/index.ts`,
  and wired into the multi-panel binder (`binders/subplots.ts`), the React hook
  (`useD3Adapter.ts`) and `<MaidrD3>` through the existing `D3PanelChartSpec`
  union.
  - *Payload* — `ChoroplethPoint[]` exactly as `src/type/grammar.ts` defines it:
    `x` (region name), `y` (value), optional `lon`/`lat` in degrees. A string
    accessor resolves on the feature and then on its `properties`, which is
    where a place name and a joined value actually live on a real map; a
    function accessor receives the whole feature.
  - *Centroids* — `lon`/`lat` are emitted only when both resolve **and** both
    fall inside ±180°/±90°. Anything else is dropped with a warning naming
    `d3.geoCentroid(d)` (unprojected degrees) against
    `d3.geoPath().centroid(d)` (projected pixels). A map that places only some
    of its regions warns too, since `ChoroplethTrace` needs the pair on every
    region before it bands the map.
  - *Highlighting* — one stamped `data-maidr-choropleth-index` selector per
    announced region (a single scoped selector for a one-region map), so the
    announced region and the lit shape stay in step even when a "no data"
    region's path sits between two announced ones. Pinned by a test that builds
    the real `ChoroplethTrace` and asserts the highlighted element after a move.
  - *No-data regions* — a region whose value join missed is omitted from the
    payload and from the selectors rather than zero-filled; `ChoroplethPoint.y`
    is required and a `y` of 0 would sonify "no data" as the lowest value on
    the map. A region the `region` accessor cannot name throws rather than
    degrading to an ordinal, since every reading of a map is by name.

**Not added**

- **`neighbors` on the choropleth layer** — deferred. Adjacency is not
  recoverable from rendered SVG paths, and deriving it needs shared-border
  topology (`topojson.neighbors`); `topojson-client` is not a dependency of this
  repository and this change does not add one. The grammar explicitly sanctions
  a layer that declares no neighbours, so the map still reads — as a spatial
  walk rather than with borders. `docs/d3.md` tells authors to hand-author the
  layer JSON in the meantime.

**Docs and examples**

- `docs/d3.md`: new "Choropleth Map" section, a binder-table row, and
  `D3ChoroplethConfig` in the type list.
- `examples/d3-bindchoropleth.html`: `d3.geoMercator` + `d3.geoPath` over inline
  GeoJSON, joined by name, with `lon`/`lat` supplied via `d3.geoCentroid`.

**Tests**

- `test/adapters/d3/choropleth.test.ts`: 14 cases covering properties-level
  defaults, function accessors, the omitted no-data region and its selectors,
  projected-coordinate rejection, half-resolved centroids, the partial-centroid
  warning, an explicit `region` accessor, both refusals (no name; no value
  anywhere), single-region scoping, the no-match error, field inference from a
  region that carries a value, and the `Figure`/highlight integration.

Nothing outside `src/adapters/d3/`, `test/adapters/d3/`, `examples/` and `docs/`
is touched. `src/type/declaration.ts` and
`src/adapters/shared/traceDeclaration.ts` are unchanged.

N/A — no visual change. `examples/d3-bindchoropleth.html` renders the map this
binder reads.

## Checklist

The manual-testing box is unticked deliberately: the automated gate was run in
full and is green, but `ManualTestingProcess.md` was not walked with a screen
reader. Verification actually run on this branch:

- `npm run lint:fix` — clean, no files changed
- `npm run type-check` (`tsc --noEmit` + `tsconfig.ci.json`) — clean
- `npm run build` — all 12 bundles built
- `npm test` — 269 suites / 3784 tests passed (unit), 7 suites / 73 tests
  passed (component). No test skipped, no snapshot written.

## Additional Notes

Notes below matter to the sibling adapter PRs on #885.

**Foundation gaps (reported, not changed).** Neither
`src/adapters/shared/traceDeclaration.ts` nor `src/type/declaration.ts` was
modified.

1. `resolveFieldRef` reads `row[name]` flat, with no notion of nesting. The D3
   row source the convention names is `element.__data__`, which for a
   choropleth is a GeoJSON `Feature` whose region name and joined value live one
   level down in `feature.properties` — so a `ChoroplethDeclaration.region` or
   `.value` could not name them through the shared resolver. D3 is exempt (it
   declares by binder), so this cost nothing here and a feature-then-properties
   lookup was implemented locally. It will bite whoever wires Vega-Lite
   `geoshape`, or any other adapter whose row is a GeoJSON feature.
2. `FIELD_REF_FALLBACKS.value = ['x', 't', 'position']` is the ridgeline chain
   and is actively wrong on a map: a region table that puts the place name in an
   `x` column resolves the **name** as the shaded value.
   `ChoroplethDeclaration.value`'s doc comment says as much and tells the author
   to name the field, but the table offers no map-appropriate chain (`rate`,
   `density`, `count`) and has no entry for `region` at all — so an undeclared
   choropleth resolves nothing in a declaration-driven adapter. This binder uses
   map-appropriate chains documented on `D3ChoroplethConfig` instead: region →
   `name`, `NAME`, `name_long`, `admin`, `state`, `id`, `label`, `x`; value →
   `y`, `rate`, `density`, `count`.
3. Nothing in the grammar or the declaration can express a region that is
   **drawn but carries no value** — `ChoroplethPoint.y` is required and there is
   no null. That is the ordinary case on a real map (the "no data" colour).
   Those regions are omitted here, per semanticsSpec §3's "never substitute a
   zero", but a declaration-driven adapter meets the same question with no
   guidance and could plausibly answer it differently. Worth a sentence in the
   spec so the eight adapters do not diverge.

**Spec corrections.**

- The D3 mandate is right that `d3.geoPath().centroid(d)` returns projected
  pixels and `d3.geoCentroid(d)` is the correct call, and right that
  `topojson-client` is not a dependency (verified absent from `package.json`).
  It understates one thing: `d3` itself is only a `devDependency`, and no D3
  binder imports d3 at runtime (`grep "from 'd3" src/` is empty) — so computing
  centroids inside the binder was never an option under any dependency
  arrangement. Accessors are the only route, which is why the geoCentroid
  instruction lives in the config's doc comment rather than in a call.
- The mandate says `x` = region name via `d.properties.name` and `y` = the
  joined value. Correct, but incomplete where it matters: `properties` is also
  where the **value** usually is, so a string accessor must resolve on the
  feature *and* on its `properties`. Every other D3 binder's
  `resolveAccessor`/`inferAccessor` is a flat key lookup on the datum, and on a
  GeoJSON feature that only ever sees `type`, `id`, `geometry`, `properties` — a
  straight reuse would have made the defaults resolve nothing on every real map.
- All line references in the mandate check out against current `main` (the
  binder export block, `DataAccessor<T>`, and the survival/forest/manhattan/
  volcano/ridgeline/hexbin/boxen config types), with one drift:
  `D3ParallelConfig.dimensions` is at `types.ts:1302`, not `:1298`.

**Implementation note for reviewers.** Field inference samples the first feature
that carries a value key rather than the first feature drawn. A no-data region
can be drawn first, and a feature the join missed names none of the keys the
join would have written — a first-datum sample would refuse an entire working
map on the strength of one region. That has its own test case.

---

### DIFF
--- a/docs/d3.md
+++ b/docs/d3.md
@@ -233,6 +233,7 @@ When you omit an accessor, the binder tries a sensible default (`x`, `y`, `fill`
 | `bindD3Ridgeline` | Ridgeline / joy plot | `<path>` density curve per group | [Ridgeline](examples.html) |
 | `bindD3Hexbin` | Hexbin density | `<path>` hexagon per bin | [Hexbin](examples.html) |
 | `bindD3Contour` | Contour / density field | `<path>` per level | [Contour](examples.html) |
+| `bindD3Choropleth` | Choropleth map | `<path>` per region | [Choropleth](examples.html) |
 
 ## Multi-Panel Charts
 
@@ -999,6 +1000,34 @@ maidrD3.bindD3Contour(svgElement, {
 
 The level is carried on every point of its curve, which is where the grammar has room for it, and the trace announces it on the `z` axis — the field's own — rather than as a series name. A level drawn as several disjoint rings is flattened into one curve in ring order, since a payload row is a single polyline: every point announced is real, but the jump from one ring to the next is inaudible. Rows keep the order the paths were drawn in, which for `d3.contours()` is ascending threshold — the order the trace measures the gap to the adjacent level in.
 
+### Choropleth Map
+
+`d3.geoPath()` draws one `<path>` per GeoJSON feature, and the feature is what is bound to it — so a string accessor names a key on the feature **or** in its `properties`, which is where a place name and a joined value live on every real map:
+
+```js
+svg.selectAll('path.region')
+  .data(topojson.feature(us, us.objects.states).features)
+  .join('path')
+  .attr('d', d3.geoPath(projection));
+
+maidrD3.bindD3Choropleth(svgElement, {
+  selector: 'path.region',
+  title: 'Unemployment by State',
+  axes: { x: 'State', y: 'Rate' },
+  value: d => rateByFips.get(d.id),   // the join, keyed by the feature's id
+  lon: d => d3.geoCentroid(d)[0],     // DEGREES
+  lat: d => d3.geoCentroid(d)[1],
+});
+```
+
+> **`lon`/`lat` are `d3.geoCentroid`, never `d3.geoPath().centroid`.** The latter returns the centre of the *drawn* shape in projected pixels. Fed those, the arrows follow the projection's paper layout instead of the compass — north and south come out swapped on a south-up projection and meaningless on an interrupted one. A coordinate outside ±180°/±90° is dropped rather than converted by guesswork, with a warning naming the count.
+
+The centroids are what make this more than a bar chart whose categories happen to be places: with them the regions are banded by latitude and ordered west to east inside each band, so Up is north and Right is east. Without them the map is read as a region list in drawn order — a poorer reading, but the one the data supports. It is all or nothing, so a map that resolves centroids for some of its regions and not the rest warns and falls back to the list.
+
+A region the value join missed is left out of the payload — it is drawn in the "no data" colour and carries no number, and a `y` of 0 would sonify it as the lowest region on the map. Its path is left out of the highlight selectors with it, so the announced region and the lit shape stay in step; the selectors are stamped per region for exactly that reason.
+
+`neighbors` is not read. Adjacency is not recoverable from rendered paths, and deriving it needs shared-border topology (`topojson.neighbors`) that this package does not depend on. A layer that declares none keeps the spatial walk and is simply told nothing about borders — to get the border readings (the neighbour rotor, the sharpest borders, the cluster of high regions), author the layer's JSON directly with a `neighbors` array per region.
+
 ### Gauge / Bullet Chart
 
 A drawn gauge binds one number, so the rest of the reading is config — and it *is* the reading: "73" means nothing without the range it sits in, the target it was aiming at, and the band it lands in:
@@ -1080,6 +1109,7 @@ import type {
   D3HistogramConfig,
   D3HeatmapConfig,
   D3CandlestickConfig,
+  D3ChoroplethConfig,
   D3ContourConfig,
   D3HexbinConfig,
   D3ParallelConfig,
--- a/examples/d3-bindchoropleth.html
+++ b/examples/d3-bindchoropleth.html
@@ -1,0 +1,115 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <title>MAIDR + D3.js Choropleth Map Example</title>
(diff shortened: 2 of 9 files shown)


jooyoungseo:
Both review observations settled in `47af859`.

**Centroid-drop denominator (obs. 2) — fixed.** The warning now divides by `data.length` rather than `elements.length`. A region whose value join missed `continue`s before the centroid check runs, so it was never eligible to be counted in the numerator; including it in the total offered a figure no region in the warning could have come from. Pinned by a new case that drops one centroid out of two shaded regions with an unshaded one drawn between them, and asserts the message says "1 of the 2".

**Symmetric inference sample (obs. 1) — declined, with a comment added instead.** The suggested tweak — also prefer a feature carrying a region key — does not reach the scenario it is aimed at. `id` is itself in `REGION_KEYS`, and every feature `topojson.feature()` emits carries one, so the feature that the described failure samples (has a value, lacks `name`, has `id`) already satisfies a region-key test. `find` returns the same element and the accessor still resolves to `id`.

Reaching it would mean ranking the region key *across* features — preferring a `name` that some other feature carries. That trades a silent fallback for a hard throw on any map whose odd feature names a key the rest do not: the accessor would be set to a key that resolves nothing on the majority, and `buildNoRegionError` refuses the whole map. Turning a map that renders today into one that throws is the worse failure of the two, and both behave identically on the uniform `properties` every real GeoJSON export has.

So the sample stays where it is, and the reasoning is now recorded in `inferenceSample`'s doc comment rather than left for the next reader to re-derive.

Full gate re-run on the branch: `lint:fix` clean, `type-check` clean, `build` 12/12 bundles, `npm test` 269 suites / 3785 unit tests and 7 suites / 73 component tests passing. All 14 checks green at `47af859`.

---

PR #890, opened 2026-08-14
feat(chartjs): read the chart types deferred out of the coverage PR

jooyoungseo:
# Pull Request

## Description

The Chart.js adapter learns to read a co-located `maidr` declaration written straight onto a dataset, and uses it to reach the trace types no heuristic can reach from the drawing alone. A Chart.js `scatter` is the same drawing whether it is a plain scatter, a volcano plot or a Manhattan plot — what separates them is meaning the author holds and the chart config does not state. This adds the slot for the author to say so, and the readings that slot unlocks.

Chart.js has no reserved place for third-party metadata, but it passes dataset properties it does not know through untouched — the same mechanism a survival curve's `censored` datum already rides on — so the block lives on the dataset:

```js
datasets: [{
  label: 'chr1',
  data: [{ x: 1e6, y: 8.2, snp: 'rs1234' }],
  maidr: { type: 'manhattan', label: 'snp', significance: 7.3 },
}]
```

The block is validated once per extraction through the shared `validateDeclaration` from #886; field names resolve through the shared `resolveFieldRef` chains. Sonification and highlighting are wired in the same change — a merged Manhattan's columns still reach the elements Chart.js actually drew, in whichever dataset they live.

## Related Issues

Part of #885

## Changes Made

**Trace types added**

- **VOLCANO** — a `scatter`/`bubble` dataset carrying `maidr: { type: 'volcano', … }` is read as `TraceType.VOLCANO`. `label`/`group` resolve off each datum through the shared field-ref chains; `significance`, `significanceDirection` and `effect` map straight into `MaidrLayer.thresholdOptions`. `merge` defaults to false, since a volcano's sibling datasets are usually up-regulated / down-regulated / unchanged — three things a reader wants told apart.
- **MANHATTAN** — the same reading with `merge` defaulting to true, so a chart drawn as one dataset per chromosome becomes a single navigable cloud rather than 22 layers to switch between. Every following sibling dataset drawn the same way that carries no block of its own is folded into the declaring layer; the merged layer records its full dataset list in `layerDatasetIndices`.
- **SURVIVAL via the dataset block** — a dataset-level `{ type: 'survival' }` on a `line` chart now reaches the existing `extractSurvivalLayer` path unchanged. No duplicate code path; the block is simply a second door onto the reading `plugins.maidr.traceType` already opened.
- **SCATTER (`point`)** — accepted as naming what the chart already is, honouring `title`, `name` and `merge`. `label` is deliberately not carried: `ScatterPoint` has nowhere to put it, as `ScatterDeclaration.label` documents.

**Trace types not added, and why**

- **Chart.js module augmentation** (`declare module 'chart.js' { … }`) — **not shipped.** `chart.js` is an optional peer dependency and is not installed, so the augmentation fails type-check with TS2664 (`module 'chart.js' cannot be found`), verified empirically. Shipping it needs `chart.js` added as a devDependency. Typed authors still get the field through MAIDR's own `ChartJsDataset.maidr`; plain-JS authors get the runtime validator, which is the defence that actually matters here. Worth a follow-up.
- **Any type with no Chart.js construct behind it** (hexbin, choropleth, and the rest of the union that belongs to other libraries) — a block declaring one warns and the chart is read as if it carried none, rather than half-building a trace the library cannot draw.

**Supporting changes**

- `ChartJsDataset.maidr?: MaidrTraceDeclaration` and a documented `[column: string]: unknown` index signature on `ChartJsPointValue`, so a TypeScript author can actually write the identity column the declaration points at (see Additional Notes).
- `declaredType` extended into the full three-level precedence: dataset block → `plugins.maidr.traceType` → heuristic, with a single warning naming both when levels 1 and 3 disagree. `MaidrPluginOptions` is structurally unchanged.
- Highlight path: `TargetMaps.scatterBuckets` (`Map<string, number[][]>`, one dataset per layer) became `pointBuckets` (`Map<string, ChartJsActiveElement[][]>`), built by one `buildPointBuckets` serving scatter, volcano and manhattan. Bucket entries carry `{ datasetIndex, index }` pairs, so a column reaches the points sharing that x in whichever dataset drew them. Plain-scatter behaviour is unchanged and its existing tests pass untouched.
- Degradation and no-inversion: a block whose type has no construct here, or whose construct does not match (a volcano on a bar chart), warns naming both what was found and what the type needs, and the chart is read exactly as if no block were present. Nothing throws. `significance` / `effect` / `significanceDirection` are never synthesised — a declared volcano with no `significance` is emitted with no `thresholdOptions` at all plus a warning naming the field.
- Tests: `test/adapters/chartjs/volcano.test.ts`, 25 cases across declaration reading, field-ref resolution and fallbacks, never-defaulting-an-inversion, merge/absorption grouping, merged-layer highlight routing, unreadable declarations, and precedence against `plugins.maidr.traceType`.
- Examples: `examples/chartjs/volcano.html` and `examples/chartjs/manhattan.html`.
- Docs: `docs/chartjs.md` gains two supported-type rows, a normative "The `maidr` Block on a Dataset" section (key table, the two rules about verbatim field names and never-inferred cutoffs, merge semantics, degradation), and Volcano/Manhattan code examples.

Nothing under `src/model/`, `src/service/`, `src/state/`, `src/ui/`, `src/command/`, `src/type/`, or `src/adapters/shared/` was modified.

Not applicable — no visual UI change. The two new example pages under `examples/chartjs/` exercise the readings by hand.

## Checklist

The third box is left unticked deliberately: the automated gate is green (`lint:fix`, `type-check`, `build`, and `npm test` at 269 + 7 suites / 3796 + 73 tests, all passing, no pre-existing test modified), but the `ManualTestingProcess.md` screen-reader pass was not performed. The merged-highlight test was mutation-checked rather than trusted — restricting `buildPointBuckets` to the first dataset makes "highlights across every dataset a merged layer was folded from" fail, so it pins the routing rather than restating it.

## Additional Notes

These matter to the seven sibling adapter PRs.

**Foundation gaps** (reported, not changed — `src/adapters/shared/traceDeclaration.ts` and `src/type/declaration.ts` were left alone as the shared foundation)

1. `validateDeclaration` gives no way to distinguish "no block was written" from "a block was written and rejected". §6(b) merge turns on whether a series *carries no declaration of its own*, which is the second question, so this adapter re-reads `dataset.maidr` itself (`carriesDeclaration`) to keep a dataset with a broken block out of a neighbour's merged layer. Every adapter implementing merge will need the same workaround, or `validateDeclaration` could return a discriminated result.
2. No helper for reading a `FieldRef` into the grammar's **string** fields (`VolcanoPoint.label`, `.group`). `resolveFieldRef<T>` returns the raw value, so each adapter decides for itself whether a chromosome authored as the number `7` becomes `'7'`, and whether `''` / `false` / an object are accepted — exactly the by-adapter drift `FIELD_REF_FALLBACKS` exists to prevent. Implemented locally as `pointText` (string kept as-is, finite number stringified, everything else omitted); a second caller would justify lifting it.
3. `warnUnresolvedRef` is documented as "call this once, after reading a series, when an explicit ref resolved on no row of it", but nothing in the module can answer that question — the adapter has to scan the emitted payload afterwards. A per-series resolver that tracked whether anything resolved would remove a fiddly, easy-to-forget step from all eight adapters.
4. No shared spelling for the §7.1 "wrong construct" warning, unlike the §3 unresolved-ref message which is centralised. The Chart.js phrasing here was written by hand, so the eight adapters will very likely word this differently.

**Spec corrections**

- `deferred-chartjs.md`'s line numbers are stale after the coverage PR: `datasetToScatterPoints` is at `extractor.ts:1683-1695` (not 930-942), `extractScatterLayers` at 1643-1668 (not 890-916), `declaredType` at 593-595 (not 585-595). `per-adapter.json`'s numbers were correct.
- `per-adapter.json` notes that `ChartJsDataset` "currently ends at the `fill?` field" — true — but does not mention that `ChartJsPointValue` is a **closed** interface. Instruction 4 (carry `label`/`group` off the datum) is not writable by a TypeScript author without a change there: `{ x: 1, y: 2, gene: 'TP53' }` is rejected by excess-property checking, so the very column the declaration points at cannot be authored. A documented `[column: string]: unknown` index signature was added. The three named riders (`censored`, `yMin`, `yMax`) stay as they are, so the file still reads as a statement about what the adapter looks for under a fixed name.
- `per-adapter.json` instruction 5 says the merged layer's id "must stay resolvable by `firstDatasetIndex` in highlightTargets.ts". `firstDatasetIndex` returns one dataset index for a whole layer, which cannot express a merged layer at all — a Manhattan merged from 22 datasets needs a dataset index *per highlighted element*. The per-dataset id convention is kept (the merged layer is identified by its declaring dataset's index, so `firstDatasetIndex` still answers correctly for anything that asks), but the bucket entries carry `{ datasetIndex, index }` pairs, which is what actually makes the merged highlight work.
- `convention.md`'s slot table says the Chart.js type change is `ChartJsDataset` plus a `declare module 'chart.js'` augmentation "shipped from the maidr/chartjs entry". That augmentation is not shippable while `chart.js` is an uninstalled optional peer dependency. **Any sibling adapter told to ship a library-module augmentation should check the library is actually resolvable at type-check time first.**
- `deferred-chartjs.md` says a Manhattan is "typically plus a threshold line dataset". Nothing was done with such a dataset: a straight line drawn at the cutoff is a scatter/line dataset like any other, and reading a cutoff out of it would be inferring `significance` from the drawing, which §5 forbids. It is absorbed into the cloud as a sibling if it carries no block, so an author drawing one should give it its own block or put the significance in the declaration — noted in the docs' merge paragraph.

**Judgement calls worth a reviewer's eye**

1. `TargetMaps.scatterBuckets` was **replaced** rather than joined by a second map. Two implementations of the same `(x asc, y asc)` sort-and-group, both of which must stay in step with `ScatterTrace`'s column order, is exactly the drift `shared/selectorUtil.ts` records as its reason for existing. `scatterBuckets` was referenced only inside `highlightTargets.ts` and no test asserted on it.
2. Chart-wide `plugins.maidr.traceType: 'volcano' | 'manhattan'` is honoured as precedence level 3 for a chart whose datasets carry no block. It buys the trace type and whatever the default name chains find, and carries no cutoff — a smaller reading than a block gives, and the one the precedence table asks for.

**What was deliberately not done:** infer `significance` from anything. Chart.js has no `plotLines` equivalent, and an annotation-plugin line is not part of the chart config — a cutoff read out of a decoration would sort every point in the figure onto the wrong side of it, silently. A declared volcano with no `significance` is emitted as a truthful smaller reading: the cloud, the identities, no findings, and a warning naming the field.

---

### DIFF
--- a/docs/chartjs.md
+++ b/docs/chartjs.md
@@ -83,8 +83,10 @@ MAIDR's Chart.js adapter is a standard Chart.js plugin:
 | Normalized Area | stacked area whose categories all total 100 (or 1) | — | [Line chart](examples.html) |
 | Bump | `'line'` with `scales.y.reverse` and ranked values | — | [Bump chart](examples.html) |
 | Dot Plot | `'line'` with `showLine: false` on a category axis | — | [Dot plot](examples.html) |
-| Survival | `'line'` with `stepped` and `plugins.maidr.traceType` | — | [Survival curve](examples.html) |
+| Survival | `'line'` with `stepped` and a `maidr` declaration | — | [Survival curve](examples.html) |
 | Scatter | `'scatter'` | — | [Scatter plot](examples.html) |
+| Volcano | `'scatter'` with a `maidr` declaration | — | [Volcano plot](examples.html) |
+| Manhattan | `'scatter'` with a `maidr` declaration | — | [Manhattan plot](examples.html) |
 | Radar | `'radar'` | — | [Radar chart](examples.html) |
 | Polar Area | `'polarArea'` | — | [Radar chart](examples.html) |
 | Box Plot | `'boxplot'` | `@sgratzl/chartjs-chart-boxplot` | [Box plot](examples.html) |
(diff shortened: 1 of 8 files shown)


jooyoungseo:
Both review observations are addressed in `aec19c1c8`, each with a correction to the finding as written.

**Conflicting dataset blocks.** The asymmetry is real, but not reachable by the route named: `dumbbell` and `gantt` have no entry in `DECLARED_TYPE_CONSTRUCTS`, so a block declaring either is rejected in `readDeclarations` before `declaredType` ever sees it. What does reach it is a **mixed chart**, where `dataset.type` overrides the chart's — a `line` chart whose survival arm declares `survival` and whose `type: 'scatter'` dataset declares `volcano`. Both blocks pass their construct check, both reach the single whole-chart reading a line chart gets, and the first won silently.

Since the gap is genuine I warned rather than only documenting it: `declaredType` now names every type found and the one that wins, in the same shape as the existing block-vs-`plugins.maidr.traceType` warning. Several datasets declaring the *same* type — a survival curve's arms each saying what they are — stays the ordinary case and passes silently. The new test is mutation-checked: forcing the branch off fails it.

**The `[23 Chart.js types]` count.** Not pre-existing. `main`'s table has exactly 23 rows, so the line was accurate until this branch added the Volcano and Manhattan rows and made it stale. Corrected to 25.

Full gate re-run at the new head: lint, type-check, build, and 269 + 7 suites / 3798 + 73 tests, all passing.

---

PR #891, opened 2026-08-14
feat(plotly): read the chart types deferred out of the coverage PR

jooyoungseo:
# Pull Request

## Description

Adds the two chart types that were deferred out of the plotly.js coverage PR: **contour** (including `histogram2dcontour`) and **mosaic / marimekko**. Both are wired end to end — detection, payload, axes, and the selector/highlight path — in this one change, so neither ships as a type that sonifies without highlighting.

Contour is detected from `trace.type` with no author involvement. Mosaic is opt-in only, declared through `trace.meta.maidr` and narrowed by the shared declaration helpers merged in #886 (`readDeclarationSlot` / `validateDeclaration` / `resolveFieldRef` / `warnUnresolvedRef`), since a marimekko is drawn through plotly's ordinary stacked-bar renderer and cannot be told apart from a stacked bar by inspection alone.

Nothing outside `src/adapters/plotly/`, `test/adapters/plotly/`, `examples/` and `docs/` is touched. The shared foundation (`src/type/declaration.ts`, `src/adapters/shared/traceDeclaration.ts`) is unmodified; two gaps found in it are reported under Additional Notes rather than patched, so the sibling adapter PRs can weigh in.

## Related Issues

Part of #885

### Trace types added

**Contour** (`TraceType.CONTOUR`) — detected with no declaration from `trace.type` of `contour` and `histogram2dcontour` in `mapTraceType`.

- Curves are recomputed by marching squares over the grid, reading `calcdata[i][0].z` first so a `histogram2dcontour`'s binned grid and a transposed `contour` both work, with `trace.z` as the fallback.
- Levels come from the ladder plotly already resolved on `trace.contours.{start,end,size}`. `setContours` mutates those on the full trace during calc even when `autocontour` chose them, so authored and auto charts read the same way.
- Coordinates are always data units, never pixels: calcdata `x`/`y` first, then the authored `x`/`y`, then grid indices.
- Saddle cells are resolved by the four-corner average. A level's disjoint rings are flattened into one row in ring order; a closed ring repeats its first vertex, matching the convention in the shipped `examples/contour.html`.
- A constraint contour, an unresolved ladder, and a level the field never crosses are each refused or skipped with a warning rather than announced at invented values.
- **Highlight:** new `contourLevelSelectors()` in `selectors.ts` emits one selector per emitted curve, scoped by panel, by the trace's position among the panel's `contourlayer` groups, and by the level's place in the ladder — plotly gives every level a `g.contourlevel` whether or not the field crosses it, so a curve index would drift.
- **Axes:** the level is announced on the field's own `z` axis, named from the colorbar title when the author gave one and left unnamed otherwise, so the model says "Level" rather than a fabricated "Value".

**Mosaic / Marimekko** (`TraceType.MOSAIC`) — opt-in only, via `trace.meta.maidr`.

- A validated mosaic declaration on one of a panel's bar traces routes the panel through `extractSegmentedBarLayer` as `TraceType.MOSAIC`, beating the gantt and stacked/dodged heuristics per the precedence rules.
- Each cell carries `width` (the declared `customdata` column via `resolveFieldRef`, else the drawn `trace.width` array normalised by its own total, so counts, percentages and fractions all read the same) and `count` (declared column only — never derived from a rounded share).
- Columns are named from `trace.text`, then `trace.hovertext`, then the axis `ticktext` at the matching `tickval`, since a marimekko's `x` is a precomputed cumulative number. Failing all three, the position stands.
- **Highlight:** `TraceType.MOSAIC` joins the bar-family case in `generatePlotlySelectors`, since plotly draws a marimekko through the bar renderer into the same `.trace.bars .point > path` elements as a stacked bar.
- **Degradation:** a mosaic declared on a non-bar trace, or on a panel whose bars plotly drew with `barmode: 'group'`, warns naming both what was found and what the type needs, then falls back to the undeclared reading. A declared type this adapter does not read warns and falls back. An explicit `FieldRef` no row carries is reported through `warnUnresolvedRef` and the field is omitted.

### Trace types not added, and why

- Nothing from the deferred set was skipped. Contour and mosaic were the two types deferred out of the coverage PR for this adapter, and both are implemented here.

### Supporting changes

- `src/adapters/plotly/types.ts` — `PlotlyTrace` gained `meta`, `width` and `contours` (plus a new `PlotlyContours` interface); `PlotlyAxis` gained `ticktext`; `PlotlyCalcData.x`/`y` widened to `number | number[]`, since a 2d-map calc entry holds the whole grid's coordinate arrays, mirroring the duality already documented on its `z`.
- `test/adapters/plotly/extractor.test.ts` — 21 new cases (0 removed), in new `contour` and `mosaic` describes.
- `examples/plotly-contour.html` and `examples/plotly-mosaic.html`.
- `docs/plotly.md` — two rows in the supported-types table, two notes, a "Declaring a mosaic" section, and the two example entries.

N/A — no visual change to MAIDR's own UI.

## Checklist

The automated gate was run in full and is green: `npm run lint:fix` (no changes), `npm run type-check`, `npm run build`, and `npm test` (268 suites / 3791 tests, plus the ESM project's 7 suites / 73 tests, all passing). `npx jest test/adapters/plotly` is 212 passing. The `ManualTestingProcess.md` box is left unticked because that manual pass was not performed; the Contributor Guidelines box is left unticked because that document was not read as part of this change.

### Foundation gaps (relevant to the sibling adapter PRs)

The shared foundation was deliberately left unmodified. Two gaps are worth recording:

1. **`MosaicDeclaration` has no way to name the column holding a cell's category label.** For plotly this is the bulk of the work: a marimekko's `x` is a precomputed cumulative *number*, so the category name is not on the axis coordinate. It is recovered here from `trace.text` / `hovertext` / axis `ticktext`, which works — but an author whose names live in a `customdata` column has no way to say so. A `label?: FieldRef` on `MosaicDeclaration`, spelled like the grammar's ordinal label as `ManhattanDeclaration` and `ScatterDeclaration` already have, would cover it.
2. **`validateDeclaration` answers "is this a valid declaration of some type", with no way for an adapter to say which types it reads.** Every adapter implementing a subset therefore hand-writes the same "this adapter does not read `<type>`" warning; one is written here in `readTraceDeclaration`. Not blocking, and probably not worth a shared export until a second adapter needs the identical sentence.

### Spec corrections

- The deferred spec's contour route names `.contourlayer path.contourlevel` as the `d`-carrying elements. Checked against plotly.js 2.35.2 (`src/traces/contour/plot.js`, `createLines`): `contourlevel` is a `g`, one per level, and the paths inside are `path.openline` and `path.closedline`. The trace group is `g.contour` — `Lib.makeTraceGroups` is called with the class string `contour`, so it is **not** `g.trace.contour` the way a bar or scatter group is. The selectors in this PR follow the verified shape.
- The spec puts `extractSegmentedBarLayer` at `extractor.ts:2020`; it was at `:4069` before this change. `extractPlotlyData` at `:78` and `mapTraceType` at `:300` were both accurate.
- The per-adapter instruction lists only `meta` and `width` as the missing `PlotlyTrace` fields. Three more were needed and none existed: `contours` on `PlotlyTrace` (without it the level ladder is unreachable and the whole detection is dead), `ticktext` on `PlotlyAxis`, and `PlotlyCalcData.x`/`y` widened to accept the coordinate *arrays* a 2d-map calc entry carries.
- `convention.md`'s shared-module surface (§"Shared module") is stale against what merged in #886: the flag helper is exported as `isFlagValue`, not `isCensoredValue`; there are six exports rather than four, the extra two being `warnUnresolvedRef` and `readDeclarationSlot`; and `resolveFieldRef`'s row parameter is `unknown`, not `Record<string, unknown> | undefined`. The merged version is the better one — `readDeclarationSlot` is exactly what plotly's `meta: unknown` needs — this is a note that the prose is stale, not a complaint.

### Two judgement calls for the reviewer

1. **Contour highlighting is approximate along a curve, and deliberately so.** Plotly draws a level as whole-curve paths with no element per vertex, so per-point highlighting can only be synthetic — `LineTrace` parses the path `d` and makes circles. Plotly's own drawn vertices are also not reproducible from the grid: `find_all_paths.js` merges vertices closer together than a fraction of the average spacing, and `smoothopen`/`smoothclosed` then add control points, so no marching-squares implementation can guarantee vertex-for-vertex correspondence. That does not affect what is *announced* (the curves are recomputed), but the data-point-to-circle mapping can be rotated or offset along the curve. The selectors are emitted anyway rather than withheld (as parcoords and an unsorted pie do), on the grounds that each selector is scoped to its own level group, so the highlight always lands on the correct curve and the error is bounded to a position along it — and withholding would leave contour with no visual modality at all. `docs/plotly.md` states the limitation. If silence is preferred over an approximate pointer, deleting the `selectors:` line in `extractContourLayer` is the whole change.
2. **A single bar trace carrying a mosaic declaration is accepted**, because a spineplot is a real chart. The `barmode` check therefore only applies when a panel has more than one bar trace.

---

### DIFF
--- a/docs/plotly.md
+++ b/docs/plotly.md
@@ -201,7 +201,10 @@ For dynamically-created charts (SPAs, notebooks), a `MutationObserver` watches f
   skipped instead of being announced as one. Plotly draws a level as whole
   curves with no element per vertex, so the visual highlight is synthesised
   along the drawn curve; each level's selector is scoped to that level's own
-  group, so a highlight always lands on the level being read.
+  group, so a highlight always lands on the level being read. A fill-only
+  contour — one that turned `contours.showlines` off — is drawn as one filled
+  path per level with no curve elements at all, so it is read and sonified but
+  not highlighted.
 
 - A mosaic (marimekko) is **declared**, not detected. Plotly draws one as
   stacked `bar` traces with a per-column `width`, but `width` is ordinary bar
@@ -256,6 +259,11 @@ bar traces — it declares the layer, not the series. Then:
   from proportions alone genuinely has no counts, and multiplying a rounded
   share back out would put a number in the announcement that the data does not
   contain.
+- **`customdata` rows must be objects**, `{ share: 0.3, count: 203 }` rather
+  than `['First', 0.3, 203]`. A declared field is a column *name*, and Plotly's
+  other canonical shape — one array per point — has no names to resolve
+  against, so a declared column is reported as unresolved rather than guessed
+  at by position.
 - **The column names** come from the trace's `text`, then its `hovertext`, and
   then the axis `ticktext` at the matching `tickvals`. A marimekko's `x` is a
   precomputed position rather than a category, so without one of those a column
--- a/docs/plotly.md
+++ b/docs/plotly.md
@@ -92,6 +92,8 @@ For dynamically-created charts (SPAs, notebooks), a `MutationObserver` watches f
 | Dot Plot | `type: 'scatter'`, `mode: 'markers'`, one marker per category | [Dot plot](examples.html) |
 | Word Cloud | `type: 'scatter'`, `mode: 'text'`, array `textfont.size` | [Word cloud](examples.html) |
 | Choropleth | `type: 'choropleth'` | [Choropleth map](examples.html) |
+| Contour | `type: 'contour'` or `type: 'histogram2dcontour'` | [Contour plot](examples.html) |
+| Mosaic / Marimekko | stacked `bar` traces declaring `meta: { maidr: { type: 'mosaic' } }` | [Mosaic plot](examples.html) |
 | Subplots / Facets | multiple `xaxis`/`yaxis` pairs, `layout.grid`, or Plotly Express facets | [Subplots](examples.html) |
 
 **Notes on chart-type detection:**
@@ -183,6 +185,32 @@ For dynamically-created charts (SPAs, notebooks), a `MutationObserver` watches f
   layer carries no selectors — audio, text, braille, and navigation all work,
   visual highlighting alone does not.
 
+- A `contour` (and a `histogram2dcontour`) is read as one curve per level.
+  Plotly computes its curves while it draws and keeps none of them — the
+  calculated data holds the grid and nothing else — so MAIDR walks the grid
+  itself by marching squares, at the levels the trace states in
+  `contours.start`, `.end` and `.size`. Those are resolved by Plotly even when
+  `autocontour` chose them, so nothing is guessed: a chart whose levels are not
+  resolved is skipped rather than read at invented ones. The level is announced
+  on the field's own axis — named from the colorbar title when there is one —
+  and each point announces the distance to the nearest point on the adjacent
+  level, which is what the crowding of the lines conveys visually. Where a
+  level is drawn as several disjoint curves, they are walked as one row, in the
+  order they were found. A `contours.type: 'constraint'` trace shades the
+  region satisfying an inequality rather than drawing a level ladder, so it is
+  skipped instead of being announced as one. Plotly draws a level as whole
+  curves with no element per vertex, so the visual highlight is synthesised
+  along the drawn curve; each level's selector is scoped to that level's own
+  group, so a highlight always lands on the level being read.
+
+- A mosaic (marimekko) is **declared**, not detected. Plotly draws one as
+  stacked `bar` traces with a per-column `width`, but `width` is ordinary bar
+  styling that any bar trace may carry, so reading a non-uniform array as data
+  would announce every column's width as a share of all observations — a number
+  the chart does not contain. Writing `meta: { maidr: { type: 'mosaic' } }` on
+  one of the panel's bar traces is what says otherwise; see
+  [Declaring a mosaic](#declaring-a-mosaic).
+
 - Plotly sorts pie slices by descending value unless the trace sets
   `sort: false`, so the authored order is not necessarily the drawn order. The
   adapter reads the slices Plotly actually drew where the rendered chart exposes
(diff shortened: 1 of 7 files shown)


jooyoungseo:
Both blocking bugs are fixed in `8185f03`, and `f86a5d5` closes the test gaps — which turned up a third defect of the same family. On the three "worth a second look" items, checked against plotly.js 2.35.2:

**3. The `1e-6` tolerance is right, and the half-step reading is not.** `src/traces/contour/empty_pathinfo.js` walks `for (var ci = start; ci < endPlus(contours); ci += size)`, and `end_plus.js` is:

```js
module.exports = function endPlus(contours) {
    return contours.end + contours.size / 1e6;
};
```

`start + k*size < end + size/1e6` is `k < (end - start)/size + 1e-6`, which is the count this computes. So `start: 0, end: 10, size: 6` draws two levels (0 and 6), not three — 12 is past `end`, and plotly's tolerance is a millionth of a step precisely so that a ladder landing *on* `end` keeps it. No change needed; the comment's claim was literal.

**4. The repeated first vertex is the existing convention, not a new one.** `examples/contour.html` shipped before this PR and its closed rings repeat their first vertex — 17 points for a 16-point ring. Dropping it here would make the plotly adapter disagree with the reference example and with whatever `ContourTrace` was written against. A closed-curve concept is worth having, but it is a grammar-and-model change, and this PR touches neither.

**5. Falling back to a heatmap for a constraint contour would announce a different chart.** A `contours.type: 'constraint'` trace shades the region satisfying an inequality; the grid behind it is not what the reader is being shown. Reading every cell of it out would be confidently answering a question the chart does not ask. A warning and no layer keeps the failure legible.

**The new one.** Writing the "hole in the grid" case the review asked for showed the authored-`z` fallback converting `null` — plotly's own documented way to write a missing cell — into a `0` via `Number()`, so curves were walked around a field claiming to be zero where it has no data. Same family as the fabricated `count`: a value invented for a cell that has none. `f86a5d5` adopts plotly's own rule (`isNumeric(v) ? +v : undefined`); the calcdata path was already clean, which is why only the fallback was wrong.

Remaining tests added: horizontal mosaic, the level-count ceiling, the grid-index coordinate fallback, and the grid hole. On the `PlotlyCalcData.x`/`y` widening — agreed it is loose, but a second calc interface would have every reader of the shared one deciding which it holds, and the `z` on it already carries the same duality with the same note. Leaving it.

---

PR #892, opened 2026-08-14
feat(anychart): read the chart types deferred out of the coverage PR

jooyoungseo:
# Pull Request

## Description

The AnyChart adapter's coverage PR deferred the chart types that needed more
than a series-type lookup. This reads two of them: **choropleth maps** and
**gantt schedules**. Both are detected, bound to the grammar's payload shapes,
and highlighted — a trace that sonifies but does not highlight is not done, so
the selector/stamper path lands in the same change as the detection.

`anychart.timeline()` is deliberately left unread; the reason is under Changes
Made.

## Related Issues

Part of #885

### Trace types added

**`CHOROPLETH`** — a `choropleth` series on `anychart.map()`.

- `'choropleth'` joins the `mapSeriesType` table, so the **series names itself**
  and no chart-level check gates it. `chart.getType()` answers `''` on a build
  without it and on a chart that has not been drawn; gating on it would drop
  working maps, and a test pins that a map reporting `''` is still read.
- Emits `ChoroplethPoint[]` as `{ x: region name, y: value }`. Rows with no
  value are dropped.
- A data row names its region by the id the geodata declared (`US.CA`), never
  by name, so the name is read from the bound feature: `point.getFeatureProp()`
  where the build offers it, else a geodata walk matched on `chart.geoIdField()`
  (defaulting to `'id'`), else the row's own `name`/`x`, else the bare id. The
  walk handles **both** shapes AnyChart accepts — GeoJSON `features[]` and
  TopoJSON `objects.<key>.geometries[]`.
- `lon`/`lat` are **omitted on purpose**: what AnyChart holds are the features'
  `middle-x`/`middle-y`, normalised coordinates in the chart's own projection,
  not degrees. Passing them through would tell a reader one region lies north
  of another when the map says nothing of the kind. `neighbors` is not
  derivable from AnyChart at all and stays absent. The map is read as a region
  list in declared order — the poorer reading the schema explicitly sanctions.
- **Highlighting:** a map is *located*, not counted. AnyChart paints every
  feature of the bound geodata (all fifty states for a table naming six) in the
  geodata's order, and the paths carry no id, so counting them off would put
  California's highlight on Alabama's shape. Each region is matched to the
  shape whose `getBBox()` equals the bounds `point.getFeatureBounds()` reports,
  within 1.5px. Zero matches, several matches, or a shape already claimed
  disables highlighting for the whole map with a warning — the word-cloud
  precedent. New stamper `data-maidr-anychart-region="<seriesIndex>-<regionIndex>"`.

**`GANTT`** — `anychart.ganttProject()` and `anychart.ganttResource()`.

- The type name is corroborated structurally before anything is read: a gantt
  has no series API and its `chart.data()` returns an `anychart.data.Tree`, so
  a chart naming itself a gantt with a flat data view behind it binds as
  nothing rather than as an empty schedule.
- Emits `GanttData` with `points` **nested one array per lane**, `lanes` naming
  every lane in the same order (so an undated task survives as an empty lane),
  `unit`, and `orientation: 'horz'` because a schedule runs left to right.
- The tree flattens depth-first — parents before children, the chart's own row
  order. A parent that states no dates takes `meta('autoStart')`/`meta('autoEnd')`;
  a resource row's `periods[]` becomes several intervals in one lane; a start
  with no end becomes the zero-length milestone AnyChart draws as a diamond.
- The ends are restated in whole days (hours below a two-day span) and the unit
  is named, plus an x-axis `format.function` that turns a position back into a
  date. This mirrors the shipped google-charts gantt; without it every task
  announces as a nine-digit millisecond figure. See Additional Notes — this
  departs from the spec's framing of `unit` as authored.
- **Highlighting:** a gantt is counted, but from more filled shapes than any
  other chart here (row stripes on both halves of the split widget, header,
  progress fills). The whole-SVG count is used when it equals the interval
  count; otherwise the search narrows to the single `ac_layer_*` holding
  exactly that many; when two layers tie — the stripes behind a
  one-bar-per-row project chart are exactly that — they are separated by the
  one property a schedule has and a backdrop does not, that its bars start and
  end in different places. Anything still ambiguous is left unstamped and
  warned about, never mislabelled. New stamper
  `data-maidr-anychart-task-bar="<laneIndex>-<intervalIndex>"`.

### Trace types not added

- **`anychart.timeline()`** — `readChartType` returns `'timeline'` for a third
  constructor unlike the two gantt ones: it has a series API, and its series
  are `range` (an interval) and `moment` (an instant). A timeline drawn mostly
  from moments has no intervals to announce, and emitting zero-length rows for
  them would describe work the chart never drew. It binds as nothing (silence),
  the acceptable failure the per-adapter mandate names. The two tree-backed
  constructors, `gantt-project` and `gantt-resource`, are both implemented.

### Collateral changes inside the adapter

- `AnyChartInstance.data()` now returns `AnyChartDataView | AnyChartTree`; the
  three existing readers go through a new `resolveChartDataView` that narrows
  on `'getIterator' in data`, so a gantt reaching a data-view reader is a no-op
  instead of an exception.
- `readRows` gained `'id'` in its field list — a choropleth row previously
  arrived carrying only `value`, with no region key at all.
- `buildLayer` gained a `chart` parameter (only the choropleth builder needs
  it, for the geodata).
- `attachAxes` now **merges** onto whatever a builder already put on an axis
  instead of replacing it, so the gantt's date formatter survives the label
  pass. No existing builder sets an axis, so behaviour is unchanged for them.
- `enableLineMarkersIfNeeded` exits early for a gantt (no series API to ask).
- `isMapChart` matches `getType() === 'map'` **exactly**, or the presence of a
  choropleth series — the tolerant substring match the pie/heatmap paths use
  would have caught `'heat-map'`.
- `src/adapters/anychart/index.ts` is untouched: neither item reads a
  declaration, so there is nothing new to re-export.

### Files

| File | |
|---|---|
| `src/adapters/anychart/converters.ts` | detection, payloads, stampers |
| `src/adapters/anychart/types.ts` | `AnyChartTree`/`AnyChartTreeItem`, optional map point API, `geoData()`/`geoIdField()` |
| `test/adapters/anychart/choropleth.test.ts` | 12 cases |
| `test/adapters/anychart/gantt.test.ts` | 15 cases |
| `examples/anychart/choropleth.html`, `examples/anychart/gantt.html` | |
| `docs/anychart.md` | supported-types table, per-type notes, stamper table, "cannot be found by counting" section |

## Checklist

`ManualTestingProcess.md` is unticked because the manual pass needs AnyChart's
map and gantt bundles from the CDN in a real browser, which was not available
here. Everything automatable was run and is green (below).

### Verification

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` were
all run on this branch and all pass: **3797 unit tests across 270 suites** plus
**73 ESM tests across 7 suites**, no failures. The 27 new cases pass. No e2e
spec was run or added — no AnyChart adapter has one, since they need the
AnyChart CDN in a browser.

### Foundation gaps — these matter to the sibling adapter PRs

`src/type/declaration.ts` and `src/adapters/shared/traceDeclaration.ts` were
**not modified**, per the shared-foundation rule. Two gaps observed there:

1. **`MaidrTraceDeclaration` (declaration.ts:133) has no `GANTT` variant**, so
   `TraceType.GANTT` cannot be declared in any adapter. Not blocking here — an
   AnyChart schedule is drawn by a dedicated constructor and is detectable —
   but a chart that draws a schedule out of ordinary series (a `range-bar`
   whose low/high are dates) has no way to say so, and the hole exists in every
   adapter. It would also be the only home for `GanttData.unit`, which the
   convention treats as authored.
2. **`ChoroplethDeclaration` (declaration.ts:585) offers `region`/`value`/
   `lon`/`lat` but no `neighbors`** — and `neighbors` is the one
   `ChoroplethPoint` field the grammar itself says cannot be recovered from a
   rendering (grammar.ts:765-772). AnyChart could not have used it (no
   per-point author slot), so nothing was blocked, but a declaration surface
   that omits the only underivable field leaves it underivable everywhere.

No gap in the parts that were used: `readDeclarationSlot`, `validateDeclaration`,
`resolveFieldRef` and `isFlagValue` were read and are sound. Neither trace type
here reads a declaration — both are `detectableWithoutOptIn` — and
`AnyChartBinderOptions` is untouched.

### Spec corrections

- **The region name is not reached "via `chart.geoData()`"** as
  `deferred-anychart.md` and `per-adapter.json` both state; that is only the
  fallback. AnyChart's map module exposes the feature's own properties on the
  point — `series.getPoint(i).getFeatureProp()` — which needs no id matching
  and survives a chart whose geodata was set by name rather than by object. The
  adapter prefers it. The geodata walk also has to handle **both** GeoJSON and
  TopoJSON; the spec mentioned only "the bound geodata feature's `properties`".
- **`per-adapter.json` says region rows "carry `id` and `value` through the
  existing `AnyChartIterator`".** True of the iterator, but not of `readRows`,
  whose field list is a fixed array that did not contain `id`. A choropleth row
  arrived carrying only `value`, with no region key. `'id'` had to be added.
- **`per-adapter.json` item 5 says `unit` is "the one authored value"** and
  that, absent, MAIDR should announce a length without naming a unit. That is
  wrong for AnyChart: a gantt's timeline is a `GanttDateTime` scale, so what
  the task tree holds is instants, and the unit is **read** rather than
  authored or guessed. Leaving it out would announce every task as a nine-digit
  millisecond figure — the failure the shipped google-charts gantt avoids at
  `converters.ts:2407-2439`. No `unit` field was added to
  `AnyChartBinderOptions`.
- **`per-adapter.json` item 6 asks for the gantt stamper to be its own commit.**
  It landed in the same commit as the choropleth work: both types edit
  interleaved regions of the single 5.5k-line `converters.ts`, and splitting
  them would have required `git checkout`/`git restore` on that file, which the
  working rules forbid. The branch carries two commits (the feature, then a
  comment correction), not one per trace type.
- Line references checked and **found accurate**: `mapSeriesType` at
  converters.ts:133, `readChartType` at :214 (spec said 208-220),
  `AnyChartBinderOptions` at types.ts:202-261, `GanttData` at grammar.ts:579-628,
  `ChoroplethPoint.lon`/`lat` at grammar.ts:754-762.

### Known limitation

The AnyChart library is not in `node_modules` and the adapter is entirely
duck-typed, so the map point API itself could not be verified against a real
build. `getFeatureProp()` and `getFeatureBounds()` are therefore declared
optional, called through `?.` inside `try`/`catch`, and a build lacking either
degrades to the id-as-name reading and to no highlighting respectively — never
to a wrong one. The gantt bar-layer search is the piece least verifiable
without the real library; every failure path is silence plus a warning naming
both counts, and `options.selectors` overrides it entirely.

---

### DIFF
--- a/docs/anychart.md
+++ b/docs/anychart.md
@@ -87,6 +87,8 @@ AnyChart must be loaded separately — the adapter does not bundle the AnyChart 
 | Radar | `anychart.radar()`, and a polar `line` / `marker` series | [Radar chart](examples.html) |
 | Polar Area | `anychart.polar()` with a `column` / `area` series | [Radar chart](examples.html) |
 | Mosaic | `anychart.mekko()`, `anychart.mosaic()`, `anychart.barmekko()` | [Marimekko chart](examples.html) |
+| Choropleth | a `choropleth` series on `anychart.map()` | [Choropleth map](examples.html) |
+| Gantt | `anychart.ganttProject()`, `anychart.ganttResource()` | [Gantt chart](examples.html) |
 
 `step-area` is the one series that still loses its fill: MAIDR has no stepped area trace, so it keeps its staircase and maps to a step trace. A console warning is emitted when that downgrade occurs.
 
@@ -118,6 +120,16 @@ AnyChart must be loaded separately — the adapter does not bundle the AnyChart 
 - **Diverging bars** — a tornado chart, or a population pyramid — are **opt-in** via `bindAnyChart(chart, { diverging: true })`. AnyChart has no diverging chart type: the idiom is a stacked `anychart.bar()` whose two series straddle zero, and nothing distinguishes that from a stacked bar chart that happens to contain negative values. Guessing would not merely rename an ordinary chart — a diverging trace replaces the sign in every announcement with the name of a side, which is the one clue a reader would have that the reading was wrong. When declared, every bar series is merged into **one** layer, because the balance MAIDR announces is a difference read down a column of one grid; the values are emitted signed, exactly as the chart draws them, and each side is named by its series. A chart with fewer than two bar series keeps its ordinary bar layers and says why.
 
 - **Tag clouds** come from `anychart-tag-cloud.min.js` and report `getType()` as `'tag-cloud'`. Axis labels fall back to `Term` and `Weight`. Highlighting pairs each term with its own `<text>` element by matching the rendered text rather than by counting DOM order: a cloud writes its words in packing order, which has no relation to the order they were declared in, so counting them off would announce one term while highlighting another. A term that does not match exactly one rendered word disables highlighting for the whole chart rather than placing a guess.
+
+- **Choropleth** maps come from `anychart-map.min.js` plus a geodata file, and the `choropleth` series is what carries the data — `marker`, `bubble` and `connector` draw *over* a map rather than colouring it. The **series' own type is the whole detection**: `chart.getType()` is deliberately not required, because it answers `''` on a build without it and on a chart that has not been drawn, and gating on it would drop working maps. A row names a region by the id the geodata declared (`US.NV`) and never by name, so the name is read from the bound geo feature — `point.getFeatureProp()` where the build offers it, and the geodata's own `properties` matched by id otherwise. A region nothing names keeps its id, which is a poor name but a true one. Axis labels fall back to `Region` and `Value`.
+
+  `lon` and `lat` are **deliberately omitted**. MAIDR's centroids are degrees east and north; what AnyChart has are its features' `middle-x` / `middle-y`, normalised coordinates in the map's own projection, which cannot be turned into degrees without inverting a projection the chart does not name. Passing them through would tell a reader that one region lies north of another when the map says nothing of the kind, so the map is read as a region list in declared order — the poorer reading the schema explicitly sanctions. `neighbors` is not derivable from AnyChart either and stays absent.
+
+- **Gantt** charts come from `anychart-gantt.min.js`. `anychart.ganttProject()` and `anychart.ganttResource()` report `'gantt-project'` and `'gantt-resource'`, and the type name is corroborated structurally before anything is read: a gantt has no series API at all and its `chart.data()` hands back an `anychart.data.Tree` of tasks rather than a data view, so a chart naming itself a gantt with no tree behind it is bound as nothing rather than as an empty schedule. The tree is flattened depth-first — parents before their children, the order the chart stacks its rows in — into one lane per row. A parent task states no dates of its own, so the pair AnyChart derived for it (`autoStart` / `autoEnd`) is read instead; a task with a start and no end is a milestone and is emitted as the zero-length interval it is; a resource chart's `periods` array becomes several intervals in one lane. A task with no dates at all becomes an **empty lane** that still carries its name, which is what MAIDR's nested gantt shape exists to express. Axis labels fall back to `Date` and `Task` (`Resource` on a resource chart).
+
+  The ends are restated in whole days — or hours, on a schedule spanning less than two days — and the unit is named alongside them, because a gantt is read for how long its intervals run and epoch milliseconds announce as an unreadable nine-digit figure. This is read rather than guessed: a gantt's timeline is a date-time scale, so what the tree holds is instants. The x axis carries a formatter that turns a position back into a date, so each end is still announced as one.
+
+  `anychart.timeline()` is **not** read. It is a third constructor with a series API of its own whose `moment` series are instants rather than intervals, and announcing one as a schedule would describe work the chart never drew; it binds as nothing instead.
 
 ## Code Examples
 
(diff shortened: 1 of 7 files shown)


jooyoungseo:
Addressed the review. All 14 checks are green at `bf16a42`.

**Observation 1 — fixed** (`bf16a42`). The reviewer is right: the trailing clause of the early-return comment in `enableLineMarkersIfNeeded` read as though a map were handled by that return. It is not in the list, it does expose the series API, and it falls through to the loop below, which no-ops only because `'choropleth'` is absent from `LINE_LIKE_SERIES_TYPES`. The comment now says that explicitly. Comment only; no behaviour change.

**Observation 2 — declined, deliberately consistent.** A weeks/months tier would make AnyChart the only adapter in the tree that announces one. `google-charts` `ganttScale` ships the identical two-tier split at the same `GANTT_HOURLY_MAX_SPAN` threshold (`converters.ts:2407-2439`), and `plotly`'s `GANTT_UNITS` also caps at `days` as its coarsest unit. Diverging here would mean the same schedule reads differently depending on which library drew it. If a coarser tier is wanted it should land across all three at once, which is its own PR rather than a one-adapter change.

**Observation 3 — declined, this is the intended degradation.** Worth recording that it is not merely coarse but deliberately atomic: `stampChoroplethAttributes` accumulates into `matched[]` and only writes attributes after every region has resolved to exactly one unclaimed shape, returning early on any ambiguity. So a partial match stamps nothing rather than stamping some regions correctly and leaving others silently unmarked at positions a reader would still navigate to. Per-region degradation would produce exactly that half-highlighted state, which is the failure mode worth avoiding. Revisiting it once a real AnyChart build is available to test against is reasonable, but it would trade the current guarantee for a weaker one.

**Foundation gaps** (missing `GANTT` variant in `MaidrTraceDeclaration`, missing `neighbors` on `ChoroplethDeclaration`) are left as flagged in the PR body. `src/type/declaration.ts` and `src/adapters/shared/traceDeclaration.ts` are the shared base for eight parallel adapter PRs, so changing either from inside one of them would destabilise the other seven; those belong in a dedicated PR.

One caveat on the review record: the second `claude-review` run posted only an unchecked progress checklist saying a deep-dive was starting, then the job exited without publishing findings. Those items were not re-reviewed, and that stub should not be read as a second approval.

---

PR #893, opened 2026-08-14
feat(highcharts): read the chart types deferred out of the coverage PR

jooyoungseo:
## Description

The Highcharts adapter now reads the three trace types deferred out of the coverage PR: **choropleth**, **survival** and **forest**. Choropleth is detected outright from the Highmaps `map` series type; survival and forest are drawn with a series type they share with something plainer, so both are *declared* through the shared `MaidrTraceDeclaration` foundation (`src/type/declaration.ts`, `src/adapters/shared/traceDeclaration.ts`) merged in #886, read out of a `custom.maidr` block on the series the author already wrote.

Every one of the three wires its highlight path in the same change as its detection — a layer that sonifies but does not highlight is not done.

Nothing outside the adapter was touched: `src/model/`, `src/service/`, `src/state/`, `src/ui/`, `src/command/`, `src/type/` and `src/adapters/shared/` are all unmodified.

## Related Issues

Part of #885

### Trace types added

**Choropleth** (`src/adapters/highcharts/adapter.ts`, `selectors.ts`) — detected with no opt-in from `resolveSeriesType(series, chart) === 'map'`; a new `case 'map'` in the `convertSeries` dispatch replaces the previous `Unsupported series type: "map"; skipping.` warning. Payload is `ChoroplethPoint[]` = `{ x: point.name ?? category ?? x, y: point.value ?? y }`. Centroids are read from the map feature's `properties['hc-middle-lon'/'hc-middle-lat']`, else from `chart.mapView.projectedUnitsToLonLat` applied to the point's `bounds` midpoint; both readings are degree-range validated (|lon| ≤ 180, |lat| ≤ 90) and omitted rather than converted when they fail. `neighbors` is not emitted — Highcharts knows which shapes it drew, not which share a border. Null-valued regions are left out of the payload. **Highlighting:** `stampPointIndices` writes `data-maidr-region-index` onto each *announced* region's graphic and the new `choroplethSelectors` addresses that stamp, so a map with a gap in it still highlights the right country. Axes are named `Region`/`Value` (a map series is bound to no titled axis). A `custom.maidr` block of type `choropleth` renames `region`/`value`/`lon`/`lat`.

**Survival** — declared with `custom.maidr = { type: 'survival', … }` on a `line`/`spline` series; produces `SurvivalPoint[][]` in place of the STEP layer the same series makes today. Companions are absorbed through Highcharts' native `linkedTo`/`linkedParent`: a linked `scatter` is the censoring ticks, a linked `arearange` is the confidence band; both are matched by x and suppressed from becoming layers of their own. `censoredSeries`/`bandSeries` accept series `id`s for charts that do not link. `censored` is read off the curve's own rows first (`resolveFieldRef` + `isFlagValue`), the tick series second. `merge` (default `true`) folds following undeclared line series in as further arms. `stepDirection` comes from `series.options.step` or the declaration. **Highlighting:** `lineSelectors`, one `path.highcharts-graph` per arm.

**Forest** — declared with `custom.maidr = { type: 'forest', … }` on the estimate series (or on the `errorbar` itself). The interval comes from the `errorbar` paired by `linkedTo`, or from `intervalSeries`; when the declaration sits on the whip, estimates come from `linkedParent` and that parent is suppressed via the existing `seriesReadAsErrorBars` rule. `weight` comes from `point.options` only — `marker.radius` is never inverted, and a weight outside 0..1 is treated as a percentage and dropped with one warning per series. `pooledIndex` counts the declaring series' rows as authored, dropped rows included; a `pooledSeries` companion is absorbed, appended last and flagged pooled. `nullValue` comes from the declaration only → `forestOptions`; the `xAxis.plotLines` fallback stays confined to `significanceThresholds`. `error` offsets are normalised to absolute bounds, losing to `yMin`/`yMax`. `orientation` from the declaration or `chart.inverted`. **Highlighting:** the whip group (`errorBarSelector`) when an interval series exists, else the estimate marks, plus the pooled companion's marks appended in row order.

### Not added

No trace type from the deferred list was skipped. One piece of *test* coverage was not added: **no e2e spec**. `.claude/rules/testing.md` asks for one per new trace type, but `e2e_tests/` was outside the paths this change was allowed to touch. That is an outstanding gap for whoever integrates the adapter branches.

### Supporting changes

- `src/adapters/highcharts/types.ts` — `custom?: { maidr?: MaidrTraceDeclaration }` on `HighchartsSeries['options']`; `properties` and `bounds` on `HighchartsPoint`; `mapView.projectedUnitsToLonLat` and the `MapLonLat` return type on `HighchartsChart`. Nothing added to `HighchartsAdapterOptions`.
- `src/adapters/highcharts/adapter.ts` — shared declaration plumbing: `declarationOf` (via `readDeclarationSlot`, memoised per series object so one binding warns once), `seriesRef`, `readSeriesField` (wraps `resolveFieldRef` + `warnUnresolvedRef` per series), `companionSeries` (linkedTo-inferred or id-named, marking the companion consumed), and a `convertDeclaredSeries` pre-pass that runs before the error-bar absorption and the type buckets so a declaration beats every heuristic.
- `src/adapters/highcharts/index.ts` — re-exports `MaidrTraceDeclaration`, `ChoroplethDeclaration`, `ForestDeclaration`, `SurvivalDeclaration`, `FieldRef`, `SeriesRef`.
- Tests: `test/adapters/highcharts/choropleth.test.ts` (8), `survival.test.ts` (8), `forest.test.ts` (9); `helpers.ts` grew `inverted` and `mapView` on `fakeChart`.
- Examples: `examples/highcharts-choropleth.html` (Highmaps US states topology), `highcharts-survival.html` (two arms + linked `arearange` + linked `scatter`), `highcharts-forest.html` (inverted scatter + linked `errorbar`, weights drawn as marker area).
- Docs: `docs/highcharts.md` gained three table rows, three notes, a `## Declaring What a Series Means` section, and a precedence note on `options.significancePlot`.

Not applicable — no UI surface changed.

## Checklist

The manual-testing box is left unticked deliberately: the automated gate is green (`lint:fix`, `type-check`, `build`, and `npm test` at 271 + 7 suites / 3795 + 73 tests, including 26 suites / 168 tests under `test/adapters/highcharts`), but `ManualTestingProcess.md` was not walked in a real browser, and the choropleth centroid path in particular has not been exercised against a real Highmaps page (see Additional Notes).

### Foundation gaps (relevant to the sibling adapter PRs — reported, not worked around in shared code)

1. **`FIELD_REF_FALLBACKS.value` is shared across types and includes `x`, `t`, `position`.** On a Highcharts map row the region name is as likely to sit in `x` as the value is, so walking the default chain for a choropleth's `value` can resolve the name as the number. `ChoroplethDeclaration.value` documents the hazard but the shared table cannot avoid it. Sidestepped here by never walking the default chain for a choropleth's `region`/`value` — those come off the point Highcharts already resolved (`point.name`, `point.value`), and `resolveFieldRef` is used only when the author declared a ref. A per-canonical-name opt-out, or a note in the shared module that map-shaped rows should always name the field, would fix it properly.
2. **`ErrorBarDeclaration.error` / `ForestDeclaration.error` has no canonical grammar field and no `FIELD_REF_FALLBACKS` entry**, so `resolveFieldRef(row, undefined, 'error')` silently reads a bare `error` column with no chain and with that behaviour undocumented. This adapter only reads it when the author declared it, which is what the field's doc implies but does not state. Not blocking.
3. **No shared way to say "valid declaration, this adapter has no reading for it yet".** `validateDeclaration` checks that a declared type is one `MaidrTraceDeclaration` covers, but the "I can't read this one" case is a Highcharts-local warning here. If all eight adapters do the same, the sentence will be spelled eight ways — exactly what the shared module exists to prevent. A `warnUnreadType(context, type)` export would fix it once a second caller exists.

### Spec corrections

- **The `hc-middle-lon` / `hc-middle-lat` claim is unverified.** `deferred-highcharts.md` states Highcharts mapdata features expose those properties; that could not be confirmed here (`highcharts` is not in the shared install, and the map collection was not fetched). What is confirmable from the map collection is `hc-middle-x`/`hc-middle-y`, which are 0..1 fractions of the bounding box and are **not** degrees. The read is implemented as specified but every reading passes a degree-range check, so on a build where the properties are absent it resolves to nothing, and on a build where they hold something else it is dropped rather than announced. **Re-verify against a real Highmaps page before trusting the docs.**
- **`MapView#projectedUnitsToLonLat` was used instead of the spec's `chart.mapView.projection.inverse([...])`.** It is the documented public method for the same conversion and, unlike `projection.inverse` (which returns its input unchanged when no projection definition is in play), it answers with nothing on a map with no geographic projection — precisely the guard needed to stop a pre-projected map's own units being announced as degrees. Both an `[lon, lat]` array and a `{ lon, lat }` object are accepted as the return value, since the installed shape could not be verified.
- **Line numbers in both spec files have drifted since the coverage PR merged.** `HighchartsSeries['options']` is at `types.ts:178-204`, not `:180-203`. The `map` fallthrough warning is around `adapter.ts:1105`, not `:1094-1096` — the behaviour claim was correct. `HighchartsPoint.options?: Record<string, unknown>` at `:263` was correct.
- **`step` was not made a precondition for SURVIVAL.** The per-adapter mandate describes it as "a `line` series with `step` set … plus `custom.maidr`". The times and probabilities announced are identical whether the curve was drawn stepped or interpolated, and refusing the reading would cost the median survival and the censoring rotor over a cosmetic difference; `stepDirection` then comes from the declaration. A non-`line` series still falls through with a wrong-construct warning.
- **`convention.md`'s draft of `SurvivalDeclaration` has no `merge` field**, but the merged `src/type/declaration.ts` and `DECLARATION_KEYS` in `traceDeclaration.ts` both ship one (default `true`, absorbing following siblings as further arms). The merged foundation was followed as authoritative; anyone reading `convention.md` alone will not know `merge` exists.

### Merge ordering (matters when the adapter branches are integrated)

`buildSubplot` now runs `convertDeclaredSeries(seriesToConvert, …)` **first** and hands the survivors to `seriesReadAsErrorBars`. That order is load-bearing: a forest plot's estimate series is exactly the parent the error-bar pass would otherwise absorb, so running the declaration pass second would make the declaring series vanish before it was read. Declared layers are pushed at the head of the subplot's layer list.

### Degradation choices

All in the direction the convention asks for: a centroid that could not be a coordinate is omitted rather than converted (`ChoroplethTrace` then keeps declared order in one band, the sanctioned poorer reading); `nullValue` is never inferred from `xAxis.plotLines`; a forest weight outside 0..1 is dropped with a warning rather than rescaled; a companion ref naming a series that does not exist emits the parent without that half; a declaration on a series that cannot back it warns naming both and falls through to the undeclared reading.

### One thing a real browser should confirm before release

The choropleth centroid path. Both of its sources are Highcharts internals that could not be exercised here (no `highcharts` install, jsdom only), so the unit tests pin the *policy* — degree-range validation, both return shapes, omission on failure — rather than the real API. The rest of the choropleth (detection, payload, stamping, selectors) is exercised end to end.

---


jooyoungseo:
Both blocking findings are fixed and pushed; the two smaller notes are deliberately left alone, and one of them rests on a premise that turns out not to hold.

**Second linked companion is not dropped.** `companionSeries` adds only the series it matched to `consumed`, so a second linked `scatter` is never removed from the panel — it falls through to the ordinary buckets and is announced as a layer of its own. Checked against the real converter: a survival curve with two linked `scatter` series emits `[survival "Treated", point "Ticks B"]`. The first tick series becomes the censoring marks, the second stays a scatter the reader can still reach. That is the right degradation for an ambiguous chart: nothing goes unannounced, and the adapter does not have to guess which of two tick styles the author meant as censoring. A warning here would need a policy for picking a winner, which is exactly the guess worth avoiding.

**`forestRows`' warning is per series by design.** A pooled companion is a separate series the author wrote separately and can misreport weight independently; two warnings naming two series is the message that locates both. Collapsing them to one per layer would name only the first.

Neither is worth a change in this PR. The declaration-cache fix carries a regression test (`survival.test.ts`, "re-reads a declaration the author corrected on the series in place") that fails without the reset — I verified it fails on the pre-fix adapter.

Still outstanding from the PR body, unchanged: the choropleth centroid path (`hc-middle-lon`/`hc-middle-lat`, `mapView.projectedUnitsToLonLat`) has never run against a real Highmaps build, and there is no e2e spec for the three new trace types because `e2e_tests/` was outside this change's allowed paths.

---

PR #895, opened 2026-08-14
feat(amcharts): read the chart types deferred out of the coverage PR

jooyoungseo:
## Description

amCharts leaves no signature for a handful of readings: a survival curve and a
step chart are the same `StepLineSeries`, a volcano and a scatter are the same
strokeless `LineSeries` with bullets, and an error bar's interval lives in
columns the chart was never bound to. This PR teaches the amCharts adapter to
read those meanings from the opt-in `maidr` declaration block on a series'
`userData`, using the shared foundation in `src/type/declaration.ts` and
`src/adapters/shared/traceDeclaration.ts` (both untouched here).

The declaration is resolved once per chart as a three-pass plan — read the
slots, absorb companion series, merge sibling arms — and the same plan backs
both entry points, so the layer a reader hears and the mark the overlay
outlines are always the same series. Detection and the selector/highlight path
land together: every type added here registers a navmap resolver, except where
the highlight contract itself cannot address the mark (see Additional Notes).

Diff: 12 files, +2558 / −5. New: `src/adapters/amcharts/declaration.ts`,
`test/adapters/amcharts/declaration.test.ts`, `examples/amcharts-declared.html`.
Changed: the adapter's `adapter.ts`, `binder.tsx`, `extractor.ts`, `navmap.ts`,
`types.ts`, `index.ts`, plus `docs/amcharts.md` and two test helpers. Nothing
outside `src/adapters/amcharts/`, `test/adapters/amcharts/`, `examples/` and
`docs/`.

## Related Issues

Part of #885

### Trace types added

- **SURVIVAL** — `{ type: 'survival' }` on a `StepLineSeries` (or any XY series
  carrying `valueY`). Reads `censored`, `yMin` and `yMax` off
  `dataItem.dataContext` through the shared fallback chains and `isFlagValue`,
  passes `stepDirection` through to the layer, absorbs a `censoredSeries`
  (ticks joined by x) and a `bandSeries` (`openValueY`→`yMin`,
  `valueY`→`yMax`, joined by x), and folds following `StepLineSeries` siblings
  in as further arms (merge defaults true). Payload `SurvivalPoint[][]`, one row
  per arm, arm name on `z`. Highlight: navmap resolver, rows = arms,
  cols = samples, kind `point`.
- **ERROR_BAR** — `{ type: 'error_bar' }` on the estimate series. Bounds resolve
  *per bound*, in the order row fields → an `error` offset (a number or
  `[lower, upper]`, read as positive magnitudes, negatives dropped, float noise
  stripped) → an absorbed `intervalSeries` companion. The companion is joined
  **by position, not by index**, and is suppressed from becoming a layer of its
  own. Orientation comes from the declaration, else from `categoryYField`.
  Payload `ErrorBarPoint[]`. Highlight: col addresses the sample, every braille
  section resolving to the same mark (the dumbbell's call), kind `column` or
  `point` depending on the estimate series' class.
- **FOREST** — `{ type: 'forest' }`: everything ERROR_BAR takes, plus `weight`
  (a resolved value above 1 or below 0 is left out rather than rescaled),
  `pooled` (strict flag reading), `pooledIndex` (counted over the declaring
  series' data items as authored, so a dropped row still counts), a
  `pooledSeries` companion (rows appended after the studies, marked pooled), and
  `nullValue` → `forestOptions`, with no default. Payload `ForestPoint[]`.
- **VOLCANO** — `{ type: 'volcano' }` on the scatter recipe (a strokeless
  `LineSeries` with bullets on two value axes), which is announced as a LINE
  layer without it. Points read straight off `valueX`/`valueY`; `label` and
  `group` off `dataContext`. `significance`, `significanceDirection` and
  `effect` become `thresholdOptions` only when declared — no block at all when
  none is. Merge defaults false.
- **MANHATTAN** — the same reading with merge defaulting true, so one series per
  chromosome becomes one navigable cloud. Siblings are absorbed only when they
  carry no declaration of their own and their *undeclared* heuristic kind
  matches the declaring series'.
- **SCATTER** (`type: 'point'`) — the plain declarable case that falls out of the
  cloud path. `label` is accepted and not read, because `ScatterPoint` has
  nowhere to put it.

### Trace types not added

- **ALLUVIAL** — skipped deliberately, on two findings the spec did not
  anticipate. (1) The per-adapter instruction ("reuse the existing extractor and
  change only the emitted type") rests on a SANKEY extractor that does not
  exist: `grep -rn 'Sankey|FlowPoint' src/adapters/amcharts/` returns nothing,
  and an `am5flow.Sankey` is a bare series in the root container that discovery
  does not find today. The item is not opt-in plumbing; it is flow discovery,
  extraction and highlight from zero. (2) The highlight cannot be wired
  truthfully without duplicating model code: amCharts renders to canvas, so
  highlighting goes through the overlay, which is handed `{layerId, row, col}`
  from `state.braille`; `FlowTrace` addresses nodes by a longest-path stage
  layering (`src/model/flow.ts` `assignStages`/`buildNodes`, with a cyclic-graph
  fallback) and its braille getter additionally transposes the pair
  (flow.ts:449-465). Turning a position back into an am5 node would mean
  re-implementing that inside the adapter, where it would drift silently and put
  the highlight box on the wrong node. Compounding both: amCharts is commercial
  and is not installed here, so `am5flow`'s data-item surface could not be
  verified against anything.

### Supporting changes

- `src/adapters/amcharts/declaration.ts` (new) — validation through
  `readDeclarationSlot`, the chart-level three-pass plan, warn-once
  bookkeeping, and the payload extractors.
- `adapter.ts` (`buildChartLayers`) and `binder.tsx` (`groupSeries`) — the
  declared short-circuit lives at these two call sites rather than at the top of
  `classifySeriesKind`, so `classifySeriesKind` stays a pure statement about the
  drawing and the sibling-merge pass can still ask it what a candidate sibling
  actually draws.
- `navmap.ts` — resolvers for the declared layers.
- `extractor.ts` — exports `isColumnSeries`, `readXValue`, `toNumber` and
  `toStringOrNumber` for the declared reader; no behaviour change.
- `types.ts` — one optional `dataContext?: unknown` on `AmDataItem`; without it
  the author's own row is unreachable in TypeScript.
- `index.ts` — re-exports the declaration types so a TypeScript author gets the
  field names checked.
- `docs/amcharts.md` — a new "Declaring What a Chart Means" section (the slot,
  the two naming laws, a per-type field table, companion series, merging, and an
  explicit statement of the cloud highlight limitation), seven new rows in the
  supported-types table, and a corrected note on scatter vs dot plot.
- `examples/amcharts-declared.html` (new) — five runnable charts.
- `test/adapters/amcharts/declaration.test.ts` (new) — 33 cases. Each declared
  reading is asserted against the *undeclared* reading of the same series (step
  vs survival, line vs volcano), so a test cannot pass by the heuristic
  happening to agree.

N/A — no visual change to existing charts. `examples/amcharts-declared.html`
exercises the new readings.

## Checklist

The manual-testing box is left unticked: amCharts is a commercial library and is
not installed in this environment, so the charts could not be driven in a
browser. Automated verification actually run, all green:
`npm run lint:fix` (clean, no files changed), `npm run type-check` (clean),
`npm run build` (exit 0), `npm test` (exit 0 — unit: 3803 passed / 269 suites;
esm: 73 passed / 7 suites). `npm run e2e` was **not** run (no browser available)
and **no e2e spec was added**; `.claude/rules/testing.md` asks for one per new
trace type, so that is a known gap in this change.

## Additional Notes

These matter to the sibling adapter PRs under #885.

### Foundation gaps (nothing in the shared foundation was changed)

1. **`validateDeclaration` warns on every call and holds no state.** An adapter
   that resolves declarations at two entry points per binding — the JSON path
   and the highlight path both re-walk the chart — then says everything twice,
   breaking the spec's own "each distinct warning at most once per chart per
   binding" clause. Solved locally with a `WeakMap` keyed on the author's
   `userData` object, which also lets a *corrected* block be re-reported since
   rewriting it produces a new object. Every adapter with two entry points will
   need the same seven lines; a memo inside the shared module, or a `once` flag
   on `DeclarationContext`, would belong there. The same applies to
   `warnUnresolvedRef`, documented as "call this once per series per field" but
   leaving the caller to arrange it.
2. **The canvas-overlay highlight contract cannot address a scatter-family
   point.** `Controller.registerNavigateCallback` (controller.ts:572-583) only
   fires when `!state.braille.empty`, and `ScatterTrace.braille`
   (scatter.ts:452-490) returns the empty state outside grid mode — so a
   volcano, Manhattan or scatter layer emits no navigation event at all in
   ordinary navigation, and in grid mode the row/col are *binned grid cells*,
   not point indices. Every SVG adapter is unaffected (it highlights through
   `selectors`); every canvas adapter — amCharts, Chart.js — is affected. The
   resolver is deliberately left unregistered here, so the overlay clears rather
   than outlining whichever point happens to sit at that ordinal, and this is
   documented in `docs/amcharts.md`. Not fixable inside an adapter: it needs
   either a point-level position in `TraceState` for scatter traces, or a
   highlight channel that is not the braille position.
3. Two smaller notes, neither blocking: `FIELD_REF_FALLBACKS` has no entry for
   `error`, so an offset column must be spelled exactly `error` or named
   explicitly (possibly intentional — `yerr`/`sd`/`sem` are ambiguous); and the
   union in `MaidrTraceDeclaration` has no way for an adapter to say "I read
   this subset", so each adapter hand-writes its own narrowing predicate
   (`isDeclarable` here) and its own "this type names no construct I draw"
   warning.

### Spec corrections

- `deferred-amcharts.md` ALLUVIAL: "The work is almost entirely the opt-in
  plumbing on top of the SANKEY extractor." There is no SANKEY extractor in
  `src/adapters/amcharts/` — no flow reading, no `am5flow` discovery, nothing.
  `per-adapter.json` repeats this. See the skip entry above.
- `per-adapter.json` item 2: "add a short-circuit at the top of
  `classifySeriesKind` … This one hook unblocks all six items." The hook alone
  is neither sufficient nor safe. Companion absorption and sibling merge are
  chart-level facts a per-series classifier cannot express, and — decisively —
  the merge rule needs the *undeclared* kind of the declaring series to decide
  what "the same drawn kind" means for its siblings, which a short-circuit at
  the top of `classifySeriesKind` destroys. The line number is also stale:
  `classifySeriesKind` is at extractor.ts:1146 on current main, not 1135.
- `per-adapter.json` item 1: "**No type changes.**" Almost right —
  `AmChartsBinderOptions` and `AmEntity` needed nothing, as claimed. But the
  normative row source for amCharts is `dataItem.dataContext`, and `AmDataItem`
  declared only `get`/`uid`/`bullets`, so one optional field had to be added to
  the adapter's own `types.ts`.
- `per-adapter.json` item 4: "the `StepLineSeries` the adapter currently emits as
  STEP (`adapter.ts:309-311`)". There is no dedicated step branch there: step
  shares one `case 'line': case 'step': case 'area': case 'radar': case 'polar':`
  arm. The behaviour described is right; the location is not.
- Verified and correct, for the record: `AmEntity.get` reads `userData` with no
  change; `AmChartsBinderOptions` needed nothing; a scatter-configured
  `LineSeries` does fall into `LINE_CLASSES` and become a LINE layer, so there
  was no pre-existing scatter path; `ErrorBarPoint`'s bounds are absolute
  (grammar.ts:380-389); MANHATTAN and VOLCANO do share `VolcanoTrace`;
  ALLUVIAL/CHORD/SANKEY do share `FlowTrace`.

### Design decisions worth a reviewer's eye

- **Bound precedence for error bar / forest**: row fields → `error` offset →
  companion series, resolved per bound, so a one-sided interval keeps its half
  and a companion fills only what the row left unsaid. The declaration's own doc
  fixes `error` as losing to `yMin`/`yMax`; it says nothing about where a
  companion sits, and it is placed last here on the grounds that a column the
  author named is the more specific statement. The opposite order is defensible.
- **Companion joins are by position, never by index**, and a companion carrying
  fewer rows than the series it decorates still lines up. There is a test for
  this specifically, because index-joining is the failure that looks correct on
  the happy path.
- **A companion ref naming a series that declares its own layer is refused with a
  warning** rather than absorbed. The spec only covers refs naming nothing;
  absorbing a declared series would silently delete a layer, so the loud smaller
  reading was chosen.
- **The pooled summary's owner is tracked** (`DeclaredSamples.owners`) so the
  highlight follows a row appended from `pooledSeries` onto the companion series
  that drew it.
- One fragile invariant to know about: `planDeclarations` gates on
  `backsDeclaration`, and `buildDeclaredLayer`'s `null` return is only reachable
  if that gate and the extractors ever disagree about what counts as a readable
  mark. If they did, the navmap's per-type queue would slip by one for a chart
  carrying two declared layers of the same type. The two are deliberately written
  from the same predicates, and `backsDeclaration` shares `declaredHorizontal`
  with the extractors for exactly this reason, but it is a coupling a future edit
  could break quietly.

---

### DIFF
--- a/docs/amcharts.md
+++ b/docs/amcharts.md
@@ -202,7 +202,9 @@ var estimate = chart.series.push(am5xy.LineSeries.new(root, {
 
 An absorbed companion is merged into the parent layer **by position**, not by index — a companion carrying fewer rows than the series it decorates still lines up — and it does not become a layer of its own. `openValueY` maps to `yMin` and `valueY` to `yMax` (`openValueX`/`valueX` on a horizontal chart); the bounds are **absolute positions** on the value axis, never offsets. An interval your data holds as an offset instead is declared with `error`, which takes a number for a symmetric interval or a `[lower, upper]` pair for an asymmetric one, both as positive magnitudes.
 
-The four roles are `intervalSeries` (error bar, forest), `censoredSeries` and `bandSeries` (survival), and `pooledSeries` (forest). A role naming no series is reported and the layer is emitted without that half.
+The four roles are `intervalSeries` (error bar, forest), `censoredSeries` and `bandSeries` (survival), and `pooledSeries` (forest). A role naming no series is reported and the layer is emitted without that half — as is a role naming a series that declares a layer of its own, or one another declaration has already absorbed, since reading a series into two layers would announce the same rows twice over.
+
+On a forest plot declaring both `intervalSeries` and `pooledSeries`, the interval companion covers the summary too: the join is by position, so a chart drawing every interval in one column series, the summary's included, has already said where the summary's interval is. The pooled series' own row fields and `error` offset still outrank it.
 
 ### Merging siblings
 
--- a/docs/amcharts.md
+++ b/docs/amcharts.md
@@ -124,6 +124,12 @@ amCharts 5 renders to an HTML5 `<canvas>`, so there are no per-element SVG nodes
 | Lollipop | `ColumnSeries` | category axis + hairline `columns.template` width + bullets |
 | Word Cloud | `am5wc.WordCloud` | series class (requires `wc.js`) |
 | Bump (rank over time) | `LineSeries` | value axis renderer `inversed: true` **and** values that are a ranking; or the `bump` option |
+| Survival (Kaplan-Meier) | `StepLineSeries` | **declared** — `userData: { maidr: { type: "survival" } }` |
+| Error bar | any XY series, with a floating column behind it | **declared** — `{ type: "error_bar" }` |
+| Forest (meta-analysis) | horizontal `openValueXField` columns plus estimate marks | **declared** — `{ type: "forest" }` |
+| Volcano | hidden-stroke `LineSeries` with bullets, two value axes | **declared** — `{ type: "volcano" }` |
+| Manhattan | the same, one series per chromosome | **declared** — `{ type: "manhattan" }` |
+| Scatter | the same | **declared** — `{ type: "point" }` |
 
 A `StepLineSeries` is piecewise constant — the value is held and then jumps — so it maps to MAIDR's step trace rather than to a line, and is announced and navigated as a step plot. amCharts positions the staircase from the axis cell rather than reporting a step convention, so the adapter emits no `stepDirection` and MAIDR's description does not name one.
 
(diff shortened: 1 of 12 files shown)


jooyoungseo:
`4694ee0` settles the three actionable review findings; CI is green on it (all 14 checks, including E2E).

- **Finding 1 — companion claimed twice.** `absorb()` now refuses a companion another declaration already took, with the same warning treatment a ref naming a declared layer gets. Resolution order is deterministic (`plan.declared` is a `Map` in chart-series order), so the first declaration in series order keeps the companion and every later claim is reported. Test: two `error_bar` layers colliding on one `intervalSeries` id.
- **Finding 2 — forest declaring both `intervalSeries` and `pooledSeries`.** Kept the interval companion in scope for the pooled summary, and pinned it with two tests plus a paragraph in `docs/amcharts.md`. Scoping it out was the other option and it is the worse one: a forest plot commonly draws every interval — the summary's included — in one column series, and the join is by position, so dropping the companion for the pooled row would silently lose a bound the author did draw. The summary's own row fields and `error` offset still outrank it, exactly as for a study.
- **Finding 3 — exhaustive switch.** `buildDeclaredLayer` names `case TraceType.SCATTER` instead of defaulting to it, so a seventh member of `AmDeclaration` fails to compile there.

**Finding 4 — no e2e spec — I am leaving open rather than fixing here.** No amCharts example is exercised by any e2e spec today (`e2e_tests/` mentions amCharts nowhere), and every amCharts example loads the library from `cdn.amcharts.com` because it is commercial and not vendored — so the first amCharts spec would make the suite depend on a third-party CDN at test time. The six trace types themselves are not new to MAIDR and already have e2e coverage against their own examples (`survival.spec.ts`, `volcano.spec.ts`, `forest.spec.ts`, `errorbar.spec.ts`); what is untested end-to-end is the amCharts reading path, and covering that properly means deciding how amCharts gets into the e2e environment at all. That is its own change, not a rider on this one.

The foundation gaps listed under Additional Notes in the description (warn-once state in `validateDeclaration`, and the canvas-overlay highlight contract for scatter-family points) are also untouched here by design — `src/type/declaration.ts` and `src/adapters/shared/traceDeclaration.ts` are shared with seven sibling PRs, so a change there belongs in its own PR.

---

PR #898, opened 2026-08-15
feat(observable): name the watcher for what it watches, and document the plain-page path

jooyoungseo:
# Pull Request

## Description

Follow-up to #887. Two things that release got wrong, both of the same kind — the adapter works on any page with an Observable Plot chart, and both its API and its documentation said otherwise.

**The names.** `initQuartoObservable` and `stopQuartoObservable` shipped with Quarto in them because Quarto is what the adapter was written for. Nothing about them is Quarto-specific: the watcher looks for the class Plot puts on every chart it draws, its root defaults to `document.body`, and it binds charts on an ordinary page exactly as it does in a `.qmd`. Anyone using Observable Plot without Quarto had to call a function named for a tool they were not using.

**The guide.** It opened on Quarto and reached plain HTML nine sections later, under a heading beginning "Outside Quarto the adapter behaves the same way" — a footnote to the real case rather than the ordinary one it is. Most people using Observable Plot are not using Quarto at all.

## Related Issues

Follow-up to #887; no issue filed.

## Changes Made

**API** — `initObservablePlots`, `stopObservablePlots`, `autoInitObservablePlots`, and the options type `ObservableWatchOptions`. The module is `watcher.ts` rather than `quarto.ts`, since a file with nothing Quarto-specific in it should not be named for Quarto either.

**Back-compat** — the released names stay exported as deprecated aliases of the same functions, and `window.maidrObservable` carries both. **Not a breaking change**: a page written against the first release goes on working. The aliases are the contract rather than a courtesy, so they are tested — identity against the new functions, plus one that drives the watcher through the old name.

**Docs** — `docs/observable.md` now opens with a table of the two paths and what actually differs between them (only how the scripts reach the page), then gives each a quick start. Quarto keeps a section of its own and says *why* it needs one: an `{ojs}` cell's value **is** the chart, so there is nowhere in the cell for a binding call to go. That is the one case where watching the page is not a convenience but the only route — and stating it is what makes the plain-page case obviously optional rather than mysteriously absent.

**Example** — `examples/observable-plain.html`, the other half of the pair to `observable-quarto.html`: no Quarto, no Observable runtime, no cell structure, just Plot charts appended to a page. Including one appended after load and replaced on a button press, which is the case a dashboard hits and the one a watcher exists for. The gallery lists both and says how they relate.

n/a — non-visual by construction.

## Checklist

## Additional Notes

**Verification.** `npm run lint:fix`, `npm run type-check`, `npm run build`, `npm test` (4051 unit + 73 esm). Both names are present in the built bundle.

The claim that motivated this was checked before the change rather than assumed. A page with **zero** Quarto or OJS markers, three Plot charts, one of them appended 400 ms after load:

```
ojsMarkersOnPage: 0   chartsDrawn: 3   bound: 3   instrumented: 3   errors: none
```

And the new example driven in Chromium, including the old API name and the redraw path:

```
before:        bound 4, instrumented 4
both names present: true    initObservablePlots === initQuartoObservable: true
after redraw:  bound 4, instrumented 4, unbind events 1, errors none
```

Four charts bound, the redraw releases exactly one and does not accumulate.

**One judgement worth flagging.** Renaming the module file (`quarto.ts` → `watcher.ts`, and its test alongside) inflates the diff for no user-visible gain, and I nearly left it. I did it because leaving it would have fixed the wart at the API level and kept it one level down, where the next person reading the adapter meets it — the same half-measure that needed a second pass twice during #887.

https://claude.ai/code/session_019Ki1N1MqVzd3Y8uY5nRrgE

---
_Generated by [Claude Code](https://claude.ai/code/session_019Ki1N1MqVzd3Y8uY5nRrgE)_

### DIFF
--- a/docs/observable.md
+++ b/docs/observable.md
@@ -2,9 +2,67 @@
 
 MAIDR provides an adapter for [Observable Plot](https://observablehq.com/plot) — the grammar-of-graphics library built on D3 — that turns its charts into accessible, navigable visualizations with audio sonification, text descriptions, braille output, and keyboard navigation.
 
-It is also how MAIDR reaches **[Quarto](https://quarto.org) documents**. Quarto renders `{ojs}` cells with the Observable runtime, which ships Plot and draws the chart in the browser. Nothing in an `{ojs}` cell can call an adapter — the cell's value *is* the chart — so this adapter watches the page instead: add two script tags to the document header and every Plot chart in it becomes navigable, including cells that redraw when a reader moves a slider.
+There are two ways it gets used, and they differ only in how the scripts get onto the page:
 
-## Quarto Quick Start
+| | You are writing | Load the scripts by | Your chart code changes |
+|---|---|---|---|
+| **[Plain page](#plain-page-quick-start)** | HTML, or an app that renders Plot charts | putting two `<script>` tags in `<head>` | not at all |
+| **[Quarto](#quarto-quick-start)** | a `.qmd` with `{ojs}` cells | `include-in-header`, or `quarto add xability/maidr` | not at all |
+
+The adapter itself does not know which one it is in. It watches the page for Plot charts and binds each one as it appears — that is the whole mechanism, and it is why neither case asks you to touch the code that draws the chart.
+
+Quarto gets its own section because it is the case that *cannot* work any other way: an `{ojs}` cell's value **is** the chart, so there is nowhere in the cell to call a binder even if you wanted to. On a plain page you have the choice, and [Manual binding](#manual-binding) covers taking it.
+
+## Plain page quick start
+
+Two script tags. Draw charts however you already do.
+
+```html
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <title>My Observable Plot chart</title>
+    <!-- 1. MAIDR core, then the Observable Plot adapter -->
+    <script src="https://cdn.jsdelivr.net/npm/maidr/dist/maidr.js"></script>
+    <script src="https://cdn.jsdelivr.net/npm/maidr/dist/observable.js"></script>
+  </head>
+  <body>
+    <div id="chart"></div>
+
+    <script type="module">
+      import * as Plot from 'https://cdn.jsdelivr.net/npm/@observablehq/plot/+esm';
+
+      const data = [
+        { day: 'Mon', count: 120 },
+        { day: 'Tue', count: 240 },
+        { day: 'Wed', count: 180 },
+      ];
+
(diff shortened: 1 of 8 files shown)


jooyoungseo:
Pushed `69dc07b`. When I opened this PR I said Observable Framework and Observable notebooks were untested and that I was guessing they would work. I have now run both, and the guess was right — so the guide states it rather than omitting it.

**Embedded notebook** — real `@observablehq/runtime` 5.9.9 + `@observablehq/inspector` 5.0.1, cells defined through the runtime and rendered by the Inspector, which is the machinery an embed actually runs on:

```
initial      charts 3, bound 3, instrumented 3, titles ["Visitors", "(untitled)", "Series 1"]
after rerun  charts 3, bound 3, instrumented 3, titles ["Visitors", "(untitled)", "Series 2"]
             unbinds 1, errors none
```

The re-run is a genuine `main.redefine('n', …)`, so the Inspector replaced the cell's node exactly as it does when a reader changes an input. One unbind, no accumulation.

**Observable Framework** — 1.13.4, a real project built with `observable build` and served as a static site:

```
initial      charts 3, bound 3, instrumented 3, titles ["Visitors", "(untitled)", "Series 1"]
after input  charts 3, bound 3, instrumented 3, titles ["Visitors", "(untitled)", "Series 3"]
             unbinds 1, errors none
first chart  [{x: "Mon", y: 20}, {x: "Tue", y: 14}, {x: "Wed", y: 23}]
```

The input was driven the way a reader would, by filling the `Inputs.range` and dispatching `input`. The exact data matters here beyond checking the numbers: it proves Plot's live `scale()` survived, so Framework is **not** pre-rendering these charts to static markup — had it been, the adapter would have dropped to the tick-derived fallback, which refuses a log axis outright. That was the specific risk worth ruling out, and it is ruled out.

One practical detail the docs now carry: put the bundles in the source root and name them in `head`, and Framework rewrites and content-hashes the paths for you — `/maidr.js` came out as `./_import/maidr.de2ef3bb.js` without any help.

The only failing request on the Framework page was `fonts.googleapis.com`, its default theme font, blocked because this sandbox has no outbound network. Nothing of ours.

**No follow-up issue filed, because nothing is broken.** The gap was documentation, and this PR is the one about where the adapter applies, so it belongs here rather than in a separate issue.

**What I did not test, and am not claiming:** a notebook hosted on observablehq.com. The adapter needs its two scripts on the page and there the head is not yours; whether a cell can inject them into its own document is a question I did not answer, so the guide says exactly that instead of implying either answer.

Gate on `69dc07b`: lint, type-check, build, 4051 unit + 73 esm.

---

Issue #413, opened 2025-08-12
feat: Go-to (g) navigation for min/max points in single & multi line plots


jooyoungseo:
# Feature Request

## Motivation
Users exploring line plots need rapid keyboard-driven navigation to statistically significant points (e.g., minima and maxima) for auditory and assistive review workflows. Currently, moving to extrema requires manual sequential navigation, which is time-consuming and error-prone—especially in dense or multi-line plots where extrema are not visually evident to blind or low-vision users. Providing a focused "go to" (g) action improves efficiency, reduces keystrokes, and aligns with accessibility goals by presenting a concise list of semantic targets.

## Proposed Solution
Introduce a keyboard shortcut 'g' (when focus is within a line plot trace) that opens a lightweight React list view (modal/overlay) showing contextually relevant jump targets for the CURRENT active line (trace):

List items (initial scope):
1. Maximum value in current line
2. Minimum value in current line

Interaction model:
- Press 'g' -> opens the list (focus placed on first item).
- Up/Down Arrow: cycles through list items (loop or bounded; recommend bounded with disabled wrap to match other menus—confirm).
- Enter: execute selected target, close list.
- Escape: close list without action.
- While list is open, non-navigation keystrokes are ignored/passed through only after closing.

Behavior when executing target:
- Compute all indices where the selected extremum occurs (may be >1 if ties).
- Maintain an internal per-trace cursor for each extremum type (max/min) so repeated execution cycles through subsequent tied occurrences.
- After jumping, move focus to the data point (consistent with existing navigation focus semantics) and trigger playback (audio + any highlight event) exactly as if the user navigated there normally.
- If multiple occurrences exist:
  - On first jump: announce (TTS/audio text service) "{n} maxima found. Jumped to 1 of {n}." (or "minima").
  - On subsequent jumps: announce "Maxima 2 of {n}" etc.
  - After last occurrence, wrap back to first (announce "Maxima 1 of {n}" again) OR stop? (Open question; propose wrap for predictability.)

Multi-line plots:
- Current active line is determined by existing navigation focus context (no new global state). The list content always pertains to that active line only.
- If 'g' is pressed with no active line (edge case), play an error/notification: "No active line selected" and do nothing.

## Architectural Placement (Respecting Layered Design)
- Core Model: No changes to fundamental data structures; extrema are derived, not stored.
- Service Layer: Add a `LineNavigationService` enhancement or dedicated `ExtremaService` that:
[...]

PR #902, opened 2026-08-15
feat(area): announce the step direction of a filled staircase

jooyoungseo:
The core half of xability/py-maidr#413.

`line.shape` and a fill are independent in every library that has both, so a band can be a staircase:

```python
go.Scatter(x=[1, 2, 3], y=[1, 2, 3], stackgroup="one", line=dict(shape="hv"))
```

A **stacked step area** is the standard way to draw a cumulative count that changes at discrete events, so this is an ordinary chart rather than a contrived one. Read as a smoothly interpolated band it tells a reader the wrong thing about every interval: that the value slid between samples, when it in fact held and then jumped.

## Why this belongs here rather than in the producer

The issue asked for exactly this to be checked before deciding, so I checked. `stepDirection` is read in exactly one place — `src/model/step.ts:64` — and `AreaTrace` extends `LineTrace`, not `StepTrace`. Neither `AreaTrace` nor `LineTrace` reads it:

```
$ grep -n "stepDirection" src/model/area.ts src/model/line.ts src/model/abstract.ts
(no matches)
```

So a producer emitting `stepDirection` on an area layer would have been a **no-op**. The adapter cannot fix this alone; the core has to learn to read it.

## Carried alongside the fill, not instead of it

The two facts are orthogonal and a reader needs both. A stepped stacked area still announces itself as a stacked area, still reports `Minimum total` / `Maximum total`, and now also says how the boundary moves. Two of the five tests assert exactly that — they're the half that a *replacement* rather than an addition would have broken.

Announced for every variant, not only a stacked one: an unstacked step area is misread the same way. That required restructuring `description`, which previously early-returned for an unstacked layer before any stat could be added.

`STEP_DIRECTION_LABEL` is exported from `step.ts` rather than copied, so the two cannot drift into describing the same convention differently — the same reasoning as the recent de-duplication in the Python binding.

## Verification

- `npm run type-check` — passes.
- `npm run lint:fix` — clean.
- `npm run build` — **all builds complete in 123.5s** (13/13 adapters).
- `npx jest test/model/` — **850 passed, 57 suites**, including 5 new.
- Falsification: dropping the announcement fails **3 of 5**; the other two are the area-reading and totals assertions that guard against it being a replacement.

## Follow-up

With this in, the py-maidr side of #413 becomes a small producer change: emit `stepDirection` on an area layer when `line.shape` says so. I'll open that separately once this lands, since until then it would emit a key nothing reads.

---

### DIFF
--- a/src/model/area.ts
+++ b/src/model/area.ts
@@ -1,8 +1,8 @@
-import type { MaidrLayer, StepDirection } from '@type/grammar';
+import type { LinePoint, MaidrLayer, StepDirection } from '@type/grammar';
 import type { DescriptionState, TextState, TraceState } from '@type/state';
 import { TraceType } from '@type/grammar';
 import { LineTrace } from './line';
-import { STEP_DIRECTION_LABEL } from './step';
+import { STEP_DIRECTION_LABEL, stepDataVertices } from './step';
 
 /** Label the running total is announced under. */
 const TOTAL = 'Total';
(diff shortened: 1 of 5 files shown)


jooyoungseo:
Both findings were real. Fixed in 8599684b — the second one especially, because shipping the announcement alone would have set exactly the trap you describe.

### 2. The highlight mismatch — confirmed, and fixed here rather than deferred

I wrote the failing test before deciding, and it fails as predicted. A four-sample `hv` band:

| highlight | expected | landed on |
|---|---|---|
| 1 | `(10, 100)` | `(10, 100)` ✅ |
| 2 | `(20, 300)` | **`(20, 100)`** — the held value at the next x |
| 3 | `(30, 60)` | **`(20, 300)`** |

The mechanism is sharper than "the same interleaved shape as a step path". A stepped band carries **both** kinds of surplus at once: the corners are interleaved *and* the return journey is trailing. So the inherited trim-from-the-end removes the baseline — correctly — and leaves the corners sitting among the samples. That's also why the plain-area case needed no help and `areaHighlight.test.ts` already passed: for a plain band the surplus is only trailing, which the existing test explicitly calls "load-bearing rather than incidental".

The fix drops the baseline first, then hands what remains to the same vertex selection `StepTrace` uses — now extracted as `stepDataVertices` and shared. Repeating it would have let the two disagree about which vertex is a sample, which is precisely the failure being fixed.

**And I walked into the trap `StepTrace` documents.** My first attempt read `this.stepDirection` in the override. That method runs during `super(layer)`, before this class assigns anything, so the field was `undefined` for every band and fell straight through to the inherited trim — the test failed identically, with no change at all. `StepTrace`'s comment warns about this for `stepPoints`; I quoted that comment in my own code and then made the same mistake one field over. `AbstractTrace` assigns `layer` before any subclass work, so the layer is read instead. Falsification pins both halves: reading the field fails 1, removing the override fails 1.

You were right to ask for it to be confirmed or explicitly scoped. Confirmed, and in scope — an announcement that promises step semantics while the highlight points elsewhere is worse than the silence it replaces.

### 1. Stale JSDoc — fixed

`MaidrLayer.stepDirection` said "Ignored by every other trace type", which my change made false, and a producer reading it would reasonably have concluded that emitting it on an area layer was a no-op. It now names both readers and keeps the "omit rather than guess" guidance.

### Nit — taken

`steppedLayer` uses the shared `StepDirection` type.

### Verification

- `npx jest test/model/` — **851 passed, 57 suites** (was 850; the new highlight test).
- `npm run type-check`, `npm run lint:fix` — clean.
- `npm run build` — **13/13 adapters, 120.1s**.

---

Issue #901, opened 2026-08-15
feat(amcharts): seven scoped chart types never landed — flow, hierarchy, map and gauge


jooyoungseo:
## What happened

#869 scoped 22 trace types for the amCharts adapter. #879 shipped it and reported 16. **The six-item difference was never reported as skipped**, and a seventh (`ALLUVIAL`) was deferred to #885 and skipped there for reasons that turn out to describe a larger gap than that one type.

Verified on `main` (`e13ba101d`):

```
$ for t in SUNBURST CHORD GAUGE SANKEY NETWORK CHOROPLETH ALLUVIAL; do
    grep -rn "TraceType.$t" src/adapters/amcharts/ | wc -l; done
0 0 0 0 0 0 0

$ # control — the same check against an adapter that does support them
$ for t in SUNBURST SANKEY GAUGE CHOROPLETH; do
    grep -rn "TraceType.$t" src/adapters/highcharts/ | wc -l; done
3 3 2 4
```

The adapter emits 31 trace types today. These seven are not among them, and `docs/amcharts.md`'s table does not claim them — so the documentation is honest; only the issue accounting was not.

## The seven

| TraceType | amCharts construct |
|---|---|
| `SANKEY` | `am5flow.Sankey` |
| `CHORD` | `am5flow.Chord`, `ChordDirected`, `ChordNonRibbon` |
| `ALLUVIAL` | `am5flow.Sankey` with nodes repeated across stages (no separate class) |
| `NETWORK` | `am5hierarchy.ForceDirected`, also `am5flow.ArcDiagram` |
| `SUNBURST` | `am5hierarchy.Sunburst` |
| `CHOROPLETH` | `am5map.MapChart` + `MapPolygonSeries` with a `valueField` and `heatRules` |
| `GAUGE` | `am5radar.RadarChart` configured as a gauge — circular `ValueAxis` with a `ClockHand` |

## Two leads worth checking before designing

[...]

PR #903, opened 2026-08-15
feat(amcharts): read flow, hierarchy, map and gauge charts

jooyoungseo:
# Pull Request

## Description

Adds the seven amCharts trace types [#901](https://github.com/xability/maidr/issues/901) found missing — `SANKEY`, `CHORD`, `ALLUVIAL`, `NETWORK`, `SUNBURST`, `CHOROPLETH` and `GAUGE`. All seven are now detected, extracted to the payload shape `src/type/grammar.ts` declares, documented and exercised by unit tests. Three of them are highlighted; four ship deliberately **without** a highlight, for a reason stated below and in `docs/amcharts.md` rather than papered over with a guessed outline.

The change is confined to `src/adapters/amcharts/`, `test/adapters/amcharts/`, `examples/` and `docs/amcharts.md`. **No core change was made** — nothing under `src/model/`, `src/service/`, `src/state/`, `src/ui/`, `src/command/`, `src/type/` or `src/adapters/shared/` is touched.

## Related Issues

Refs #901 — deliberately **not** `Closes`.

All seven types are accounted for and none is silently dropped, which is the accounting failure #901 exists to correct. But #901's own acceptance criterion says a type that sonifies and does not highlight "is not done", with an escape only where a highlight genuinely *cannot* be addressed. For `SANKEY`, `CHORD`, `ALLUVIAL` and `NETWORK` the honest position is narrower than that escape: the highlight *can* be addressed, but only by a `src/model/` change outside this PR's scope. So those four are shipped as reading-without-outline with the shortfall documented, and the issue is left open for a maintainer to close once the follow-up lands. Auto-closing it here would overclaim.

### Reconciliation — all seven, individually

| TraceType | Detection + payload | Highlight | Notes |
|---|---|---|---|
| `SUNBURST` | ✅ landed | ✅ wired | `am5hierarchy.Sunburst`. Wedge measured from radius/sweep, not the degenerate box a `Slice` reports (#774). |
| `GAUGE` | ✅ landed | ✅ wired | `RadarChart` + `ClockHand`. Payload is a **single `GaugePoint` object**, not an array, per the acceptance criterion. |
| `CHOROPLETH` | ✅ landed | ✅ wired | `am5map.MapChart` + `MapPolygonSeries`. Both detected and declarable. |
| `SANKEY` | ✅ landed | ❌ **not wired** | Reading is complete; overlay clears as the cursor moves. Reason below. |
| `CHORD` | ✅ landed | ❌ **not wired** | Same. All three `Chord` classes plus `ArcDiagram`. |
| `ALLUVIAL` | ✅ landed | ❌ **not wired** | Same. Declared (`{ type: "alluvial" }`), as no amCharts class distinguishes it. |
| `NETWORK` | ✅ landed | ❌ **not wired** | Same. `am5hierarchy.ForceDirected`. |

**Why the four flow/network types have no highlight.** #901 hoped #899's `highlightedPointIndices` had removed this class of problem. It has — but only from the *model* side, and the getter does not yet exist on `FlowTrace` or `NetworkTrace`. Without it the adapter would have to re-implement the longest-path stage layering in `src/model/flow.ts` and the component walk in `src/model/network.ts`, and a copy of model logic drifts and then outlines a confidently wrong node. So no resolver is registered: `NavMap.resolve` answers `[]` and the overlay clears. Audio, text, braille and keyboard navigation all work normally. An empty outline is truthful; a box drawn from a guessed layering is not — that is the failure #895 refused, and this PR refuses it too.

The follow-up is a small `highlightedPointIndices` getter on each trace, after which these four are resolvable by registering resolvers and nothing else. **It still needs filing by hand** — I had no issue-filing tool in the sessions that produced this work. One correction it must carry: the shape originally proposed (`[...node.out, ...node.in].map(e => e.at)`) is **wider** than what the SVG path actually highlights — `FlowTrace.mapToSvgElements` takes `touching[0]` and `NetworkTrace.mapToSvgElements` uses `node.linkAt[0]`, i.e. one ribbon per node. Shipping the wider shape verbatim would make the canvas highlight disagree with the SVG highlight for the same trace.

## Changes Made

**Detection and extraction** (`extractor.ts`, `adapter.ts`)
- `STANDALONE_KINDS` gains `Sunburst`, `Sankey`, `ArcDiagram`, the three `Chord` classes and `ForceDirected`; `SeriesKind` gains `sunburst`, `sankey`, `chord`, `network`, `choropleth`.
- New extractors: `extractFlowPoints`, `extractNetworkPoints`, `extractGaugePoint`/`extractGaugeBands`, `extractChoroplethPoints`. New layer builders for each, wired into `buildChartLayers`.
- New chart gate `asMapPanel` for `am5map.MapChart`, alongside the existing XY and percent gates.
- Gauge is asked for only when the series loop produced **zero** layers, so it can never steal a `RADAR` or `POLAR_AREA` reading.

**Highlight** (`navmap.ts`, `overlay.ts`)
- `buildHierarchyResolver` gains a `kind` parameter; sunburst passes `slice`, treemap and icicle keep `column`.
- `buildGaugeResolver` synthesises a duck-typed data item so the `ClockHand` reaches the existing column path with no `overlay.ts` change.
- `buildChoroplethResolver` mirrors `ChoroplethTrace.arrange`'s lat/lon banding, registered only when the filtered polygon count matches the layer, and **pinned by a contract test against the real trace** built through `TraceFactory` — the first mirror in this adapter shipped with one.
- New `kind: 'region'` + `regionRect` in `overlay.ts`: prefers `globalBounds()` for an irregular projected shape and returns `null` when neither source reports a box **with area**, so a degenerate `Graphics` clears the overlay instead of drawing a confident 1px mark.
- `groupSeries` moved from `binder.tsx` to `navmap.ts` and **exported**, so it is unit-testable for the first time — it is the edit site whose omission is silent (audio and text keep working; only the highlight vanishes).

**Declarations** (`declaration.ts`) — `AlluvialDeclaration` and `ChoroplethDeclaration` join the `AmDeclaration` union. `backsDeclaration` (not `isDeclarable`) carries the series-class check, so the failure message matches the actual failure.

**Correction to the original design, worth a reviewer's eye:** the gauge's min/max were specified to fall back to the axis' `start`/`end`. On an am5 `Axis` those are **relative zoom positions** (0–1), not extremes — that fallback would have reported every computed-extreme gauge as a confident 0-to-1 dial, and `GaugeTrace` pitches its tone against the range. `getPrivate('min')`/`getPrivate('max')` is used instead. A dial with no finite ends emits no layer rather than a dial of `NaN`s.

**Tests** — 249 amCharts tests pass. Each new test was verified to fail without its hunk by reverting the source and re-running, except a small set of deliberate absence-guards (e.g. "an ordinary radar chart stays a radar, hand or no hand"; "no resolver is registered for a sankey, so the overlay clears") which are non-discriminating by construction but are what turns a future regression into a red build. Those are called out in the batch notes rather than passed off as discriminating.

**Docs and examples** — `docs/amcharts.md` gains classification bullets, a support-table row per type, declared-field rows, sections for Gauge, Sunburst, the flow family and Choropleth, and a `#### Flow and network are not outlined` subsection stating the shortfall and its mechanism. New `examples/amcharts-flow.html`, `amcharts-gauge.html`, `amcharts-choropleth.html`; `amcharts-treemap.html` gains a sunburst panel.

Not applicable — amCharts is commercial and not installed in this repo, so the example pages cannot be rendered here. They load amCharts from CDN and are intended to be opened by a reviewer who has access.

## Checklist

- [ ] I have tested the changes locally following `ManualTestingProcess.md`, and all tests related to this pull request pass. — *Unit tests pass (see below); the manual process was **not** run, because amCharts is not installed here and the example pages cannot be exercised in this environment.*

Gate, run in order on this branch, all passing:

| Check | Result |
|---|---|
| `npm run lint:fix` | ✅ clean, no files changed |
| `npm run type-check` | ✅ clean (`tsc --noEmit` + `tsconfig.ci.json`) |
| `npm run build` | ✅ all 13 bundles built |
| `npm test` | ✅ 4237 passed / 4237 (283 suites) + 73 passed / 73 (ESM), 0 failures |

No SIGKILL and no flake occurred; the suite was run twice with identical results.

## Additional Notes

**amCharts is commercial and is not installed in this repository, so none of the following could be verified against the real library.** The adapter is duck-typed throughout for exactly this reason. A reviewer with amCharts access should check these specifically — they are listed roughly worst-consequence first.

*Assumptions where a wrong guess produces a **wrong reading** (highest priority):*

1. **The class-name lists are exhaustive and literal** — `Sankey`, `ArcDiagram`, `Chord`, `ChordDirected`, `ChordNonRibbon`, `ForceDirected`, `Sunburst`, and the five map classes (`MapPolygonSeries`, `MapLineSeries`, `MapPointSeries`, `ClusteredPointSeries`, `GraticuleSeries`). This is the sharpest risk in the PR: `classifySeriesKind` answers `'bar'` for anything it does not recognise, so a **misspelled or missing class name is announced as a bar chart of its links or shapes**, not skipped. Each name is pinned by a test, but a test cannot know the real spelling.
2. **`MapPolygon.geoCentroid()` returns `{ longitude, latitude }` in degrees.** A wrong unit would reverse the compass. Mitigated three ways: the call is wrapped in `try`/`catch`, both numbers must be finite *and* within ±180/±90, and the author's own row is read **first**, so a table stating its coordinates never depends on this.
3. **An `am5flow` series' `dataItems` are the links (one per ribbon), not the nodes.** The whole flow payload rests on this. If a build files nodes there too they are dropped for want of a resolvable end — quiet and in the safe direction, but the link count would then be right only by luck.

*Assumptions where a wrong guess degrades to silence or a poorer reading (still worth checking):*

4. `am5radar.ClockHand`'s `className` is literally `'ClockHand'` — the entire discriminating half of the gauge signature. A miss means no gauge is ever detected (silent no-op). It cannot steal `RADAR`/`POLAR_AREA`, because zero layers is required first.
5. The hand is reachable as `axisDataItem.get('bullet').get('sprite')`, and its value is `get('value')` on that same item. Four probes cover the sprite (accessor and property, plus `item.bullets[]` both ways); the value is unmitigated — if it lives elsewhere the layer is dropped.
6. A made axis data item lands in `axis.axisRanges.values` or `axis.dataItems`. Both are read, ranges first.
7. Gauge band ranges expose `get('endValue')`, with a name via `get('label')` (a `Label` with `get('text')`, or a plain string) or `get('category')`. Unnamed bands fall back to `Band n`.
8. `ClockHand.globalBounds()` reports a box **with area**. A `ClockHand` is a `Container`, so unlike a `Slice` it should — but if it does not, the overlay clears rather than boxing the whole dial. Pinned by a test.
9. A `Sunburst` node hangs its wedge on `slice` or `graphics`, and that `Slice` carries `radius`/`innerRadius`/`startAngle`/`arc` as a pie's does. Both slots are probed; if the settings are absent the chain falls back and ultimately clears.
10. Flow link ends answer `get('sourceId')`/`get('targetId')`, then `get('source')`/`get('target')` as node items with `name`/`id`, then `dataContext[sourceIdField ?? 'from']` — three probes in order. A link no probe resolves is dropped, never guessed.
11. The link weight is `item.get('value')`, else `dataContext[valueField ?? 'value']`; and `'from'`/`'to'` are amCharts' default end-field names.
12. `am5hierarchy.ForceDirected` exposes its graph as a hierarchy (a root whose `children` are the nodes), as `Treemap` and `Partition` do; and `linkWithField` names a `dataContext` column holding an **array** of node references. Non-arrays are ignored and unresolvable entries skipped, leaving the parent-child edges — a correct if smaller network.
13. A polygon data item exposes its shape as `item.get('mapPolygon')`, and `MapPolygon.globalBounds()` reports a box with area. The latter is doubtful — a `MapPolygon` is a `Graphics` painted through a draw callback, exactly the shape that reports a degenerate point in #774 — so `regionRect` clears rather than drawing a hairline. Pinned by a test.
14. The default am5map row is `{ id: 'US-NV', value: N }`. If a build keeps the place name out of the chain's reach, regions are announced by code rather than name — poorer, not wrong, and what the `region` declaration exists to fix.
15. `series.get('heatRules')` answers with an array. Only used to widen the shaded-series gate; if not, a map shaded by heat rules alone is skipped and is fixable by declaring `{ type: "choropleth" }`.

**Deliberate omissions** (decisions, not oversights, each documented in `docs/amcharts.md`): the gauge's `target` field — amCharts has no duck-typable construct separating a bullet-chart target from a second hand, and a guessed target is a number the chart never stated; choropleth `neighbors` — adjacency is not recoverable from rendered geometry or from centroids, and the grammar sanctions the absence; and `MapPointSeries`/`MapLineSeries` as readings — a bubble map and a route map are different charts, and reading either through `ChoroplethPoint` would announce a region list the chart does not draw.

**One follow-up is outstanding and unfiled:** the `highlightedPointIndices` getter on `FlowTrace` and `NetworkTrace`, described under Related Issues. It needs filing by hand alongside this PR.

---

### DIFF
--- a/docs/amcharts.md
+++ b/docs/amcharts.md
@@ -91,6 +91,7 @@ Series are classified by their amCharts class name and field configuration:
 - `am5wc.WordCloud` → **word cloud**
 - `am5flow.Sankey` → **sankey**; `Chord`, `ChordDirected` and `ChordNonRibbon` → **chord**; `ArcDiagram` → **sankey**, because it carries weights and a network has nowhere to put one
 - `am5hierarchy.ForceDirected` → **network**
+- `am5map.MapPolygonSeries` bound to a `valueField` (or carrying `heatRules`) → **choropleth**; a polygon series with neither, and every `MapLineSeries`, `MapPointSeries` or `GraticuleSeries`, is the base geography and is skipped
 - `LineSeries` on a value axis whose renderer is `inversed`, carrying a genuine ranking → **bump** (rank over time); the `bump` option settles the cases the axis cannot
 
 ### Visual Highlighting
@@ -131,6 +132,7 @@ amCharts 5 renders to an HTML5 `<canvas>`, so there are no per-element SVG nodes
 | Sankey † | `am5flow.Sankey`, `ArcDiagram` | series class (requires `flow.js`) |
 | Chord † | `am5flow.Chord`, `ChordDirected`, `ChordNonRibbon` | series class (requires `flow.js`) |
 | Network † | `am5hierarchy.ForceDirected` | series class (requires `hierarchy.js`) |
+| Choropleth | `am5map.MapPolygonSeries` inside a `MapChart` | series class + a bound `valueField` or `heatRules` (requires `map.js`) |
 | Bump (rank over time) | `LineSeries` | value axis renderer `inversed: true` **and** values that are a ranking; or the `bump` option |
 | Gauge | *no series* — an `am5radar.ClockHand` on a `RadarChart` axis | chart class `RadarChart` + a `ClockHand` bullet, asked only when the chart's series produced no layer (requires `radar.js`) |
 | Survival (Kaplan-Meier) | `StepLineSeries` | **declared** — `userData: { maidr: { type: "survival" } }` |
@@ -140,6 +142,7 @@ amCharts 5 renders to an HTML5 `<canvas>`, so there are no per-element SVG nodes
 | Manhattan | the same, one series per chromosome | **declared** — `{ type: "manhattan" }` |
 | Scatter | the same | **declared** — `{ type: "point" }` |
 | Alluvial † | `am5flow.Sankey` | **declared** — `{ type: "alluvial" }` |
+| Choropleth (renamed fields) | `am5map.MapPolygonSeries` | **declared** — `{ type: "choropleth", value: "…" }` |
 
 † Read, announced, brailled and navigated — but **not outlined**. See [Flow and network are not outlined](#flow-and-network-are-not-outlined).
 
@@ -191,6 +194,9 @@ A field you leave out falls back to its canonical name and then to a short list 
 | `"volcano"` | the same, with `merge` defaulting to `false` |
 | `"point"` | `label`, `merge` (default `false`) — note the value is `"point"`, not `"scatter"` |
 | `"alluvial"` | nothing — the block is the whole declaration |
+| `"choropleth"` | `region`, `value`, `lon`, `lat` |
+
+`"choropleth"` is the one declared type the adapter also detects on its own. A `MapPolygonSeries` bound to a `valueField` needs no block; the block is for a map whose region name, shaded value or centroid pair lives in a column amCharts was never told about — which is also the only way to get the centroids, and so the arrow keys walking north and south, out of a table that spells them `east`/`north` or keeps them beside the value. It is accepted only on a `MapPolygonSeries`; a block on anything else is reported and the chart is read as what it was drawn as.
 
 `"alluvial"` takes no fields at all, and that is the point: an alluvial is the same weighted flow a sankey carries, drawn without a left-to-right budget, so the nodes, the links and their weights are already in the drawing and the only thing missing is which of the two readings it stands for. It is accepted only on an `am5flow` series that carries at least one readable link — a block on anything else is reported and the chart is read as what it was drawn as.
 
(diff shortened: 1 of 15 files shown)


jooyoungseo:
Two corrections for anyone reading the checks on this PR.

**The `claude-review` check is green but there is no review behind it.** Its comment is a progress checklist with six of seven items still unticked — the job exited after "Gather context" without reviewing `extractor.ts`, `navmap.ts`, `declaration.ts`, the tests or the docs, and without posting findings. The other review on the PR is a Copilot quota notice. So this PR currently has **zero** substantive automated review; the green tick should not be read as an approval. The Additional Notes list fifteen unverified amCharts API assumptions, and those still need a reviewer with library access — assumption 1 (the literal class-name lists) most of all, since `classifySeriesKind` defaults to `'bar'` and a misspelled class is announced as a bar chart of its links rather than skipped.

**The follow-up is filed.** The description says the `highlightedPointIndices` getter on `FlowTrace`/`NetworkTrace` "still needs filing by hand" — that is stale. It is #904, and it carries the correction about `touching[0]` / `node.linkAt[0]` rather than the wider `[...node.out, ...node.in]` shape. The four flow types staying un-outlined here is deliberate and stays that way in this PR: registering resolvers requires the `src/model/` change #904 owns, and mirroring the stage layering in the adapter is the drift #895 refused.

Checks at `8ba4c14`: 13 of 14 green (unit tests, type-check, build, lint, docs, CodeQL, preview). E2E (chromium) is still running after ~38 minutes; no E2E spec touches amCharts and this branch changes only `src/adapters/amcharts/`, `test/adapters/amcharts/`, `examples/` and `docs/amcharts.md`, so it should be unaffected — but it has not gone green yet and should be confirmed before merge.

---

PR #906, opened 2026-08-15
fix(highlight): let a flow or network trace publish the ribbon it highlighted

### LINKED ISSUE #904: a11y: let FlowTrace and NetworkTrace publish their highlights, so a canvas adapter can outline a ribbon
The last piece of #901. #903 lands SANKEY, CHORD, ALLUVIAL and NETWORK for amCharts as **readings without a highlight**: all four sonify, describe, braille and navigate correctly, but the overlay clears as the cursor moves.

## Why the adapter cannot fix it

amCharts renders to canvas, so there are no `selectors` to target and the highlight goes through the overlay, which is handed `{layerId, row, col}`. Turning that back into an am5 node means re-implementing:

- `FlowTrace`'s longest-path stage layering (`src/model/flow.ts` — `assignStages`/`buildNodes`, plus its cyclic-graph fallback), and
- `NetworkTrace`'s component walk (`src/model/network.ts`).

A copy of model logic in an adapter drifts, and the failure mode is an outline confidently on the wrong node. #895 refused to ship that, and #903 kept the same line.

## Why it is now cheap

#899 landed the mechanism: `PointCloudHighlightable` / `highlightedPointIndices` in `src/type/navigation.ts` lets a trace **publish** the marks it highlighted so the adapter never reconstructs them. `ScatterTrace` already does this and both canvas adapters consume it. `FlowTrace` and `NetworkTrace` implementing the same interface is the whole fix.

## One correction to carry — this is the part worth getting right

The obvious shape is wrong:

```ts
// WIDER than what the SVG path highlights — do not ship this
[...node.out, ...node.in].map(e => e.at)
```

The SVG path highlights **one ribbon per node**, not all of them:

- `FlowTrace.mapToSvgElements` takes `touching[0]`
- `NetworkTrace` uses `node.linkAt[0]`

Publishing the wider set would make the canvas highlight disagree with the SVG highlight **for the same trace and the same cursor position** — amCharts would outline every ribbon touching a node while Vega-Lite outlined one. That is a worse bug than the current honest blank, because it looks correct.

So `highlightedPointIndices` must return exactly what `mapToSvgElements` selects. Deriving both from one shared expression is the way to keep them from drifting again.

## Acceptance criteria

- [ ] `FlowTrace` and `NetworkTrace` implement `PointCloudHighlightable`, returning exactly the marks `mapToSvgElements` selects — ideally from a single shared expression rather than two parallel ones.
- [ ] amCharts registers the resolvers #903 deliberately left out, and the four types outline the right ribbon.
- [ ] A test asserts the canvas and SVG highlights agree for the same trace and cursor position — the specific disagreement described above.
- [ ] SVG adapters are provably unaffected.
- [ ] `docs/amcharts.md`'s "Flow and network are not outlined" subsection is removed once it is no longer true.
- [ ] `lint:fix`, `type-check`, `build` and `test` pass.

Related: #901 (the seven types), #903 (the PR that ships them without this), #899 (the mechanism), #897 (the same problem for scatter).

jooyoungseo:
# Pull Request

## Description

A sankey, chord, alluvial or network layer drawn on a canvas library produced **no visual highlight at all**. A canvas chart has no SVG elements for `selectors` to reach, so its only path to an overlay is the navigate callback — and that callback carried the *braille* position. For `FlowTrace` that position is deliberately transposed (`row` is the stage, `col` the node within it), and for neither trace does it name the ribbon or line the chart actually outlines. Turning it back into an am5 node meant re-implementing `FlowTrace`'s longest-path stage layering (`assignStages`/`buildNodes`, plus the cyclic-graph fallback) or `NetworkTrace`'s component walk inside an adapter, where a copy drifts and the failure mode is an outline confidently on the wrong node. #895 refused to ship that and #903 kept the same line.

#899 landed the mechanism that makes it cheap: `PointCloudHighlightable` / `highlightedPointIndices` lets a trace **publish** the marks it highlighted, so the adapter never reconstructs them. `FlowTrace` and `NetworkTrace` implementing the same interface is the model-side fix; registering four resolvers that invert those published indices is the adapter-side one. Both halves are here, so all four types now outline.

**The correction the issue asks for is the substance of this change.** The obvious shape —

```ts
[...node.out, ...node.in].map(e => e.at)   // do not ship this
```

— is **wider** than what the SVG path outlines. Verified against the code rather than taken on trust: `FlowTrace.mapToSvgElements` builds `touching = [...node.out, ...node.in]` and then returns `touching[0]`, and `NetworkTrace.mapToSvgElements` reads `node.linkAt[0]`. One ribbon per node, not all of them. Publishing the wider set would make amCharts outline every ribbon touching a junction while Vega-Lite outlined one — the two renderers disagreeing about the same trace at the same cursor position, which is worse than the old honest blank because it looks correct.

So both channels are derived from **one shared expression** — `FlowTrace.ribbonOf` and `NetworkTrace.lineOf` — rather than from two parallel ones. `mapToSvgElements` indexes its element list with it, and `highlightedPointIndices` returns it. They cannot name different marks without the shared function being wrong for both at once.

## Related Issues

Closes #904

All five acceptance criteria are covered: the model change, the agreement test, the SVG-unaffected proof, the four amCharts resolvers, and the docs caveat. The branch has since been rebased onto `main` with #903 in it, so the flow and network readings the adapter half needs — the declaration cases, the series kinds, `extractFlowPoints` / `extractNetworkPoints` — exist here now, and the `docs/amcharts.md` subsection that said these four are not outlined exists to be removed.

## Changes Made

**Model** (`src/model/flow.ts`, `src/model/network.ts`)

- Both classes now `implements PointCloudHighlightable` and expose `highlightedPointIndices`. Each gains one private static — `FlowTrace.ribbonOf(node)` / `NetworkTrace.lineOf(node)` — that answers "which declared record drew this node's mark", and `mapToSvgElements` is rewritten to read it instead of open-coding the selection. That rewrite is behaviour-preserving: the old flow branch filtered `undefined` out of the mapped elements before taking `[0]`, but `flat` comes from `Array.from(querySelectorAll(...)).map(...)` and the method has already returned `null` unless `flat.length === declared`, so every `edge.at` is in range and the filter could never drop anything.
- **No type change.** `PointCloudHighlightable`, its guard and `NavigateCallback.pointIndices` all shipped in #899 unchanged.
- **No wiring change.** `createNavigateObserver` feature-tests the trace it is handed, so both traces move onto the cloud branch by implementing the interface. `src/controller.ts` is untouched.

**Ordering, and why it is that way.** `ribbonOf` prefers `out[0]` and falls back to `in[0]`, which is the order `touching` was built in — outgoing is the direction the chart is drawn in and the ribbon `FORWARD` follows, and a sink has nothing leaving it. Both edge lists are sorted largest-first at construction, so this is the widest ribbon touching the node. `lineOf` reads `linkAt[0]`, parallel to the degree-sorted `links`, so it is the line to the biggest neighbour — and the line the rotor's first step travels, which is the agreement `networkHighlight.test.ts` already pinned. Both statics take the same `Node | null` the grid holds, so they read as the one expression they are.

**The `isInitialEntry` gate.** `highlightedPointIndices` returns `[]` until the cursor has entered the trace, because the inherited `AbstractTrace.highlight` reports out-of-bounds there. Without it the two channels genuinely disagree at one position — the freshly-constructed one — and the agreement claim would be true only *almost* everywhere. It is deliberately **not** gated on SVG availability: a canvas chart has no elements by definition, and that is the case this exists to serve.

**Adapter** (`src/adapters/amcharts/`)

- `navmap.ts` registers resolvers for SANKEY, CHORD, ALLUVIAL and NETWORK, consuming the published indices exactly the way `buildCloudResolver` (#899) does: a build-time alignment guard, an index lookup at resolve time, and an **empty answer rather than a guess** when either fails. No model logic — no stage layering, no component walk — is reconstructed in the adapter; the resolver only inverts an index against the list this adapter itself emitted.
- New `flowSeriesList` and `networkSeriesList` buckets in `SeriesGroups`/`groupSeries`. Sankey and chord share one bucket because they are one weighted graph drawn two ways, the way treemap and icicle already share theirs. A **declared** alluvial is never classified, so it keeps travelling in `declaredList` and reaches the same `buildFlowResolver` through `nextDeclared(ALLUVIAL)`.
- `extractor.ts`: `extractFlowLinks` is now the single walk over an am5flow series and `extractFlowPoints` is that list with the data items dropped — so the n-th emitted point and the n-th ribbon are aligned **by construction** rather than by two walks agreeing about the same drop rules. `flowRibbonOf` reads the drawn band off the link data item.
- `extractor.ts`: `networkLinkKey` is extracted as the one expression both `extractNetworkPoints`' de-duplication and the new `findNetworkLink` use, so a line drawn with its ends the other way round still matches. `findNetworkLink` walks `series.links` and matches by node category name, because a `ForceDirected`'s links are **not** data items — it is a hierarchy series, and amCharts keeps the drawn lines in a list of its own.
- `NavTarget` became a discriminated union (`NavItemTarget | NavRibbonTarget`): every other mark is named by the data item it hangs on, a ribbon by the sprite. `overlay.ts` gained a `'ribbon'` kind and `ribbonRect`, which measures the reported box, falls back to the sprite's own corners, and pads a perfectly straight link whose box has zero thickness — otherwise `intersectRect` collapses it. A sprite reporting nothing in either direction still clears.
- `types.ts`: `AmXYSeries` gains the optional `links` list.
- `adapter.ts`: the now-false "no highlight, on purpose" comments on `buildFlowLayer`, `buildNetworkLayer` and the declared-alluvial case are corrected.

**Docs.** `docs/amcharts.md`'s `#### Flow and network are not outlined` subsection is replaced by `#### Flow and network highlighting`, the summary sentence near line 101 is rewritten, and the four `†` markers on the Sankey/Chord/Network/Alluvial table rows and their footnote are removed. `grep -n 'not outlined\|un-outlined' docs/amcharts.md` returns nothing and no dangling anchor to the old section remains.

**Still no service, ViewModel or React change**, and the model still only calls `notifyStateUpdate()`, imports no service, and touches no DOM beyond the existing `Svg` helpers.

N/A — no UI surface of its own. The overlay is the existing amCharts highlight rectangle, now drawn on four more types.

## Checklist

The third box is unticked deliberately: the automated gate is fully green (below), but `ManualTestingProcess.md` was not walked by hand and no live canvas page was driven. amCharts is commercial and is not installed in this environment, so the adapter is duck-typed throughout and tested against plain fakes.

## Additional Notes

**What a reviewer should check hardest**

1. **That `ribbonOf` and `lineOf` really are the single source both channels read.** This is the point of the model half. If a later edit reintroduces a second expression on either side, the canvas and SVG highlights can diverge again and nothing about either announcement will look wrong. The two are adjacent to `mapToSvgElements` in each file so a drift is visible in a side-by-side read.
2. **That `mapToSvgElements` still outlines exactly what it outlined before.** The refactor is the one place an SVG regression could hide. Argued above from `Svg.selectAllElements` being dense plus the `flat.length !== declared` guard, and pinned by the existing `flowHighlight.test.ts` / `networkHighlight.test.ts` plus a new walk over *every* addressable cell of both traces.
3. **The two amCharts reads that cannot be verified here** — `dataItem.get('link')` for a flow band, and `series.links` with each link's `source`/`target` node data items for a network line. These are in the same class as every other sprite read in this adapter (`mapPolygon`, `geoCentroid`, `slice`/`graphics`): amCharts is commercial and not installed, so neither can be checked against the real library. Both are duck-typed and both degrade to `undefined` → empty resolve → cleared overlay, which is exactly the behaviour these four types shipped with. If a real build hangs them elsewhere, the failure mode is the old honest blank rather than a wrong outline. **Worth confirming against a live chart before release.**
4. **Whether `highlightedPointIndices` should follow the rotor.** It does not: it is derived from the grid position alone, exactly as `highlightValues[row][col]` is. Following `arrivedVia` or `flowWalk` would put the index channel somewhere the element channel is not, which is the bug this change exists to prevent — but it is the decision most worth arguing about.
5. **The alignment guards.** Each resolver re-extracts at build time and refuses to resolve at all if its list is a different length from `layer.data`. A mismatch means the chart moved underneath the extraction, and clearing is strictly better than a band picked by a stale index. Both guards have a dedicated drift test.

**Two design decisions worth naming**

- **`NavTarget` is a union now.** A flow ribbon hangs on the link's own data item, but a force-directed network's lines are not data items at all, so a ribbon target names a sprite instead. Keeping a single interface would have meant either an optional `dataItem` — breaking every existing assertion with a possibly-undefined error — or passing a data item the target has nothing to do with. The union costs ~37 mechanical narrowings in existing tests, all routed through one `itemOf()` helper in `helpers.ts`; type-only, no assertion changed meaning.
- **The ribbon padding rule.** A network link drawn perfectly horizontally or vertically reports a box with zero thickness, which `intersectRect` would collapse to nothing; it is padded to a 2px hairline band instead. That is real geometry, and it is distinct from the degenerate point of #774, which still clears.

**Test approach.** Every new test was proven to fail without the fix by mutating the source, not by trusting the diff:

- Reverting `src/model/flow.ts` and `src/model/network.ts` to their pre-change state fails **all 14** new model tests.
- Replacing the getters with the wider set the issue warns against — `[...node.out, ...node.in].map(e => e.at)` and `node.linkAt.filter(at => at >= 0)` — fails **5**, including both `the canvas and the SVG channels agree at every position` cases and both junction/hub cases. That is the specific disagreement, pinned specifically.
- Deleting the `isInitialEntry` gate alone fails the two `nothing is published before the cursor has entered` cases, so the gate is load-bearing rather than defensive.
- On the adapter side: unregistering the four resolvers fails **9** of the new navmap tests; dropping the flow bucket fails **6**; unsorting `networkLinkKey` fails the network test; neutering either alignment guard fails exactly its drift test; requiring area in `hasExtent` fails the straight-link test.

Every blank-answer assertion is paired with a positive control in the same test, so no test in this PR passes with *and* without the change.

**How the agreement is asserted.** `test/model/flowNetworkPointIndices.test.ts` walks every addressable cell of both traces and, at each, compares the *names* of the marks the two channels select: the SVG channel via the `data-mark` attribute on the outlined element, the canvas channel by inverting each published index against `layer.data` — which is exactly the inversion the adapter performs. Both traces are given a chart authored out of value/degree order, so a pairing by rank rather than by declared position fails too. `test/controller/navigateCallbackFlowNetwork.test.ts` runs the same claim through the shipped `createNavigateObserver` and `Context`, on layers carrying **no selectors at all** — the reported canvas case. `test/adapters/amcharts/navmap.test.ts` then drives a real `FlowTrace` and a real `NetworkTrace` end to end into the resolvers, with the network's drawn lines declared out of order and one pair reversed.

**Left undone, and why**

- **No e2e Playwright spec.** `e2e_tests/` has no canvas-chart harness — amCharts is bound from live JS objects, not from a static page MAIDR loads — and this adds no trace type, so `.claude/rules/testing.md`'s per-trace-type e2e requirement is not triggered.
- **No manual verification against a live page**, for the reason in point 3 above.

**Gate**, run on the final tree:

- `npm run lint:fix` — clean, no files changed.
- `npm run type-check` — clean (`tsc --noEmit` and `tsconfig.ci.json`).
- `npm run build` — all 13 builds complete.
- `npm test` — 285/285 unit suites (4266 tests) and 7/7 esm suites (73 tests), fully green.

### DIFF
--- a/docs/amcharts.md
+++ b/docs/amcharts.md
@@ -98,7 +98,7 @@ Series are classified by their amCharts class name and field configuration:
 
 amCharts 5 renders to an HTML5 `<canvas>`, so there are no per-element SVG nodes for MAIDR's usual highlighting. `bindAmCharts` instead draws an absolutely-positioned outline box over the canvas at the active data point's pixel geometry (computed via am5's `sprite.toGlobal()`) — the same overlay approach the Chart.js adapter uses. The overlay re-anchors on resize. Call the returned `dispose()` to unmount MAIDR, remove the overlay, and restore the chart. Highlighting is unavailable on the `fromAmCharts` JSON/attribute path.
 
-**Four types are read but not outlined**: sankey, alluvial, chord and network. They sonify, describe, braille and navigate correctly, and the overlay simply clears as the cursor moves. This is deliberate, and [the reason is written out below](#flow-and-network-are-not-outlined) — an empty outline is truthful, and a box drawn around a guessed node is not.
+A sankey, alluvial, chord or network layer is outlined one **ribbon** at a time — the band or line the cursor's node is read through — rather than by a box around the node itself; see [Flow and network highlighting](#flow-and-network-highlighting).
 
 ## Supported Chart Types
 
@@ -129,9 +129,9 @@ amCharts 5 renders to an HTML5 `<canvas>`, so there are no per-element SVG nodes
 | Dot Plot (Cleveland) | `LineSeries` | category axis + `strokes.template` hidden + bullets |
 | Lollipop | `ColumnSeries` | category axis + hairline `columns.template` width + bullets |
 | Word Cloud | `am5wc.WordCloud` | series class (requires `wc.js`) |
-| Sankey † | `am5flow.Sankey`, `ArcDiagram` | series class (requires `flow.js`) |
-| Chord † | `am5flow.Chord`, `ChordDirected`, `ChordNonRibbon` | series class (requires `flow.js`) |
-| Network † | `am5hierarchy.ForceDirected` | series class (requires `hierarchy.js`) |
+| Sankey | `am5flow.Sankey`, `ArcDiagram` | series class (requires `flow.js`) |
+| Chord | `am5flow.Chord`, `ChordDirected`, `ChordNonRibbon` | series class (requires `flow.js`) |
+| Network | `am5hierarchy.ForceDirected` | series class (requires `hierarchy.js`) |
 | Choropleth | `am5map.MapPolygonSeries` inside a `MapChart` | series class + a bound `valueField` or `heatRules` (requires `map.js`) |
 | Bump (rank over time) | `LineSeries` | value axis renderer `inversed: true` **and** values that are a ranking; or the `bump` option |
 | Gauge | *no series* — an `am5radar.ClockHand` on a `RadarChart` axis | chart class `RadarChart` + a `ClockHand` bullet, asked only when the chart's series produced no layer (requires `radar.js`) |
@@ -141,10 +141,8 @@ amCharts 5 renders to an HTML5 `<canvas>`, so there are no per-element SVG nodes
 | Volcano | hidden-stroke `LineSeries` with bullets, two value axes | **declared** — `{ type: "volcano" }` |
 | Manhattan | the same, one series per chromosome | **declared** — `{ type: "manhattan" }` |
 | Scatter | the same | **declared** — `{ type: "point" }` |
-| Alluvial † | `am5flow.Sankey` | **declared** — `{ type: "alluvial" }` |
+| Alluvial | `am5flow.Sankey` | **declared** — `{ type: "alluvial" }` |
 | Choropleth (renamed fields) | `am5map.MapPolygonSeries` | **declared** — `{ type: "choropleth", value: "…" }` |
-
-† Read, announced, brailled and navigated — but **not outlined**. See [Flow and network are not outlined](#flow-and-network-are-not-outlined).
 
 A `StepLineSeries` is piecewise constant — the value is held and then jumps — so it maps to MAIDR's step trace rather than to a line, and is announced and navigated as a step plot. amCharts positions the staircase from the axis cell rather than reporting a step convention, so the adapter emits no `stepDirection` and MAIDR's description does not name one.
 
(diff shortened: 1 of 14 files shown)


jooyoungseo:
Two follow-ups on the review notes above.

**The "outstanding" item is closed.** The description was synced before the last review ran, so its final section is reading a stale snapshot: the body now says `Closes #904`, the documentation checkbox is ticked, and both the adapter half and the retired `docs/amcharts.md` caveat are described. Nothing in the description still claims the amCharts side is deferred.

**The non-blocking nit is fixed rather than declined.** `77b40b9` narrows `NetworkTrace.lineOf` to `NetworkNode | null` — the grid is `(NetworkNode | null)[][]` and `current` is `NetworkNode | null`, so the `undefined` arm described an unreachable input and made the getter read unlike `FlowTrace.ribbonOf`, which it is meant to mirror. No behaviour change; the body already reached `linkAt[0]` through an optional chain.

Gate re-run on the final tree (`77b40b9`): `lint:fix` clean and no files changed, `type-check` clean, `build` all 13 targets, `npm test` 285/285 unit suites (4266 tests) and 7/7 esm suites (73 tests). All 14 CI checks are green at this head.

One thing genuinely left open for a human, repeated from the description because it is the only claim here that testing cannot settle: `dataItem.get('link')` and `series.links` (with each link's `source`/`target` node data items) are reads against a library that is commercial and not installed, so they are duck-typed like every other sprite read in this adapter and unverified against a real build. Both degrade to `undefined` → empty resolve → cleared overlay, which is the behaviour these four types shipped with, so a wrong guess would show as the old blank rather than a misplaced outline — but worth a look at a live chart before release.

---

PR #908, opened 2026-08-15
fix(line): read every path command, so a staircase highlights its samples

### LINKED ISSUE #907: fix(line): a staircase highlights nothing — the path parser drops H/V commands
Found while measuring plotly's stepped-area geometry for xability/py-maidr#413. Independent of that issue and larger than it: **every Plotly step chart currently gets no visual highlight at all.**

## Root cause

`LineTrace.extractPathCoordinates` (`src/model/line.ts:872`) reads a rendered path with two regexes:

```ts
const SVG_PATH_ML_REGEX = /[ML]\s*(-?\d+(?:\.\d+)?)[,\s]+(-?\d+(?:\.\d+)?)/g;
const SVG_PATH_C_REGEX  = /C\s*(?:...){2}(-?\d+(?:\.\d+)?)[,\s]+(-?\d+(?:\.\d+)?)/g;
```

`M`, `L` and `C` only. The SVG path grammar also has `H` (horizontal lineto), `V` (vertical lineto), `Q`/`T` (quadratic), `S` (smooth cubic) and their relative forms, and none of them are read.

That is not a theoretical gap. A staircase is *exactly* the shape a renderer draws with `H`/`V`, because every one of its segments is axis-aligned.

## Measured

Real Plotly.js output in Chromium, four samples per trace, reading `path.js-line` — the element py-maidr's selectors address:

| `line.shape` | `d` | vertices the parser recovers |
|---|---|---|
| `hv` | `M0,399H246.67V147H493.33V273H740V21` | **1** |
| `vh` | `M0,399V147H246.67V273H493.33V21H740` | **1** |
| `hvh` | `M0,399H123.34V147H246.67H370V273H493.33H616.67V21H740` | **1** |
| `vhv` | `M0,399V273H246.67V147V210H493.33V273V147H740V21` | **1** |
| `linear` | `M0,399L246.67,147L493.33,273L740,21` | 4 |
| `spline` | `M0,399Q158.85,161.34 246.67,147C324.5,...,493.33,273Q581.15,258.66 740,21` | **2** |

Only the `M` survives on a staircase — every subsequent vertex is an `H` or a `V`.

## What the reader gets

Confirmed with a jsdom test driving the real path strings through `StepTrace`:

```
LINEAR  highlightValues: [4]   circles: [(0,399), (246.67,147), (493.33,273), (740,21)]
HV      highlightValues: null  circles: [(0,399)]
```

With 1 coordinate against 4 expected, `stepDataVertices` matches neither `2N-1` nor `2N` and returns `null`; `LineTrace.reconcilePathCoordinates` then takes its `coordinates.length < 2` branch and pads with `NaN`; `mapViaPathParsing` sees the `NaN`, marks the series failed, and — with every series failing — `mapToSvgElements` returns `null`.

So: correct audio, correct braille, correct text, **no highlight, and nothing anywhere saying why**. Same silent-failure class as py-maidr#400 and #412.

`spline` is the milder version of the same bug: 2 vertices for 4 samples, so the highlights are interpolated along the `M`→`C`-endpoint chord instead of sitting on the samples.

## Fix

Replace the two regexes with a command tokenizer that walks the path, tracks the current point, and emits each drawing command's endpoint. That covers `H`/`V`/`Q`/`S`/`T` and the relative forms in one pass, and reproduces today's behaviour exactly for `M`/`L`/`C` — for a cubic it is still the endpoint that is emitted.

Two things the fix has to get right beyond parsing:

1. **`hvh` must not regress from "no highlight" to "wrong highlight."** Tokenized, plotly's `hvh` path yields 10 vertices for 4 samples, not the `2N = 8` `stepDataVertices` expects, because plotly emits consecutive same-direction moves (`H246.67H370` with no `V` between). Those are collinear artifacts of the renderer. Collapsing consecutive vertices that continue an axis-aligned run brings it to 8 and lands the highlights correctly. Without that step the count falls through to the trim-from-the-end branch and bunches every highlight at the left edge — worse than the current silence, since a wrong highlight is believed.
2. **Arcs (`A`/`a`) stay unhandled.** Their flag arguments are legally written without separators (`a1 1 0 011 1`), so a naive number scan mis-tokenizes them. Line-family geometry never contains an arc; leaving them out keeps the parser honest rather than confidently wrong.

## Scope

Everything that routes through `LineTrace.mapViaPathParsing`: `LineTrace`, `StepTrace`, `AreaTrace`, `SmoothTrace`. Any producer whose renderer emits `H`/`V` is affected, not only Plotly — this is a core parser gap, not a Plotly quirk.

jooyoungseo:
Closes #907.

## The defect

`LineTrace.extractPathCoordinates` read a rendered path with two regexes over `M`, `L` and `C`. The SVG path grammar also has `H`, `V`, `Q`, `S`, `T` and every relative form, and none of them were read.

That is not a corner. **Every segment of a staircase is axis-aligned**, so a renderer draws one with `H` and `V` — which meant a Plotly step chart parsed to its opening `M` and stopped.

Measured in Chromium, four samples per trace, off the `path.js-line` element py-maidr's selectors address:

| `line.shape` | `d` | vertices recovered |
|---|---|---|
| `hv` | `M0,399H246.67V147H493.33V273H740V21` | **1** |
| `vh` | `M0,399V147H246.67V273H493.33V21H740` | **1** |
| `hvh` | `M0,399H123.34V147H246.67H370V273H493.33H616.67V21H740` | **1** |
| `linear` | `M0,399L246.67,147L493.33,273L740,21` | 4 |
| `spline` | `M0,399Q158.85,161.34 246.67,147C…493.33,273Q581.15,258.66 740,21` | **2** |

With one vertex against four samples, `stepDataVertices` matches neither shape, `reconcilePathCoordinates` pads with `NaN`, `mapViaPathParsing` drops the series, and `mapToSvgElements` returns `null`. Confirmed by driving the real path strings through `StepTrace`:

```
LINEAR  highlightValues: [4]   circles: [(0,399), (246.67,147), (493.33,273), (740,21)]
HV      highlightValues: null  circles: [(0,399)]
```

Correct audio, correct braille, correct text, **no highlight, and nothing anywhere saying why** — the same silent class as py-maidr#400 and #412.

## The fix

Walk the path, tracking the current point, and take each drawing command's endpoint. That reaches `H`/`V`/`Q`/`S`/`T` and the relative forms in one pass, and `M`/`L`/`C` behave exactly as they did — a cubic still contributes its endpoint alone.

Arcs stay unread on purpose. Their flag arguments are legally written without separators (`a1 1 0 011 1` is three flags and a coordinate pair), so a plain number scan mis-tokenizes them and would push invented vertices. Line-family geometry never contains one.

Parsing more surfaced two consequences that had to be handled, or `hvh` would have gone from *no* highlight to a *wrong* one — which is worse, because a wrong highlight is believed:

**Redundant collinear vertices.** Plotly splits a `mid` staircase's horizontal runs into consecutive same-direction moves — `…H246.67H370…` with no `V` between — giving 10 vertices where the convention has `2N = 8`. Collapsing vertices that only continue an axis-aligned run recovers it.

This runs **as a fallback, never as a first pass**, and that ordering is load-bearing. A step chart holds its value across an interval, so two equal samples put three consecutive vertices on one horizontal line in a path that is *already* exactly `2N - 1`; collapsing first would drop the middle one and break a path that was right. `a well-formed staircase with a flat run is left alone` pins it.

The collapse also requires the middle vertex to lie *between* its neighbours, not merely on their line. `vhv` goes out along an axis and comes back, so its middle vertex shares an x with both while sitting outside their span — dropping it would erase a corner the renderer drew.

**The area baseline test.** `AreaTrace` dropped a stepped band's return journey when the path was "longer than a top edge" (`> 2N - 1`). A `mid` edge is `2N`, already longer than an `hv` one, so that fired on a path with **no baseline at all** and sliced away real samples. Since Plotly strokes the edge as its own `path.js-line`, that is the common case, not the exotic one. Now matched by exact count: a closed staircase is `3N - 1` (`hv`/`vh`) or `3N` (`mid`).

## Verification

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass: **284 unit suites / 4187 tests, 7 ESM suites / 73 tests**, plus 18 new tests in `test/model/pathCommands.test.ts`. Every `d` attribute in that file was read out of Chromium from real Plotly.js output rather than hand-written.

Falsification — each mechanism reverted in turn, and the tests that should fail do:

| reverted | failures |
|---|---|
| parser back to the original two regexes | **15 of 18** (the 3 survivors are the `M`/`L`/`C` pins, as intended) |
| collapse fallback removed | **2** — both `mid` cases |
| baseline rule back to `> 2N - 1` | **1** — the `mid` bare-edge band |

## Follow-on

This unblocks the highlight half of py-maidr#413: a stepped area could not have highlighted correctly no matter what the producer emitted, because its geometry never survived parsing. The producer change there is separate and small.

The `spline` row above is fixed as a by-product — quadratics were as invisible as `H`/`V`, so a Plotly spline's highlights were interpolated along a chord instead of sitting on the samples.

---

### DIFF
--- a/src/model/area.ts
+++ b/src/model/area.ts
@@ -41,6 +41,39 @@ function variantOf(type: TraceType): AreaVariant {
 }
 
 /**
+ * Drop a stepped band's return journey along the baseline, if it has one.
+ *
+ * A closed band walks its top edge and then comes back along the baseline,
+ * adding exactly one vertex per sample. So a closed staircase is `3N - 1`
+ * vertices for `hv`/`vh` and `3N` for `mid`, against `2N - 1` and `2N` for a
+ * top edge stroked on its own.
+ *
+ * Matched exactly rather than by "longer than a top edge", which is what this
+ * was and which misread a `mid` band: its `2N` top edge is already longer than
+ * the `2N - 1` an `hv` one has, so the test fired on a path with no baseline
+ * at all and sliced away real samples. Plotly strokes the edge separately —
+ * `path.js-line` carries no baseline — so that case is the common one, not the
+ * exotic one.
+ *
+ * A count matching neither is left whole: {@link stepDataVertices} answers
+ * `null` for it and the caller falls back to interpolation, which is the right
+ * recovery for geometry this does not recognise.
+ *
+ * @param coordinates - Vertices parsed from the path
+ * @param expected - How many samples the series has
+ * @returns The vertices with any baseline removed
+ */
+function withoutBaseline(coordinates: LinePoint[], expected: number): LinePoint[] {
+  if (expected <= 0) {
+    return coordinates;
+  }
+  const closedLengths = [3 * expected - 1, 3 * expected];
+  return closedLengths.includes(coordinates.length)
+    ? coordinates.slice(0, coordinates.length - expected)
+    : coordinates;
+}
+
+/**
  * Whether a magnitude is a real number the totals may include.
  *
  * A gap travels as `NaN`, and `NaN` spreads: summing one into a column total
@@ -144,10 +177,12 @@ export class AreaTrace extends LineTrace {
    * band, the second highlight sat on the held value at the next x rather than
    * on the sample, and every one after it was displaced too.
    *
-   * So the baseline is dropped first, and what remains is handed to the same
-   * {@link stepDataVertices} that `StepTrace` uses. Sharing that rather than
-   * repeating it keeps the two from disagreeing about which vertex is a
-   * sample, which is the failure this method exists to prevent.
+   * So {@link withoutBaseline} drops the return journey first, and what
+   * remains is handed to the same {@link stepDataVertices} that `StepTrace`
+   * uses. Sharing that rather than repeating it keeps the two from disagreeing
+   * about which vertex is a sample, which is the failure this method exists to
+   * prevent — and it is why a band whose edge is stroked on its own, carrying
+   * no baseline to drop, needs no separate handling here.
    *
    * @param coordinates - Vertices parsed from the path, mutated in place
    * @param row - Index of the series these coordinates belong to
@@ -170,12 +205,10 @@ export class AreaTrace extends LineTrace {
     }
 
     const expected = this.points[row]?.length ?? 0;
-    const closed
-      = expected > 0 && coordinates.length > 2 * expected - 1
-        ? coordinates.slice(0, coordinates.length - expected)
-        : coordinates;
-
-    const dataVertices = stepDataVertices(closed, expected);
+    const dataVertices = stepDataVertices(
+      withoutBaseline(coordinates, expected),
+      expected,
+    );
     if (dataVertices !== null) {
       coordinates.length = 0;
       coordinates.push(...dataVertices);
--- a/src/model/line.ts
+++ b/src/model/line.ts
@@ -12,18 +12,37 @@ import { MovableGraph } from './movable';
 
 const TYPE = 'Group';
 /**
- * Regex for extracting data point coordinates from SVG path `d` attribute.
+ * Splits a path `d` attribute into commands, each with its argument text.
  *
- * Matches M/L commands (direct endpoints) with comma or whitespace separators,
- * and C (cubic bezier) commands — extracting the endpoint (last coordinate pair).
- *
- * M/L: `M65,231.42` or `L 100 200` → captures (65, 231.42) or (100, 200)
- * C:   `C81,215,97,199,113,182`    → captures endpoint (113, 182)
+ * `A`/`a` are deliberately absent. An arc's flag arguments are legally
+ * written without separators — `a1 1 0 011 1` is three flags and a coordinate
+ * pair — so a plain number scan mis-tokenizes them and would push invented
+ * vertices. Line-family geometry never contains an arc, so an unread one is
+ * the honest answer; an arc in the middle of a path leaves the current point
+ * stale, which is no worse than the whole command being invisible as it was
+ * before.
  */
-const SVG_PATH_ML_REGEX
-  = /[ML]\s*(-?\d+(?:\.\d+)?)[,\s]+(-?\d+(?:\.\d+)?)/g;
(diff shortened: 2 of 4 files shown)


jooyoungseo:
Both findings were real. Fixed in `3b578e7`.

**1. The arc handling was worse than I documented — you were exactly right.** I reproduced it before changing anything:

```
M 0 0 L 10 10 A 5 5 0 0 1 20 20 L 30 30
→ [[0,0], [10,10], [5,5], [0,0], [1,20], [30,30]]
                   ^^^^^^^^^^^^^^^^^^^^ fabricated
```

Three invented vertices, exactly as you described: the argument group runs to the next command letter, so with `A` outside the class the arc's flags and coordinates were swept into the preceding `L` and consumed as further linetos. "Unread" was not what the code did.

The fix is to put `A`/`a` **in** the command class but leave it **out** of `SVG_PATH_ARITY`. It is then recognised — so its tokens stay inside it — and contributes nothing, because the walk's bound `i + arity <= args.length` is false immediately when `arity` is `undefined`. That is what I meant the comment to describe, and the comment now says which half does what and why. The stale-current-point consequence is stated as what it actually is: it misplaces only a *relative* command following an arc, and line-family geometry contains neither.

Two tests pin it — the plain form above, and `a5 5 0 001 1`, the packed-flag serialisation that motivated skipping arcs in the first place. Falsified by dropping `A` back out of the class: **2 of 22 fail**, and they are these two.

**2. `S`/`T` are now exercised.** `S 25 -20 30 0` after a `C`, and `T 20 0` after a `Q`, each asserting the endpoint rather than the control point.

Full run after both: `npm run lint:fix`, `npm run type-check` clean; **284 unit suites / 4191 tests** and **7 ESM suites / 73 tests** pass, with `pathCommands.test.ts` now at 22.

On the verification note — CI has since gone green on the previous commit (Unit Tests, Type Check, Build, Docs, CodeQL, E2E), and this push re-runs it.

---

PR #909, opened 2026-08-15
fix(trace): answer empty rather than throwing when the cursor has no point

### LINKED ISSUE #905: fix(line): a layer with an empty series throws when its state is read
Found while investigating xability/py-maidr#421, where a producer can emit such a layer.

## What happens

A `LINE` layer whose only series is empty throws as soon as its state is read:

```ts
TraceFactory.create({
  id: 'empty', type: TraceType.LINE, title: 'T',
  axes: { x: { label: 'X' }, y: { label: 'Y' } },
  data: [[]],
}).state;
```

```
TypeError: Cannot read properties of undefined (reading 'z')

  423 |       zData = point.z ? { z: { label: zLabel, value: point.z } } : {};
      at LineTrace.text        (src/model/line.ts:423:21)
      at LineTrace.state       (src/model/abstract.ts:507:18)
      at LineTrace.get state   (src/model/line.ts:906:29)
```

`point` is `this.points[this.row]?.[this.col]`, which is `undefined` for an empty series. The optional chaining stops one step short of the dereference — the guard is on the *property* (`point.z ? … : {}`, which correctly treats `""` as absent) and there is none on the *point*.

## Why it matters beyond the one layer

A throw out of trace construction propagates out of `new Figure(...)`, so a single malformed layer takes the whole figure with it rather than degrading to one silent layer. That is the outcome `rules/model.md` describes for an unregistered trace type, and it applies here too.

A producer can always emit an empty series — py-maidr does today for a subplot whose only trace is empty, which is what surfaced this. The core cannot assume the producer got it right.

## What a fix needs

Return the empty trace state rather than dereferencing. `AbstractTrace` already has the shape for it (`outOfBoundsState` / the `empty: true` branch), and other traces already answer that way, so this is about reaching the existing answer rather than inventing one.

Worth checking the same pattern in the sibling traces while there — anything that indexes `points[row][col]` and reads a property off the result without checking the point exists.

## Reproduction

Constructing the layer above and reading `.state` is enough; no DOM or adapter needed.

jooyoungseo:
Closes #905.

## The defect

Every populated branch of `AbstractTrace.state` reaches `points[row][col]` through one accessor or another, and none of them guarded the *point*. The optional chaining stops one step short — `this.points[this.row]?.[this.col]` is `undefined` for an empty series, and the property read off it throws:

```
TypeError: Cannot read properties of undefined (reading 'z')
  at LineTrace.text   (src/model/line.ts:423)
  at LineTrace.state  (src/model/abstract.ts:507)
```

The issue asked me to check the siblings while I was in there, and that paid out — it is not a `LineTrace` bug:

| type | before |
|---|---|
| `line` / `smooth` / `step` / `area` | throws on `point.z` |
| `bar` | throws on **`point.x`** — same defect, different accessor |
| `point` (scatter) | no throw |

## Why it is worse than one bad layer

A throw out of trace construction propagates out of `new Figure(...)`, so a single malformed layer takes the whole render with it rather than costing the one layer — the outcome `rules/model.md` describes for an unregistered trace type. And a producer can always emit an empty series; py-maidr did, for a subplot whose only trace had nothing to plot (xability/py-maidr#421, guarded on the producer side in py-maidr#422). The core cannot assume the producer got it right.

## The fix

One guard at the single funnel the accessors are reached through:

```ts
if (this.isEmptyAtCursor) {
  return this.outOfBoundsState;
}
```

Keyed off `dimension` rather than off any one trace's data, because that is the one shape every trace already answers in — which is why it covers `bar` and the line family together without touching either.

**It is evaluated at the cursor, and that is load-bearing.** `LineTrace.dimension` reports the *current series'* length as `cols`, so this covers a **ragged** layer — one series with points, another with none, which is what a producer emits when a group has no data for the range — and not only a layer that is empty throughout. Three tests pin that the populated series is still described, the empty one reports empty, and moving back describes the first again (the guard must not latch).

`outOfBoundsState` is the existing `empty: true` shape, so this reaches an answer the codebase already has rather than inventing one.

## Scatter is deliberately untouched

`ScatterTrace.dimension` floors at `Math.max(1, …)` so out-of-bounds audio panning matches the in-bounds geometry. It therefore never reports `cols: 0` and the guard cannot fire — but it also never threw, so nothing here regresses. It does report `empty: false` with `freq.raw: [null]` for an empty layer, which is inconsistent but is a separate question from this crash, and changing its floor would disturb panning that was set that way on purpose. Flagging rather than folding it in.

## Verification

`npm run lint:fix`, `npm run type-check`, `npm run build` clean; **285 unit suites / 4277 tests** and **7 ESM suites / 73 tests** pass, including 12 new ones in `test/model/emptySeriesState.test.ts`. No existing test depended on the throw or on an empty trace reporting `empty: false`.

Falsification — guard removed: **8 of 12 fail**. The 4 survivors are the populated-trace and populated-row cases, which is what they are there to hold still.

---

### DIFF
--- a/src/model/abstract.ts
+++ b/src/model/abstract.ts
@@ -472,6 +472,23 @@ export abstract class AbstractTrace extends AbstractPlot<TraceState> implements 
   }
 
   /**
+   * Whether the cursor has no point to report.
+   *
+   * Read from {@link dimension} rather than from any one trace's data,
+   * because that is the one shape every trace already answers in. It is also
+   * evaluated *at the cursor* — `LineTrace` returns the current series'
+   * length as `cols` — so this covers a ragged layer, where one series has
+   * points and another has none, and not only a layer that is empty
+   * throughout.
+   *
+   * @returns True when there is nothing at `(row, col)` to describe
+   */
+  protected get isEmptyAtCursor(): boolean {
+    const { rows, cols } = this.dimension;
+    return rows <= 0 || cols <= 0;
+  }
+
+  /**
    * Gets the current state of the trace including audio, braille, text, and highlight information.
    * @returns The current TraceState
    */
@@ -489,6 +506,21 @@ export abstract class AbstractTrace extends AbstractPlot<TraceState> implements 
         },
         warning: true,
       };
+    }
+    // Answer "nothing here" rather than describing a point that does not
+    // exist. Every populated branch below reaches `points[row][col]` through
+    // one accessor or another -- `LineTrace.text` reads `point.z`,
+    // `BarTrace`'s reads `point.x` -- and none of them guards the point
+    // itself, so an empty series threw a TypeError. That throw leaves trace
+    // construction, propagates out of `new Figure(...)`, and takes the whole
+    // render with it, so one malformed layer silences a figure that is
+    // otherwise fine (#905).
+    //
+    // Guarded here rather than in each accessor because this is the single
+    // funnel they are all reached through, and because a producer can always
+    // emit an empty series -- the core cannot assume otherwise.
+    if (this.isEmptyAtCursor) {
+      return this.outOfBoundsState;
     }
     return {
       empty: false,
--- a/test/model/emptySeriesState.test.ts
+++ b/test/model/emptySeriesState.test.ts
@@ -91,7 +91,7 @@ describe('a ragged layer', () => {
 
   test('the empty series reports empty rather than throwing', () => {
     const trace = raggedTrace();
-    (trace as unknown as { row: number }).row = 1;
+    trace.row = 1;
 
     expect(() => trace.state).not.toThrow();
     expect(trace.state.empty).toBe(true);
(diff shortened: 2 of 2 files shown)


jooyoungseo:
Both points actioned.

**The cast** — you were right, `row` is a public accessor on `Movable` and the cast bought nothing. Dropped in `7f08179`; type-check is clean and the 12 tests still pass.

**The `dimension` gap** — real, and **broader than diagnosed**. I reproduced it before filing, and it is not limited to heatmap/box/pie:

```
heatmap row=5 → TypeError: Cannot read properties of undefined (reading 'length')
line    row=5 → TypeError: Cannot read properties of undefined (reading 'length')
```

`LineTrace.dimension` has the same unchecked `this.lineValues[this.row].length`. So the trace family this PR's tests exercise is *not* covered for the out-of-range case either — the tests pass because they park the cursor on a row that exists and is empty, which is the case this guard is for.

That distinction is worth stating precisely, because the two failures look alike and are not: an **empty series** is a cursor sitting somewhere real with nothing in it, and `dimension` answers honestly with `cols: 0` — which is exactly what `isEmptyAtCursor` reads. An **out-of-range row** is a cursor sitting somewhere that does not exist, and `dimension` throws before the guard can read anything. No amount of guarding at the shared funnel fixes the second, which is why I agree it should not be folded in here: it has to be each trace's own indexing that is checked.

Filed as #910 with the reproduction, the reachability note about `getStateAt`, and the observation that the traces which *are* safe today (`BarTrace` wrapping its data as one row, `ScatterTrace` flooring at `Math.max(1, …)`) are safe by accident of their data shape rather than by a check — so it wants one pass over every `dimension` that indexes by `row`, not a patch of the four named.

On verification: CI has gone green on this branch — Unit Tests, Type Check, Build Project, Docs Build, CodeQL, E2E (chromium) and the lint jobs all passed on `05df170`, and this push re-runs them.

---

Issue #911, opened 2026-08-15
docs(examples): 127 example pages are invisible on the site, and 4 trace types have no hand-authored example


jooyoungseo:
Audited every `TraceType` in `src/type/grammar.ts` against `examples/` and `docs/SCHEMA.md` on `main` (`da131aa76`).

**The good news first:** all 50 trace types are exercised by at least one example page, and `docs/SCHEMA.md` documents all 50. No trace type is undocumented.

Two gaps behind that headline.

## 1. The gallery misses 127 of 199 example pages

`scripts/build-site.js` builds `examples.html` from a **hardcoded list of `loadHTML(...)` calls**. It names 88 pages. The directory holds 199.

```
example .html pages: 199
linked in gallery  : 88
NOT linked         : 127
```

So the great majority of the examples — including essentially every chart type added recently: `area`, `boxen`, `bump`, `choropleth`, `contour`, `dot`, `dumbbell`, `errorbar`, `forest`, `funnel`, `gantt`, `gauge`, `hexbin`, and the whole amCharts, Highcharts, Vega-Lite and Google Charts families — exist in the repo and are **unreachable from the site's own gallery**. A page nobody can find does not document anything.

This has already been noticed once and waved through: the Google Charts adapter PR recorded "the five new example pages are not linked from the example gallery" as a known, declined finding. That was the right call for that PR and the wrong outcome overall, because nothing tracked it.

**The hardcoded list is the root cause, not the symptom.** It has drifted by 111 entries and will drift again on the next example anyone adds. Generating the gallery from the directory — with grouping and labels derived from a small curated map plus a filename fallback — makes the drift structurally impossible. A test asserting every `examples/*.html` appears in the built gallery would keep it that way.

## 2. Four trace types have no hand-authored MAIDR JSON example

Every trace type has *some* example, but for these four the only examples are adapter-bound — the chart library renders and the adapter emits the type at runtime, so nothing in the page shows an author what the MAIDR JSON looks like:

| TraceType | value | only covered by |
|---|---|---|
| `CHORD` | `chord` | `highcharts-chord.html` |
| `NORMALIZED` | `stacked_normalized_bar` | `highcharts-normalized.html`, Recharts TSX |
| `NORMALIZED_AREA` | `stacked_normalized_area` | Recharts TSX |
| `POLAR_AREA` | `polar_area` | Recharts TSX, Victory TSX |

[...]

PR #912, opened 2026-08-15
docs(examples): add hand-authored pages for chord, normalized and polar area

jooyoungseo:
# Pull Request

## Description

Four trace types had no hand-authored MAIDR JSON example. Their only coverage was adapter-bound — the chart library renders and the adapter emits the type at runtime — so nothing in the page showed an author what the schema looks like. This adds one page each for `chord`, `stacked_normalized_bar`, `stacked_normalized_area` and `polar_area`, following the conventions of the existing hand-authored set (`area.html`, `dot.html`, `forest.html`, `sankey.html`, `radar.html`): an inline SVG whose element ids the layer's `selectors` actually resolve against, the schema carried literally in an HTML-entity-encoded `maidr='{...}'` attribute, and an opening comment saying what the type is *for* and how it differs from its nearest neighbour.

It also writes down why `candlestick_delta` gets no example, so a future audit of example coverage does not close that gap by inventing one.

## Related Issues

Refs #911 (the gallery-generation half of the issue lands separately, and that PR closes it).

## Changes Made

- **`examples/chord.html`** — twelve weighted flows between four regions, every pair drawn in both directions. The nearest neighbour is `sankey`, which shares the trace class and the `FlowPoint` payload; what makes a chord its own type is that it is cyclic, so the stage assignment does not sort and every node lands in one stage. The values are chosen so that following the widest ribbon out walks Americas → Europe → Asia → Africa → Americas — four presses and you are back where you started, which a sankey cannot do. `selectors` is one entry per flow, in declared order.
- **`examples/normalized-barplot.html`** — `stacked_normalized_bar`, four quarters × three channels, every column totalling 100. The derived summary row announces `Sum` = 100 at every column, which is the clearest statement of what the chart has normalised away. Rects are emitted bottom-segment-first within each column, the order MAIDR expects without a `domMapping`.
- **`examples/normalized-area.html`** — the same idea as bands over a continuous x. Against a plain `stacked_area` the top edge is flat at 100 by construction and carries nothing, so all the movement is in the bands. Each band is drawn as top edge out and baseline back, and MAIDR keeps only the first N vertices.
- **`examples/polar-area.html`** — a wind rose, where the arrangement around the circle genuinely is the data. Shares `RadarTrace` with `radar`; the difference is entirely the mark, and the stereo position follows the spoke's angle rather than its index.
- **`docs/SCHEMA.md`** — a short note beside the trace-type list explaining that `candlestick_delta` has no example on purpose: it is derived at runtime from a `candlestick` layer (`src/model/candlestickDelta.ts`, Alt+L), `TraceFactory.create` has no case for it, and a layer declaring it throws `Invalid trace type: candlestick_delta` rather than binding.

### One change beyond the task

`docs/SCHEMA.md` on `main` carries committed merge-conflict markers (`<<<<<<< HEAD` … `>>>>>>> 97ef5574`) around the `type` bullet — exactly where the note belongs. Placing a note inside a broken conflict block would have produced nonsense, so the markers are resolved, keeping the variant whose list includes `network` (a real declarable type with its own example page). The two sides differed only in that entry. Flagging it here because it is not strictly part of the requested change.

Not attached. The four pages were checked by rendering them in Chromium; what was verified is listed below rather than shown.

## Verification

All four pages were driven in a real browser (Chromium, `file://`, against a fresh `npm run build`). For each: MAIDR binds with no page errors, the layer reports the declared type, every `selectors` entry resolves, and navigation highlights the element the announcement names.

| Page | Announced type | Selectors resolve | Highlight lands where it should |
|---|---|---|---|
| `chord.html` | `chord` | 12 selectors → 12 ribbons | Each node highlights its widest outgoing ribbon (`ribbon-0`, `-6`, `-10`, `-5` for Americas, Europe, Asia, Africa) |
| `normalized-barplot.html` | `vertical stacked_normalized_bar` | 1 selector → 12 rects | Organic → top rect, Referral → middle, Paid → bottom, checked against each rect's geometry |
| `normalized-area.html` | `100% stacked area with 3 groups` | 3 selectors → 1 path each | Marker sits on the band's own cumulative top edge (62% → y 174, 92% → y 84, 100% → y 60 at 2019) |
| `polar-area.html` | `polar area` | 1 selector → 8 wedges | Each direction highlights its own wedge; box geometry matches the wedge's radius and angle |

Also confirmed on each page: braille renders real cells (one row per series; an unregistered encoder is the one failure that is silent), and the announcements are the ones the type exists to give — the normalized bar's summary row reads 100 at all four quarters, and every normalized-area point reads `Total is 100` with its own value as the share.

The `candlestick_delta` claim in `docs/SCHEMA.md` was checked the same way, with a throwaway page that declared the type: it throws `Invalid trace type: candlestick_delta` and the figure never binds. That probe page was not committed.

Not verified: nothing was checked by ear — sonification and stereo panning were not listened to, only the code path that sets them was read. No e2e specs were added for these pages; the existing suite was not run.

## Checklist

- [ ] I have added appropriate unit tests, if applicable. — no new code paths; the pages exercise existing traces, and the gallery-coverage test belongs to the sibling PR.

## Additional Notes

Checks run on this branch, all passing: `npm run lint:fix`, `npm run type-check`, `npm run build`, `npm test` (286 suites / 4288 tests, plus 7 suites / 73 tests in the second project). `examples/` and `docs/` are both in eslint's ignore list, so lint neither reformatted nor inspected the new files.

These four pages are not linked from `examples.html`, which is built from a hardcoded list — that is the other half of #911 and is fixed by the sibling PR that generates the gallery from the directory.

---

### DIFF
--- a/docs/SCHEMA.md
+++ b/docs/SCHEMA.md
@@ -123,16 +123,20 @@ Use the following to define the object properties:
 
 - `type`: the type of plot. The declarable types are `alluvial`, `area`, `bar`, `box`, `boxen`, `bump`, `candlestick`, `chord`, `choropleth`, `contour`, `diverging_bar`, `dodged_bar`, `dot`, `dumbbell`, `error_bar`, `forest`, `funnel`, `gantt`, `gauge`, `heat`, `hexbin`, `hist`, `icicle`, `line`, `lollipop`, `manhattan`, `mosaic`, `network`, `parallel_coordinates`, `pie`, `point`, `polar_area`, `radar`, `ridgeline`, `sankey`, `smooth`, `stacked_area`, `stacked_bar`, `stacked_normalized_area`, `stacked_normalized_bar`, `step`, `sunburst`, `survival`, `treemap`, `violin_box`, `violin_kde`, `volcano`, `waterfall`, `word_cloud`. `TraceType` in `src/type/grammar.ts` is the source of truth; `candlestick_delta` appears there but is built at runtime from a candlestick and a reference line, so it is not something a page declares.
 
-> **`candlestick_delta` has no example page, on purpose.** Every declarable
-> trace type above has a hand-authored example under `examples/` that carries
-> its schema literally in a `maidr='{...}'` attribute. `candlestick_delta` has
-> none and should never be given one: it is a reading mode the model derives at
-> runtime from a `candlestick` layer (`src/model/candlestickDelta.ts`, reached
-> with Alt+L), not a value a page declares. `TraceFactory.create` in
-> `src/model/factory.ts` has no case for it, so a layer declaring it throws
+> **`candlestick_delta` has no example page, on purpose.** It should never be
+> given one: it is a reading mode the model derives at runtime from a
+> `candlestick` layer (`src/model/candlestickDelta.ts`, reached with Alt+L), not
+> a value a page declares. `TraceFactory.create` in `src/model/factory.ts` has
+> no case for it, so a layer declaring it throws
 > `Invalid trace type: candlestick_delta` and the figure never binds — an
 > example would document something that does not exist. An audit of example
 > coverage that finds this gap has found the correct state of affairs.
+>
+> It is the only deliberate gap. Other declarable types that lack a
+> hand-authored example are genuinely missing one, and adding it is a normal
+> contribution — at the time of writing that is `alluvial`, `area` and
+> `manhattan`, which appear under `examples/` only as adapter-bound pages where
+> a chart library emits the schema at runtime.
 
 - `id`: the id that you added as an attribute of your main SVG.
 - `title`: the title of the plot. (optional)
--- a/docs/SCHEMA.md
+++ b/docs/SCHEMA.md
@@ -121,11 +121,19 @@ Or multiple plots:
 
 Use the following to define the object properties:
 
-<<<<<<< HEAD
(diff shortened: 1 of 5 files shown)


jooyoungseo:
Fixed the `docs/SCHEMA.md` overclaim in `0d8fe09`. Recording the audit, because the sentence was wrong in a second way the review did not reach.

I enumerated the declarable-type list against every `.html` under `examples/`, counting a type as hand-authored if the page carries the schema literally in any of the three forms actually in use — `maidr='{...}'`, `maidr-data='{...}'`, or an inline `const maidr = {...}` — and not counting adapter-bound pages where a chart library emits the type at runtime. Result: three declarable types have no hand-authored example, not one. `alluvial` (only `amcharts-flow.html`), `area` (only adapter-bound; `area.html` declares `stacked_area`, a different type), and `manhattan` (only `highcharts-manhattan.html`).

The second error: the claim named the `maidr='{...}'` attribute specifically, but a large share of covered types use one of the other two forms — `hist`, `dodged_bar` and `stacked_bar` are `maidr-data='{...}'`; `bar`, `box`, `heat`, `pie` and `candlestick` are inline-script. So the sentence was false about the mechanism as well as the coverage.

The universal claim was never load-bearing — the note exists to keep an audit from inventing a `candlestick_delta` example — so it is dropped rather than patched, and the three real gaps are now named as gaps. Filling them is ordinary work, not part of this PR.

Unrelated and deliberately untouched: the committed conflict markers in `docs/BRAILLE.md` are #917, a content decision.

---

PR #916, opened 2026-08-15
fix(trace): answer for a row that does not exist instead of throwing

### LINKED ISSUE #910: fix(trace): `dimension` throws on an out-of-range row, before any guard can run
Raised in review on #909 and confirmed — with one correction to the original diagnosis: it is not limited to heatmap/box/pie. `LineTrace` throws too.

## What happens

Several traces' `dimension` getters index the outer array with `this.row` and read `.length` off the result, with no bounds check:

```ts
// line.ts
cols: this.lineValues[this.row].length
// heatmap.ts:132
cols: this.heatmapValues[this.row].length
// also box.ts:275, pie.ts:321
```

So an out-of-range `row` throws inside `dimension` itself:

```
heatmap row=5 → TypeError: Cannot read properties of undefined (reading 'length')
line    row=5 → TypeError: Cannot read properties of undefined (reading 'length')
```

## Why the #909 guard does not cover it

#909 added `isEmptyAtCursor` to `AbstractTrace.state`, keyed off `dimension`. That fixes an **empty series** — where `dimension` answers honestly with `cols: 0` — but it cannot fix an **out-of-range row**, because `dimension` throws before the guard reads it. The two look similar and are not: one is a cursor sitting somewhere real that has nothing in it, the other is a cursor sitting somewhere that does not exist.

That is also why this wants its own change rather than an extension of #909. Fixing it properly means guarding `dimension` per trace — it is each trace's own indexing that is unchecked — not adding another test at the shared funnel.

## How it is reachable

`AbstractTrace.getStateAt(row, col)` assigns `this.row` / `this.col` directly with no bounds validation. It is what monitor mode uses to sonify a newly appended point without moving the user's cursor.

In practice it is called with indices that correspond to real appended data, so this is unlikely to fire today. It is a latent sharp edge rather than a live bug — but it is the same sharp edge as #905, one level up: the outer index is unchecked where the inner one now is, and the blast radius is the same, since a throw here propagates the same way.

Ordinary keyboard navigation cannot reach it: `MovableGraph`/`MovableGrid` bound the cursor to real positions.

## What a fix needs

Return a `Dimension` for a row that does not exist rather than throwing — `{rows, cols: 0}` is the natural answer, and it composes with #909's guard, which then reports the empty state for it.

Worth doing for every trace whose `dimension` indexes by `row`, in one pass, rather than the four named above: the ones that are safe today are safe by accident of their data shape (`BarTrace` wraps its data as a single row, so `rows` is always ≥ 1; `ScatterTrace` floors both at `Math.max(1, …)`), not by a check.

## Reproduction

```ts
const trace = TraceFactory.create({
  id: 'l', type: TraceType.LINE, title: 'T',
  axes: { x: { label: 'X' }, y: { label: 'Y' } },
  data: [[{ x: 1, y: 1 }]],
});
trace.row = 5;
trace.state; // throws
```

jooyoungseo:
Closes #910, which came out of review on #909.

## The defect

Five traces read their outer array with no bounds check in `dimension`:

```ts
cols: this.lineValues[this.row].length      // line.ts
cols: this.heatmapValues[this.row].length   // heatmap.ts
cols: this.barValues[this.row].length       // bar.ts
cols: this.boxValues[this.row].length       // box.ts
cols: this.sliceValues[this.row].length     // pie.ts
```

So a cursor parked past the end throws inside `dimension` itself:

```
line    row=5 → TypeError: Cannot read properties of undefined (reading 'length')
heatmap row=5 → TypeError: Cannot read properties of undefined (reading 'length')
```

## Why #909's guard does not cover it

`isEmptyAtCursor` is keyed **off** `dimension`, so `dimension` throwing happens first. The two failures look alike and are not:

| | cursor is | `dimension` says |
|---|---|---|
| **empty series** (#905) | somewhere real with nothing in it | `cols: 0` — honestly |
| **out-of-range row** (this) | somewhere that does not exist | *throws* |

That is why this could not be folded into #909 and why the fix is in each trace's own indexing rather than at the shared funnel.

## The fix

The optional index that **six** other traces already use — `choropleth`, `network`, `treemap`, `violin`, `violinBox`, and `Subplot`, the last of which already carries a comment explaining exactly this. So the idiom and its rationale were already established in the file; five sites had just missed it.

Worth naming: the traces that were *safe* were safe by accident of their data shape, not by a check. `BarTrace` wraps `layer.data` as a single row so `rows` is always ≥ 1, and `ScatterTrace` floors both at `Math.max(1, …)` for audio panning. Neither is a bounds test, which is why this went one pass over every `dimension` that indexes by `row` rather than patching the four the issue named.

## Scope: `Figure` is deliberately left out

`Figure.dimension` indexes `subplots[this.row]` the same way, and I did patch it first — then found it changes nothing observable. `Figure.state` reaches `activeSubplot`, which does its own unguarded `subplots[this.row][this.col]` and throws before `dimension` is consulted:

```
at Figure.get activeSubplot (src/model/plot.ts:245:35)
```

That is a separate read wanting a separate fix, and unlike a trace, `Figure` has no `getStateAt` to reach it — navigation bounds the cursor. I reverted the change rather than ship a guard I could not write a passing test for. Recorded in the test file's header so the next person does not re-derive it.

## Reachability

Through `AbstractTrace.getStateAt(row, col)`, which assigns `this.row`/`this.col` directly with no validation — what monitor mode uses to sonify a newly appended point without moving the user's cursor. In practice it is called with indices that correspond to real appended data, so this is a latent sharp edge rather than a live crash. It is the same edge as #905 one level up: the outer index unchecked where the inner one now is, with the same blast radius, since a throw here propagates out of `new Figure(...)` the same way.

## Verification

`npm run lint:fix`, `npm run type-check`, `npm run build` clean; **288 unit suites / 4310 tests** and **7 ESM suites / 73 tests** pass, including 10 new ones in `test/model/outOfRangeRow.test.ts` covering line, smooth, step, area, bar and heatmap, plus that the guard does not latch when the cursor comes back into range.

Falsification — all five guards reverted: **8 of 10 fail**. The 2 survivors are the in-range controls, which is what they are there to hold still.

---

### DIFF
--- a/src/model/bar.ts
+++ b/src/model/bar.ts
@@ -219,7 +219,7 @@ export abstract class AbstractBarPlot<T extends BarPoint> extends AbstractTrace 
   protected get dimension(): Dimension {
     return {
       rows: this.barValues.length,
-      cols: this.barValues[this.row].length,
+      cols: this.barValues[this.row]?.length ?? 0,
     };
   }
 
--- a/src/model/box.ts
+++ b/src/model/box.ts
@@ -272,7 +272,7 @@ export class BoxTrace extends AbstractTrace {
     // navigation col), so rows/cols map directly to its shape.
     return {
       rows: this.boxValues.length,
-      cols: this.boxValues[this.row].length,
+      cols: this.boxValues[this.row]?.length ?? 0,
     };
   }
 
--- a/src/model/heatmap.ts
+++ b/src/model/heatmap.ts
@@ -129,7 +129,7 @@ export class Heatmap extends AbstractTrace {
   protected get dimension(): Dimension {
     return {
       rows: this.heatmapValues.length,
-      cols: this.heatmapValues[this.row].length,
+      cols: this.heatmapValues[this.row]?.length ?? 0,
     };
   }
 
--- a/src/model/line.ts
+++ b/src/model/line.ts
@@ -473,7 +473,7 @@ export class LineTrace extends AbstractTrace {
   protected get dimension(): Dimension {
     return {
       rows: this.lineValues.length,
-      cols: this.lineValues[this.row].length,
+      cols: this.lineValues[this.row]?.length ?? 0,
     };
   }
 
--- a/src/model/pie.ts
+++ b/src/model/pie.ts
@@ -318,7 +318,7 @@ export class PieTrace extends AbstractTrace {
   protected get dimension(): Dimension {
     return {
       rows: 1,
-      cols: this.sliceValues[this.row].length,
+      cols: this.sliceValues[this.row]?.length ?? 0,
     };
   }
 
--- a/test/model/outOfRangeRow.test.ts
+++ b/test/model/outOfRangeRow.test.ts
@@ -83,6 +83,49 @@ describe('a cursor parked past the end of the data', () => {
     expect(() => trace.state).not.toThrow();
     expect(trace.state.empty).toBe(true);
   });
+
+  // `box` and `pie` are patched by the same change but do not fit the table
+  // above: a box point is a five-number summary rather than an {x, y} pair,
+  // and a pie is a single row of slices. Given separately rather than left
+  // untested, since both are in the diff.
+
+  test('a box trace reports empty rather than throwing', () => {
+    // Its own index, deliberately: `boxValues` is section-major, so a single
+    // box is seven rows (the five-number summary plus the outlier groups) and
+    // `PAST_THE_END` would still be inside it. Picking a row that is only
+    // past the end of *some* traces is how a guard gets a green test it never
+    // exercised.
+    const trace = TraceFactory.create(
+      layer(TraceType.BOX, [
+        {
+          z: 'a',
+          lowerOutliers: [],
+          min: 1,
+          q1: 2,
+          q2: 3,
+          q3: 4,
+          max: 5,
+          upperOutliers: [],
+        },
+      ]),
+    );
+    trace.row = 99;
+
+    expect(() => trace.state).not.toThrow();
+    expect(trace.state.empty).toBe(true);
+  });
+
+  test('a pie trace reports empty rather than throwing', () => {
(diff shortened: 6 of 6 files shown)


jooyoungseo:
Blocking finding actioned in `a53b2996`, and it turned up something worth recording.

**`box` and `pie` now have cases.** You were right that shipping two of the five patched files without a failing-before test breaks the repo's own rule.

Adding them caught a second-order mistake in my own test design: the box case **passed with the guard reverted**. `boxValues` is section-major, so a single box is *seven* rows — the five-number summary plus the outlier groups — and `PAST_THE_END = 5`, which is past the end of every other trace here, is still comfortably inside it. The test was green without ever exercising the thing it named.

```
box, row 0 → dimension {rows: 7, cols: 1}
box, row 5 → dimension {rows: 7, cols: 1}   ← in range
```

So the box case uses its own index, with the reason written down, because "a row past the end of *some* traces" is exactly how a guard gets a test that never runs it.

Falsified against those two guards alone: **both new cases fail** without them. Full suite now **288 suites / 4312 tests** plus 73 ESM.

**On `search_in_row` / `search_in_col`** — I traced reachability rather than filing it speculatively, and it does not need one. `moveToNextCompareValue` has a single caller, `RotorService` (`src/service/rotor.ts:135`), which acts on the *user's* cursor; that stays in bounds via `MovableGrid`. And `getStateAt` has no caller outside the traces' own overrides — nothing in `src/service/` or `src/state/` reaches it — so monitor mode cannot park the cursor out of range and then enter the rotor. It is the same latent shape, but there is no path to it today. Worth revisiting only if `getStateAt` ever gains an external caller; I would rather leave that unfiled than add an issue nobody can reproduce.

**On the E2E check** — it came back `cancelled` rather than failed on the previous run (superseded by this push), and this push re-runs it.

---

Issue #917, opened 2026-08-15
docs: unresolved merge-conflict markers committed in BRAILLE.md and SCHEMA.md


jooyoungseo:
Found while fixing #913 (heading ids for in-page anchors). Reporting rather than fixing, because resolving these is a content decision rather than a mechanical one.

Three unresolved conflict regions are committed on `main`:

```
docs/BRAILLE.md:838  <<<<<<< HEAD
docs/BRAILLE.md:870  =======
docs/BRAILLE.md:893  >>>>>>> eae4da3f (feat(network): reach every node, which following links cannot)
docs/BRAILLE.md:898  <<<<<<< HEAD
docs/BRAILLE.md:902  =======
docs/BRAILLE.md:906  >>>>>>> eae4da3f (feat(network): reach every node, which following links cannot)
docs/SCHEMA.md:124   <<<<<<< HEAD
docs/SCHEMA.md:126   =======
docs/SCHEMA.md:128   >>>>>>> 97ef5574 (fix(choropleth): let the border walk be entered leftwards)
```

They came in with 8de03546a (`feat(choropleth): read a map as a map, not as a list of places`, #850).

## Why this is worse than three stray lines

The `=======` marker in `docs/BRAILLE.md` lands **inside a fenced code block**, so it flips the fence parity for the rest of the file. From that point on the file has an odd number of ` ``` ` fences and everything after line 935 is swallowed into an unterminated code block — on GitHub and on maidr.ai alike.

That silently loses the last ~90 lines of the document, including two headings that are otherwise fine:

- `## Volcano and Manhattan` (line 956)
- `## Multiline Braille Display Support` (line 979)

`marked`'s lexer confirms it: the last heading it sees in the file is `Sankey, alluvial and chord`, and nothing after it is parsed as markdown at all.

The knock-on effect is that `docs/BRAILLE.md:14` —

```markdown
For multiline braille display setup, see [Multiline Braille Display Support](#multiline-braille-display-support).
```

— cannot resolve, because its target heading no longer exists as a heading. This is the one in-page anchor that the fix for #913 could not repair, and it is pinned there as a failing test (`test/scripts/siteAnchors.esm-test.ts`) referencing this issue.

## The content question

The two sides are not competing edits to one passage — they look like two sections that should both survive:

- `docs/BRAILLE.md` HEAD side describes `## Sankey, alluvial and chord`; the other side describes `## Network`. Both read as complete, and each has its own `### Multiline Displays` subsection in the second conflict region.
[...]

PR #914, opened 2026-08-15
fix(docs): generate the examples gallery so it cannot miss a page

jooyoungseo:
# Pull Request

## Description

The gallery on `examples.html` was a hand-written list of `loadHTML(...)` calls
inside `scripts/build-site.js`. It named **88** pages. `examples/` holds **199**
at the top level. Nothing failed when the two drifted apart, so 127 examples —
`area`, `boxen`, `bump`, `choropleth`, `contour`, `dot`, `dumbbell`, `errorbar`,
`forest`, `funnel`, `gantt`, `gauge`, `hexbin`, and the whole amCharts,
Highcharts, Vega-Lite and Google Charts families — shipped inside the site and
were reachable from nothing on it. A page nobody can open documents nothing.

The list is now read off the directory, so it cannot drift again, and a test
fails if it ever does.

### Before / after

| | links in `_site/examples.html` |
|---|---|
| before | **92** — 88 `loadHTML` calls plus 4 hand-written entries |
| after | **244** — 241 `loadHTML` calls plus 3 hand-written entries |

241 = 244 pages found under `examples/` (199 top level, 21 `anychart/`,
21 `chartjs/`, 3 Vite entry points) − 3 deny-listed. Verified by building the
site with `npm run build:preview` and counting the generated page; every
`loadHTML` target resolves to a real file under `_site/examples/`.

## Related Issues

Closes #911 — the first of the two gaps in that issue (the gallery). The four
missing hand-authored trace-type examples are not part of this change.

## Changes Made

**`scripts/examplesGallery.js`** (new) — enumerates `examples/` two levels deep
and turns it into gallery sections:

- **Grouping** is the one the hand-written list already had, by integration,
  because that is how someone arrives at the gallery: they know which charting
  library they use, not what MAIDR calls the chart type. Membership is derived
  from the filename prefix — `plotly-`, `highcharts-`, `vegalite-`, `amcharts-`,
  `anychart-`, `frappe-`, `google-charts-`, `d3-bind`, `observable-` — plus the
  `anychart/` and `chartjs/` subdirectories, with everything else falling into
  the hand-authored **HTML / Vanilla JS** group.
- **Labels** come from a shared chart vocabulary (`bar` → "Bar Chart",
  `scatter` → "Scatter Plot"), a curated per-page map, or a de-slugified
  filename, in that order. The curated map carries **every** title the old list
  had, so generating the gallery did not cost the wording someone chose:
[...]

PR #918, opened 2026-08-15
fix(docs): give built pages heading ids so in-page links jump

### LINKED ISSUE #913: a11y: every in-page anchor link on maidr.ai is dead — marked v15 emits no heading ids
Reported: clicking a table-of-contents link on maidr.ai does not jump to the section. Reproduced, and the cause is one line of configuration.

## Root cause

`scripts/build-site.js` renders every page with `marked`:

```js
marked.setOptions({
  gfm: true,
  breaks: false,
});
```

**`marked` v15 does not emit heading `id` attributes.** The `headerIds` option was removed from marked core and moved to the `marked-gfm-heading-id` extension, which is not installed here. So a TOC link and its target render like this:

```
$ node -e "…marked.parse('- [Declaring What a Chart Means](#declaring-what-a-chart-means)\n\n## Declaring What a Chart Means\n')"

<li><a href="#declaring-what-a-chart-means">Declaring What a Chart Means</a></li>
<h2>Declaring What a Chart Means</h2>          ← no id
```

The link resolves to nothing, so the browser does not move. Nothing is logged and nothing looks broken, which is why this survived.

## Scope

**50 in-page anchor links** across the README and the docs that get built into maidr.ai:

| source | links |
|---|---|
| `README.md` → `index.html` (the home page) | 12 |
| `docs/amcharts.md` | 6 |
| `docs/observable.md` | 5 |
| `docs/vegalite.md`, `docs/frappe.md` | 2 each |
| `docs/anychart.md`, `chartjs.md`, `highcharts.md`, `plotly.md`, `recharts.md`, `victory.md` | 1 each |
| `docs/VIOLIN_PLOT_SPEC.md` | 16 |
| `docs/BRAILLE.md` | 1 |

Every one of them is dead on the built site.

## Why this matters more here than on most sites

This is an accessibility project, and in-page anchors are disproportionately load-bearing for the people it serves. A sighted user who clicks a dead TOC link sees the page not move and scrolls manually. A screen-reader user gets no announcement, no focus move, and no signal that anything failed — the reading position simply stays put. Skipping to a section is the whole reason the TOC exists.

## The fix

Give headings ids whose slugs **match what the markdown already links to**. The docs are read on GitHub as well as on maidr.ai, so the slug algorithm has to be GitHub-compatible or the same TOC would work in one place and not the other: lower-case, spaces to hyphens, punctuation dropped, and a `-1`/`-2` suffix for duplicates in document order.

Either `marked-gfm-heading-id` or a small custom renderer is fine. A custom renderer avoids a dependency and makes the slug rule explicit and testable, which is worth something given the GitHub-compatibility requirement — but either is acceptable if the slugs are right.

## Also worth fixing while in here

`docs/template.html` has a `<main>` element but **no skip-to-content link**. That is the other half of what a keyboard user needs on a page whose nav precedes the content, and it is a conspicuous omission on this project's own site. Small and adjacent; include it if it stays small.

## Acceptance criteria

- [ ] Every heading in a built page carries an `id` matching the GitHub slug of its text.
- [ ] Duplicate heading texts get GitHub's `-1`, `-2` disambiguation in document order.
- [ ] **A test asserts that every in-page anchor link in every built page resolves to an element with that id.** This is what stops it regressing — the bug was invisible precisely because nothing checked it.
- [ ] Verified in a real browser: click a TOC link on the built `index.html` and on `amcharts.html` and confirm the viewport moves and focus lands on the heading. A Chromium binary is at `/opt/pw-browsers/chromium-1194/chrome-linux/chrome`.
- [ ] `lint:fix`, `type-check`, `build` and `test` pass.

Related: #911 also changes `scripts/build-site.js` (generating the examples gallery), so the two will need sequencing.

jooyoungseo:
# Pull Request

## Description

Every in-page anchor link on maidr.ai was dead.

`marked` v15 emits no heading `id` attributes — the `headerIds` option left marked core for the `marked-gfm-heading-id` extension, which this repository does not depend on. `scripts/build-site.js` crossed that boundary and nothing failed, because an unresolvable fragment is not an error: the browser simply does not move. Reproduced before changing anything:

```console
$ node -e "…marked.parse('- [Declaring What a Chart Means](#declaring-what-a-chart-means)\n\n## Declaring What a Chart Means\n')"
<li><a href="#declaring-what-a-chart-means">Declaring What a Chart Means</a></li>
<h2>Declaring What a Chart Means</h2>          ← no id
```

That is worse here than on most sites. A sighted user who clicks a dead table-of-contents link sees the page not move and scrolls manually. A screen-reader user gets no announcement, no focus move, and no signal that anything failed — the reading position simply stays put, and skipping to a section is the whole reason the table of contents exists.

## Related Issues

Closes #913.

Surfaced #917 — see "Additional Notes".

### The slug rule, and why it is GitHub's

New `scripts/markdown.js` renders the site's markdown and gives every heading its GitHub slug as an `id`.

`marked-gfm-heading-id` is not installed and could not be added, so this is a small explicit renderer — which the issue notes is the better answer anyway, since it makes the rule testable, and the GitHub-compatibility requirement is exactly the kind of thing worth pinning.

The rule has to be GitHub's rather than a tidier one of our own: these files are read on github.com *and* on maidr.ai, and they carry one set of `[…](#…)` links for both. Anything else would work in one place and not the other.

1. Lower-case.
2. Drop anything that is not a letter, a number, a combining mark, a space, `_`, or an ASCII hyphen.
3. Turn **each** remaining space into a hyphen — each, not each run.
4. Number repeats in document order, and keep searching when the numbered form is itself taken.

Three details are load-bearing, and each is the kind of thing a naive slugifier gets wrong. All three are pinned by tests against real headings in this repository:

| Heading | Slug | Why it is easy to get wrong |
|---|---|---|
| `## Live & Streaming Data` | `live--streaming-data` | Dropping `&` leaves two spaces, and **both** become hyphens. Collapsing runs is the most tempting simplification here and it breaks the README's link to this heading. |
| ``## 4. KDE Layer — `violin_kde` `` | `4-kde-layer--violin_kde` | The em dash and the ASCII hyphen are both `\p{Pd}`. Slugging by that category drops the wrong one and gives a single hyphen. `docs/` has nine em-dashed headings. |
| ``### `embed(target, spec, options?)` `` | `embedtarget-spec-options` | Characters are **deleted**, not replaced — the bracket closes the gap while the comma leaves its following space behind. Reads like a typo; is what GitHub does. |

Code spans, links and `&` are slugged from the heading's rendered *text* rather than its raw markdown, so backticks and URLs never reach the slug and `&` arrives as itself rather than as the escaped `&amp;` (which would give `live-amp-streaming-data`).

The rule was verified character for character against `github-slugger` — the library GitHub's own pipeline uses — across **all 579 headings** in `README.md` and `docs/`, plus adversarial cases (emoji, CJK, accented letters, `C++ / C#`, `100%`, padding). Zero mismatches, including the duplicate-numbering path where a literal `Bar Chart 1` heading forces `bar-chart-1-1`. `github-slugger` is only present transitively, so it is **not** used by the shipped code or the tests — it was a one-off oracle.

The module is separate from `build-site.js` because that script writes files and shells out to `npm run build:react-example` on import, so nothing could load it to ask what a heading becomes. `scripts/markdown.d.ts` sits beside it, the same arrangement as `scripts/testArgs.d.ts`.

### The regression test

`test/scripts/siteAnchors.esm-test.ts` is the criterion that stops this coming back — the bug survived precisely because nothing checked it. For every page the site builds, it collects every `href="#…"` and asks the DOM whether an element with that id exists. It catches a broken slug rule, a renamed heading, **and** a link written against a slug GitHub would never generate; a unit test of `slugify` alone would miss the last two.

It lives in `test/scripts/` per the convention that directory documents, and in the `esm` project because the module under test is ESM and the CommonJS project cannot require it — the same split, and the same reasoning, as `testArgs.esm-test.ts`. It reuses the real template so the skip link is checked too, and does not shell out to the full build.

### Skip-to-content link

`docs/template.html` had a `<main>` but no skip link. Added: first focusable element in the body, offscreen until focused, targeting `<main id="main-content" tabindex="-1">`. Two CSS rules and one line of markup.

### Two links that were wrong on GitHub too

`docs/vegalite.md` pointed at `#bindvegalite-view-spec-options` and `#embed-target-spec-options`. GitHub deletes `(` without closing the gap, so the real slugs are `#bindvegaliteview-spec-options` and `#embedtarget-spec-options`. Those two links were dead on github.com as well, and no heading-id change could have fixed them.

Not a visual change; the browser verification below is the equivalent evidence.

## Checklist

### Verified in a real browser

The whole bug is "the page does not move", which no unit test fully captures. Built the site and drove it in Chromium (`/opt/pw-browsers/chromium-1194/chrome-linux/chrome`) via Playwright:

| Page | TOC link clicked | `scrollY` | Result |
|---|---|---|---|
| `index.html` | `#braille-generation` | 0 → 4164 | `<h2>` "Braille Generation" at top of viewport |
| `index.html` | `#live--streaming-data` | 0 → 3932 | `<h2>` "Live & Streaming Data" at top of viewport |
| `amcharts.html` | `#declaring-what-a-chart-means` | 0 → 6581 | `<h2>` "Declaring What a Chart Means" at top of viewport |
| `amcharts.html` | `#multi-panel-charts` | 0 → 9826 | `<h2>` "Multi-Panel Charts" at top of viewport |

In every case the viewport actually moved, `location.hash` updated, and the target heading landed at `top: 0px`. Pressing <kbd>Tab</kbd> immediately after each click moved focus to an element *below the heading* (y = 127, 152, 502, 283) rather than back to the top of the document — so the browser set the sequential focus navigation starting point at the heading, which is the correct behaviour for a non-focusable fragment target and the thing a screen reader follows.

Skip link: from a fresh load, the first <kbd>Tab</kbd> focuses `<a class="skip-link">` and it becomes visible (`left: 0`); <kbd>Enter</kbd> sets `#main-content` and moves focus to `<main id="main-content">`.

Full sweep in the browser across all 20 built pages: **70 in-page anchors, 1 dead** — the pinned `docs/BRAILLE.md` case below. Before this change, 50 of them were dead.

### One anchor is left dead, deliberately

`docs/BRAILLE.md` carries **unresolved merge-conflict markers committed on `main`** (filed separately as #917, along with a third in `docs/SCHEMA.md`). One `=======` marker lands inside a fenced code block, flipping the fence parity so the last ~90 lines of the file are swallowed into an unterminated code block. `marked`'s lexer stops seeing headings after `Sankey, alluvial and chord`, which means the `## Multiline Braille Display Support` heading that line 14 links to no longer exists as a heading at all.

No slug rule can fix that, and resolving it is a content decision — the two sides look like two sections that should both survive — so it belongs to a different change. Per the repo's testing rule it is pinned with `it.failing` and the issue number rather than skipped: the case keeps running, the suite stays green while the bug stands, and the day #917 lands it turns red as the reminder to remove the pin.

### Checks

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass — `unit` 4288/4288, `esm` 112/112.

Each new test was verified to fail without the fix: reverting the renderer hunk fails 17 of them. The slug rule was additionally mutation-tested — collapsing space runs fails 6, slugging by `\p{Pd}` fails 3, dropping duplicate numbering fails 2 — and the skip-link cases fail when the template hunk is reverted. Three tests pass either way by design and are disclosed as such: the hand-written-declarations guard (the convention `testArgs.esm-test.ts` establishes), the guard asserting the page list has not silently emptied, and the per-page cases for the six docs that currently contain no table-of-contents anchors — those are generated by `it.each` over the real `docs/` listing so a newly added anchor is covered without hand-maintaining a list.

### Note on #911

#911 also touches `scripts/build-site.js`, replacing the hardcoded examples gallery with a generated one. This change is deliberately confined to the `marked` configuration and the markdown-to-HTML path (the import, `markdownToHtml`, and the 13 per-page `marked.parse(…)` → `renderMarkdown(…)` call sites). It does not touch the examples-gallery section at all, so the two should merge cleanly.

Squash title is `fix(...)` on purpose: this is a user-facing defect on the published site, so it should cut a patch release rather than the no-op a `docs(...)` title would give.

---

### DIFF
--- a/docs/template.html
+++ b/docs/template.html
@@ -271,10 +271,31 @@
       background: rgba(255, 255, 255, 0.2);
     }
 
+    /* Skip link: first thing in the tab order, offscreen until focused. */
+    .skip-link {
+      position: absolute;
+      left: -10000px;
+      top: 0;
+      z-index: 1100;
+      padding: 0.75rem 1rem;
+      background: white;
+      color: #1a1a1a;
+      font-weight: 600;
+    }
+
+    .skip-link:focus {
+      left: 0;
+    }
+
     main {
       max-width: 1200px;
       margin: 0 auto;
       padding: 2rem;
+    }
+
+    /* Focused only to move the reading position from the skip link. */
+    main:focus {
+      outline: none;
     }
 
     /* Markdown content styles */
@@ -407,6 +428,7 @@
 </head>
 
 <body>
+  <a class="skip-link" href="#main-content">Skip to content</a>
   <nav>
     <div class="container">
       <a href="{{BASE_PATH}}index.html" class="logo">
@@ -456,7 +478,7 @@
     </div>
   </nav>
 
-  <main>
+  <main id="main-content" tabindex="-1">
     {{CONTENT}}
   </main>
 
--- a/docs/vegalite.md
+++ b/docs/vegalite.md
@@ -53,7 +53,7 @@ Load Vega, Vega-Lite, vega-embed, MAIDR core, and the Vega-Lite adapter, then ca
 
 `embed()` returns a `Promise<{ view }>` — `await` it (or use `.then()` / `.catch()`) when you need the underlying Vega view or want to handle render failures.
 
-If you already drive `vegaEmbed` yourself (e.g. you need the `view` for additional logic before MAIDR mounts), use [`bindVegaLite`](#bindvegalite-view-spec-options) instead. **You must `await result.view.runAsync()`** between embedding and binding so the SVG is in the DOM by the time MAIDR mounts:
+If you already drive `vegaEmbed` yourself (e.g. you need the `view` for additional logic before MAIDR mounts), use [`bindVegaLite`](#bindvegaliteview-spec-options) instead. **You must `await result.view.runAsync()`** between embedding and binding so the SVG is in the DOM by the time MAIDR mounts:
 
 ```js
 // renderer must be 'svg' — MAIDR navigates the SVG output, not canvas.
@@ -574,7 +574,7 @@ Lower-level entry point that mounts MAIDR on the SVG produced by an existing `ve
 | `options.id` | `string` (optional) | Unique ID for the MAIDR instance. Defaults to the view container's `id`, or a timestamp fallback. |
 | `options.title` | `string` (optional) | Override for the chart title used in announcements. |
 
-**Callers must `await view.runAsync()` before calling `bindVegaLite()`.** `vegaEmbed(...)` resolves when the view is *constructed*, not when it has *rendered* — calling `bindVegaLite()` immediately after `vegaEmbed()` resolves can race the first paint and produce a "No SVG found" error on slow or aggregated specs (histograms, complex transforms). The view must also be created with `renderer: 'svg'`; MAIDR cannot navigate canvas output. **Prefer [`embed()`](#embed-target-spec-options)** which performs both steps for you.
+**Callers must `await view.runAsync()` before calling `bindVegaLite()`.** `vegaEmbed(...)` resolves when the view is *constructed*, not when it has *rendered* — calling `bindVegaLite()` immediately after `vegaEmbed()` resolves can race the first paint and produce a "No SVG found" error on slow or aggregated specs (histograms, complex transforms). The view must also be created with `renderer: 'svg'`; MAIDR cannot navigate canvas output. **Prefer [`embed()`](#embedtarget-spec-options)** which performs both steps for you.
 
 ### `vegaLiteToMaidr(spec, view?, options?)`
 
--- a/scripts/build-site.js
+++ b/scripts/build-site.js
@@ -12,8 +12,8 @@ import { execSync } from 'node:child_process';
 import fs from 'node:fs';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { marked } from 'marked';
 import { buildGallery, listExamplePages, renderGallery } from './examplesGallery.js';
+import { renderMarkdown } from './markdown.js';
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
@@ -30,21 +30,18 @@ if (!fs.existsSync(SITE_DIR)) {
 // Read template
 const template = fs.readFileSync(TEMPLATE_PATH, 'utf-8');
 
-// Configure marked for safe HTML output
-marked.setOptions({
-  gfm: true,
-  breaks: false,
-});
-
 /**
- * Convert markdown to HTML using marked library
+ * Convert markdown to HTML for the site.
+ *
+ * `renderMarkdown` carries the marked configuration and the heading ids the
+ * table-of-contents links need; see `scripts/markdown.js` for why the ids are
+ * not marked's own.
  */
 function markdownToHtml(md) {
   // Remove the centered logo div at the top of README
   const content = md.replace(/<div align="center">[\s\S]*?<\/div>\s*/, '');
 
-  // Convert markdown to HTML
-  return marked.parse(content);
+  return renderMarkdown(content);
 }
 
 // Per-page SEO descriptions.
@@ -191,7 +188,7 @@ if (fs.existsSync(reactMdPath)) {
   const reactMd = fs.readFileSync(reactMdPath, 'utf-8');
   const reactHtml = `
 <div class="content">
-  ${marked.parse(reactMd)}
+  ${renderMarkdown(reactMd)}
 </div>
 `;
   const reactPage = generatePage({ title: 'React', content: reactHtml, activePage: 'react', slug: 'react.html', ogType: 'article' });
@@ -205,7 +202,7 @@ if (fs.existsSync(rechartsMdPath)) {
   const rechartsMd = fs.readFileSync(rechartsMdPath, 'utf-8');
   const rechartsHtml = `
 <div class="content">
-  ${marked.parse(rechartsMd)}
+  ${renderMarkdown(rechartsMd)}
 </div>
 `;
   const rechartsPage = generatePage({ title: 'Recharts', content: rechartsHtml, activePage: 'recharts', slug: 'recharts.html', ogType: 'article' });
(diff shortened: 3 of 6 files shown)


jooyoungseo:
Thanks for the review. Both observations were non-blocking; I checked each and am leaving the code as-is. No new commit.

**1. Empty-slug collision — the premise does not hold.** The suggestion was that two emoji-only headings give `""` then `"-1"`, where GitHub might number differently. It does not: `github-slugger`'s `slug()` tracks occurrences in a `Map` and tests with `.has()`, so a zero count still counts as taken, and the empty base numbers exactly like any other. Three `🎉` headings, ours vs. the library:

| # | github-slugger | `createHeadingSlugger` |
|---|---|---|
| 1 | `""` | `""` |
| 2 | `"-1"` | `"-1"` |
| 3 | `"-2"` | `"-2"` |

I also ran a differential fuzz rather than reasoning about it — 10,399 headings across 3,000 synthetic documents, drawn from an alphabet seeded with the characters most likely to split the two implementations (em/en dash, non-breaking hyphen, combining marks, CJK, emoji, brackets, `&`, runs of spaces, tabs): **zero mismatches**, plus zero on the emoji, empty-string and duplicate-numbering cases specifically. Note the empty case never reaches the HTML anyway — `renderMarkdown` omits the attribute rather than emitting `id=""`.

**2. `main:focus { outline: none }` — deliberate, and the alternative is worse.** `<main tabindex="-1">` is not in the tab order; it takes focus only when the skip link is activated. WCAG 2.4.7 governs keyboard-operable components, which a programmatic fragment target is not, so suppressing the ring is not a failure — it is the standard skip-link-target pattern. Following the `:focus` / `:focus-visible` pair used by the nav items above it would actively regress this: browsers do match `:focus-visible` after a keyboard-activated skip link, which would draw a ring around a container several viewport-heights tall, most of it offscreen. The perceivable confirmation is the one that already exists and is verified in the PR body — the viewport scrolls to the content and `location.hash` updates.

Two things outside this PR, for whoever sequences the merges: the `docs/BRAILLE.md` conflict markers stay untouched (#917 owns that; the one resulting dead anchor is pinned with `it.failing` and turns red when #917 lands), and the `scripts/build-site.js` diff is confined to the `marked` import, its config, and the 13 `parse` call sites — it touches no line of the examples gallery, so it should not contend with #914.

---

PR #921, opened 2026-08-15
fix(docs): resolve the committed conflict markers and guard against more

### LINKED ISSUE #917: docs: unresolved merge-conflict markers committed in BRAILLE.md and SCHEMA.md

jooyoungseo:
# Pull Request

## Description

Resolves the two merge-conflict regions committed in `docs/BRAILLE.md`, and adds a check that stops the next one reaching `main`.

The markers were not three stray lines. The `=======` at line 870 landed **between the fences of a code block**, which flipped the fence parity for the rest of the file — so everything after it was swallowed into an unterminated block, on GitHub and on maidr.ai alike. `## Volcano and Manhattan` and `## Multiline Braille Display Support` stopped existing as headings, and `docs/BRAILLE.md:14`'s link to `#multiline-braille-display-support` pointed at nothing.

`docs/SCHEMA.md`, the third region in the report, was already resolved by #912.

## Related Issues

Closes #917

### The two regions, and what each was checked against

The sides were not competing edits to one passage, so neither was discarded. **Both sections are kept**, each with its own code block and its own `### Multiline Displays` note — which is what the second region turned out to be: one `### Multiline Displays` heading shared between two sections that each needed their own.

| Region | HEAD side | Other side | Resolution |
| --- | --- | --- | --- |
| 838–893 | `## Sankey, alluvial and chord` — prose plus a `Stage / Cells` table whose opening fence is inside the region | `## Network` — prose plus a `Group / Cells` table, sharing HEAD's closing fence | Both sections kept, in that order, each with a complete fenced block of its own |
| 898–906 | The multiline note for sankey ("one line per stage") | The multiline note for network ("one line per group") | Both kept, each under the `### Multiline Displays` of its own section |

Each description was read against the model rather than picked by position:

- **Sankey** — `src/model/flow.ts`. `braille` returns `values` grouped by stage with `min: this.stageMin` / `max: this.stageMax`, and swaps the cursor out as `row: this.col, col: this.row`. That is exactly the "one row per stage, each scaled against its own stage, transposed on its way out" the HEAD prose describes. The arrow geometry it claims (left/right along a ribbon, up/down within a column) is what `buildGraph` wires, and the rotor units it mentions are `OUT_ROTOR_UNIT` / `IN_ROTOR_UNIT`.
- **Network** — `src/model/network.ts`. `braille` returns one row per connected component with `min`/`max` set to the chart-wide `minDegree`/`maxDegree` on *every* row — the whole-chart scaling the prose calls the deliberate opposite of the treemap's, and the reason it gives (a hub of three should not feel like a hub of thirty) matches the comment on `audio`. `findComponents` sorts components largest first and members by degree, so "ordered most connected first" holds; no layout coordinate is read anywhere, so "where a node is drawn is not encoded" holds too.

### Heading counts, before and after

Measured with `marked.lexer` on the file itself:

| | headings in file | headings parsed | last heading parsed |
| --- | --- | --- | --- |
| before | 73 | **64** | `Sankey, alluvial and chord` |
| after | 74 | **74** | `Setup` |

`Volcano and Manhattan` and `Multiline Braille Display Support` are both back. The file's heading count goes 73 → 74 because the resolution gives the network section the `### Multiline Displays` it had been sharing with the sankey one; the point is that parsed now equals present, where it was nine short.

Fences: 21 (odd) → 22 (even).

### The `it.failing` pin

`test/scripts/siteAnchors.esm-test.ts` excluded `docs/BRAILLE.md` from the page loop and pinned it with `it.failing` referencing this issue. Both are gone: the page is now checked by the same `it.each(pages)` case as every other page, asserting positively. `docs/BRAILLE.md` is named explicitly in the "should be checking the pages this bug was reported against" case so it cannot quietly drop out of the list again.

### The guard

`scripts/conflictMarkers.js`, wired to `npm run check:conflicts` and to a `No Conflict Markers` job in `ci.yml`.

- **Every tracked file, not just markdown.** `git ls-files` is the file list, so `.gitignore` keeps itself in step and nothing has to enumerate `node_modules`/`_site`/`dist`. A marker in a `.ts` file would at least fail the type check; in `.html` or `.json` it would sail through exactly as this one did. Binary files are skipped on a NUL byte rather than by an extension list.
- **All four markers**, including the diff3 base marker a contributor with `merge.conflictStyle = diff3` would produce. The separator is anchored at both ends, since a bare run of `=` is also a setext underline.
- **Its own job, gating nothing** — same reasoning as `commitlint` in this workflow (#763): a marker should fail its own check, not grey out the build and the tests. No `npm ci`; it is node and git only, so it costs seconds.
- **It does not match itself.** The check and its test build the marker strings with `repeat` instead of writing them down, so no run of seven ever hits disk. The proof is that the repository-wide case — which reads `scripts/conflictMarkers.js` and `test/scripts/conflictMarkers.esm-test.ts` along with everything else — comes back empty.

One real bug surfaced while proving that: `|` is regular-expression alternation, so the first version's base pattern was seven empty alternatives and reported **all 1120 tracked files as conflicted**. The characters are escaped into the pattern now, and the note is kept in the source.

### Failure evidence

Against `origin/main`'s `docs/BRAILLE.md`, the check finds every region the issue reported and nothing else:

```
origin/main docs/BRAILLE.md -> ["838:ours","870:separator","893:theirs","898:ours","902:separator","906:theirs"]
resolved   docs/BRAILLE.md -> []
```

And reintroduced live, in a scratch `.html` file staged into the index — the non-markdown case the guard exists for:

```
$ npm run check:conflicts
Merge-conflict markers in 1 tracked file(s):

  docs/scratch-guard-demo.html:2  separator  =======

Resolve the conflict and commit the resolved file.
exit=1
```

Scratch file removed, same command:

```
$ npm run check:conflicts
No merge-conflict markers in 1126 tracked files.
exit=0
```

## Checklist

## Additional Notes

The manual-testing box is left unticked deliberately: `ManualTestingProcess.md` walks a chart with a screen reader, and nothing here changes runtime behaviour — the change is prose, a test, a script and a CI job. The automated suites were all run and all passed:

| | result |
| --- | --- |
| `npm run lint:fix` | passes (four `node/prefer-global` errors on the new files were fixed) |
| `npm run type-check` | passes |
| `npm run build` | passes |
| `npm test` | passes — unit 4312/4312, esm 136/136 |
| `npm run sync:copilot:check` | in sync |

Nothing under `src/` is touched.

One judgement call worth flagging for review: the sections are ordered sankey-then-network, which keeps the HEAD section where it already sat and reads into `## Treemap` (the chart the network prose compares its scaling against). If the section owner would rather they ran the other way, it is a block move with no content change.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,37 @@ jobs:
       - name: Check generated instruction files
         run: npm run sync:copilot:check
 
+  # Three unresolved conflict regions sat on `main` for weeks and shipped to
+  # maidr.ai (#917). They passed commitlint, the linter, the type check, the
+  # build and the whole test suite, because none of those read prose — and one
+  # `=======` inside a fenced code block silently swallowed the last ninety
+  # lines of docs/BRAILLE.md.
+  #
+  # Every tracked file rather than markdown: the same mistake in a .ts file
+  # would at least fail the type check, but in .html or .json it would sail
+  # through exactly as this one did.
+  #
+  # Pure Node and git with no dependencies, so it skips npm ci, and it gates
+  # nothing for the same reason commitlint below does not.
+  conflict-markers:
+    name: No Conflict Markers
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 'lts/*'
+
+      - name: Check tracked files for conflict markers
+        run: npm run check:conflicts
+
   # Gates nothing, and must not: `needs` does not just order jobs, it *skips*
   # them when the dependency fails, so a commit subject that does not parse
   # used to take the build, the type check, the whole unit suite and every e2e
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -835,7 +835,6 @@ Contours use one line per level on a multiline display, which is the case where
 braille gives back what the announcement carries: two fingers on adjacent rows
 feel exactly where the curves crowd together and where they open out.
 
-<<<<<<< HEAD
 ## Sankey, alluvial and chord
 
 One row per **stage** -- one column of the drawing -- with one cell per node in
@@ -867,7 +866,14 @@ Stage   Cells
 0       ⣿⠶
 1       ⠂⣿⡶
 2       ⣿⡖
-=======
+```
+
+### Multiline Displays
+
+One line per stage, so a hand across the display holds the whole chart and the
+narrowing from source to sink is felt directly -- the one thing here braille
+conveys better than it conveys the ribbons.
+
 ## Network
 
 One row per **connected group**, one cell per node in it, ordered most
@@ -890,20 +896,13 @@ data, so there is nothing there to put in a cell.
 Group   Cells
 1       ⣿⡶⡶⠆⠆
 2       ⡶⡶⡶
->>>>>>> eae4da3f (feat(network): reach every node, which following links cannot)
 ```
 
 ### Multiline Displays
 
-<<<<<<< HEAD
-One line per stage, so a hand across the display holds the whole chart and the
-narrowing from source to sink is felt directly -- the one thing here braille
-conveys better than it conveys the ribbons.
-=======
 One line per group, so the split into groups -- the structure a reader
 following links can never discover, because links do not cross between them --
 is felt directly as separate rows of different lengths.
->>>>>>> eae4da3f (feat(network): reach every node, which following links cannot)
 
 ## Treemap
 
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
     "lint": "eslint .",
     "lint:fix": "eslint . --fix",
     "check-models": "node scripts/check-model-catalog.mjs",
+    "check:conflicts": "node scripts/conflictMarkers.js",
     "sync:copilot": "node scripts/sync-copilot-instructions.mjs",
     "sync:copilot:check": "node scripts/sync-copilot-instructions.mjs --check",
     "semantic-release": "semantic-release",
--- a/scripts/conflictMarkers.d.ts
+++ b/scripts/conflictMarkers.d.ts
@@ -1,0 +1,34 @@
+/**
+ * Hand-written declarations for `scripts/conflictMarkers.js`.
+ *
+ * The same arrangement as `scripts/markdown.d.ts`: the module is plain ESM run
+ * directly by node, and `tsconfig.json` sets `allowJs: false`, so a test can
+ * only import it with declarations beside it.
+ * `test/scripts/conflictMarkers.esm-test.ts` checks these against what the
+ * module actually returns, since `tsc` never sees the JavaScript.
+ */
+
+/** One marker line, as found in a file. */
+export interface ConflictMarker {
+  /** 1-based line number. */
+  line: number;
+  /** Which of the four markers this is. */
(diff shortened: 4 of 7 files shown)


jooyoungseo:
The review's one actionable note — that a bare marker with no trailing label would not match, since `ours`/`base`/`theirs` require the space — I checked rather than took on trust, and the pattern stays as it is.

Git writes a space and a label for all three under every conflict style (default, `diff3` and `zdiff3` all confirmed against a real merge). The separator is the only marker git writes bare, and that is precisely the one anchored at both ends. The tails already match git's own output exactly.

Loosening the tail to `(?: |$)` would cost more than it buys: a markdown table row of six empty cells, and a seven-deep blockquote, both begin matching. That is a live false positive in a docs-heavy repository, traded against a shape only non-git tools emit.

Re-ran the separator against setext underlines too, since a bare run of `=` is the one genuine ambiguity: runs of 6, 8 and 20 all pass clean, only an exactly-seven run matches — indistinguishable from a real separator by construction — and no tracked file carries a setext `=` heading. Repository scan stays clean over all 1126 tracked files, with the check and its own test among them.

No code changed as a result; all 15 checks are green at `3509fe9`. The section-order question in the PR body is still open for the docs owner.

---

Issue #438, opened 2025-09-16
feat: toggle settings for hover


ellvix:
# Feature Request

## Motivation

Users should be able to turn the hover feature on and off. 

## Proposed Solution

Add a settings field, letting users choose between off, hover, and click. Default / initial behavior is hover.

ellvix:
fixed with PR #446

PR #929, opened 2026-08-16
feat(scatter): let a point say which category it is in

### LINKED ISSUE #927: A scatter on a category axis cannot say which category a point is in
## What is missing

`ScatterPoint.x` is typed `number`:

```ts
export interface ScatterPoint {
  x: number;
  y: number;
  z?: number;
}
```

`BarPoint` is not — it takes `x: string | number` — so the categorical case is expressible for a bar layer and not for a point layer. There is no way for a producer to say "this observation is in category *a*".

## Why it matters

A categorical scatter is not a rare shape. It is what `seaborn.stripplot`, `seaborn.swarmplot` and `ggplot2::geom_jitter` draw, and what any `geom_point` on a discrete scale draws. In all of them one axis carries names and the other carries the measurement.

Today a producer has two options and both lose something:

- **emit the drawn coordinate** — which on a strip plot is the *jitter*, an offset the renderer chose at random so overlapping points stay separable. Measured on a three-category `sns.stripplot` against an axis labelled `g` whose ticks read `a`, `b`, `c`:

  ```json
  {"x": -0.03994198890364789, "y": 0.1257302210933933}
  {"x":  0.06297755824455262, "y": -0.1321048632913019}
  {"x": -0.07398610980531882, "y": 0.6404226504432821}
  ```

  A reader is given a precise number for a quantity that does not exist.

- **emit the tick position** — `0`, `1`, `2`. Honest about position and still silent about the name. This is what xability/py-maidr#439 does as the available half of the fix, and it is a real improvement (it also restores column grouping, see below), but a reader still hears "g is 0" where the chart says "a".

## The navigation half

`ScatterTrace` groups points into columns by exact `x` equality:

```ts
if (!currentX || currentX.x !== point.x) { … }
```

With jittered x every point gets its own column, so a 90-point three-category strip plot builds 90 columns of one point instead of 3 columns of 30. Column navigation then steps through individual observations and never through categories. Snapping to the tick fixes that much without any core change.

## Why a string cannot just be allowed today

`ScatterTrace` does arithmetic on `x`:

```ts
const sortedByX = [...data].sort((a, b) => a.x - b.x || a.y - b.y);
this.xIndexByValue = new Map(this.xValues.map((x, index) => [x, index]));
const distance = Math.hypot(center.x - _x, center.y - _y);
```

`'a' - 'b'` is `NaN`, so widening the type alone would give an unstable sort, a broken column index and a highlight that resolves to nothing. The position has to stay numeric internally.

## Possible shapes

Recorded as options rather than a recommendation — the choice is a grammar decision.

1. **A name alongside the position.** Keep `x: number` and add an optional label, so the trace sorts and pans on the number and announces the name. Smallest change, and it matches what the chart is: a category *at* a position.
2. **Category labels on `AxisConfig`.** A `categories?: string[]` indexed by tick position. Serves every trace whose x is a slot rather than a measurement, not just scatter.
3. **A `strip` trace type of its own.** Justified if the reading should differ from a scatter's — a strip plot is a distribution per category, so a reader may want to move category to category and hear the spread, rather than point to point. That is a different navigation model, not just a different label.

Option 1 is the least disruptive; option 3 is the one that would make the chart read as what it is.

## Related

- xability/py-maidr#439 — stops the jitter being announced, which is the part that needs no core change
- xability/py-maidr#251, xability/py-maidr#252 — the seaborn strip and swarm requests
- `BoxenPoint.z` and `BarPoint.x` — the two places the grammar already carries a category name

jooyoungseo:
Closes #927.

## What was missing

`ScatterPoint.x` was typed `number` while `BarPoint.x` is `string | number`, so the categorical case was expressible for a bar layer and not for a point layer. A producer had no way to say "this observation is in category *a*".

That is not a rare shape. It is what `seaborn.stripplot`, `seaborn.swarmplot` and `ggplot2::geom_jitter` draw, and what any `geom_point` on a discrete scale draws — one axis carries names, the other carries the measurement.

Both available answers lost something:

- **emit the drawn coordinate** — on a strip plot that is the *jitter*, an offset chosen at random so overlapping points stay separable. A reader is handed a precise number for a quantity that does not exist.
- **emit the tick position** — `0`, `1`, `2`. Honest about position, silent about the name. A reader hears "g is 0" where the chart says "a".

## The shape

Option 1 from the issue: the name travels **alongside** the position.

```json
{ "x": 0, "xLabel": "a", "y": 1.4 }
```

The coordinate has to stay numeric — `ScatterTrace` sorts on it (`a.x - b.x`), measures distance with it (`Math.hypot(center.x - _x, …)`), and resolves the column index stereo panning uses from it. `'a' - 'b'` is `NaN`, so widening the field alone would give an unstable sort, a broken column index and a highlight that resolves to nothing. Keeping it numeric and naming it separately is also what the chart *is*: a category **at** a slot.

**Both axes get one.** Either can be the categorical one — `stripplot(x='g', y='v')` puts the names on x, `stripplot(y='g', x='v')` puts them on y. Measured on py-maidr, which emits the tick position today:

```
vertical   (x=g, y=v)   axes={x: 'g', y: 'v'}   first point {x: 0.0, y: 0.1257}
horizontal (y=g, x=v)   axes={x: 'v', y: 'g'}   first point {x: 0.1257, y: 0.0}
```

A single un-suffixed `label` would leave a consumer guessing which coordinate it names. (`LinePoint.label` gets away with being un-suffixed because a line's x can simply *widen* to `string` — it never subtracts one. Only its y needs the alongside treatment.)

**One name per numeric slot.** Every point sharing an `x` must carry the same `xLabel`, since the slot *is* the category. A second, different name for the same `x` is dropped rather than splitting the category into two columns — recorded in the field's doc comment and in `SCHEMA.md`.

## Every mode, not just the default one

The trace announces a coordinate from five places, and a mode left behind would announce a slot index as though it were a measurement — reachable with one keystroke. All are named:

| mode | what it announces |
|---|---|
| COL | the column's category, and its y values |
| ROW | the row's category, and **every x sharing that y** |
| point | that point's x and y |
| intersection | the anchor, and the stack value the cursor is on |
| grid-cell point | as COL, within the cell |
| description table | the same names the cursor gave |

`TextState.main.value` and `.cross.value` widen to admit `string[]`, which ROW mode needs. `TextService` already casts arrays to `(number | string)[]` and joins them, so this widens the *type* to what the service has always accepted at runtime rather than adding a shape it has to learn.

## Tests

`test/model/scatterCategoryLabel.test.ts` — 15 tests across both orientations and all five modes, plus the description table.

The guards matter as much as the feature:

- a **continuous** scatter still announces its numbers, on both axes
- the **measurement** axis of a strip plot stays numeric — naming everything would be the same defect mirrored
- an **empty** `xLabel` counts as absent, so a producer emitting `''` gets the number rather than a blank where a value belongs
- **sorting and column grouping still key off the number** — labels arriving out of order must not disturb the ascending column order, and two points in one category must stay one column. This is the reason `x` stays numeric, asserted rather than assumed.
- a **partially labelled** axis resolves the unlabelled slots to their numbers rather than to blanks

### Falsification

| reverted | tests that fail |
|---|---|
| `named()` returns the value, ignoring the label | **8** |
| `namedAll()` returns the values, ignoring the labels | **3** |
| the label in `enterGridCell()`'s own grouping pass | **1** |
| nothing | **0 / 15 pass** |

Disjoint — the scalar sites, the array sites and the grid cell are separate failures, so none is carrying another. The grid-cell case is a distinct pass because `enterGridCell()` groups the cell's points itself rather than reusing `xPoints`; naming the other four sites genuinely does not name it.

## Verification

- `npm run lint:fix` — clean
- `npm run build` — 13/13 builds
- `npm test` — **4357 passed** (292 suites) and **136 passed** (10 ESM suites)

`docs/SCHEMA.md` gains a worked entry, including the warning a producer most needs: emit the tick position, not the drawn coordinate.

## Found during review, filed separately

A numeric `axes.x.format` applied to a category name announces **`$NaN`** — `formatters.currency('USD', 2)('a')` returns `"$NaN"`. Measured, and confirmed pre-existing rather than introduced here: the model emits the string intact (a `BarTrace` with `data: [{x: 'a', y: 1}]` and a currency x-format emits `{"label":"g","value":"a"}`), and the `NaN` appears in `TextService`. `BarPoint.x` has been `string | number` all along, so any bar layer already reaches it.

Filed as **#930** rather than folded in — different defect, different fix, and it affects bar today with no scatter involved.

## Follow-on

This is the grammar half. `xability/py-maidr#439` already snaps strip/swarm x to the tick — which fixes column grouping on its own, since `ScatterTrace` groups by exact `x` equality and jittered x gave every point its own column. Emitting `xLabel` alongside is the producer-side change this unblocks.

Option 3 from the issue — a `strip` trace with category-to-category navigation — stays open. It is a different navigation model rather than a different label, and it is worth deciding on its own rather than folding into this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,51 @@
+# [4.3.0](https://github.com/xability/maidr/compare/v4.2.0...v4.3.0) (2026-08-16)
+
+
+### Bug Fixes
+
+* **ci:** size the e2e report's code fence so its content cannot close it ([#924](https://github.com/xability/maidr/issues/924)) ([e4afdd9](https://github.com/xability/maidr/commit/e4afdd9c9ffe5f76e604dce8b4416b89cbc36c48)), closes [#715](https://github.com/xability/maidr/issues/715)
+* **docs:** generate the examples gallery so it cannot miss a page ([#914](https://github.com/xability/maidr/issues/914)) ([f5aecad](https://github.com/xability/maidr/commit/f5aecad934944c3f36413e23403b81493b3428f0)), closes [#911](https://github.com/xability/maidr/issues/911)
+* **docs:** give built pages heading ids so in-page links jump ([#918](https://github.com/xability/maidr/issues/918)) ([9ed5a39](https://github.com/xability/maidr/commit/9ed5a396e2e4668c3d00e6d91861995223d73607)), closes [#917](https://github.com/xability/maidr/issues/917) [#913](https://github.com/xability/maidr/issues/913)
+* **docs:** resolve the committed conflict markers and guard against more ([#921](https://github.com/xability/maidr/issues/921)) ([0c04513](https://github.com/xability/maidr/commit/0c04513a25cb96afbff3d15195cbb435716140b1)), closes [#918](https://github.com/xability/maidr/issues/918) [#917](https://github.com/xability/maidr/issues/917)
+* **highlight:** let a flow or network trace publish the ribbon it highlighted ([#906](https://github.com/xability/maidr/issues/906)) ([da131aa](https://github.com/xability/maidr/commit/da131aa76099a2730119e9bdf9c5391f99a72cd9)), closes [#895](https://github.com/xability/maidr/issues/895) [#903](https://github.com/xability/maidr/issues/903) [#899](https://github.com/xability/maidr/issues/899) [#904](https://github.com/xability/maidr/issues/904) [#903](https://github.com/xability/maidr/issues/903) [#904](https://github.com/xability/maidr/issues/904) [#904](https://github.com/xability/maidr/issues/904)
+* **line:** read every path command, so a staircase highlights its samples ([#908](https://github.com/xability/maidr/issues/908)) ([035c5bd](https://github.com/xability/maidr/commit/035c5bdd95d076385d7e26b815fb0e9503811005))
+* **plotly:** count a box's group across the whole box layer ([#883](https://github.com/xability/maidr/issues/883)) ([78375cb](https://github.com/xability/maidr/commit/78375cb4587b1468fb14d4dc5381b0529f172d3a)), closes [#882](https://github.com/xability/maidr/issues/882)
+* **plotly:** scope a candlestick's selector to its own panel and trace ([#882](https://github.com/xability/maidr/issues/882)) ([1cf0470](https://github.com/xability/maidr/commit/1cf0470be356e5d9adcb97d938d0795695367042)), closes [#881](https://github.com/xability/maidr/issues/881)
+* **trace:** answer empty rather than throwing when the cursor has no point ([#909](https://github.com/xability/maidr/issues/909)) ([893d6ea](https://github.com/xability/maidr/commit/893d6ead3d7b6078f4e0bdf798c0b1b537522a48))
+* **trace:** answer for a row that does not exist instead of throwing ([#916](https://github.com/xability/maidr/issues/916)) ([4b6f731](https://github.com/xability/maidr/commit/4b6f7315fca7f2de45d417c95fe4a2a1fe18f449)), closes [#905](https://github.com/xability/maidr/issues/905)
+
+
+### Features
+
+* **adapters:** add the co-located maidr trace declaration ([#886](https://github.com/xability/maidr/issues/886)) ([bbd62b6](https://github.com/xability/maidr/commit/bbd62b690c4e2bd46ce7ae8532303be557abe026))
+* **adapters:** close the declaration-foundation gaps the adapters reported ([#900](https://github.com/xability/maidr/issues/900)) ([ec47b0e](https://github.com/xability/maidr/commit/ec47b0eb12200b8d50e70d964e6aab31f57a93e9)), closes [#896](https://github.com/xability/maidr/issues/896) [#896](https://github.com/xability/maidr/issues/896)
+* **amcharts:** read flow, hierarchy, map and gauge charts ([#903](https://github.com/xability/maidr/issues/903)) ([7cf4c5c](https://github.com/xability/maidr/commit/7cf4c5ce3f64dcedbaf39553f213e29ee4f7385d))
+* **amcharts:** read the chart types deferred out of the coverage PR ([#895](https://github.com/xability/maidr/issues/895)) ([5a39999](https://github.com/xability/maidr/commit/5a39999581a6f033bc5de762a68bcd7d776c97e0))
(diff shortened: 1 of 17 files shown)


jooyoungseo:
All three addressed in `c60f7cae`. Finding 1 is a fair catch on my own claim — I listed grid-cell point mode in the PR table and wrote no test for it.

## 1. Grid-cell point mode — fixed

You're right that it is the one mode the falsification table did not cover, and right about *why* it matters: `enterGridCell()` groups the cell's points in its own pass rather than reusing `xPoints`, so naming the other four sites genuinely does not name this one.

Test added, and falsified on its own — reverting the label in that grouping pass alone (`scatter.ts:1670`) fails **exactly this test and nothing else**:

```
✕ grid-cell point mode announces the category
Tests: 1 failed, 14 passed, 15 total
```

Which is the demonstration that it is a separate code path rather than a redundant assertion. **15 tests now.**

## 2. Last-write-wins on inconsistent labels — documented

Agreed it is low risk and agreed nothing said what happens. Recorded in both places you suggested — the `ScatterXPoint.label` doc comment and `SCHEMA.md`:

> One name per numeric slot: every point sharing an `x` must carry the same `xLabel`, since the slot *is* the category. A second, different name for the same `x` is dropped rather than splitting the category into two columns.

Dropping is the right answer rather than merely the current one: inventing a second column would split one category in half, and there is nothing better to do with a name for a slot that already has one.

## 3. `AxisFormat` on a categorical axis — confirmed, and filed as #930

Measured rather than reasoned about, because the "pre-existing" half is the part worth being sure of:

```
formatters.currency('USD', 2)('a')  ->  "$NaN"
formatters.percent(1)('a')          ->  "NaN%"
formatters.currency('USD', 2)(12)   ->  "$12.00"
```

And your reading of where it happens is exactly right — the model is not the problem. A `BarTrace` built with `data: [{x: 'a', y: 1}]` and a currency x-format emits `{"label":"g","value":"a"}`; the string leaves the model intact and the `NaN` appears in `TextService`. So it is pre-existing and identical for bar, not a regression here.

Filed as **#930** rather than folded in. It is a different defect with a different fix (guard the formatters on `Number.isFinite`), it affects bar today with no scatter involved, and mixing it in would make this PR two things. Same split as #441 out of #438 earlier.

Worth saying plainly though: it is the sharpest of your three findings. "g is $NaN" is not a degraded announcement, it is a confident one that is false — the class this library exists to prevent.

## Note on the sandbox

Understood on not being able to re-run the gates. For the record, on this commit: `npm run type-check` clean, `npm run lint:fix` clean, `npm test` **4357 + 136 passed**, `npm run build` 13/13.

---

PR #931, opened 2026-08-16
fix(format): announce a value a numeric format cannot express, not NaN

### LINKED ISSUE #930: A numeric axis format applied to a category name announces "$NaN"
## The defect

`AxisFormat`'s numeric formatters coerce with `Number.parseFloat(String(value))` and format the result unconditionally. Given a category *name* they produce a formatted `NaN`:

```
formatters.currency('USD', 2)('a')  ->  "$NaN"
formatters.percent(1)('a')          ->  "NaN%"
formatters.currency('USD', 2)(12)   ->  "$12.00"
```

Measured directly against `src/util/format.ts`. `number` and `scientific` take the same `parseFloat` path (`format.ts:75`, `:96`, `:139`, `:150`).

## How it is reached

A layer whose x carries names *and* declares a numeric `axes.x.format`. The model emits the name and `TextService` formats it:

```ts
const mainValue = Array.isArray(text.main.value)
  ? this.formatArrayValue(...).join(', ')
  : this.formatSingleValue(text.main.value as number | string, mainAxisType);
```

Confirmed the model's half is not the problem — a `BarTrace` built with `data: [{x: 'a', y: 1}]` and a currency x-format emits `{"label":"g","value":"a"}`. The string is intact leaving the model; the `NaN` appears in the service.

**Pre-existing, not new.** `BarPoint.x` has been `string | number` since long before this was noticed, so any bar layer on a named axis with a numeric x-format already announces `$NaN`. #929 makes it reachable for scatter as well (`ScatterPoint.xLabel`), which is how it surfaced, but it introduces neither the coercion nor the shape.

## Why it matters

This is the class of failure MAIDR exists to prevent: not silence, but a confident announcement of something false. A reader hears "g is $NaN" for a bar the chart labels "Cherries". Nothing in the audio, the braille or the text says the number is an artefact of a formatter meeting a value it cannot parse — it reads exactly like a real value would.

`NaN` also does not survive `JSON.stringify` in the places the payload is serialised, which is a separate hazard the same coercion feeds.

## Possible fixes

1. **Pass the value through unformatted when it does not parse.** One guard per formatter, or one in a shared helper: `Number.isFinite(num) ? format(num) : String(value)`. A name announced as itself is right, and it is what the axis meant.
2. **Ignore a numeric format on an axis whose values are names**, decided once per axis rather than per value. Cleaner conceptually, but the axis' type is not currently something the service knows independently of the values it is handed.
3. **Reject the combination at validation time.** Loudest, and probably wrong: a mixed axis is legal today (`BarPoint.x` is a union), and refusing to render is worse than announcing the name.

Option 1 is the smallest and covers the mixed case, which the other two do not.

## Reproducing

```ts
import { formatters } from '@util/format';
formatters.currency('USD', 2)('a');  // "$NaN"
```

## Related

- #929 — where this surfaced; adds `ScatterPoint.xLabel` / `yLabel`, so a scatter can now reach the same combination a bar always could
- #925 — the same principle for a *missing* value: `Number(null)` is `0`, and announcing a gap as a real reading of zero was the bug

jooyoungseo:
Closes #930.

## What was wrong

Every numeric formatter coerced with `Number.parseFloat(String(value))` and formatted the result unconditionally. Given a category *name*, all six produce something false:

```
currency    -> $NaN
percent     -> NaN%
number      -> NaN
scientific  -> NaN
fixed       -> NaN
date        -> THREW RangeError
```

Measured directly against `src/util/format.ts` with `'Cherries'`.

**A named axis with a numeric `AxisFormat` is a legal combination.** `BarPoint.x` is `string | number` and has been all along; #927's `ScatterPoint.xLabel` now reaches the same place. So this is not a malformed payload being punished — it is a formatter meeting a value it cannot express.

And "g is $NaN" is not a degraded announcement. It is a confident one that is false, which is the failure this library exists to prevent. Nothing in the audio, the braille or the text says the number is an artefact of a coercion; it reads exactly like a real value would.

## The `date` case is worse, and was not in the original report

`Intl.DateTimeFormat.format` **throws** a `RangeError` on an invalid date rather than returning a NaN-ish string. I traced the path looking for a guard and there is none — `FormatUtil.applyFormat`, `FormatterService.formatSingleValue` and `TextService.formatSingleValue` all call straight through. So a `date` format on a named axis took the **whole reading** out, not one value.

That is the sharper half of this fix, and I only found it by probing all six rather than the three the issue named.

## The fix

One shared `asFiniteNumber(value): number | null`, and each formatter returns `String(value)` when it gets `null`. The `date` formatter checks `Number.isNaN(date.getTime())` before reaching `Intl`.

The rule matches one `TextService` already applies a layer up — a non-finite *number* is announced as missing rather than as a value. The difference here is that the value is real and only the formatting does not apply, so it is announced **as itself** rather than as "missing".

**The coercion is kept for what it was there for.** A producer emitting numbers as strings is ordinary, and `'1234.5'` still formats as `$1,234.50`. That is asserted, not assumed — a guard that threw away `parseFloat`'s actual purpose would trade one defect for another.

## Tests

Six added to `test/util/format.test.ts`:

- a category name comes back unchanged from all five numeric formatters
- the date formatter **does not throw**, and returns the name
- `NaN` and `Infinity` are treated the same way — announcing `$∞` would be the same confident falsehood
- **numeric strings still format** (`'1234.5'` → `$1,234.50`, `'0.156'` → `15.6%`, `'3.14159'` → `3.14`)
- real numbers still format, across all five
- a real date still formats (`'2023-01-15'` → `Jan 15`)

### Falsification

| reverted | tests that fail |
|---|---|
| `asFiniteNumber` returns the raw coercion | **2** |
| the `date` validity check | **1** |
| nothing | **0 / 16 pass** |

Disjoint — the numeric guard and the date guard are separate failures.

## Verification

- `npm run lint:fix` — clean
- `npm test` — **4364 passed** (292 suites) and **136 passed** (10 ESM suites)

## Scope

Deliberately at the formatter rather than at the axis. Deciding once per axis "these values are names, ignore the numeric format" would be cleaner conceptually, but the axis' type is not something the service knows independently of the values it is handed — and a *mixed* axis (`BarPoint.x` being a union) would still need a per-value answer. Rejecting the combination at validation time is the third option and probably wrong: refusing to render is worse than announcing the name.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,51 @@
+# [4.3.0](https://github.com/xability/maidr/compare/v4.2.0...v4.3.0) (2026-08-16)
+
+
+### Bug Fixes
+
+* **ci:** size the e2e report's code fence so its content cannot close it ([#924](https://github.com/xability/maidr/issues/924)) ([e4afdd9](https://github.com/xability/maidr/commit/e4afdd9c9ffe5f76e604dce8b4416b89cbc36c48)), closes [#715](https://github.com/xability/maidr/issues/715)
+* **docs:** generate the examples gallery so it cannot miss a page ([#914](https://github.com/xability/maidr/issues/914)) ([f5aecad](https://github.com/xability/maidr/commit/f5aecad934944c3f36413e23403b81493b3428f0)), closes [#911](https://github.com/xability/maidr/issues/911)
+* **docs:** give built pages heading ids so in-page links jump ([#918](https://github.com/xability/maidr/issues/918)) ([9ed5a39](https://github.com/xability/maidr/commit/9ed5a396e2e4668c3d00e6d91861995223d73607)), closes [#917](https://github.com/xability/maidr/issues/917) [#913](https://github.com/xability/maidr/issues/913)
+* **docs:** resolve the committed conflict markers and guard against more ([#921](https://github.com/xability/maidr/issues/921)) ([0c04513](https://github.com/xability/maidr/commit/0c04513a25cb96afbff3d15195cbb435716140b1)), closes [#918](https://github.com/xability/maidr/issues/918) [#917](https://github.com/xability/maidr/issues/917)
+* **highlight:** let a flow or network trace publish the ribbon it highlighted ([#906](https://github.com/xability/maidr/issues/906)) ([da131aa](https://github.com/xability/maidr/commit/da131aa76099a2730119e9bdf9c5391f99a72cd9)), closes [#895](https://github.com/xability/maidr/issues/895) [#903](https://github.com/xability/maidr/issues/903) [#899](https://github.com/xability/maidr/issues/899) [#904](https://github.com/xability/maidr/issues/904) [#903](https://github.com/xability/maidr/issues/903) [#904](https://github.com/xability/maidr/issues/904) [#904](https://github.com/xability/maidr/issues/904)
+* **line:** read every path command, so a staircase highlights its samples ([#908](https://github.com/xability/maidr/issues/908)) ([035c5bd](https://github.com/xability/maidr/commit/035c5bdd95d076385d7e26b815fb0e9503811005))
+* **plotly:** count a box's group across the whole box layer ([#883](https://github.com/xability/maidr/issues/883)) ([78375cb](https://github.com/xability/maidr/commit/78375cb4587b1468fb14d4dc5381b0529f172d3a)), closes [#882](https://github.com/xability/maidr/issues/882)
+* **plotly:** scope a candlestick's selector to its own panel and trace ([#882](https://github.com/xability/maidr/issues/882)) ([1cf0470](https://github.com/xability/maidr/commit/1cf0470be356e5d9adcb97d938d0795695367042)), closes [#881](https://github.com/xability/maidr/issues/881)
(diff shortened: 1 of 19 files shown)


jooyoungseo:
Style nit taken in `258db422` — all five numeric formatters now use the early-return form.

I nearly left it, since `currency` and `number` can't take a clean ternary anyway (the `Intl` call spans lines) and converting the other three costs two lines each. What changed my mind is what the guard *is*: the thing a seventh formatter must not be written without. One that looks identical in all five is harder to leave out than one with two spellings, so the uniformity buys something beyond conformity.

Two things in your review worth confirming back:

**`wrapFormat` intercepting numeric `NaN` upstream** — thanks for tracing that, I hadn't. It means the `NaN`/`Infinity` half of `asFiniteNumber` is defence-in-depth rather than load-bearing *in the production pipeline*. Still worth keeping and still worth testing, for the reason you give: `formatters.*` are exported and reached directly by `GoToExtremaViewModel` and by anything else that resolves a formatter without going through `wrapFormat`. The category-name half is load-bearing everywhere.

**The `date` case being the more important half** — agreed, and it is the part I'd have missed by fixing only what the issue named. I filed #930 off three formatters the previous review mentioned; probing all six is what turned up that `Intl.DateTimeFormat` is the only one that *throws* rather than returning a NaN-ish string. A wrong value is bad; an exception between the model and the announcement takes the whole reading out.

On not being able to run the gates: CI is green on this branch, and locally on this commit `npm test` gives **4364 passed** (292 suites) and **136 passed** (10 ESM suites), with `type-check` and `lint:fix` clean.

---

Issue #937, opened 2026-08-17
feat(tableau): settle stacked vs dodged bar detection, or document it as undecidable


jooyoungseo:
Follow-up to #932.

Tableau's summary data gives each segment's own value whether the view drew the bars stacked or side by side. Nothing in the rows distinguishes the two, so the adapter defaults to `dodged_bar` and reaches `stacked_bar` / `stacked_normalized_bar` only through an explicit `overrides.traceType`.

`dodged_bar` is the honest default — it announces each group's own value and never claims a total the view may not have drawn — but it does mean a stacked chart is announced as a grouped one unless the page says otherwise.

**Work**

- Once the visual specification is wired up (see the `getVisualSpecificationAsync` issue), check whether the marks card's `Color` encoding plus the mark type can settle it.
- If it can, add the inference and the tests.
- If it cannot, say so in `docs/tableau.md` in as many words and document `overrides.traceType` as the only answer — a documented limit beats a confident guess.

**Blocked by:** the `getVisualSpecificationAsync` verification issue.

PR #939, opened 2026-08-17
fix(tableau): correct the visual specification against the shipped declarations

### LINKED ISSUE #935: feat(tableau): verify getVisualSpecificationAsync against the real payload and use it for mark-type classification
Follow-up to #932.

`reader.ts` already feature-detects `worksheet.getVisualSpecificationAsync` and hands the result to the extractor, which prefers it over the heuristic ladder when present (`markTypeToTrace`). That path is written from the documentation prose and has never seen a real payload — no test can exercise it against live Tableau, and the Embedding API surface does not expose the method at all today, so it is currently dead code that will wake up under the Extensions binder.

**Work**

- Verify the actual shape against the shipped `@tableau/extensions-api-types` declarations rather than the docs prose: `activeMarksSpecificationIndex`, `marksSpecifications[].primitiveType`, the `encodings` array and `rowFields`/`columnFields`.
- Confirm whether `activeMarksSpecificationIndex` is meaningful in view mode.
- Work out how a dual-axis viz populates `marksSpecifications` — the current mapping assumes one card and falls back to `marksSpecifications[0]`, which is a guess for multi-card charts.
- Extend the `MarkType` → `TraceType` table to whatever the real enum contains, and keep the `default:` fall-through to the ladder rather than throwing.

**Why now:** the mapping table is the primary trace-type evidence once the Extensions binder lands, so a wrong reading here silently mislabels every chart it touches.

jooyoungseo:
# Pull Request

## Description

Closes #935.

The adapter's mark-type path was written from documentation prose and had never been checked against the contract Tableau actually ships. This PR reads the real declarations — `@tableau/embedding-api@3.12.1`, which vendors `@tableau/api-external-contract-js` — and corrects everything they contradict.

Two of them mattered:

**`getVisualSpecificationAsync` is declared on the Embedding worksheet, not only the Extensions one.** `ExternalContract/Embedding/SheetInterfaces.d.ts` declares it inside `interface Worksheet`. #932 assumed it was Extensions-only and said so in the docs, which made the whole spec path look like dead code waiting on a future binder. It is live on the surface that already shipped — when the host page's embedding library is new enough, the mark type outranks the heuristics today.

**`MarkType` has exactly thirteen members and no `Automatic`.** Tableau resolves "Automatic" to whatever primitive it drew, which is what we want; the mapping now handles all thirteen explicitly rather than assuming the set.

## Related Issues

Closes #935. Unblocks part of #937 — see *Additional Notes*.

## Changes Made

- **`types.ts`** — the structural interfaces now mirror the real ones member for member (`VisualSpecification`, `MarksSpecification`, `Encoding`, `FieldInstance`/`FieldBase`), each carrying a doc comment naming the declaration file and API version it was checked against, so the next reader can re-verify instead of trusting this commit. Notably every member is required and `readonly`; none is optional, which our previous guesswork had wrong.
- **`extractor.ts`** — `markTypeToTrace` handles all thirteen `MarkType` values. The ones with an honest MAIDR equivalent map to it; `text`, `map`, `polygon`, `gantt-bar` and `viz-extension` skip the worksheet with a warning rather than being approximated. The `default:` arm still falls through to the heuristic ladder, for a member a later Tableau version adds.
- **`activeMarksSpecificationIndex`** is a plain `number` the declarations never bound, so an out-of-range index now falls back to the first marks card and then to the ladder, instead of reading `undefined` off the array.
- **`reader.ts`** — feature detection stays. An older embedding library on the host page genuinely will not have the method, so this is not defensive clutter.
- **`docs/tableau.md`** — corrected everywhere it described the visual specification, including the claim that the mark type is unavailable on this surface.

The heuristic ladder is unchanged. It remains the fallback, and `overrides.traceType` still outranks the spec.

Not applicable.

## Checklist

Locally: `npm run type-check`, `npx eslint`, `npm test` (297 suites / 4501 tests — 51 new — plus 10 / 137 ESM), `npm run build` (14 bundles) and `node scripts/build-site.js` all pass.

## Additional Notes

**What this does not do.** Everything here is verified against the shipped *type declarations*, not against a live Tableau payload. The declarations settle the shape; they cannot settle behaviour. Two questions stay open and are deliberately left to #935's sibling issues rather than guessed at here:

- Whether `activeMarksSpecificationIndex` is meaningful in view mode — the declarations do not say, hence the fallback rather than a claim.
- How a dual-axis viz populates `marksSpecifications`. The declarations permit several cards; which one a dual-axis chart marks active is not documented.

**For #937 (stacked vs dodged).** The real `EncodingType` enum does expose `Color`, so a colour encoding is distinguishable in principle. But the *internal* contract that actually builds the payload carries only ten of the public enum's sixteen members, and nothing in either declaration says whether a stacked bar and a side-by-side bar differ in the spec at all. So #937 is not settled by this PR — it is now answerable only by inspecting a real payload from both layouts.

---
_Generated by [Claude Code](https://claude.ai/code/session_01661Y35cni4efW36rCQNkfL)_


jooyoungseo:
Thanks — point 1 is the right thing to insist on, since the whole diff rests on it. I've now verified it independently rather than from the copy the change was written against.

**Provenance first.** Re-fetched the package from the registry and checked it against the registry's own metadata before reading a byte of it:

```
$ npm view @tableau/embedding-api@3.12.1 dist.integrity dist.shasum
dist.integrity = 'sha512-fHDrh+sEgSdOyC2sRDSCxGm2o92TvX1YN3Z/8JnaZKfPRxCXgVmrZy2LIsUnnkB4cTo376eZVNJk3dYJ34KLiw=='
dist.shasum    = '89fc342973f1be6c4fecdc2bb9a1b6f95ac4a2bf'

$ npm pack @tableau/embedding-api@3.12.1
$ openssl dgst -sha512 -binary tableau-embedding-api-3.12.1.tgz | openssl base64 -A
fHDrh+sEgSdOyC2sRDSCxGm2o92TvX1YN3Z/8JnaZKfPRxCXgVmrZy2LIsUnnkB4cTo376eZVNJk3dYJ34KLiw==

$ sha1sum tableau-embedding-api-3.12.1.tgz
89fc342973f1be6c4fecdc2bb9a1b6f95ac4a2bf
```

Both digests match the registry. Vendored contract version, read from the tarball: `@tableau/api-external-contract-js@1.211.0` — the version the PR claims.

**The three load-bearing facts, checked against that tarball:**

1. `getVisualSpecificationAsync` is declared **inside `export interface Worksheet extends Sheet`** in `ExternalContract/Embedding/SheetInterfaces.d.ts` — confirmed by extracting the interface body and finding it there, not merely by grepping the file (which would also have matched an unrelated declaration elsewhere in it):

   ```
   462:    getVisualSpecificationAsync(): Promise<VisualSpecification>;
   ```

2. `MarkType` in `ExternalContract/Shared/Namespaces/Tableau.d.ts` has **exactly 13 members**:

   ```
   Bar="bar" Line="line" Area="area" Square="square" Circle="circle" Shape="shape"
   Text="text" Map="map" Pie="pie" GanttBar="gantt-bar" Polygon="polygon"
   Heatmap="heatmap" VizExtension="viz-extension"
   ```

3. `Automatic` does not appear anywhere in that enum file, which is why the mapping treats "Automatic" as resolved-by-Tableau rather than as a value to handle.

**On point 2** — I grepped `README.md`, all of `docs/`, and `src/` for surviving "Extensions-only" phrasing about the visual specification. Nothing stale outside the diffed file; `reader.ts:136` already states it is declared on **both** public `Worksheet` interfaces.

**On point 3** — agreed, and to be explicit about why `markTypeToTrace` takes `string`: the value arrives as JSON across an iframe boundary from a library build this code did not choose, so a closed union would be a claim about the payload rather than about our own code. The exhaustive `Record<TableauMarkType, …>` in the test is what keeps the mapping honest at compile time; the runtime signature stays open on purpose.

One thing your review is right to leave standing: this verifies the *declarations*, not runtime *behaviour*. Whether `activeMarksSpecificationIndex` is meaningful in view mode, and what a dual-axis viz puts in `marksSpecifications`, still need a real payload — which is why they are called out in the PR body as open rather than resolved here.

---

PR #941, opened 2026-08-17
feat(tableau): lay dashboard worksheets into a geometry-aware subplot grid

### LINKED ISSUE #936: feat(tableau): lay dashboard worksheets into a geometry-aware subplot grid
Follow-up to #932.

The adapter currently lays a dashboard's worksheets into an **N×1** subplot column, in `dashboard.worksheets` order. That order is defensible — Tableau's own accessibility guidance says screen readers read dashboard objects in the order they were added, so it matches what a reader already hears — but it throws away the dashboard's actual two-dimensional layout. Two worksheets sitting side by side are announced as if one were below the other.

**Work**

- Band the worksheets by their vertical position and sort by horizontal position within each band, producing a real `MaidrSubplot[][]` grid.
- **First confirm the API actually offers the geometry.** `DashboardObject.position` / `.size` are documented on the *Extensions* `DashboardObject`; whether the Embedding API's equivalent carries them is unverified. Feature-detect and fall back to the current N×1 column when absent — never assume.
- Keep the invariant that a skipped worksheet contributes no subplot and never shifts the `layer.id` mapping the selection index depends on.

**Care needed:** a wrong grid is worse than a flat column, because Page Up / Page Down would then move in an order that contradicts what the reader was told.

jooyoungseo:
# Pull Request

## Description

Closes #936.

A Tableau dashboard's worksheets were laid into an **N×1** subplot column in add-order. That order was defensible — it matches the order a screen reader already narrates the dashboard — but it discarded the layout the author actually drew: two worksheets side by side were announced as if one were below the other.

This reads each dashboard object's `position` and `size`, bands the worksheets into rows by vertical position, and orders them by horizontal position within a band, so paging through the subplots follows the drawn layout.

## Related Issues

Closes #936.

## Changes Made

**Geometry is captured in the binder, not the reader.** `reader.ts` reads one worksheet at a time and never sees the dashboard; geometry is a sheet-level fact on `dashboard.objects`, and the binder already owns sheet discovery and already re-runs it on `tabswitched`. `WorksheetSnapshot` gains an optional `geometry` of six primitives (x, y, width, height, isFloating, isVisible) — flattened rather than holding `Point`/`Size` so the snapshot stays plain JSON, which is what will let an Extensions binder fill it in unchanged. The extractor stays pure: it receives four numbers and two booleans, never a Tableau object.

**Banding is a vertical-overlap ratio, not a pixel threshold.** Two worksheets join a band when their vertical intervals overlap by at least half of the shorter extent. A fixed pixel tolerance is the wrong shape of rule — 10px is generous on a 600px dashboard and meaningless on a 4000px one, and it moves again under browser zoom and device-pixel-ratio. The ratio is scale-free, and 0.5 is not a tuned constant: it is the point where "these sit side by side" stops being more true than "one is above the other". It is nowhere near the case it exists to absorb — per-object padding on objects hundreds of px tall gives an overlap ratio around 0.98 — so it only bites on layouts that are genuinely ambiguous, which is exactly where falling back is right.

**It falls back rather than guessing.** Any survivor without geometry (an older host library), anything floating or invisible, or a banding that does not come out confident, and the layout reverts to the previous N×1 column. A wrong grid is worse than a flat one, because Page Up / Page Down would then move in an order that contradicts what the reader was told.

Not applicable — the change is in navigation order, which is non-visual.

## Checklist

Locally: `npm run type-check`, `npx eslint`, `npm test` (297 suites / 4531 tests — 30 new — plus 10 / 137 ESM), `npm run build` (14 bundles) and `node scripts/build-site.js` all pass, re-run after the rebase onto `main`.

### The invariant this change could have broken silently

`layer.id` was `String(subplots.length)` — the count of surviving subplots — and it keys the `SelectionIndex` that maps a layer back to its live worksheet. That worked only because survivor count and row count were the same number under an N×1 layout. **The moment a row holds two subplots, that identity breaks and the ids collide**, which would send a highlight to the wrong worksheet without any error. The id is now the survivor index explicitly, never derived from grid position, and a test pins it. Nothing in the model derives row/col from `layer.id` (it is stored opaquely and echoed into `NavigateCallback`), so keeping it flat while the grid reorders around it is sound.

### Verified rather than assumed

- **Ragged rows are supported.** `MovableGrid.moveOnce`/`.moveToExtreme` re-clamp the column after a row change, `isMovable` bounds-checks per row, and every service that walks subplots iterates `subplots[r].length`. Nothing requires rectangularity. An *empty* row, by contrast, crashes `Figure.state`, so the banding never emits one.
- **Row 0 is the bottom of the figure** for navigation. Tableau layers emit no `selectors`, so `resolveSubplotLayout` cannot read geometry from the DOM and every Tableau figure takes `buildFallbackLayout` (`invertVertical: false`). The banding is ordered to match.

### Two things the declarations could not settle

- `Size` carries no "pixels" wording where `Point` does. `size` is therefore used only inside ratios, so a unit mismatch could change a confidence verdict but never produce a wrong ordering.
- Whether `object.worksheet` is reference-identical to the matching entry in `dashboard.worksheets` is not stated, so worksheets are matched by `name` — which the `worksheets` doc comment does guarantee. Note this is deliberately *not* `DashboardObject.name`, which is the object's authoring name and can differ.

---
_Generated by [Claude Code](https://claude.ai/code/session_01661Y35cni4efW36rCQNkfL)_

### DIFF
--- a/docs/tableau.md
+++ b/docs/tableau.md
@@ -65,8 +65,35 @@ The adapter never imports a Tableau package. It duck-types the live `<tableau-vi
 1. **Find the worksheets.** `viz.workbook.activeSheet` is read at bind time, and again on every refresh — a `tabswitched` event changes which sheet the worksheets come from. A `worksheet` sheet is a single worksheet; a `dashboard` contributes `dashboard.worksheets`, in the order the author added them — which is also the order Tableau's own documentation says a screen reader narrates a dashboard. A `story` sheet is skipped with a warning (see [Limitations](#limitations)).
 2. **Read the summary data.** For each worksheet the adapter calls `getSummaryColumnsInfoAsync()` for the columns *in view order*, then opens a `getSummaryDataReaderAsync()` and pages through it. The reader hands its columns back **alphabetically**, so the adapter builds a view-order-to-alphabetical index map by `fieldId` and remaps every row; every index downstream is a view index. The reader is always released in a `finally` block, and every read goes through a per-viz promise chain, because Tableau supports **only one active summary-data reader at a time** and a leaked one blocks the next read.
 3. **Classify the columns.** Each column becomes a measure or a dimension — see [Supported Chart Types](#supported-chart-types).
-4. **Decide the trace type and build the layer.** One worksheet produces exactly **one** MAIDR layer inside its own subplot; a dashboard of four worksheets is a four-row, one-column subplot grid.
+4. **Decide the trace type and build the layer.** One worksheet produces exactly **one** MAIDR layer inside its own subplot. How those subplots are arranged depends on what the dashboard reports about itself — see [Dashboard Layout](#dashboard-layout).
 5. **Mount.** A wrapper `<div>` is inserted before the viz element and a React root renders MAIDR into it. The viz element itself is never moved — `<tableau-viz>` is a custom element and re-parenting re-runs its `connectedCallback`, with undocumented consequences for the iframe.
+
+### Dashboard Layout
+
+A dashboard's worksheets become a **two-dimensional subplot grid that follows the dashboard's own layout** when the layout is readable and unambiguous, and a **one-worksheet-per-row column** whenever it is not. Both paths are always available and nothing on the page has to opt in.
+
+The geometry comes from `dashboard.objects`: each object reports a `position`, a `size`, and — for a worksheet object — the worksheet inside it. The binder reads it once per re-read and hands four numbers and two flags per worksheet to the extractor, which stays pure.
+
+**How the grid is worked out.** Worksheets are swept top to bottom into *bands*. A worksheet joins the band being built when its vertical extent overlaps the band's by at least **half of the shorter of the two**, and starts a new band otherwise; within a band the order is left to right. The threshold is a ratio rather than a pixel tolerance on purpose — ten pixels is generous on a 600px dashboard and meaningless on a 4000px one, and it moves again under browser zoom. Real tiled dashboards miss perfect alignment by the few pixels of an object's padding and border, which is an overlap of about 98%, so the threshold only bites on layouts that are genuinely ambiguous.
+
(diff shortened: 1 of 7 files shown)


jooyoungseo:
Both observations addressed in `15cd7e0` — comments only, no behaviour change.

**On the double discovery.** You're right that it reads oddly, but it isn't redundant: the bind-time call exists so an empty sheet returns `null` **before** a wrapper is mounted, leaving the page untouched, while the call inside `refresh()` has to re-read because a `tabswitched` refresh describes a different sheet entirely. Collapsing them would mean either mounting a wrapper and then tearing it down, or teaching `refresh()` to skip its re-discovery on the first pass — both worse than one extra synchronous property walk. So I've kept the two calls and written down why, which is the outcome your "in case it's worth a comment noting it's intentional" points at.

**On the duplicate worksheet name.** Agreed it isn't worth guarding, so I've recorded the assumption instead of adding a branch that can never be taken: `Dashboard.worksheets` is documented as corresponding one-for-one with the `worksheet`-typed objects, so a collision cannot arise from authoring. The comment also pins down the thing that *is* easy to get wrong there — the key is `object.worksheet.name`, never `object.name`, since the latter is the zone's authoring name and can differ from the sheet it holds.

**On verification.** Fair to want CI rather than my word for it — the checks on this commit are the independent run. For what it's worth locally: `type-check`, `eslint` and the Tableau suites (5 suites / 167 tests) pass on `15cd7e0`, and the full run on the previous commit was 297 suites / 4531 tests plus 10 / 137 ESM, with `build` (14 bundles) and `build-site` green.

Thanks for tracing the staircase case by hand — that all-pairs re-check after the sweep is exactly the part I'd have wanted a second reader on, since the running-union sweep alone merges `A/B/C` and only the re-check catches it.

---

PR #943, opened 2026-08-17
feat(error-bar): give a grouped interval chart a shape in the grammar

### LINKED ISSUE #942: feat: an error bar chart with groups has no shape in the grammar
## Description

`error_bar` takes a **flat** `ErrorBarPoint[]`, and `ErrorBarPoint` has no field naming which group a point belongs to:

```ts
data: … | ErrorBarPoint[] | …          // src/type/grammar.ts:1312

export interface ErrorBarPoint {
  x: number | string;
  y: number;
  yMin?: number;
  yMax?: number;
}
```

Every other trace that can be grouped carries the group. `bar`, `line`, `smooth`, `step`, `segmented` and `contour` are all `…Point[][]` — a series per group — and their points carry `z`, with `axes.z` naming the grouping variable. `error_bar` is the one that cannot say it.

A grouped error bar chart is not an exotic figure. It is the ordinary way to show two treatments across several categories, and the comparison it exists to support — *does group p's interval overlap group q's?* — is precisely the one that becomes unavailable.

## Both bindings are already bent around this

**r-maidr** flattens the groups into one series, silently. Measured on `main`:

```r
d <- data.frame(g = rep(c("a","b","c"), each = 2), grp = rep(c("p","q"), 3),
                y = c(2,3,4,5,3,6), lo = c(1.5,2.4,3.2,4.1,2.7,5.2),
                hi = c(2.9,3.6,5.5,6.0,3.2,6.9))

ggplot(d, aes(g, y, colour = grp)) +
  geom_errorbar(aes(ymin = lo, ymax = hi), position = position_dodge(width = 0.5))
```

```
type=error_bar   one series, 6 points, axes keys: x,y
  x=a  y=2  yMin=1.5  yMax=2.9
  x=a  y=3  yMin=2.4  yMax=3.6
  x=b  y=4  yMin=3.2  yMax=5.5
  x=b  y=5  yMin=4.1  yMax=6
  x=c  y=3  yMin=2.7  yMax=3.2
  x=c  y=6  yMin=5.2  yMax=6.9
```

Every number is right and nothing is missing — which is what makes it the bad case. A reader hears six points where each category name arrives twice with different values and nothing saying which is `p` and which is `q`. Recovering the grouping means assuming the emission order is stable, which nothing states.

The same chart as bars keeps its groups, so the loss is specific to this trace type rather than to the binding:

```
dodged bar     2 series, points carry z=p / z=q, axes keys: x,y,z
grouped line   2 series, points carry z=p / z=q, axes keys: x,y,z
dodged errorbar  1 series, no z,                 axes keys: x,y
```

**py-maidr** has already met this and chose to lose the intervals rather than mis-assign them — `maidr/patch/pointplot.py`, on `sns.pointplot(..., hue=…)`:

> more than one estimate means a `hue` split the chart into groups while the MAIDR error bar layer carries a single series — so those intervals are dropped rather than mis-assigned.

That is the right call against today's grammar, and it means a hued seaborn point plot currently reads as a plain `line`: the estimates survive, the intervals do not.

So the same gap costs one binding its group identity and the other its intervals.

## Scope

- [ ] `ErrorBarPoint[][]` in the trace data union, alongside the flat form — a single-group chart must keep emitting `ErrorBarPoint[]` unchanged
- [ ] A group field on the point (`z`, matching `BarPoint`/`LinePoint`) and `axes.z` naming the variable
- [ ] `ErrorBarTrace` navigating group-wise, the way the multi-series traces do
- [ ] Announcement and braille carrying the group, as `z` already does elsewhere

The three-section grid (`lower` / `value` / `upper`) is the wrinkle worth planning for: `ErrorBarTrace` already spends its row axis on the sections, so groups need a third dimension rather than more rows — which is what makes this a design question rather than a widening of a type.

## Follow-on

Once this lands, both bindings can stop working around it:

- r-maidr — `Ggplot2ErrorbarLayerProcessor` splits on the grouping aesthetic instead of flattening
- py-maidr — `_register_point_layer` keeps the intervals for a hued `pointplot` instead of falling back to `line`

## Related

Interval trace: #789. Coverage roadmap: #814.

jooyoungseo:
# Pull Request

## Description

`error_bar` took a flat `ErrorBarPoint[]` with no field naming a group, while every other groupable trace is `…Point[][]` carrying `z`. A dodged chart over two treatments therefore arrived as two estimates at each category with nothing telling them apart:

```
dodged bar       2 series, z=p / z=q, axes.x,y,z
grouped line     2 series, z=p / z=q, axes.x,y,z
dodged errorbar  1 series, no z,      axes.x,y     ← a,a,b,b,c,c
```

Every number right, no point missing — which is what made it worse than a visible failure. The reading sounded complete, and the comparison the chart is drawn for, *does one group's interval clear the other's*, was the part that had gone.

**Both bindings were already bent around this.** r-maidr flattens the groups into one unlabelled series (xability/r-maidr#183). py-maidr drops a hued `pointplot`'s intervals rather than mis-assign them (xability/py-maidr#462) — the right call against the old grammar, and it means one binding loses group identity while the other loses its uncertainty, from the same missing shape.

## Related Issues

Closes #942. Unblocks xability/r-maidr#183 and xability/py-maidr#462.

## Changes Made

**`src/type/grammar.ts`** — `ErrorBarPoint.z?: string`, named and documented to match `LinePoint.z`; `ErrorBarPoint[][]` added to the `MaidrLayer.data` union alongside the flat form.

**`src/model/errorBar.ts`** — rows become groups × sections.

### The ordering is the design decision

Rows run **group-major**: each group's sections contiguous and ordered bottom to top, then the next group's.

```
row 0  control lower      row 3  treated lower
row 1  control value      row 4  treated value
row 2  control upper      row 5  treated upper
```

That is what makes the chart's own question answerable by ear. Pressing up at one category walks control's lower, value and upper and then treated's — so control's upper (5.5) and treated's lower (4.1) are **adjacent moves**, and a reader hears the overlap directly. Interleaving by section instead (every group's lower, then every group's value) would put the two ends of one interval three moves apart and answer nothing.

Crossing from a group's upper bound to the next group's lower is a step *down* in value, which the shared pitch scale renders honestly as a drop. The announcement names the group on every move, so it is heard as arriving somewhere else rather than as the data falling.

### What deliberately does not change

* **The flat shape.** A single-series chart is the common case and must not have to wrap itself in an array to keep working; `toGroups` makes a flat payload the one group it is. An empty payload becomes one *empty* group rather than no groups, so a layer with no data keeps the single-row grid it had instead of one with no row for the cursor to be on.
* **`points` stays the flattened list**, so `ForestTrace` — which reads each study's weight and pooled flag off it — and the selector-count check both go on meaning what they meant. For an ungrouped chart it is byte-for-byte what it was.
* **A single unnamed group announces no group.** One group has nothing to disambiguate, and "Group is Error bar 1" on every move is noise in place of a reading. The rule `LineTrace` already applies to its series.
* **Columns come from the current row**, not the flattened total — the rule `LineTrace.dimension` follows. Using the total would let the cursor walk off the end of every row.
* **Selectors slice per group.** The flat element list runs in the same group-major order the rows do, so each group's rows take that group's slice; handing every row the whole list would light a second group's whips while the cursor was in the first.

### The description

Gains a group column and a `Number of groups` stat when there is more than one group. Without the column the table has two rows headed `a` differing only in their numbers — the announcement's loss, written down. An ungrouped table keeps the four columns it had.

Not applicable — the change is in what is announced and sonified.

## Checklist

- [ ] I have updated the documentation, if applicable — `docs/BRAILLE.md` needs no change: the braille encoder takes rows generically and the extra rows encode as they did.

### Tests

`test/model/errorBarGroups.test.ts`, 16 new, in four groups: the grid shape and the group-major ordering; the announcement naming the group (including the authored `axes.z.label`, so a chart grouped by treatment does not say "Group is control"); the ungrouped path unchanged, including a single group wrapped in an array reading identically to a flat one; and the description table.

The existing `errorBar.test.ts` and `forest.test.ts` (41 tests) pass untouched, which is the claim that the flat path is unaffected.

### Falsification

| reverted | failures |
|---|---|
| group-major ordering → interleave by section | **2** — `rows are group-major…`, `pressing up at one category crosses…` |
| the `z` on the announcement | **4** — the four naming tests |
| the description's group column | **2** — `the table names the group…`, `the group count is reported` |

Disjoint, and each fails only the tests written for it.

### Verification

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass.

```
Test Suites: 299 passed, 299 total     Tests: 4524 passed  (unit)
Test Suites:  10 passed,  10 total     Tests:  137 passed  (esm)
```

Baseline before the change was 298/4508 and 10/137 — the delta is the one new suite and its 16 tests.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,51 @@
+# [4.3.0](https://github.com/xability/maidr/compare/v4.2.0...v4.3.0) (2026-08-16)
+
+
+### Bug Fixes
+
+* **ci:** size the e2e report's code fence so its content cannot close it ([#924](https://github.com/xability/maidr/issues/924)) ([e4afdd9](https://github.com/xability/maidr/commit/e4afdd9c9ffe5f76e604dce8b4416b89cbc36c48)), closes [#715](https://github.com/xability/maidr/issues/715)
+* **docs:** generate the examples gallery so it cannot miss a page ([#914](https://github.com/xability/maidr/issues/914)) ([f5aecad](https://github.com/xability/maidr/commit/f5aecad934944c3f36413e23403b81493b3428f0)), closes [#911](https://github.com/xability/maidr/issues/911)
+* **docs:** give built pages heading ids so in-page links jump ([#918](https://github.com/xability/maidr/issues/918)) ([9ed5a39](https://github.com/xability/maidr/commit/9ed5a396e2e4668c3d00e6d91861995223d73607)), closes [#917](https://github.com/xability/maidr/issues/917) [#913](https://github.com/xability/maidr/issues/913)
+* **docs:** resolve the committed conflict markers and guard against more ([#921](https://github.com/xability/maidr/issues/921)) ([0c04513](https://github.com/xability/maidr/commit/0c04513a25cb96afbff3d15195cbb435716140b1)), closes [#918](https://github.com/xability/maidr/issues/918) [#917](https://github.com/xability/maidr/issues/917)
+* **highlight:** let a flow or network trace publish the ribbon it highlighted ([#906](https://github.com/xability/maidr/issues/906)) ([da131aa](https://github.com/xability/maidr/commit/da131aa76099a2730119e9bdf9c5391f99a72cd9)), closes [#895](https://github.com/xability/maidr/issues/895) [#903](https://github.com/xability/maidr/issues/903) [#899](https://github.com/xability/maidr/issues/899) [#904](https://github.com/xability/maidr/issues/904) [#903](https://github.com/xability/maidr/issues/903) [#904](https://github.com/xability/maidr/issues/904) [#904](https://github.com/xability/maidr/issues/904)
+* **line:** read every path command, so a staircase highlights its samples ([#908](https://github.com/xability/maidr/issues/908)) ([035c5bd](https://github.com/xability/maidr/commit/035c5bdd95d076385d7e26b815fb0e9503811005))
+* **plotly:** count a box's group across the whole box layer ([#883](https://github.com/xability/maidr/issues/883)) ([78375cb](https://github.com/xability/maidr/commit/78375cb4587b1468fb14d4dc5381b0529f172d3a)), closes [#882](https://github.com/xability/maidr/issues/882)
+* **plotly:** scope a candlestick's selector to its own panel and trace ([#882](https://github.com/xability/maidr/issues/882)) ([1cf0470](https://github.com/xability/maidr/commit/1cf0470be356e5d9adcb97d938d0795695367042)), closes [#881](https://github.com/xability/maidr/issues/881)
+* **trace:** answer empty rather than throwing when the cursor has no point ([#909](https://github.com/xability/maidr/issues/909)) ([893d6ea](https://github.com/xability/maidr/commit/893d6ead3d7b6078f4e0bdf798c0b1b537522a48))
+* **trace:** answer for a row that does not exist instead of throwing ([#916](https://github.com/xability/maidr/issues/916)) ([4b6f731](https://github.com/xability/maidr/commit/4b6f7315fca7f2de45d417c95fe4a2a1fe18f449)), closes [#905](https://github.com/xability/maidr/issues/905)
+
+
+### Features
+
+* **adapters:** add the co-located maidr trace declaration ([#886](https://github.com/xability/maidr/issues/886)) ([bbd62b6](https://github.com/xability/maidr/commit/bbd62b690c4e2bd46ce7ae8532303be557abe026))
+* **adapters:** close the declaration-foundation gaps the adapters reported ([#900](https://github.com/xability/maidr/issues/900)) ([ec47b0e](https://github.com/xability/maidr/commit/ec47b0eb12200b8d50e70d964e6aab31f57a93e9)), closes [#896](https://github.com/xability/maidr/issues/896) [#896](https://github.com/xability/maidr/issues/896)
+* **amcharts:** read flow, hierarchy, map and gauge charts ([#903](https://github.com/xability/maidr/issues/903)) ([7cf4c5c](https://github.com/xability/maidr/commit/7cf4c5ce3f64dcedbaf39553f213e29ee4f7385d))
+* **amcharts:** read the chart types deferred out of the coverage PR ([#895](https://github.com/xability/maidr/issues/895)) ([5a39999](https://github.com/xability/maidr/commit/5a39999581a6f033bc5de762a68bcd7d776c97e0))
+* **amcharts:** support the chart types added since the pie chart ([#879](https://github.com/xability/maidr/issues/879)) ([1763eec](https://github.com/xability/maidr/commit/1763eec9667d0fb26bc59de9e8bc9eb8a5545562))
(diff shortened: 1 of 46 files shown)


jooyoungseo:
Good catch — the bug is real, and both codepaths were wrong exactly as you traced them. Fixed in `006569bb`.

## Reproduced first

Your static trace was right, and I turned it into failing assertions before touching anything:

```
✕ the largest estimate is found wherever it is      Expected: 6   Received: 4
✕ the smallest estimate is found wherever it is     Expected: 0   Received: undefined
✕ the label names the group the extreme belongs to  Expected substring: "treated"
                                                    Received string:    "Max value at b"
✕ navigating to an extreme lands on that group row  Expected: 6   Received: 4
```

The root cause is the one you named: `valueRow` is the estimate row *within* a group, and rows are group-major, so anything resolving to "the estimate row" landed in group 0 whatever it had been asked for.

There was a third consequence you didn't list, on the same line: `this.points[highest.index]` indexes the **flattened** list, so even the label's `x` would have come from group 0 for a target found elsewhere.

## The fix

`valueRowOf(group)` is now the single answer to "which row carries this group's estimates", and both paths iterate:

- **`getExtremaTargets`** searches every group's value row and carries `groupIndex` on the target — the field the grammar already had for group-based plots. The label names the group when there is more than one (`Max value at c, treated`), because the groups share their category names and the menu is read without the chart in view. The min/max equality guard now compares *both* coordinates, since two groups can share a column index without being the same reading.
- **`navigateToExtrema`** resolves to `valueRowOf(target.groupIndex ?? 0)`.
- **`findNearestPoint`** searches every group's slice and reports that group's row.

## The pointer case is tested, not just reasoned about

I wanted this one behind a test rather than an argument, so `test/model/errorBarGroupPointer.test.ts` drives the real pointer path (`moveToPointAndGetPointerGuidance`) in jsdom against four whips at known positions.

Getting there turned up something worth writing down, because my first attempt passed for the wrong reason and then failed for a misleading one: **the highlight mapping consults *clones*.** `Svg.selectAllElements` clones the elements it resolves, so a `getBoundingClientRect` stub installed on the elements in the document is never the one asked. Every whip answered jsdom's zeroed rect, they all sat at the origin, and the nearest was always the first — including on the *ungrouped* control case, which my change cannot affect. That the untouched path failed too is what said the harness was wrong rather than the code. The stub now lives on `SVGElement.prototype` and reads an index attribute, which survives cloning.

Falsification, with the fix reverted to `highlightValues[0]`: **2 fail** (`lands on the second group…`, `resolves to the estimate…`), and the two first-group cases keep passing — which is precisely why the flat suite never noticed.

## On your two other notes

**`mapToSvgElements` order is an implicit contract.** Agreed, and it stays one. The length check catches a producer emitting the wrong *number* of whips; nothing can catch the wrong *order*, because there is no independent fact to check it against — the elements carry no group. Worth the flag, and I'd rather it be documented than silently assumed, which it now is.

**The toolchain you couldn't run.** Re-run here after this change:

```
Test Suites: 300 passed, 300 total     Tests: 4534 passed  (unit)
Test Suites:  10 passed,  10 total     Tests:  137 passed  (esm)
lint:fix ✓   type-check ✓   build ✓
```

`errorBar.test.ts` and `forest.test.ts` still pass untouched, including their existing extrema tests — the ungrouped label format is unchanged, since `groupNameAt` returns undefined for a single group.

---

PR #944, opened 2026-08-17
fix(autoplay): play a chart at one tempo whichever way it is read

### LINKED ISSUE #614: fix: different speeds during vertical and horizontal autoplay
# Bug Report

## Description

Unless the speed has been explicitly set with ctrl+comma or ctrl+period, autoplay moves at different speeds vertically and horizontally.
## Steps to Reproduce

1. Open the example for candlesticks or vertical boxplot.
2. Before making any change to speed, press ctrl+shift+right/left arrow for horizontal, and ctrl+shift+up/down arrows for vertical navigation.
3. Compare the difference in speed.

## Actual Behavior

In candlesticks, vertical autoplay is faster. In vertical boxplot, Horizontal autoplay is faster.
## Expected Behavior

In both plots, there should be no difference in speed between horizontal and vertical movements.
## Environment

- Operating System: Windows 11
- Browser: Chrome
- Version: 148

jooyoungseo:
# Pull Request

## Description

Autoplay's rate was `duration / count-in-this-direction`, so the same chart played at two different speeds depending on which way the reader went.

Measured on the two charts the report names, at the default 4000 ms:

| chart | vertical count | horizontal count | vertical step | horizontal step |
|---|---|---|---|---|
| candlestick, 12 samples | 5 sections | 12 samples | **800 ms** | **334 ms** |
| vertical box, 3 boxes | 7 sections | 3 boxes | **572 ms** | **1334 ms** |

The swing reverses between them, which is exactly why the report describes one as faster vertically and the other horizontally — it is the same defect presenting in opposite directions.

Tempo is a cue a reader uses to judge how much chart is left and how fast it is going by. It can carry nothing if it means something different on each axis of the same chart.

## Related Issues

Closes #614

## Changes Made

`getAutoplayRate` now derives one rate for the trace and takes no direction at all — the rate is a property of the chart rather than of the pass being played, which is what makes the two directions agree by construction rather than by arithmetic that happens to match.

**The longest axis, not the shortest.** That choice is the substance:

* the longest keeps `autoplayDuration` naming something real — a full pass along the longest direction takes the configured duration, and a pass along a shorter one finishes sooner *because there is less of it*, which is the honest rendering of a chart wider than it is tall;
* the shortest would make every wide chart crawl: a 12-sample candlestick would take `12 × 800 ms` across, three times the configured duration.

An explicit speed set with <kbd>Ctrl</kbd>+<kbd>,</kbd> / <kbd>Ctrl</kbd>+<kbd>.</kbd> still wins, unchanged — `userSpeed` is checked before any of this, which matches the report's own "unless the speed has been explicitly set".

Not applicable — the change is in pacing.

## Checklist

- [ ] I have updated the documentation, if applicable — the setting's UI label ("Autoplay Duration (ms)") is unchanged and now describes the behaviour more accurately than it did.

### Tests

Two added to `test/service/autoplay.test.ts`, over a new `oblongState(rows, cols)` helper — the existing `traceState` sets all four directions to the same count, so nothing in the file could have caught this.

* **`steps at the same rate whichever way the reader goes`** drives the candlestick's shape (5 × 12) vertically and horizontally and asserts both move on the same tick.
* **`takes its tempo from the longest axis, not whichever is being played`** uses the box plot's shape (7 × 3), where the longest axis is the *vertical* one — so a rate taken from the played direction, or hard-coded to either axis, gives a different answer.

### Falsification

Restoring the original per-direction lookup fails **both** new tests, and the seven existing pacing and starting-point tests keep passing — which is the point: every one of them uses a square state, so the defect was invisible to the whole file.

Worth noting my first falsification attempt was too weak. Pinning the rate to `FORWARD` failed only the second test, because the first test's horizontal count *is* the longest axis and the two answers coincide there. The second test exists precisely because that coincidence is possible.

### Verification

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass.

```
Test Suites: 300 passed, 300 total     Tests: 4566 passed  (unit)
Test Suites:  10 passed,  10 total     Tests:  137 passed  (esm)
```

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,51 @@
+# [4.3.0](https://github.com/xability/maidr/compare/v4.2.0...v4.3.0) (2026-08-16)
+
+
+### Bug Fixes
+
+* **ci:** size the e2e report's code fence so its content cannot close it ([#924](https://github.com/xability/maidr/issues/924)) ([e4afdd9](https://github.com/xability/maidr/commit/e4afdd9c9ffe5f76e604dce8b4416b89cbc36c48)), closes [#715](https://github.com/xability/maidr/issues/715)
+* **docs:** generate the examples gallery so it cannot miss a page ([#914](https://github.com/xability/maidr/issues/914)) ([f5aecad](https://github.com/xability/maidr/commit/f5aecad934944c3f36413e23403b81493b3428f0)), closes [#911](https://github.com/xability/maidr/issues/911)
+* **docs:** give built pages heading ids so in-page links jump ([#918](https://github.com/xability/maidr/issues/918)) ([9ed5a39](https://github.com/xability/maidr/commit/9ed5a396e2e4668c3d00e6d91861995223d73607)), closes [#917](https://github.com/xability/maidr/issues/917) [#913](https://github.com/xability/maidr/issues/913)
+* **docs:** resolve the committed conflict markers and guard against more ([#921](https://github.com/xability/maidr/issues/921)) ([0c04513](https://github.com/xability/maidr/commit/0c04513a25cb96afbff3d15195cbb435716140b1)), closes [#918](https://github.com/xability/maidr/issues/918) [#917](https://github.com/xability/maidr/issues/917)
+* **highlight:** let a flow or network trace publish the ribbon it highlighted ([#906](https://github.com/xability/maidr/issues/906)) ([da131aa](https://github.com/xability/maidr/commit/da131aa76099a2730119e9bdf9c5391f99a72cd9)), closes [#895](https://github.com/xability/maidr/issues/895) [#903](https://github.com/xability/maidr/issues/903) [#899](https://github.com/xability/maidr/issues/899) [#904](https://github.com/xability/maidr/issues/904) [#903](https://github.com/xability/maidr/issues/903) [#904](https://github.com/xability/maidr/issues/904) [#904](https://github.com/xability/maidr/issues/904)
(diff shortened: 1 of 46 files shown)


jooyoungseo:
Addressed in `51fafc3e`, and thank you for the trace through `AbstractTrace.autoplay` and `violin.ts` — that is the check that makes "longest axis" a claim about every trace type rather than about the two in the report.

## The unreachable guard — now says why it is there

You're right that `typeof count === 'number'` cannot fail through any producer today. The comment now says what it is actually guarding, which is the other half of the filter:

> Both halves of the filter guard a value that arrives from parsed chart input rather than from this codebase. Every trace populates all four directions with numbers today, so neither is reachable through `AbstractTrace.autoplay` — but a zero would have divided the duration into `Infinity` and hung the run on its first step, which the previous `duration / pointCount` had nothing to stop.

Worth saying you found that side effect before I wrote it down: the `count > 0` half is a real behavioural improvement, not tidiness. The old expression had no guard, and `Math.ceil(4000 / 0)` is `Infinity`, which `scheduleStep` would hand straight to `setTimeout` — a run that starts and never takes a step, with nothing to indicate why. An empty trace reaching autoplay is the shape that would produce it; the commands gate on `!state.empty`, so it is not reachable *today* either, which is the same reason the type-level guard is not.

## CI

Green before this commit — 15/15, including E2E and the full unit suite. Re-running now for the comment change.

```
Test Suites: 300 passed, 300 total     Tests: 4566 passed  (unit)
Test Suites:  10 passed,  10 total     Tests:  137 passed  (esm)
lint:fix ✓   type-check ✓   build ✓
```

---

Issue #971, opened 2026-08-17
A plotly heatmap navigates upside down: Up moves toward the bottom of the chart


jooyoungseo:
## Summary

`Heatmap`'s constructor reverses both `data.y` and `data.points`:

```ts
this.y = [...data.y].reverse();
this.heatmapValues = [...data.points].reverse();
```

so that internal row 0 is the **bottom** of the drawn grid — the Cartesian
convention its own highlight code names, and the one that makes <kbd>↑</kbd>
(which increments the row index) move visually upward. That means a producer
must emit `y` and `points` **top-first**, in visual reading order.

The plotly adapter emits plotly's `trace.y` and `trace.z` untouched, and
plotly draws `y[0]` at the **bottom**. After the constructor's reversal
internal row 0 is therefore the *top* row, and the whole vertical axis is
inverted: the cursor enters the chart at the top and <kbd>↑</kbd> walks down
it.

Nothing errors, and — importantly — **no value is announced against the wrong
label**. Both arrays are reversed together, so the pairing survives. What is
wrong is the reader's model of *where they are*: someone who navigates up
through a heatmap and reports what "the top row" contains has it exactly
backwards.

## Measured

Which end plotly draws row 0 at, read from a real render (plotly.js 3.7.0 in
Chromium) via `_fullLayout.yaxis.c2p`, where a smaller pixel is higher on
screen:

```
y: ['first', 'second', 'third'],  z: [[1,2], [3,4], [5,6]]

c2p(0) = 266.67   ← 'first'  is LOWER  on screen
c2p(2) =  53.33   ← 'third'  is HIGHER on screen
```

So plotly's `y[0]`/`z[0]` is the **bottom** row.

Driving that layer through the real `TraceFactory` and pressing <kbd>↑</kbd>:

| emitted order | cursor enters at | ↑ moves to | direction |
|---|---|---|---|
| `['first','second','third']` — what the adapter emits | `third` (value 5) — the **top** row | `second` (value 3) | **downward** |
| `['third','second','first']` — top-first | `first` (value 1) — the **bottom** row | `second` (value 3) | upward |

Values track their labels correctly in both (`third`↔5, `second`↔3,
`first`↔1). Only the direction and the entry corner differ.

## Root cause

`HeatmapData` in `src/type/grammar.ts` is:

```ts
/**
 * Data structure for heatmap charts with x/y labels and 2D point values.
 */
export interface HeatmapData {
  x: string[];
  y: string[];
  points: number[][];
}
```

[...]

PR #980, opened 2026-08-17
feat(d3): let a heatmap say which of its rows is the top one

### LINKED ISSUE #978: bindD3Heatmap gives a caller no way to say which row is the top one
Last of the heatmap row-order family — after plotly (#972), Highcharts (#975),
Chart.js (#976) and Vega-Lite (#977) — and the only one that cannot be fixed by
reading the chart, because with d3 there is nothing to read.

## Summary

`HeatmapData` runs top-first, and `Heatmap` reverses it so its own row 0 is the
bottom of the drawn grid. `buildHeatmapLayer` collects its rows in **order of
appearance** in the DOM:

```ts
// Build unique x and y labels (preserving order of appearance)
if (!seenY.has(cell.y)) { seenY.add(cell.y); yLabels.push(cell.y); }
```

For a d3 chart that is the order the author's `.data().join()` iterated, which
has no particular relationship to the order their own y scale draws. A
`scaleBand` whose domain was sorted, or reversed, or taken from a `Map`'s
insertion order, all produce the same DOM and different pictures.

Every value still travels with its own label, so no cell is mis-valued. What is
wrong is the reader's sense of which row sits above which — and, when the two
happen to be opposites, <kbd>↑</kbd> walking down the chart.

## Why this one is different

The other four adapters could each be fixed by consulting something the chart
library owns: plotly's resolved axis range, Highcharts' `yAxis.reversed`,
Chart.js's `scales.y.reverse`, Vega's compiled scale domain. A d3 binder is
handed an `<svg>` and a selector — the scale lives in the author's own closure
and is never passed in. So the binder genuinely **cannot** work the order out,
and no amount of cleverness in the adapter will change that.

That leaves two honest options, and I think both are worth doing:

1. **Let the caller say it.** Add an optional `yOrder?: string[]` to
   `D3HeatmapConfig` — the row labels top-first, which for the common case is
   literally `yScale.domain()` (or its reverse, for a bottom-up band scale).
   One line at the call site, and it makes the chart's own scale the source of
   truth rather than its DOM order.
2. **Document the contract either way.** `bindD3Heatmap`'s docblock is
   otherwise unusually thorough — it explains the render-timing trap at length
   — and says nothing about row order. Someone reading it has no way to learn
   that appearance order is being taken as visual order, or that getting it
   wrong is silent.

`D3HeatmapConfig` currently exposes only `selector`, `x`, `y` and `value`, so
there is nowhere for a caller to put this today even if they knew to.

## Not a claim about any particular chart

Unlike the other four, I have not measured a d3 heatmap being wrong, and I
cannot: whether a given one is right depends entirely on how its author wrote
their join. A chart built with `yScale.domain().forEach(...)` over a
top-down band scale is already correct and would stay correct. The defect is
that nothing tells the author which of those they need, and nothing complains
when they pick the other one.

jooyoungseo:
Closes #978. Last of the heatmap row-order family, after plotly (#972),
Highcharts (#975), Chart.js (#976) and Vega-Lite (#979) — and the only one
that cannot be fixed by reading the chart.

## The bug

The schema orders rows **top-first**, and `Heatmap` turns them over so its own
row 0 is the bottom of the drawn grid — which is what makes <kbd>↑</kbd> move
visually up. `buildHeatmapLayer` took its order from the cells' **order of
appearance in the DOM**: the order the author's `.data().join()` ran in, which
bears no particular relation to the order their scale draws.

## Measured

Through the binder and then the real trace. A join over an ascending band
domain — `['first','second','third']`, which draws `'first'` at the bottom, the
most natural way to write a d3 heatmap:

| join order | cursor enters at | ↑ moves to | |
|---|---|---|---|
| `first, second, third` (bottom-up) | `third` — the **top** row | `second` | walking **down** |
| `third, second, first` (top-down) | `first` — the bottom row | `second` | up |

Nothing errors, and every value is still announced against its own label.

## Why this one needs a config field

The four siblings each had something to consult — plotly's resolved range,
Highcharts' `reversed`, Chart.js's `scales.y.reverse`, Vega's compiled domain.
A d3 binder is handed an `<svg>` and a selector; the scale lives in the
author's own closure and never reaches it. So the caller says:

```ts
bindD3Heatmap(svg, { selector: 'rect.cell', /* … */ yOrder: yScale.domain() });
```

(or its reverse, for a band scale whose domain ascends up the page). The
contract is documented on both `D3HeatmapConfig.yOrder` and `bindD3Heatmap`'s
own docblock, which was otherwise thorough — it explains the render-timing trap
at length — and said nothing about row order.

## The selectors had to move with it

Same cancelling pair as #971, where the two errors agreed with each other and
nothing looked broken. A single selector resolves in **DOM order**, which is
the very join order the rows were taken off, so reordering the rows alone would
announce one cell and outline another — correct navigation, wrong highlight.

The layer now emits **one selector per cell**, laid out in the model's own row
order, via `stampOrderedSelectors` — the helper the gantt, hexbin and
choropleth binders already use for exactly this case, "traces whose payload
order is the binder's own rather than the DOM's". So the DOM-stamping decision
is the repo's existing one, not a new one I introduced.

With per-cell selectors there is nothing left to flatten, so `domMapping` goes
with them, kept only on the fallback path where one selector still has to be
interpreted.

A test pins the property that matters: **row 0's selector resolves to the cell
row 0 announces**.

## Two order-mismatch cases

- naming rows the cells do not carry — a scale's domain outliving a filter is
  ordinary — drops the extras;
- an order **missing** a row the chart draws is declined in favour of
  appearance order, since honouring it would lose a row the reader can see.

## Verification

`npm run lint:fix`, `npm run type-check` (both tsconfigs), `npm run build`, and
`npm test` — 316 suites / 4722 tests, plus 10 / 137 esm — all pass.

9 new tests in `test/adapters/d3/heatmapRowOrder.test.ts`. Falsification, each
half reverted separately:

| reverted | failures |
|---|---|
| `yOrder` ignored | 4 |
| selector rows not laid out in the model's order | 1 — the announce/highlight agreement case |

No existing test changes; all 263 pre-existing d3 tests still pass.

One test deliberately **pins the unfixed default**: a bottom-up join with no
`yOrder` is still read upside down. That is not an oversight — without the
scale there is nothing the binder could consult, so appearance order remains
the only available default, and pinning it makes the limitation visible rather
than latent.

## The family is now complete

| producer | verdict |
|---|---|
| plotly, highcharts, chartjs, vegalite | fixed (#972, #975, #976, #979) |
| d3 binder | fixed here |
| py-maidr plotly | fixed (xability/py-maidr#488) |
| py-maidr matplotlib, r-maidr ggplot2, r-maidr base R | **already correct** — measured |
| tableau | **not established** — see my correction on #971 |
| anychart, amcharts | **not established** |

amCharts remains the likeliest untested instance: its extractor takes
`series.dataItems` order and never consults the axis, and the adapter already
has an `inversed` reader used only for the bump heuristic — the shape Chart.js
had. I could not render amCharts 5 headlessly here (ESM-only, CDN unreachable
through the proxy).

Not verified: no E2E coverage of a d3 heatmap's vertical navigation, and no
check with a screen reader.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,51 @@
+# [4.3.0](https://github.com/xability/maidr/compare/v4.2.0...v4.3.0) (2026-08-16)
+
+
+### Bug Fixes
+
+* **ci:** size the e2e report's code fence so its content cannot close it ([#924](https://github.com/xability/maidr/issues/924)) ([e4afdd9](https://github.com/xability/maidr/commit/e4afdd9c9ffe5f76e604dce8b4416b89cbc36c48)), closes [#715](https://github.com/xability/maidr/issues/715)
+* **docs:** generate the examples gallery so it cannot miss a page ([#914](https://github.com/xability/maidr/issues/914)) ([f5aecad](https://github.com/xability/maidr/commit/f5aecad934944c3f36413e23403b81493b3428f0)), closes [#911](https://github.com/xability/maidr/issues/911)
+* **docs:** give built pages heading ids so in-page links jump ([#918](https://github.com/xability/maidr/issues/918)) ([9ed5a39](https://github.com/xability/maidr/commit/9ed5a396e2e4668c3d00e6d91861995223d73607)), closes [#917](https://github.com/xability/maidr/issues/917) [#913](https://github.com/xability/maidr/issues/913)
+* **docs:** resolve the committed conflict markers and guard against more ([#921](https://github.com/xability/maidr/issues/921)) ([0c04513](https://github.com/xability/maidr/commit/0c04513a25cb96afbff3d15195cbb435716140b1)), closes [#918](https://github.com/xability/maidr/issues/918) [#917](https://github.com/xability/maidr/issues/917)
+* **highlight:** let a flow or network trace publish the ribbon it highlighted ([#906](https://github.com/xability/maidr/issues/906)) ([da131aa](https://github.com/xability/maidr/commit/da131aa76099a2730119e9bdf9c5391f99a72cd9)), closes [#895](https://github.com/xability/maidr/issues/895) [#903](https://github.com/xability/maidr/issues/903) [#899](https://github.com/xability/maidr/issues/899) [#904](https://github.com/xability/maidr/issues/904) [#903](https://github.com/xability/maidr/issues/903) [#904](https://github.com/xability/maidr/issues/904) [#904](https://github.com/xability/maidr/issues/904)
+* **line:** read every path command, so a staircase highlights its samples ([#908](https://github.com/xability/maidr/issues/908)) ([035c5bd](https://github.com/xability/maidr/commit/035c5bdd95d076385d7e26b815fb0e9503811005))
+* **plotly:** count a box's group across the whole box layer ([#883](https://github.com/xability/maidr/issues/883)) ([78375cb](https://github.com/xability/maidr/commit/78375cb4587b1468fb14d4dc5381b0529f172d3a)), closes [#882](https://github.com/xability/maidr/issues/882)
+* **plotly:** scope a candlestick's selector to its own panel and trace ([#882](https://github.com/xability/maidr/issues/882)) ([1cf0470](https://github.com/xability/maidr/commit/1cf0470be356e5d9adcb97d938d0795695367042)), closes [#881](https://github.com/xability/maidr/issues/881)
+* **trace:** answer empty rather than throwing when the cursor has no point ([#909](https://github.com/xability/maidr/issues/909)) ([893d6ea](https://github.com/xability/maidr/commit/893d6ead3d7b6078f4e0bdf798c0b1b537522a48))
+* **trace:** answer for a row that does not exist instead of throwing ([#916](https://github.com/xability/maidr/issues/916)) ([4b6f731](https://github.com/xability/maidr/commit/4b6f7315fca7f2de45d417c95fe4a2a1fe18f449)), closes [#905](https://github.com/xability/maidr/issues/905)
+
+
+### Features
+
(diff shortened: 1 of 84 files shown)


jooyoungseo:
Finding #1 is right, and I've fixed it in 685d3e9 rather than leaving it — the triggering input is unlikely, but the failure mode is a row silently vanishing from the chart, which is the same class of quiet wrongness this whole PR family is about.

The hole was counting mentions rather than distinct rows. `yOrder: ['a', 'a']` over a grid of `a` and `b` filters to `['a', 'a']`, whose length is 2, which matched — so the layer went out with `a` emitted twice and `b` gone. The filter now drops repeats as well as unknowns, so the length comparison really does mean "names every row exactly once":

```ts
for (const row of declared) {
  if (present.has(row) && !taken.has(row)) {
    taken.add(row);
    ordered.push(row);
  }
}
return ordered.length === appearance.length ? ordered : appearance;
```

Two tests, both falsified against the old filter (each fails without the dedupe):

- `['third', 'first', 'first']` over three rows is **declined** — it names three but covers two.
- `['third', 'second', 'third', 'first']` is **used** — the stray repeat does not stop it naming every row, so the order is still recoverable and discarding it would lose good information.

That second case is the one I would have got wrong by simply rejecting anything with a duplicate.

On finding #2 — agreed, and I left it inline for the same reason you give. If a second binder ever needs the schema-order-to-model-row-order flip, that is the moment to extract it.

On the note about not being able to run the suite: `npm run lint:fix`, `npm run type-check` (both tsconfigs), `npm run build` and `npm test` all pass locally — **316 suites / 4724 tests**, plus 10 / 137 esm — and CI is running the same set on this push.

---

Issue #981, opened 2026-08-17
An amCharts heatmap navigates upside down


jooyoungseo:
Fifth and last measurable instance of #971, after plotly (#972), Highcharts
(#975), Chart.js (#976) and Vega-Lite (#979). I listed this as "not
established" on those PRs because I could not render amCharts headlessly;
bundling it with esbuild settled it.

## Summary

`HeatmapData` runs top-first and `Heatmap` turns it over so its own row 0 is
the bottom of the drawn grid, which is what makes <kbd>↑</kbd> move visually
up. `extractHeatmapData` takes its rows from `series.dataItems` order and never
consults the axis:

```ts
if (!yIndex.has(y)) {
  yIndex.set(y, yLabels.length);
  yLabels.push(y);
}
```

amCharts draws a `CategoryAxis`'s first category at the **bottom**, so the layer
is emitted bottom-first and the chart is navigated upside down.

## Measured

**The library.** amCharts 5 bundled to an IIFE with esbuild and rendered in
Chromium, reading each drawn column's own vertical position (amCharts' y grows
downward, so a larger number is lower on screen), for
`categories: ['first','second','third']` with `renderer.inversed` at its
default:

| row | column centre y | |
|---|---|---|
| `first` (category index 0) | 437.8 | **bottom** |
| `second` | 289.8 | middle |
| `third` | 141.8 | top |

`yRenderer.get('inversed')` reads `false`.

An earlier attempt via `yAxis.getDataItemCoordinateY` returned `null` for every
row, and the comparison `null < null` produced a plausible-looking "BOTTOM"
that was not a measurement at all. The numbers above come from the columns'
own geometry.

**The adapter.** Through the real `extractHeatmapData`, with data items in the
order amCharts lays them out:

```
emitted y = ["first","second","third"]      ← bottom-first
points    = [[1,2],[3,4],[5,6]]
```

So `first` — the row drawn at the bottom — leads the payload, and after the
core's reversal the cursor enters on `third`, the row at the top.

## The fix is one-sided here

`buildHeatmapLayer` emits **no `selectors`** — amCharts paints to canvas — so
there is no highlight index keyed to the row order, and reversing the payload
cannot desynchronise anything. Unlike plotly (#972) and Highcharts (#975),
which each needed a matching second change, this is the Chart.js case: reverse
the rows and stop.

```ts
const inversed = /* renderer.inversed on the series' y axis */;
// amCharts counts a category axis from the bottom, so turn the rows over
[...]

Issue #977, opened 2026-08-17
A Vega-Lite heatmap's rows are ordered by the data, not by the axis


jooyoungseo:
Fourth in the family after plotly (#972), Highcharts (#975) and Chart.js (#976)
— but a different failure mode. Vega-Lite draws its first domain value at the
top, so there is no inversion to undo. What goes wrong is that the adapter
never asks the axis what its order *is*.

## Summary

`extractHeatmapData` builds both axes from the order the data rows happen to
arrive in:

```ts
for (const row of rows) {
  xLabelsSet.add(String(row[xField] ?? ''));
  yLabelsSet.add(String(row[yField] ?? ''));
}
const yLabels = [...yLabelsSet];
```

Vega-Lite sorts a nominal domain ascending by default, so unless the author
listed their rows already sorted, the emitted axis is in a different order from
the drawn one. Every value still travels with its own label, so no cell is
mis-valued — but the reader's picture of which row sits above which is simply
not the chart's.

## Measured

Real Vega-Lite compiled and run through Vega, with rows deliberately listed out
of order:

```js
// listed:  zebra, apple, mango
view.scale('y').domain()   // ['apple', 'mango', 'zebra']   ← sorted
y('apple') = 0     ← top
y('mango') = 100
y('zebra') = 200   ← bottom
```

So the drawn order top-to-bottom is `apple, mango, zebra`, and `domain[0]` is
the **top** row — which is exactly the order `HeatmapData` wants, no reversal
needed.

What the adapter emits for that chart, through the real `vegaLiteToMaidr`:

```
type=heat  y=["zebra","apple","mango"]  points=[[1,2],[3,4],[5,6]]
```

The core then reverses, putting model row 0 on `'mango'` — while the row
actually drawn at the bottom is `'zebra'`.

## Proposed fix

The adapter already has the right tool and does not point it at this layer.
`readScaleDomain(view, name, markGroupPrefix)` exists precisely to "align data
ordering with the SVG DOM order", handling Vega's scoped scale names for
composite children and Vega's habit of throwing on an unknown scale name.

So: order the heatmap's rows (and columns) by `readScaleDomain(view, 'y', …)`
when a view was supplied, falling back to today's data-encounter order when it
was not. `vegaLiteToMaidr(spec, view?, options?)` already takes the view; it is
simply not threaded down to `extractHeatmapData`.

Using the view rather than sorting in the adapter matters because the domain
order is not always "sorted ascending": `sort` on the encoding can name another
[...]

Issue #974, opened 2026-08-17
A Chart.js matrix chart navigates upside down, and takes its row order from the data rather than the axis


jooyoungseo:
Third instance of #971, after plotly (#972) and Highcharts (#973). Same
inversion, plus a second problem specific to this adapter — and, unusually, a
one-sided fix.

## Summary

`HeatmapData` runs top-first and `Heatmap` turns it over so its own row 0 is
the bottom of the drawn grid. `buildHeatmapLayers` derives `yLabels` from the
**order the data points happen to be listed in**:

```ts
if (!ySet.has(y)) {
  ySet.add(y);
  yLabels.push(y);
}
...
const points: number[][] = yLabels.map(y => xLabels.map(x => valueMap.get(...) ?? 0));
```

Two things follow. Chart.js's matrix controller defaults the y category scale
to `reverse: true`, so data listed in scale order comes out **bottom-first**
and the chart is navigated upside down — the cursor enters at the top row and
<kbd>↑</kbd> walks down. And because the order comes from the data rather than
the scale, a chart whose points are listed in any other order gets a y axis in
that order instead of the drawn one.

The adapter never reads `scales.y.reverse`. Its only mentions of `reverse` are
the bump-chart heuristic.

## Measured

Real Chart.js 4.5.1 with `chartjs-chart-matrix`, in Chromium, reading
`chart.scales.y` after a non-responsive layout (scale spans y 32..122, so a
larger pixel is lower on screen). Data listed `['first','second','third']`:

| `scales.y.reverse` | resolved | `px('first')` | `px('third')` | `'first'` drawn at |
|---|---|---|---|---|
| not declared | **`true`** | 121.6 | 32.0 | **bottom** |
| `false` | `false` | 32.0 | 121.6 | top |

So the matrix chart's own default puts the first-listed row at the bottom.

## The highlight needs no change here

Unlike plotly (#971, index-keyed overlay rects) and Highcharts (#973,
index-keyed selectors), this adapter's highlight looks the cell up **by
label**, and derives that label from `hd.y` itself:

```ts
const originalYi = (hd.y.length - 1) - row;
const yLabel = hd.y[originalYi];
const flatIndex = maps.heatmapIndices.get(layer.id)?.get(`${xLabel}\0${yLabel}`);
```

Reversing `y` and `points` together leaves the resolved label unchanged, so
the lookup still lands on the same cell. This one is a one-sided fix.

## Proposed fix

Take the row order from the **scale**, not from the data, and honour
[...]

PR #984, opened 2026-08-17
test(e2e): pin that a heatmap climbs on Up

### LINKED ISSUE #983: Heatmap e2e spec never makes a vertical move, so six row-order bugs passed it
## What

`e2e_tests/specs/heatmap.spec.ts` covers loading, the four modes, axis titles, the menus, autoplay speed, and left/right movement. It never presses Up or Down.

That is the one direction the row-order family of bugs got wrong.

## Why it matters

Six adapters emitted a heatmap's rows bottom-first when the grammar asks for top-first — #971 (plotly), #973 (Highcharts), #974 (Chart.js), #977 (Vega-Lite), #978 (d3), #981 (amCharts). Every one of them was invisible to this suite, and would still be:

- the chart loaded
- every mode worked
- the highlight tracked the cursor
- every value was announced against its own label

Only the *direction* was false. Up walked down the grid. There is no way to catch that except by pressing the key and looking at where you land, which is exactly what an e2e suite is for and exactly what this one did not do.

`Heatmap`'s constructor is the contract in question:

```ts
this.y = [...data.y].reverse();
this.heatmapValues = [...data.points].reverse();
```

Internal row 0 is the **bottom** of the drawn grid, so Up increments the row index. Nothing pins that from the outside.

## Proposed

Two cases in the existing `Navigation Controls` block:

1. entry lands on the row drawn at the bottom — the payload's **last** `y` — and Up climbs to the one above it;
2. Down comes back.

Both should read the expected label out of `heatmapData.y`'s own order rather than hard-coding the example's rows, so they keep their meaning if the example changes, and both should skip on a single-row chart, which has no direction to get wrong.

The test is only worth adding if it fails when the behaviour is wrong, so it should be falsified by dropping the reversal above — that is what an inverted heatmap looks like from the outside — and the rest of the spec should stay green while it does.

jooyoungseo:
Closes #983.

## The gap

Six fixes across the row-order family — #971 (plotly), #973 (Highcharts), #974 (Chart.js), #977 (Vega-Lite), #978 (d3), #981 (amCharts) — each shipped with the same line under "not verified": no end-to-end coverage of a heatmap's vertical navigation.

The spec had none. It covers loading, the four modes, axis titles, the menus, autoplay speed, and left/right movement, and never presses Up or Down. So every one of those six bugs could have walked straight through it, and would still.

They were exactly the kind this suite exists for. The chart loaded, it navigated, it highlighted, and it announced every value against its own label. Only the direction was false — Up walked down the grid. Nothing but driving the keys catches that.

## The change

Two cases in the existing `Navigation Controls` block, +33 lines, no source changes:

- **enters at the bottom row and climbs on Up** — entry lands on the payload's last `y`, which is the row drawn at the bottom, and Up moves to the one above it;
- **comes back down on Down**.

Both read the expected label out of `heatmapData.y`'s own order rather than hard-coding the example's rows, so they keep their meaning if the example changes. Both `test.skip` on a single-row chart, which has no direction to get wrong.

The contract they pin is `Heatmap`'s constructor:

```ts
this.y = [...data.y].reverse();
this.heatmapValues = [...data.points].reverse();
```

Internal row 0 is the bottom of the drawn grid, so Up increments the row index. That is what makes top-first input the right way round, and it was unpinned from the outside until now.

## Verification

```
npx playwright test e2e_tests/specs/heatmap.spec.ts --project=chromium   → 23 passed
npm run lint:fix                                                        → clean
npm run type-check                                                      → clean (both tsconfigs)
```

**Falsified.** Removed the two `.reverse()` calls from `Heatmap`'s constructor — which is what an inverted heatmap looks like from the outside — rebuilt, and re-ran: exactly these 2 failed and the other 21 passed. Restored the file, rebuilt, re-ran: 23 passed again.

A test that cannot fail is worse than no test, and on a suite this size it is easy to add one by accident, so the falsification is the part of this worth trusting.

## Not verified

- Only the chromium project was run locally. The full `npm run e2e` matrix runs in CI.
- This pins the direction on the bundled heatmap example. It does not exercise any adapter — the six fixes above have their own unit tests for that. What it adds is the last link none of them cover: that a correctly-ordered payload is actually *navigated* the right way up.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,51 @@
+# [4.3.0](https://github.com/xability/maidr/compare/v4.2.0...v4.3.0) (2026-08-16)
+
+
+### Bug Fixes
+
+* **ci:** size the e2e report's code fence so its content cannot close it ([#924](https://github.com/xability/maidr/issues/924)) ([e4afdd9](https://github.com/xability/maidr/commit/e4afdd9c9ffe5f76e604dce8b4416b89cbc36c48)), closes [#715](https://github.com/xability/maidr/issues/715)
+* **docs:** generate the examples gallery so it cannot miss a page ([#914](https://github.com/xability/maidr/issues/914)) ([f5aecad](https://github.com/xability/maidr/commit/f5aecad934944c3f36413e23403b81493b3428f0)), closes [#911](https://github.com/xability/maidr/issues/911)
+* **docs:** give built pages heading ids so in-page links jump ([#918](https://github.com/xability/maidr/issues/918)) ([9ed5a39](https://github.com/xability/maidr/commit/9ed5a396e2e4668c3d00e6d91861995223d73607)), closes [#917](https://github.com/xability/maidr/issues/917) [#913](https://github.com/xability/maidr/issues/913)
+* **docs:** resolve the committed conflict markers and guard against more ([#921](https://github.com/xability/maidr/issues/921)) ([0c04513](https://github.com/xability/maidr/commit/0c04513a25cb96afbff3d15195cbb435716140b1)), closes [#918](https://github.com/xability/maidr/issues/918) [#917](https://github.com/xability/maidr/issues/917)
+* **highlight:** let a flow or network trace publish the ribbon it highlighted ([#906](https://github.com/xability/maidr/issues/906)) ([da131aa](https://github.com/xability/maidr/commit/da131aa76099a2730119e9bdf9c5391f99a72cd9)), closes [#895](https://github.com/xability/maidr/issues/895) [#903](https://github.com/xability/maidr/issues/903) [#899](https://github.com/xability/maidr/issues/899) [#904](https://github.com/xability/maidr/issues/904) [#903](https://github.com/xability/maidr/issues/903) [#904](https://github.com/xability/maidr/issues/904) [#904](https://github.com/xability/maidr/issues/904)
+* **line:** read every path command, so a staircase highlights its samples ([#908](https://github.com/xability/maidr/issues/908)) ([035c5bd](https://github.com/xability/maidr/commit/035c5bdd95d076385d7e26b815fb0e9503811005))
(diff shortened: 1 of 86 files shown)


jooyoungseo:
Both taken, in `5d0ecc4`. The nit turned out to be sitting next to a real hole.

**The comment.** Trimmed to the invariant and dropped the issue roll call — agreed it belongs in the history, not the file. It also said *five* adapters where six were fixed, which I hadn't noticed until I went to edit it.

**The collision check, and where it doesn't hold.** You checked the current fixture and were right about it: `T5` and `RoBERTa` are unambiguous. But the assertion is weaker than that check suggests, in one arrangement it does not rule out.

`toContain` on a row name passes while standing on a row whose name *extends* it. That doesn't bite on the pair the example happens to end with, but the same `y` contains `BiLSTM`, `BiLSTM+Attn`, `BiLSTM+CoVe` and `BiLSTM+ELMo`. Reorder the example so the bottom two rows are `BiLSTM+Attn` (bottom) and `BiLSTM` (above), and:

- entry asserts `toContain('BiLSTM+Attn')` — specific, still sound;
- after Up asserts `toContain('BiLSTM')` — **passes without the cursor having moved**, since the announcement it is still sitting on says `BiLSTM+Attn`.

That is the one assertion the test exists for, so it would go vacuous silently. Not hypothetical for a fixture already four rows deep in a shared prefix.

Fixed by requiring the announcement to have changed:

```ts
expect(above).not.toBe(entry);
expect(above).toContain(rows[rows.length - 2].toString());
```

and Down now asserts it lands back on *exactly* the announcement it started from (`toBe(entry)`), with its own `not.toBe` on the intermediate step so it stands up without leaning on the first test. I used the announcement rather than the cell's value from `points` deliberately — a value assertion would have to match how the number is rendered (the announcement reads `Score is 71.6%`), which is a second thing able to break for an unrelated reason.

Re-falsified with the stronger assertions: dropping the reversal in `Heatmap`'s constructor still fails exactly these two, and the other 21 pass. Locally after the change: 23 passed, lint and type-check clean.

---

PR #988, opened 2026-08-17
fix(plotly): read a bar chart in the order plotly draws it

### LINKED ISSUE #987: A plotly bar chart with categoryorder is read and panned in the data's order
## What

`extractBarLayer` walks the trace's own arrays:

```ts
for (let i = 0; i < len; i++) {
  data.push(barPoint(calcdata[i], x[i], y[i], isHorizontal));
}
```

`categoryorder` re-sorts the drawn axis and leaves `trace.x` / `trace.y` alone, so the bars are emitted in an order the chart does not show.

Sibling of #985, which fixed the heatmap. **The failure is not the same shape, and neither is the fix** — see the trap below before touching this.

## Measured

plotly.js in Chromium. `x = ['charlie','alpha','bravo']`, `y = [3,1,2]`. Bar heights read from each rendered `<path>`'s own bounding box, and their drawn position from its centre.

| layout | `_categories` | `calcdata` `p` | heights in DOM order | heights left→right |
|---|---|---|---|---|
| *(none)* | `charlie, alpha, bravo` | `0, 1, 2` | `304, 101, 203` | `304, 101, 203` |
| `categoryorder: 'category ascending'` | `alpha, bravo, charlie` | `2, 0, 1` | `304, 101, 203` | **`101, 203, 304`** |
| `categoryorder: 'total descending'` | `charlie, bravo, alpha` | `0, 2, 1` | `304, 101, 203` | **`304, 203, 101`** |
| `autorange: 'reversed'` | `charlie, alpha, bravo` | `0, 1, 2` | `304, 101, 203` | **`203, 101, 304`** |

Two things to take from this:

1. **`calcdata[i].p` is already the resolved axis position.** The drawn order is in hand at the call site — no `_categories` lookup needed, unlike the heatmap.
2. **The DOM order is the trace's order, not the drawn order.** In the sorted rows the rendered `<path>` elements still come out `304, 101, 203`. Plotly renders bars trace-by-trace and positions them by transform.

## What a reader gets

Every label keeps its own value, and — because the DOM order matches the data order — the highlight lands on the right bar too. What is wrong is everything that treats the index as a **position**:

- arrowing right walks `charlie(3), alpha(1), bravo(2)` while the chart reads `alpha(1), bravo(2), charlie(3)`;
- `BarTrace.audio` pans by `this.col` (`src/model/bar.ts:179`), so `charlie` plays hard **left** while it is drawn hard **right**;
- `BarTrace.braille` renders `barValues` in index order, so the braille line — a spatial medium — shows a chart that is not on screen;
- autoplay sweeps in that same false order.

So a reader comes away with a confident and wrong picture of the chart's shape: which bar is tallest is right, *where* it is is not.

## The trap

**A data-order fix alone would make this worse, not better.** Today three things agree — data order, DOM order, and therefore the highlight. Reordering `data` without reordering `selectors` would announce `alpha` while outlining `charlie`, turning a correct highlight into a wrong one.

This is the third time in this family that the two halves have to move together (#971 and #978 were the others), and the first where the *current* state is the consistent one.

## Proposed

Both halves, in one change:

1. Order the points by `calcdata[i].p`, and reverse when the category axis' resolved range descends (a reversed axis is not in `p`, exactly as it is not in `_categories` — same finding as #985).
2. Emit `selectors` as a **`string[]`, one per bar, permuted the same way**, instead of the single `.trace.bars .point > path` string that resolves in DOM order.

Measured feasible: `.trace.bars .point:nth-child(k) > path` addresses one bar each — `g.points` holds only `g.point` children, and a bar's text label lives inside its own `.point` group rather than beside it, so the count stays clean. Returns heights `285, 95, 190` for k = 1, 2, 3, matching the DOM order exactly.

**Emit the array only when the order actually differs from the trace's.** With no `categoryorder` and no reversal the single selector should stay exactly as it is — there is no reason to take on `nth-child`'s brittleness for the overwhelmingly common chart that does not need it.

## Scope

Bar-family types share `extractBarLayer` and the same `.trace.bars .point > path` selector — `BAR`, `DODGED`, `STACKED`, `NORMALIZED`, `DIVERGING`, `MOSAIC`, `HISTOGRAM`, `FUNNEL`, `DOT`. Only what is measured should be claimed: this was measured on a plain vertical `bar`. Whether a stacked or dodged layer's DOM order survives the same way needs its own measurement before the fix is extended to them, since those draw more than one `.point` per category.

Needs a test that would have failed on both halves: the emitted order under `categoryorder`, and a selector that resolves to the bar whose value is announced.

jooyoungseo:
Closes #987. Sibling of #986, which fixed the heatmap — **the failure is not the same shape, and neither is the fix.**

## The bug

`extractBarLayer` walked the trace's own arrays. `categoryorder` sorts the category axis and leaves `x` and `y` alone, so the bars were emitted in an order the chart does not show.

Every label kept its own value, and — because plotly renders bars in the trace's order too — the highlight kept landing on the right bar. What was wrong was everything that reads the index as a **position**:

- arrowing right walked `charlie(3), alpha(1), bravo(2)` while the chart read `alpha(1), bravo(2), charlie(3)`;
- `BarTrace.audio` pans by `this.col` (`src/model/bar.ts:179`), so a bar drawn hard **right** played hard **left**;
- `BarTrace.braille` renders `barValues` in index order, so the braille line — a spatial medium — showed a chart that is not on screen;
- autoplay swept in that same false order.

A reader came away sure of which bar was tallest and wrong about where it was.

## The trap, and why both halves move together

That agreement between data order, DOM order and highlight is exactly what makes this dangerous. **Reordering the points alone would have turned a correct highlight into a wrong one** — announcing `alpha` while outlining `charlie`. This is the third time in this family the two halves have to move together (#971 and #978 were the others), and the first where the *current* state is the consistent one.

So the change does both or neither:

- points ordered by `calcdata[i].p`, the resolved axis position, then reversed when the category axis' range descends;
- the layer's one selector narrowed to a per-bar `nth-child` list, permuted the same way.

If the selector is not the shape this knows how to narrow, the **whole** reordering is declined rather than half of it.

## Measured

plotly.js in Chromium. `x = ['charlie','alpha','bravo']`, `y = [3,1,2]`, heights from each rendered path's own bounding box.

| layout | `calcdata` `p` | heights in DOM order | heights left→right |
|---|---|---|---|
| *(none)* | `0, 1, 2` | `304, 101, 203` | `304, 101, 203` |
| `categoryorder: 'category ascending'` | `2, 0, 1` | `304, 101, 203` | **`101, 203, 304`** |
| `categoryorder: 'total descending'` | `0, 2, 1` | `304, 101, 203` | **`304, 203, 101`** |
| `autorange: 'reversed'` | `0, 1, 2` | `304, 101, 203` | **`203, 101, 304`** |

`calcdata[i].p` is already the resolved position, so the drawn order is free here — no category list to consult, unlike the heatmap. The DOM order stays the trace's, which is the fact the selector half rests on. A reversed axis is not in `p` any more than it is in `_categories`, so it is still asked separately.

Selector feasibility was measured too, not assumed: a bar's `.point` group is one of `g.points`' only children, and its text label sits **inside** that group rather than beside it, so counting `.point` siblings names exactly one bar.

## Verification

```
npx jest --selectProjects unit   → 4756 passed, 319 suites (12 new)
npm run lint:fix / type-check / build → clean
```

**End-to-end against a real render.** The unit tests drive a hand-built `gd`, so I also ran the shipped logic against actual plotly output in Chromium and resolved each emitted selector:

```
=== categoryAscending    order [1,2,0]
  slot 0: announces alpha=1    → nth-child(2)  resolves to height 101 at x=90
  slot 1: announces bravo=2    → nth-child(3)  resolves to height 203 at x=270
  slot 2: announces charlie=3  → nth-child(1)  resolves to height 304 at x=450
  bars actually drawn, left to right: 101@90, 203@270, 304@450

=== reversed             order [2,1,0]
  slot 0: announces bravo=2    → nth-child(3)  resolves to height 203 at x=90
  slot 1: announces alpha=1    → nth-child(2)  resolves to height 101 at x=270
  slot 2: announces charlie=3  → nth-child(1)  resolves to height 304 at x=450
  bars actually drawn, left to right: 203@90, 101@270, 304@450
```

In every case the announced order matches the bars' own left-to-right positions, and each selector resolves to the bar being announced.

**Falsified twice**, which matters more than usual here:

1. Reverting the reordering entirely → **8 of the 12 new cases fail.**
2. Reordering the points while leaving the selector alone — the half-fix that would break the highlight → **exactly the 4 selector cases fail and the 4 data cases pass.** That split is the evidence the suite catches the trap rather than just the bug.

## Not verified

- **The single selector is kept whenever nothing has moved** — deliberately. There is no reason to take on `nth-child`'s brittleness for the overwhelmingly common chart that does not need it, and a test pins that.
- **Only a plain `bar` is reordered.** `extractBarLayer` is shared with `FUNNEL` and `DOT`, and the bar-family selector is shared more widely still (`DODGED`, `STACKED`, `NORMALIZED`, `DIVERGING`, `MOSAIC`, `HISTOGRAM`). Only a plain vertical bar was measured; those draw more than one `.point` per category or into different elements, so the selector arithmetic needs its own measurement before it can be trusted on them. Recorded in #987 rather than extended on a guess.
- A horizontal bar takes its order from the y axis, which follows from `orientation` and is covered by the code path but was not separately measured in a browser.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,51 @@
+# [4.3.0](https://github.com/xability/maidr/compare/v4.2.0...v4.3.0) (2026-08-16)
+
+
+### Bug Fixes
+
+* **ci:** size the e2e report's code fence so its content cannot close it ([#924](https://github.com/xability/maidr/issues/924)) ([e4afdd9](https://github.com/xability/maidr/commit/e4afdd9c9ffe5f76e604dce8b4416b89cbc36c48)), closes [#715](https://github.com/xability/maidr/issues/715)
+* **docs:** generate the examples gallery so it cannot miss a page ([#914](https://github.com/xability/maidr/issues/914)) ([f5aecad](https://github.com/xability/maidr/commit/f5aecad934944c3f36413e23403b81493b3428f0)), closes [#911](https://github.com/xability/maidr/issues/911)
+* **docs:** give built pages heading ids so in-page links jump ([#918](https://github.com/xability/maidr/issues/918)) ([9ed5a39](https://github.com/xability/maidr/commit/9ed5a396e2e4668c3d00e6d91861995223d73607)), closes [#917](https://github.com/xability/maidr/issues/917) [#913](https://github.com/xability/maidr/issues/913)
+* **docs:** resolve the committed conflict markers and guard against more ([#921](https://github.com/xability/maidr/issues/921)) ([0c04513](https://github.com/xability/maidr/commit/0c04513a25cb96afbff3d15195cbb435716140b1)), closes [#918](https://github.com/xability/maidr/issues/918) [#917](https://github.com/xability/maidr/issues/917)
+* **highlight:** let a flow or network trace publish the ribbon it highlighted ([#906](https://github.com/xability/maidr/issues/906)) ([da131aa](https://github.com/xability/maidr/commit/da131aa76099a2730119e9bdf9c5391f99a72cd9)), closes [#895](https://github.com/xability/maidr/issues/895) [#903](https://github.com/xability/maidr/issues/903) [#899](https://github.com/xability/maidr/issues/899) [#904](https://github.com/xability/maidr/issues/904) [#903](https://github.com/xability/maidr/issues/903) [#904](https://github.com/xability/maidr/issues/904) [#904](https://github.com/xability/maidr/issues/904)
+* **line:** read every path command, so a staircase highlights its samples ([#908](https://github.com/xability/maidr/issues/908)) ([035c5bd](https://github.com/xability/maidr/commit/035c5bdd95d076385d7e26b815fb0e9503811005))
+* **plotly:** count a box's group across the whole box layer ([#883](https://github.com/xability/maidr/issues/883)) ([78375cb](https://github.com/xability/maidr/commit/78375cb4587b1468fb14d4dc5381b0529f172d3a)), closes [#882](https://github.com/xability/maidr/issues/882)
+* **plotly:** scope a candlestick's selector to its own panel and trace ([#882](https://github.com/xability/maidr/issues/882)) ([1cf0470](https://github.com/xability/maidr/commit/1cf0470be356e5d9adcb97d938d0795695367042)), closes [#881](https://github.com/xability/maidr/issues/881)
+* **trace:** answer empty rather than throwing when the cursor has no point ([#909](https://github.com/xability/maidr/issues/909)) ([893d6ea](https://github.com/xability/maidr/commit/893d6ead3d7b6078f4e0bdf798c0b1b537522a48))
+* **trace:** answer for a row that does not exist instead of throwing ([#916](https://github.com/xability/maidr/issues/916)) ([4b6f731](https://github.com/xability/maidr/commit/4b6f7315fca7f2de45d417c95fe4a2a1fe18f449)), closes [#905](https://github.com/xability/maidr/issues/905)
+
+
+### Features
+
+* **adapters:** add the co-located maidr trace declaration ([#886](https://github.com/xability/maidr/issues/886)) ([bbd62b6](https://github.com/xability/maidr/commit/bbd62b690c4e2bd46ce7ae8532303be557abe026))
+* **adapters:** close the declaration-foundation gaps the adapters reported ([#900](https://github.com/xability/maidr/issues/900)) ([ec47b0e](https://github.com/xability/maidr/commit/ec47b0eb12200b8d50e70d964e6aab31f57a93e9)), closes [#896](https://github.com/xability/maidr/issues/896) [#896](https://github.com/xability/maidr/issues/896)
+* **amcharts:** read flow, hierarchy, map and gauge charts ([#903](https://github.com/xability/maidr/issues/903)) ([7cf4c5c](https://github.com/xability/maidr/commit/7cf4c5ce3f64dcedbaf39553f213e29ee4f7385d))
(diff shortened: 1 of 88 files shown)


jooyoungseo:
Fixed in `663289b` — the docstring is back above `extractBarLayer`. You were right that it was stranded: the two new functions went in between the comment and the function it describes, so it ended up documenting `drawnBarOrder` and `extractBarLayer` lost its own.

On the horizontal-bar gap you noted: agreed it is a follow-up rather than a blocker, and it is now recorded on #987 alongside the rest of the bar-family scope, which I have since measured. Short version, in case it saves anyone the browser time:

- **dodged / stacked / normalized** — same bug (`p` comes back `[2,0,1]`, DOM order stays the trace's), but the fix does not port directly: each trace gets its own `.trace.bars` group and the shared selector spans all of them, so an `nth-child` would name the nth bar *of every group*. Scoping to one group first is a step this PR did not need.
- **histogram** — **no bug.** Its calcdata is built after binning, so `p` is already `[0,1,2]` and the DOM order matches the drawn order.
- **funnel** — draws into its own layer, zero `.trace.bars` groups, so `nthBarSelector` declines it outright, which is what the `TraceType.BAR` gate already does.

Thanks for tracing `BAR_SELECTOR_TAIL` against `selectors.ts:70` rather than taking the `endsWith` guard on trust — that is the assumption the whole selector half rests on, and it is the one that would fail silently if the generated selector ever changed shape.

---

Issue #989, opened 2026-08-17
A dodged or stacked plotly bar layer is read in the data's order, not the drawn one


jooyoungseo:
The half of #987 that #988 deliberately left out, now measured well enough to implement.

## What

`extractSegmentedBarLayer` builds each series by walking the trace's own arrays:

```ts
for (let i = 0; i < len; i++) {
  const point: SegmentedPoint = { ...barPoint(cd[i], x[i], y[i], isHorizontal), z };
  ...
}
```

`categoryorder` sorts the category axis and leaves `x`, `y` and `z` alone, exactly as it does for a plain bar. So every series is emitted in an order the chart does not show.

## Measured

plotly.js in Chromium, two traces over `x = ['charlie','alpha','bravo']` with `categoryorder: 'category ascending'`:

```
dodged   trace groups in DOM: 2   calcdata p per trace: [[2,0,1],[2,0,1]]
  group 0  bars in DOM order: h=30@491, h=10@64, h=20@277
  group 1  bars in DOM order: h=304@576, h=101@149, h=203@363

stacked  trace groups in DOM: 2   calcdata p per trace: [[2,0,1],[2,0,1]]
  group 0  bars in DOM order: h=28@533, h=9@107, h=18@320
  group 1  bars in DOM order: h=276@533, h=92@107, h=184@320
```

`p` is `[2, 0, 1]` — the resolved positions — while each group's DOM order is still the order its `y` was written in (group 0 comes out `30, 10, 20`). Same bug, same evidence as #987.

## Why #988's fix does not port over

Its selector narrowing assumes one `.trace.bars` group. A segmented layer has one **per trace**, and the shared selector spans all of them:

```
.barlayer children classes: ["trace bars", "trace bars"]

UNSCOPED  .trace.bars .point:nth-child(2) > path   → 2 elements
          [{h:10, cx:64}, {h:101, cx:149}]          one per group
```

So `nthBarSelector` as it stands would name the nth bar *of every group*. The group has to be pinned first.

## Measured feasible

```
.barlayer > .trace:nth-child(1) .point:nth-child(2) > path        → 1 element  {h:10, cx:64}
.barlayer > .trace:nth-of-type(1) .point:nth-child(2) > path      → 1 element  {h:10, cx:64}
.barlayer .trace.bars:nth-child(2) .point:nth-child(1) > path     → 1 element  {h:304, cx:576}
```

[...]

PR #991, opened 2026-08-17
fix(bar): let a per-bar selector list reach the highlight

### LINKED ISSUE #990: A reordered plotly bar chart has no highlight at all — #988's per-bar selectors never reach the DOM
Regression from #988, which is merged. Reporting it against myself.

## What

#988 made a `categoryorder` bar chart announce its bars in the order plotly draws them, and narrowed the layer's single selector into a per-bar `string[]` so the highlight would move with the data. **The array never reaches the DOM.**

`AbstractBarPlot.mapToSvgElements` (`src/model/bar.ts:286`) takes a `string`:

```ts
protected mapToSvgElements(selector?: string): SVGElement[][] | null {
  if (!selector) return null;
  const queried = Svg.selectAllElements(selector);
  if (queried.length === 0) return null;
```

and `Svg.selectAllElements` guards on the type (`src/util/svg.ts:112`):

```ts
private static isUsableSelector(query: unknown): query is string {
  return typeof query === 'string' && query.trim() !== '';
}
```

An array is not a string, so it returns `[]`, `queried.length === 0`, and the layer reports **no highlight**. The call site casts the lie into place — `this.mapToSvgElements(layer.selectors as string)` — so nothing complains at build time.

## Confirmed, not inferred

Driving the real `TraceFactory.create` over a layer shaped exactly as #988 emits:

```
✕ still has a highlight
    expect(received).not.toBeNull()
    Received: null
```

## What it costs

Before #988, a `categoryorder` bar chart had a **correct** highlight: the data order, the DOM order and the trace order all agreed. After #988 the announcement is right and the highlight is gone.

That is better than the failure #988 was written to avoid — announcing one bar while outlining another — but it is still a regression against the chart as it was, and for a low-vision user relying on the visual highlight it is the difference between a working chart and a broken one.

The irony is exact: #988's whole design was "both halves move together or neither does", falsified twice to prove the second half was covered. Both falsifications tested what the **extractor emitted**. Neither tested what the **model does with it**, and that is where the second half was dropped. A test can pin a contract on one side of a boundary and still say nothing about the other.

## Fix

Teach `AbstractBarPlot.mapToSvgElements` to accept `string | string[]`, resolving an array as one element per point in the given order. No alignment heuristics apply on that path — an adapter that names one element per bar has already resolved the order, so the zero-skipping and count-matching logic below it has nothing to infer.

Also worth doing, so this class of bug cannot recur silently: drop the `as string` cast at the call site, so a shape the consumer does not handle is a build error rather than an empty highlight.

## Scope

`SegmentedTrace.mapToSvgElements` has the same `string`-only signature and much more elaborate chunking (`skipZeros`, row/column-major, `groupsRunForward`). #989 proposes per-bar selectors for dodged and stacked layers and would hit this same wall — worth fixing this first, and pinning it with a test that goes through the model rather than stopping at the payload.

jooyoungseo:
Closes #990. **Regression from #988, which is merged** — reporting it against myself.

## What broke

#988 narrowed a reordered bar layer's single selector into a per-bar `string[]`, so the highlight would move with the points rather than stay in the trace's order. The array never reaches the DOM.

`AbstractBarPlot.mapToSvgElements` takes a `string`, and `Svg.selectAllElements` guards on the type:

```ts
private static isUsableSelector(query: unknown): query is string {
  return typeof query === 'string' && query.trim() !== '';
}
```

So an array returns `[]`, the `queried.length === 0` branch fires, and the layer reports **no highlight at all**. The call site's `as string` cast kept the build quiet about it.

## Confirmed, not inferred

Driving the real `TraceFactory.create` over a layer shaped exactly as #988 emits:

```
✕ still has a highlight
    expect(received).not.toBeNull()
    Received: null
```

## What it cost

Before #988, a `categoryorder` bar chart's highlight was **correct** — data order, DOM order and trace order all agreed. After #988 the announcement is right and the outline is gone.

Better than the failure #988 existed to prevent (announcing one bar while outlining another), but still a regression against the chart as it was, and for someone relying on the visual highlight it is the difference between a working chart and a broken one.

Worth naming the shape of the mistake, because it is not the one I was watching for. #988's design was "both halves move together or neither does", and it was falsified **twice** to prove the second half was covered. Both falsifications tested what the *extractor emitted*. Neither tested what the *model does with it* — and that is exactly where the second half was being dropped. A test can pin a contract on one side of a boundary and say nothing at all about the other side.

## The fix

`src/model/bar.ts` only.

- `mapToSvgElements` accepts `string | string[]`. An array resolves one element per point, in the order given.
- None of the alignment logic below applies on that path: an adapter that names one element per bar has already settled which element is which, so there is nothing to infer, and a zero-height bar it named was named deliberately rather than omitted by the renderer.
- A partial resolution (any selector matching nothing) declines the whole thing and removes the clones it already inserted, so it cannot leak hidden elements into the DOM — the same contract the string path already keeps.
- The call site's cast is widened to the shapes the function now handles, so the array path is reachable at all. **This is not type-level enforcement across the whole union** — `layer.selectors` is `string | string[] | string[][] | BoxSelector[] | CandlestickSelector | undefined`, and an assertion narrows it rather than proving it, so a producer sending `string[][]` for a bar layer still compiles. It degrades safely rather than silently: `Svg.selectElement` rejects a non-string, the partial-resolution branch fires, and the layer declines with its clones cleaned up. Real narrowing at the call site is a separate change.

## Verification

```
npx jest --selectProjects unit   → 4760 passed, 320 suites (2 new)
npm run type-check / lint:fix / build → clean
```

The new test lives in `test/model/`, goes through `TraceFactory.create`, and asserts on the resolved `highlightValues` elements — that row 0 holds the bars the layer announces, in the announced order — rather than stopping at the payload. It fails against `main` today (`Received: null`), which is what makes it worth having.

## Follow-on

`SegmentedTrace.mapToSvgElements` has the same `string`-only signature and much more elaborate chunking (`skipZeros`, row/column-major, `groupsRunForward`). #989 proposes per-bar selectors for dodged and stacked layers and would hit this same wall — it should land on top of this, and its tests should go through the model for the same reason.


jooyoungseo:
Both findings taken; the second is a correction to something I wrote, not to the code.

**Test location** — moved to `test/model/barSelectorConsumption.test.ts` in `400c76a`. You are right that being plotly-shaped does not make it an adapter test: it imports `@model/factory` and `@type/grammar` and nothing from `src/adapters/plotly/`, and `test/model/` already holds `barGaps`, `barInputMutation` and `barMagnitudeGuard` for this same file. I put it next to the change that caused it rather than next to the code it covers.

**The cast** — your correction stands, and it is a correction to my PR body rather than a nit. I wrote that "a shape the consumer does not handle is a build error". That is true only for the two shapes the function's own signature now names. `layer.selectors` is `string | string[] | string[][] | BoxSelector[] | CandlestickSelector | undefined`, and `as string | string[] | undefined` is an assertion over that union, not a guard — so a producer sending `string[][]` for a bar layer still compiles.

I have edited the PR body to say what the change actually buys, which is narrower: the signature no longer *lies*, so the array path is reachable at all. Type-level enforcement across the whole union would need a real narrowing at the call site, and that is a different change.

Your trace of the degradation is right too, and worth having on the record: `Svg.selectElement` guards on `typeof query === 'string'`, so a `string[][]` entry resolves to `null`, the partial-resolution branch fires, and the layer declines with its clones cleaned up. Safe, not silent-wrong.

**The `points.length !== 1` guard** — read exactly as you describe. It is a no-op for both current callers and deliberately scopes the fast path to the single-row case, since "there is nothing to align" only holds there. #989's segmented work is where multi-row per-element selectors get handled, and it needs `SegmentedTrace`'s own chunking dealt with rather than this branch widened.

---

jooyoungseo:
One more, in `e20dd61` — a hole this PR **introduced** rather than one it found.

Widening `AbstractBarPlot.mapToSvgElements` to `string | string[]` left `SegmentedTrace.mapToSvgElements` still declaring `string`. TypeScript accepts that override by parameter bivariance, so `tsc` said nothing: the base constructor now hands a possible array to a method that says it only takes strings, and at runtime it would reach `Svg.selectAllElements`, get `[]` back, and report no highlight — **the exact failure #990 is about, one class down, created by the fix for it.**

Nothing emits a list for a segmented layer today, so it is latent rather than live. But it is also precisely the wall #989 walks into, and an explicit decline with a comment is a better thing to walk into than a silent empty. Honouring a list there needs `SegmentedTrace`'s chunking dealt with first — `skipZeros`, row versus column major, and `groupsRunForward` all exist to *infer* what a per-cell list would simply state, so the two cannot just coexist.

Worth noting what this says about the finding you already made. You pointed out that the widened cast is an assertion rather than a guard, and that the PR body overclaimed. This is the same gap seen from the other end: the type system did not object to a genuine mismatch here either, for a different reason. Two ways for a selector shape to reach a consumer that cannot use it, both quiet.

```
npx jest --selectProjects unit → 4760 passed, 320 suites
npm run type-check             → clean
```

---

jooyoungseo:
Added in `a9c7a64`, and falsifying it twice gave two different answers — worth reporting both, because only one of them is the one you were asking for.

| what I broke | the test |
|---|---|
| removed the `Array.isArray` guard entirely | **still passes** |
| half-honoured a list (`selector = selector[0]`) | **fails** |

The first is your own observation turned around on me: `selectAllElements` answers `[]` for an array too, so with or without the guard the observable is identical. The test therefore pins the **contract** — a segmented layer handed a list reports no highlight, and does not resolve part of one — not the guard that implements it.

That is still the case worth having, and it is the one #989 needs to turn red before it can turn green. But it does not do the thing you asked for, which was to catch someone widening the override back. Nothing observable distinguishes those two states today, so no test at this level can; catching that would need a type-level assertion, not a behavioural one. I have said so in the commit rather than let the test imply more than it does — the same overclaim this PR already had to correct once in its own body.

Second case pins that the plain-string path is untouched. It needed `SVGPathElement` stubbed: jsdom has no specialised SVG constructors, and `SegmentedTrace` branches on it before it chunks — which incidentally confirms the guard short-circuits, since the array case never reaches that line.

```
npx jest --selectProjects unit → 4762 passed, 320 suites
npm run type-check             → clean
```

---

PR #996, opened 2026-08-18
fix(plotly): scope a bar-family selector to its own trace group

### LINKED ISSUE #993: A plotly panel with two bar-layer traces read as separate layers loses every highlight
Found while measuring #989, which needed to know exactly what `.barlayer` holds. Unrelated to ordering, so it is filed rather than folded in.

## What

`generatePlotlySelectors` gives every bar-family layer the same panel-wide selector:

```ts
case TraceType.BAR:
case TraceType.HISTOGRAM:
  return `${prefix}.trace.bars .point > path`;
```

That is right when the panel draws one bar-layer trace, and it is what the overwhelmingly common chart does. When a panel draws **two or more** and they are emitted as separate layers, every one of those layers gets a selector matching **all** of their bars — and `AbstractBarPlot.mapToSvgElements` rejects a row with more elements than points, so each layer ends up with no highlight at all.

A dodged or stacked panel is not affected: its traces become one layer whose point count matches, and after #989 its selectors are per cell anyway.

## Measured

plotly.js in Chromium, counting what `.subplot.xy .trace.bars .point > path` resolves to:

```
panel                       .barlayer groups   points per group   shared selector matches
one bar alone               1                  [3]                3     ← fine
one bar + one histogram     2                  [3, 2]             5
two bars, barmode overlay   2                  [3, 3]             6
two histograms              2                  [2, 2]             4
```

And end to end through `TraceFactory.create`, on a two-group DOM with a three-point `bar` layer carrying that selector:

```
document.querySelectorAll('.trace.bars .point > path').length  →  6
trace.highlightValues                                          →  null
```

`mapToSvgElements` takes the `svgElements[row].length > this.points[row].length` branch, discards the clones it just inserted, and returns `null`. Every affected layer is silent for highlighting; audio, text and braille are unaffected, so nothing else about the chart looks wrong.

The failure is always a **missing** highlight, never a wrong one: the shared selector's count is the sum over the groups, so it strictly exceeds any one layer's point count and the alignment branch is never reached. That is the better of the two failures, and still a chart where outlining does nothing.

## Which panels reach it

The layers are emitted individually — rather than combined — in these cases, all from `extractPanelLayers`:

- `barmode: 'overlay'` (or an unrecognised barmode) with more than one bar trace, which pushes each onto `otherTraces`;
- any panel mixing a `bar` with a `histogram`, since a histogram is never in `barTraces`;
- more than one histogram on a panel.

`funnel`, `waterfall`, `box` and `scatter` are **not** involved — measured: they draw into their own layers, and only `bar` and `histogram` land in `.barlayer`.

## Proposed

Scope the selector to the trace's own group, the way the gantt and error-bar paths already do:

```ts
`${prefix}.barlayer > g.trace.bars:nth-of-type(${barLayerPosition(gd, traceIndex, trace) + 1}) .point > path`
```

Both helpers exist and both already count histograms into the position, which is the part that is easy to get wrong — measured, a histogram declared before the bars takes group 1 and shifts everything after it.

Worth scoping unconditionally rather than only when the panel has several groups: a one-group panel resolves the same either way, and a selector that is right only in the common case is the shape of bug this is.

jooyoungseo:
Closes #993. Found while measuring #989, which needed to know exactly what `.barlayer` holds.

## The bug

Every bar-family layer shared one layer-wide selector — `.subplot.xy .trace.bars .point > path` and the `.funnellayer` / `.waterfalllayer` equivalents. That is right when the panel draws one trace into that layer, which is the overwhelmingly common chart. It is wrong whenever it draws more and MAIDR emits a layer per trace.

Measured on plotly.js in Chromium, counting what each layer-wide selector resolves to:

```
panel                       layer            groups   points each   matches
one bar alone               .barlayer        1        [3]           3   ← fine
one bar + one histogram     .barlayer        2        [3, 2]        5
two bars, barmode overlay   .barlayer        2        [3, 3]        6
two histograms              .barlayer        2        [2, 2]        4
two funnels                 .funnellayer     2        [3, 3]        6
two waterfalls              .waterfalllayer  2        [3, 3]        6
```

**The two failures that produces are not the same size.** A plain layer sees more elements than points, so `AbstractBarPlot` discards its clones and reports no highlight — driven through `TraceFactory.create`, a three-point bar layer on a two-group DOM came back `null`.

A **segmented** layer does not compare counts, it *chunks*. So the extra elements are not ignored — they shift every cell:

```
highlight  [["H-bin1", "H-bin2", "A-charlie"],
            ["A-alpha",  "A-bravo",  "B-charlie"]]
```

Series A at 'charlie' announces its own value and outlines the **histogram's first bin**. That is a wrong highlight, not a missing one, and it is the more serious half. My original issue body claimed the failure was always a missing highlight; that was wrong for segmented layers and is [corrected on the issue](https://github.com/xability/maidr/issues/993#issuecomment-5321731391).

## What this changes

**Each layer names its own group.** `barLayerPosition` already counted the traces plotly drew into `.barlayer` before this one — including histograms, which share it. Generalised from `barlayer` to *the layer the type is drawn into*, so a funnel counts among funnels and a waterfall among waterfalls. Measured on one panel carrying all four types at once:

```
_fullData             [bar, funnel, histogram, waterfall, funnel, waterfall]
.barlayer children    ["trace bars", "trace bars"]   ← bar + histogram
.funnellayer          ["trace bars", "trace bars"]   ← the two funnels
.waterfalllayer       ["trace bars", "trace bars"]   ← the two waterfalls
```

No type strays into another's layer, so a bar sitting before a funnel does not shift it.

Scoped **unconditionally**, not only when the panel holds several groups: a one-group panel resolves the same either way, and a selector right only in the common case is the shape of this bug.

**A segmented layer takes a list.** It spans several traces, so no single group names its bars. `querySelectorAll` answers a selector list in *document* order, which across those groups is series-major — exactly the order the chunking already expects. So the chunking is untouched; only its input changes:

```
.subplot.xy .barlayer > g.trace.bars:nth-of-type(2) .point > path,
.subplot.xy .barlayer > g.trace.bars:nth-of-type(3) .point > path
    →  A's three bars, then B's three, skipping the histogram's group
```

**`generatePlotlySelectors` declines the segmented types** rather than answering with the panel-wide selector. `extractSegmentedBarLayer` is its only caller for those — `mapTraceType` never returns a segmented type, so `extractLayer` cannot reach them — and it builds its own. Leaving the old answer in the switch would hand a future caller the very mismatch this closes.

## Verification

| Step | Result |
| --- | --- |
| `npx jest --selectProjects unit` | 4802 passed, 322 suites |

Every guard falsified independently, with the failing test named:

| Removed | Fails |
| --- | --- |
| the scoping, back to the panel-wide selector | 5 of 12 |
| the group index, hardcoded to 0 | 3 |
| `barLayerPosition` for the segmented list, replaced by the series index | `leaves out a histogram it does not cover` |
| the segmented list's group order, reversed | 3 |
| the per-layer map, all four types into `barlayer` | 3 |
| the funnel/waterfall scoping, back to layer-wide | 3 |

Three existing assertions pinned the old selector strings and were updated with a note saying why they changed.

The model side is pinned separately, in `test/model/` — #990's lesson: an adapter change is worth nothing unless the model consumes what it emits. One of those cases asserts what the model does with a panel-wide selector on a multi-group DOM. It is not a bug being enshrined: the chunking is handed a flat list and told the layer owns all of it, and nothing in that list says which group an element came from. That is precisely why the adapter has to scope, and the case is there so a later attempt to make the model clever about it turns red and reads the reason.

## Also covered

The seam where #988's per-bar narrowing meets this scoping. `nthBarSelector` narrows whatever head it is given, so on a reordered overlay panel it now counts `nth-child` *inside* the trace's own group; narrowed from the panel-wide head it would have named the nth bar of both. Pinned by `narrows within its own group rather than across both`.

## Not in scope

Only the bar-renderer layers. A `dot` plot is scoped by trace uid already, and the box, violin and scatter layers were not measured for this.

---


jooyoungseo:
Both notes acted on in `b77157cc` — they were right on both counts.

**The module header.** It said subplot-level scoping is sufficient, which my previous commit made false for the bar family without updating it. It still describes the scatter and line families accurately, so it now says which is which, and why bars need one level finer: they carry no uid to scope by, so the group is counted rather than named.

**The hardcoded `'barlayer'`.** Your trace is right — `barTraces` is filled from `TraceType.BAR` entries alone (`extractor.ts:920`), and a funnel or waterfall maps to its own type and reaches `extractLayer` instead, so no non-bar can arrive. Since that is an invariant of the *caller* rather than of the function, I wrote it down at the hardcoding rather than leaving it to be re-derived. I did consider reading `BAR_FAMILY_LAYER` there instead, and decided against: `barLayerPosition` in the extractor counts `bar`/`histogram` within the panel and would need the same generalisation to match, which is machinery in service of a case that cannot occur.

Re-verified after the change: `type-check` clean, `lint:fix` clean, 307 plotly + selector tests passing.

On the unverified section — that is the right thing to say rather than assume, and worth noting it has been accurate every time: the commands are run locally before each push, and the numbers in the body are what they printed.

---

jooyoungseo:
Housekeeping, not a code change: I cancelled and re-queued run [32084494577](https://github.com/xability/maidr/actions/runs/32084494577). `E2E (chromium)` had been in progress for **35+ minutes** while the other fourteen checks all went green in under four; the same job took ~5.5 minutes on #991, #992 and the previous commit here. It had produced no logs to read, so there was nothing to diagnose — it looked stalled rather than failing.

I did not merge past it, and I'm noting the reasoning rather than leaving a silent re-run in the history.

Two independent reasons to think the stall is infrastructure rather than this diff:

- **No E2E spec touches plotly.** The suite drives MAIDR's own example schemas; there is no plotly spec under `e2e_tests/specs/`, so a change to the plotly adapter's selector strings has no path into it.
- **The bar specs pass locally.** Running the three that could plausibly care, on Chromium against this branch:

```
npx playwright test stackedBarplot.spec.ts dodgedBarplot.spec.ts barplot.spec.ts --project=chromium
  71 passed (51.9s)
```

(The unfiltered run also reports six failures, all `browserType.launch: Executable doesn't exist at /opt/pw-browsers/firefox-1511/…` — firefox and webkit are not installed in this environment. CI runs Chromium only.)

If the re-run stalls the same way, that is worth looking at on its own rather than being read as anything about this PR.

---

PR #998, opened 2026-08-18
fix(vegalite): read a segmented bar&#39;s columns in the order they are drawn

### LINKED ISSUE #994: A Vega-Lite stacked or dodged bar reads its categories in the data&#39;s order, not the drawn one
Found by taking #989's question — "does this adapter read the drawn order or the data's?" — to the next library. Vega-Lite has the same defect, and unlike plotly's it does not need an opt-in sort: it is the **default**.

## What

`extractSegmentedData` groups the rows by fill and pushes each series' points in data-flow order:

```ts
const groups = new Map<string, SegmentedPoint[]>();
for (const row of rows) {
  const fill = String(row[colorField] ?? '');
  …
  groups.get(fill)!.push(pt);
}
return [...groups.values()];
```

Vega **sorts a nominal domain** rather than keeping the rows' order — the same fact `orderedHeatmapAxis` already documents for the heatmap (#977), and the same fact `sortSimpleBarsByVisualOrder` already fixes for a plain bar. Neither reaches a segmented layer. So the columns of every stacked, dodged, normalized and diverging Vega-Lite bar are announced in whatever order the rows happened to be listed in.

## Measured

Compiled with vega-lite 6 and run through vega, rows listed `charlie, alpha, bravo`:

```
                             x scale domain            drawn left → right
stacked, no sort             alpha, bravo, charlie     alpha@5, bravo@105, charlie@205
stacked, sort: '-y'          charlie, bravo, alpha     charlie@5, bravo@105, alpha@205
stacked, sort: [b, c, a]     bravo, charlie, alpha     bravo@5, charlie@105, alpha@205
stacked, sort: 'descending'  charlie, bravo, alpha     charlie@5, bravo@105, alpha@205
dodged,  sort: '-y'          charlie, bravo, alpha     charlie@19, bravo@113, alpha@206
```

and end to end through `vegaLiteToMaidr`, with the `x` scale domain available on the view:

```
type       stacked_bar
columns    [["charlie","alpha","bravo"], ["charlie","alpha","bravo"]]
selectors  "g.mark-rect.role-mark.marks path, g.mark-rect.role-mark.layer_0_marks path"
```

The first row is the one that matters most: **no `sort` anywhere in the spec**, and the announced order is still not the drawn one. Any spec whose rows are not already listed alphabetically is affected.

Every label keeps its own value and the highlight still lands on the right bar, so nothing reads as broken. What is wrong is everything treating the index as a position: which way arrowing travels, the stereo pan (`BarTrace.audio` pans by `col`), the braille line, the autoplay sweep, and which cells the summary row sums together.

## Why the existing passes do not cover it

`src/vegalite-entry.ts` has three reorder passes, and none of them touches columns:

- `sortSimpleBarsByVisualOrder` — plain `BAR` only, and says so: *"Segmented layers (stacked / dodged / normalised) use `layer.domMapping` instead and must NOT be re-ordered here."*
- `reorderDodgedBarsByVisualPosition` — sorts **series**, so `data[0]` is the leftmost subgroup within a category.
- `reorderSegmentedSeriesByVisualBottom` — the vertical equivalent for stacked.

Rows are handled; columns are not.

## Suggested shape

Two options, and I think the first is clearly right *for this adapter*:

**Extend the existing geometry pass to columns.** `sortSimpleBarsByVisualOrder` already reads `getBoundingClientRect()`, permutes `layer.data`, and calls `reappendInOrder` so `querySelectorAll` comes back in the new order. Doing the same per series for a segmented layer keeps the single selector, needs no grammar change, and reads the drawn positions directly rather than re-deriving them — which is stronger evidence than any scale-domain lookup. `isLaidOutForSort` already guards the case where the chart has no layout to measure.

The care needed is that `domMapping` describes how the DOM flattens (`order: 'row'` for stacked, `order: 'column'` for dodged), so a re-append has to preserve that grouping — exactly the arithmetic `reorderDodgedBarsByVisualPosition` already does when it reorders series.

**Not** the #989 grid. Emitting `string[][]` per cell works — `SegmentedTrace` reads that shape as of #992 — but it is the right answer for plotly, where the adapter never touches the DOM and the drawn order has to be reconstructed from `calcdata`. Here the elements are in hand and movable, and copying the grid would add a second mechanism where one already exists.

## Related

- #977 — the same Vega fact, fixed for the heatmap via `orderedHeatmapAxis`
- #989 / #992 — the same defect in the plotly adapter, fixed the other way round because that adapter cannot move the DOM

jooyoungseo:
Closes #994. Found by taking #989's question — "does this adapter read the drawn order or the data's?" — to the next library.

## The bug

`extractSegmentedData` groups the rows by fill and pushes each series' points in data-flow order. Vega **sorts a nominal domain** rather than keeping that order, so the columns of every stacked, dodged and normalized Vega-Lite bar were announced in whatever order the rows happened to be listed in.

Not a `sort`-only case, which is what makes it worth a pass of its own. Compiled with vega-lite 6 and run through vega, rows listed `charlie, alpha, bravo`:

```
spec                          x scale domain          drawn left → right
no sort at all                alpha, bravo, charlie   alpha@5,    bravo@105,  charlie@205
sort: '-y'                    charlie, bravo, alpha   charlie@5,  bravo@105,  alpha@205
sort: ['bravo','charlie',…]   bravo, charlie, alpha   bravo@5,    charlie@105, alpha@205
dodged, sort: '-y'            charlie, bravo, alpha   charlie@19, bravo@113,  alpha@206
```

The first row is the one that matters: **no `sort` anywhere in the spec**, and the announced order is still not the drawn one. Any spec whose rows are not already listed in the scale's order was affected.

Every label kept its own value and the highlight landed correctly, so nothing read as broken. What was wrong is everything treating the index as a position: which way arrowing travels, the stereo pan, the braille line, the autoplay sweep, and which cells the summary row sums together.

## The missing member of a family

This is not a new mechanism. `bindVegaLite` already runs a visual-order pass for almost every layer type:

| pass | covers | aligns |
| --- | --- | --- |
| `sortSimpleBarsByVisualOrder` | `BAR` | data + DOM, by mark geometry |
| `sortHistogramBinsByVisualOrder` | `HISTOGRAM` | data + DOM, by mark geometry |
| `sortHeatmapCellsByVisualOrder` | `HEATMAP` | labels, matrix and DOM |
| `sortLinesByVisualOrder` | `LINE` | data, by rendered tick order |
| `reorderSegmentedSeriesByVisualBottom` | `STACKED`, `NORMALIZED` | **series** only |
| `reorderDodgedBarsByVisualPosition` | `DODGED` | **series** only |

The two segmented passes handle rows; nothing handled columns. `sortLinesByVisualOrder`'s own docstring describes this exact bug — *"Fixes 'highlight column != announced x-value' when Vega-Lite renders nominal/ordinal X scales alphabetically while the dataset is in insertion order"* — fixed for lines and not for segmented bars.

**It follows the geometry passes, not the line one.** `sortLinesByVisualOrder` reorders data alone, which works only because a line series is a single `<path>` with no per-point element. A segmented bar has one element per cell, so data and DOM must move together — otherwise the layer announces one bar and outlines another, which is #990's lesson applied before the fact rather than after.

Every series is permuted **the same way**, because a segmented layer is read by column: column `c` is one category across every series, and the summary row sums down it. Declined for a ragged layer, whose series do not line up at all.

## Verified against a real render, not only a fixture

The synthetic DOM in the test is the obvious risk, so I checked it against Vega itself. Rendering the spec in Chromium and reading the marks before binding:

```
charlie@205  alpha@5  bravo@105  charlie@205  alpha@5  bravo@105
```

— series-major, each series in the rows' order, `charlie` drawn rightmost. The fixture reproduces that exactly, down to `<path>` rather than `<rect>` (Vega's SVG renderer draws rect marks as paths). `vegaLiteToMaidr` on the same real view still returns `[["charlie","alpha","bravo"], …]`, confirming the converter is not where this is fixed.

## Verification

| Step | Result |
| --- | --- |
| `npx jest --selectProjects unit` | 4808 passed, 323 suites |
| `npx playwright test barplot dodgedBarplot stackedBarplot --project=chromium` | 71 passed |

Every guard falsified independently:

| Removed | Fails |
| --- | --- |
| the pass, not called | `announces its categories…`, `moves the DOM with the data` |
| `reappendInOrder` — data moves, DOM does not | `moves the DOM with the data` |
| the shared permutation, each series sorted alone | `announces…`, `permutes every series the same way` |
| the horizontal branch, always comparing x | `reads a sideways chart off the axis its categories are on` |
| `domMapping`, everything treated row-major | `permutes inside a subject-major DOM…` |
| the column-major re-append | same |
| the sort direction, reversed | 4 of 6 |

One test I corrected rather than shipped: I first named a case *"leaves a chart already in the drawn order alone"*, implying it pinned the already-sorted short circuit. Falsifying showed it does not — re-appending in identity order is observationally a no-op, so removing that branch leaves everything green. It is renamed and re-commented to say what it does catch: a sort in the wrong direction on a chart that starts correct, which no other case here distinguishes from no sort at all.

Coverage gap I found the same way: the `domMapping: 'column'` branch was untested, because my fixtures' marks had no `fill` and `detectSegmentedDomOrder` reports nothing without one. The marks now carry fills, and a subject-major case exercises that branch.

## Not in scope

`orderedHeatmapAxis` already covers the heatmap (#977). The tick-label route `sortLinesByVisualOrder` uses stays available as a fallback if geometry is ever unavailable, but it would still need the DOM half bolted on.

---


jooyoungseo:
Good catch — that is a real inconsistency, and it is fixed in `44d692b6`.

Confirmed the three defaults were out of step:

```
src/vegalite-entry.ts:400   reorderSegmentedSeriesByVisualBottom   ?? 'row'      STACKED, NORMALIZED
src/vegalite-entry.ts:574   reorderSegmentedColumnsByVisualOrder   ?? 'row'      all three     ← wrong for DODGED
src/vegalite-entry.ts:689   reorderDodgedBarsByVisualPosition      ?? 'column'   DODGED
```

The two existing passes are each right for the types they handle; mine handles all three under one default, so it was wrong for exactly the one whose default differs. Now type-keyed, as you suggested, so all three agree per type.

**Pinned rather than argued away.** The trigger is narrow, which is a reason to write a test rather than a reason to skip one — a narrow case is the kind nobody reproduces by hand later. The new case builds a dodged chart whose marks carry **no `fill`**, which is what actually makes `detectSegmentedDomOrder` decline (its own limitation #2/#3) rather than merely disagree, and emits them subject-major:

```
falls back to a dodged chart's own grouping when detection declines
```

Falsified by reverting the default to `'row'` — that case fails and the other six stay green, so it pins this and nothing else.

Worth noting the failure mode you described is exactly right and worse than it might sound: `reorderDodgedBarsByVisualPosition` re-appends on the `'column'` assumption immediately before this pass runs, so the two would not merely disagree in the abstract — the second would regroup the DOM the first had just arranged, leaving the highlight pointing into cells neither pass believed in.

Re-verified after the change: `npx jest --selectProjects unit` → **4809 passed, 323 suites**; `type-check` and `lint:fix` clean.

On the sandbox note — that is the right thing to state rather than assume. For what it is worth the commands are run locally before each push, and the numbers in the body and here are what they printed.

---

Issue #997, opened 2026-08-18
A Highcharts `type: &#39;bar&#39;` chart with `inverted: true` is announced vertical while drawn horizontal


jooyoungseo:
Found while measuring the scope of #995, which needed to know exactly when Highcharts sets `xAxis.reversed` by itself.

## What

`convertBarSeries` resolves orientation by flipping the series' default when the chart is inverted:

```ts
const isInverted = chart.options.chart?.inverted === true;
const defaultOrientation = seriesType === 'bar' ? Orientation.HORIZONTAL : Orientation.VERTICAL;
const orientation = isInverted
  ? (defaultOrientation === Orientation.VERTICAL ? Orientation.HORIZONTAL : Orientation.VERTICAL)
  : defaultOrientation;
```

The premise is that `inverted` toggles the drawing, so a `bar` series (already sideways) plus `inverted` returns to vertical. **Highcharts does not work that way.** `type: 'bar'` *is* `type: 'column'` with `inverted: true` — it sets the same flag rather than composing with it — so `bar` and `bar + inverted` are the same chart.

## Measured

Highcharts 12 in Chromium, `categories: ['alpha','bravo','charlie']`, one series `[10, 20, 30]`, reading each rendered point's own box:

```
chart options                     chart.inverted   left → right                      top → bottom
{type:'column'}                   false            alpha@105, bravo@313, charlie@522  charlie@210, bravo@263, alpha@315
{type:'column', inverted:true}    true             charlie@59, bravo@174, alpha@290   charlie@333, bravo@431, alpha@529
{type:'bar'}                      true             charlie@59, bravo@174, alpha@290   charlie@333, bravo@431, alpha@529
{type:'bar', inverted:true}       true             charlie@59, bravo@174, alpha@290   charlie@333, bravo@431, alpha@529
```

The last three are **geometrically identical** — same coordinates to the pixel. There is no double flip.

And through the adapter itself, `highchartsToMaidr` on those same charts:

```
chart options                  orientation   data[0]
{type:'bar'}                   horz          {x: 10,      y: 'alpha'}   ← correct
{type:'bar', inverted:true}    vert          {x: 'alpha', y: 10}        ← wrong
{type:'column'}                vert          {x: 'alpha', y: 10}        ← correct
{type:'column', inverted:true} horz          {x: 10,      y: 'alpha'}   ← correct
```

Three of the four are right. The fourth describes a chart nobody is looking at.

## Why it matters

[...]

PR #1000, opened 2026-08-18
fix(highcharts): order a reversed bar axis the way the chart draws it

### LINKED ISSUE #995: A Highcharts bar on a reversed x axis is announced left to right while it is drawn right to left
Third result from taking #989's question — "does this adapter read the drawn order or the data's?" — through the remaining bar-capable adapters.

## What

`isReversedAxis` exists in `src/adapters/highcharts/adapter.ts:1719` and has exactly three callers, all of them asking about **`series.yAxis`**: bump detection, heatmap row order (#973), and one more. Nothing asks it about the category axis.

`buildSegmentedRows` places each point by its own category index:

```ts
const index = indexForX(p.x);
row[index] = { x: categoryLabels[index], y: value, z: series.name };
```

which is right — it is what makes Highcharts immune to the data-order bug plotly (#987) and Vega-Lite (#994) have. But `categoryLabels[0]` is `xAxis.categories[0]`, and on a reversed axis Highcharts draws that at the **right-hand end**. So column 0 is the rightmost bar and the announced order runs opposite to the drawn one.

## Measured

Highcharts 12 in Chromium, `categories: ['alpha','bravo','charlie']`, two series:

```
chart                          xAxis.reversed   drawn left → right
column, plain                  undefined        alpha, bravo, charlie
column, xAxis.reversed: true   true             charlie, bravo, alpha
column stacked, reversed:true  true             charlie, bravo, alpha
```

MAIDR announces `alpha, bravo, charlie` in all three.

The DOM does **not** move with the axis — reading each rendered bar's own box, `.highcharts-series-0 .highcharts-point` comes back in the series' data order either way, only the x coordinates change:

```
column plain      h105@105  h210@313  h315@522
column reversed   h105@522  h210@313  h315@105
```

So the highlight lands on the right bar and every label carries its own value. What is wrong is only the direction — which is exactly the failure that made #971–#978 worth a sweep: the chart loads, navigates, highlights and announces a complete reading, and the reading is false.

## Suggested

`isReversedAxis(series.xAxis)` already answers this; the categories and each row simply need reversing together when it is true, in the plain bar path and `buildSegmentedRows` alike. No selector change: the DOM order is the *data's*, so reversing the data means the per-index highlight follows it.

## The part that is NOT this issue

An **inverted** chart (`type: 'bar'`, Highcharts' horizontal bars) has `xAxis.reversed === true` set by Highcharts itself, with `options.reversed` left `undefined`:

```
bar (inverted), plain              inverted: true   xAxis.reversed: true   options.reversed: undefined
bar (inverted), reversed: false    inverted: true   xAxis.reversed: false  options.reversed: false
                                   both draw top → bottom: charlie, bravo, alpha
```

So a fix that reverses on `isReversedAxis(xAxis)` alone would flip every horizontal Highcharts bar as a side effect. Worse, it is not clear it *would* be a side effect: `MaidrLayer.orientation`'s contract documents which field holds the category and which the magnitude, and says nothing about which end of a horizontal bar's category axis `data[0]` should sit at. Setting that convention is a core question, not an adapter one, and it wants its own issue rather than being decided by whichever adapter is patched first.

Scope this one to `chart.inverted !== true`, where the answer is unambiguous.

## Adapter sweep, for the record

Same question asked of every adapter that draws bars:

| adapter | data order vs drawn order |
| --- | --- |
| plotly | **affected** — `categoryorder` sorts the axis, not the trace (#987, #989) |
| Vega-Lite | **affected** — Vega sorts a nominal domain by default (#994) |
| Highcharts | **affected via `reversed` only** — indexes by the point's own category, this issue |
| Chart.js | not affected — `labels[i]` *is* the drawn order, and `data[i]` aligns to it |
| Recharts, Google Charts, Victory, Frappe, AnyChart, Observable, amCharts, D3 | not affected — each draws in the order the data is given |
| r-maidr | not affected — reads ggplot's resolved scale (`scale$limits` / `panel_params$x$get_labels()`) |

The pattern: only libraries with a *separate* axis-order concept — a sort, a domain, a reversal — can disagree with their own data, and those are the three above.

Only the bar family was measured. Whether a Highcharts line or scatter on a reversed category axis has the same problem is untested.

jooyoungseo:
Closes #995. Lands on top of #999 (#997), which is what makes the scope gate expressible.

## What was wrong

`buildSegmentedRows` places each value under its own category index — which is exactly what makes Highcharts immune to the data-order bug plotly (#987) and Vega-Lite (#994) have. But the categories belong to the **axis**, and `xAxis.reversed` draws `categories[0]` at the right-hand end.

So the chart loads, navigates, highlights and announces a complete reading — and the reading runs the opposite way from the bars. Measured, Highcharts 12 in Chromium, `categories: ['alpha','bravo','charlie']`:

| chart | `xAxis.reversed` | drawn left → right | announced |
| --- | --- | --- | --- |
| `{type:'column'}` | `undefined` | alpha, bravo, charlie | alpha, bravo, charlie |
| `{type:'column'}` + reversed | `true` | **charlie, bravo, alpha** | alpha, bravo, charlie |
| stacked column + reversed | `true` | **charlie, bravo, alpha** | alpha, bravo, charlie |

## Why reversing the data alone would have been worse

The issue's own "Suggested" section said no selector change was needed. That is wrong, and wrong in the direction that ships a defect — the same trap as plotly #988/#990.

The DOM does **not** move with the axis. Reading each element's own box, in document order:

```
column plain      h105@105  h210@313  h315@522
column reversed   h105@522  h210@313  h315@105
```

`barSelector` matches every point of a series at once and leaves the pairing to document order, which is data order. Reverse the rows and `data[0] = charlie` starts pointing at `DOM[0] = alpha`'s bar — a correct highlight traded for a wrong one. Both halves have to move together, so each bar is now named outright.

## A DOM position is not a category index

Measured, and it decides the addressing:

```
series [10, null, 0, 40]  →  3 elements, at the alpha, charlie and delta slots
```

A `null` renders nothing; a genuine `0` renders an element. So the per-cell selectors count **bars actually drawn**, and the two cases are deliberately counted apart.

Where some cell has no element at all, the group is left exactly as it was rather than half-fixed: a grid names an element per cell, and `SegmentedTrace` rejects the whole grid when one cell resolves to nothing — which would cost the layer its highlight entirely. Reading backwards is the lesser failure of the two.

## Scope

Vertical layers only. Highcharts sets `xAxis.reversed` **itself** on everything it draws sideways, so the flag alone cannot separate those from a column chart whose author asked for a reversal:

```
{type:'column'}                 inverted false   reversed undefined
{type:'column', reversed:true}  inverted false   reversed true,  options.reversed true
{type:'bar'}                    inverted true    reversed true,  options.reversed undefined
{type:'column', inverted:true}  inverted true    reversed true,  options.reversed undefined
```

The resolved orientation can separate them, but only since #999 — before it, `bar + inverted` resolved *vertical* and carried an auto-set `xAxis.reversed`, so the gate would have reversed a chart nobody asked to reverse. Measured against the fixed adapter, the gate fires on exactly one row, and it is the row where drawn-first and announced-first disagree.

Every sideways Highcharts chart is *also* drawn `charlie, bravo, alpha` while its data starts at `alpha` — recorded on #995 and deliberately not folded in. `MaidrLayer.orientation` documents which field holds the category and which the magnitude, and says nothing about which end of a horizontal bar's category axis `data[0]` sits at. That is a core convention question, not an adapter one.

## Changes

- `selectors.ts` — new `barPointSelector(containerId, seriesIndex, drawnIndex)`. Counted with `:nth-child`, not `:nth-of-type`: Highcharts v11+ renders points as `<path>` for rounded corners, so `rect.highcharts-point:nth-of-type(k)` matches nothing on a default chart. Measured, the series group holds its points and nothing else, so each `nth-child` resolves to exactly one element.
- `adapter.ts` — `categoryIndexer` extracted out of `buildSegmentedRows` so the rows and the DOM positions computed for them cannot drift apart; `drawnCellIndices` counts drawn bars and declines on a gap; `orderSegmentedByAxis` reverses rows and emits the `string[][]` grid, wired into the stacked, dodged, normalized and diverging paths; `convertSingleBar` emits the `string[]` form. Both core-side shapes exist as of #991 and #992.

## Tests

`test/adapters/highcharts/reversedBarOrder.test.ts`, 12 cases.

Falsified against the unfixed adapter: **8 fail**. The 4 that pass are the "leaves it alone" cases, which describe unchanged behaviour — so each was falsified separately by removing the guard it names:

| guard removed | test that goes red |
| --- | --- |
| the `VERTICAL` half of the scope gate | leaves a sideways chart alone |
| the gap guard in `drawnCellIndices` | leaves a group with a gap as it was |
| counting a genuine zero as drawn | counts a genuine zero, which Highcharts does draw |
| the reversal condition entirely (always fire) | both "in the data's order" cases |

Every case pins something. `npm run build` clean, `npm run type-check` clean, unit suite 325 suites / 4829 tests green.

---


jooyoungseo:
Pushed `e2f824bb`, two tests. The 12 already here assert the selector **strings** the adapter emits, which says the payload is right and says nothing about whether the reader sees the right bar — and #990 was exactly a layer whose selector shape looked sound and whose highlight was gone.

These drive the emitted grid through the model against a document laid out the way Highcharts lays one out, and assert the id of the element each cell actually outlines:

```
plain,   reversed   →  S0-charlie  S0-bravo  S0-alpha
stacked, reversed   →  S0-charlie  S0-bravo  S0-alpha  |  S1-charlie  S1-bravo  S1-alpha
```

Falsified with a one-off in the `nth-child` narrowing: both go red, along with 5 of the string assertions. Full unit suite 325 suites / 4831 tests green.

One note on the fixture, because it caught me out elsewhere today. jsdom builds every SVG child as a plain `SVGElement` and defines neither `SVGPathElement` nor `SVGRectElement`, so which branch of `SegmentedTrace.mapToSvgElements` a test exercises is decided by what it binds those globals to, **not** by whether its markup says `<path>` or `<rect>`. The repo's existing idiom pins the rect branch. A grid is resolved before that branch and so is unaffected either way, but this fixture binds both deliberately rather than inheriting the choice. I filed #1003 for the underlying asymmetry — the two branches default oppositely when nothing is declared — and closed #1001, which I had raised off a measurement that went through the wrong one.

---

Issue #1007, opened 2026-08-18
A Highcharts line or scatter on a reversed axis is read left to right while it is drawn right to left


jooyoungseo:
Answers the open question left at the end of #995: *"Only the bar family was measured. Whether a Highcharts line or scatter on a reversed category axis has the same problem is untested."*

It does. Measured, both.

## Measured

Highcharts 12.6 in Chromium, `categories: ['alpha','bravo','charlie']`, one series `[10, 20, 30]`, reading each point's own resolved position:

```
line, plain      alpha@104   bravo@313   charlie@522
line, reversed   alpha@522   bravo@313   charlie@104
d (plain)        M 104 305 L 313 183 L 521 61
d (reversed)     M 521 305 L 313 183 L 104 61      ← the path itself runs right to left
```

Scatter is the same; only the marker coordinates move.

And what the adapter emits for those charts, through the real `highchartsToMaidr`:

```
line,    plain      type=line   x=["alpha","bravo","charlie"]
line,    reversed   type=line   x=["alpha","bravo","charlie"]     ← unchanged
scatter, plain      type=dot    x=["alpha","bravo","charlie"]
scatter, reversed   type=dot    x=["alpha","bravo","charlie"]     ← unchanged
```

`convertLineSeries` maps `series.data` and never consults the axis:

```ts
const data: LinePoint[][] = seriesList.map(series =>
  series.data.filter(p => p.y !== null).map(p => ({ x: pointLabel(p), y: p.y as number, ... })));
```

`convertScatterSeries` does the same.

## What is and is not wrong

Every point keeps its own label and value, and the highlight lands on the right point — `LineTrace` derives its highlight from the path's `d`, whose vertices come out in path order, which is the series' data order either way. So this is the #987 shape rather than the #995 one: nothing is mis-valued and nothing is mis-outlined.

What is false is the **direction**, and everything that reads an index as a position:

- arrowing right walks `alpha, bravo, charlie` while the chart reads `charlie, bravo, alpha` left to right;
- `AudioService` pans by column, so the point drawn hard **right** plays hard **left**;
- the braille line — a spatial medium — renders a chart that is not on screen;
- autoplay sweeps in that same false order.

A reader comes away right about every value and wrong about the shape.

## The adapter cannot fix this on its own

Worth stating, because the bar fixes in this family were all adapter-side and this one cannot be.

[...]

Issue #1001, opened 2026-08-18
Every cell of a Highcharts stacked or dodged bar chart highlights the wrong bar


jooyoungseo:
Found while building #995's fix, and unrelated to it. Not an edge case: it fires on an ordinary two-series stacked column chart with no gaps, no zeros and no reversal.

## What

`convertDivergingBar` emits `domMapping: { order: 'row' }`, with a docstring saying exactly why:

```ts
// Highcharts lays one series group out after another, so the elements the
// selector list resolves to run series by series. `SegmentedTrace` assumes
// the opposite for `<rect>` marks (the shape Highcharts draws whenever a
// bar has square corners), and would pair the first side's announcements
// with alternating bars from both sides.
domMapping: { order: 'row' },
```

`convertStackedBar` and `convertDodgedBar` build their rows with the *same* `buildSegmentedRows` and emit the *same* joined `barSelector` list — and do not set it. Confirmed on the real adapter:

```
stacked:     type=stacked_bar             domMapping=undefined
normalized:  type=stacked_normalized_bar  domMapping=undefined
dodged:      type=dodged_bar              domMapping=undefined
diverging:   type=diverging_bar           domMapping={"order":"row"}
```

The reasoning in that docstring is right, and it applies to all four.

## Measured

Highcharts 12.6 in Chromium, `categories: ['alpha','bravo','charlie']`, `series: [A [10,20,30], B [1,2,3]]`, `stacking: 'normal'`. Querying the adapter's own joined selector and identifying each element through `point.graphic`:

```
document order   A-alpha  A-bravo  A-charlie  B-alpha  B-bravo  B-charlie
x                    127      335        544      127      335       544
tag              path
```

Series-major, exactly as the diverging docstring says.

Driving the real `SegmentedTrace` over that DOM order, reading which element each cell highlights:

```
no domMapping (what the adapter emits today)
  A/alpha  -> A-bravo     A/bravo  -> B-alpha     A/charlie -> B-charlie
  B/alpha  -> A-alpha     B/bravo  -> A-charlie   B/charlie -> B-bravo

domMapping {order:'row'}
  A/alpha  -> A-alpha     A/bravo  -> A-bravo     A/charlie -> A-charlie
  B/alpha  -> B-alpha     B/bravo  -> B-bravo     B/charlie -> B-charlie
```

[...]

PR #1016, opened 2026-08-18
fix(chartjs): order a chart's bar categories by the axis, not by the data

### LINKED ISSUE #1015: A Chart.js bar chart on a reversed category axis is read exactly backwards
A Chart.js bar chart with `scales.x.reverse: true` draws its categories from the far end, but the adapter emits them in the order they were written. The chart loads, navigates, highlights nothing wrong (there is nothing to highlight — Chart.js is canvas), and announces a complete reading that is the mirror image of what is on screen.

This is the same defect #1010 found in the matrix chart, in the family next door. #1011 fixed the matrix chart by asking the scale instead of the data; the bar builders were not part of that change.

## Measured

Chart.js 4.5.1 in Chromium, `labels: ['alpha', 'bravo', 'charlie']`, reading each bar's own resolved position off `getDatasetMeta(0).data[i].x`:

```
vertical, plain            left→right: alpha, bravo, charlie
vertical, x reverse        left→right: charlie, bravo, alpha
horizontal (indexAxis y)   top→bottom: alpha, bravo, charlie
horizontal, y reverse      top→bottom: charlie, bravo, alpha
```

`chart.data.labels` stays `["alpha","bravo","charlie"]` in every one of those. `chart.options.scales.<axis>.reverse` resolves readably after construction — `false` when undeclared, `true` when set — which is the same signal `drawnCategoryOrder` already reads for the matrix chart.

The flip is not confined to the simple bar. Same measurement, same result:

```
dodged bar, x reverse      left→right: charlie, bravo, alpha
stacked bar, x reverse     left→right: charlie, bravo, alpha
floating bar, x reverse    left→right: charlie, bravo, alpha
dot (showLine: false)      left→right: charlie, bravo, alpha
```

## What the adapter emits

Driving `extractChartData` on the same three charts:

```
bar,        scales.x.reverse: true   →  x: ["alpha","bravo","charlie"]
dodged_bar, scales.x.reverse: true   →  x: ["alpha","bravo","charlie"]
bar (horz), scales.y.reverse: true   →  y: ["alpha","bravo","charlie"]
```

Exactly the reverse of the drawn order in each case.

The cause is that both bar builders index the labels directly and never consult the scale:

`singleDatasetToBarLayer` (`src/adapters/chartjs/extractor.ts:907`):

```ts
dataset.data.forEach((value, i) => {
  ...
  points.push(isHorizontal
    ? { x: num, y: labels[i] ?? i }
    : { x: labels[i] ?? i, y: num });
});
```

`extractSegmentedBarLayers` (`src/adapters/chartjs/extractor.ts:943`):

```ts
for (let j = 0; j < numCategories; j++) {
  ...
  groupPoints.push(isHorizontal
    ? { x: num, y: labels[j] ?? j, z: dataset.label ?? '' }
    : { x: labels[j] ?? j, y: num, z: dataset.label ?? '' });
}
```

## Why the fix is one-sided

Two things that had to be checked before calling this adapter-side, because the same reasoning failed in #1001:

- **`BarTrace` and `SegmentedTrace` do not re-sort.** Both take `layer.data` as given, unlike `ScatterTrace`, which sorts by ascending x in its constructor and so cannot be steered from an adapter at all (#1007). So the emitted order really is the announced order here.
- **Chart.js paints to canvas**, so no bar layer emits `selectors`. There is no highlight index keyed to the category order, and therefore none of the both-halves-must-move hazard that made #988/#990 and #1000 delicate. Reordering the payload is the whole fix.

## Scope

`extractDotLayers` builds its layers by calling `singleDatasetToBarLayer`, so a Cleveland dot plot is fixed by the same change rather than separately.

The floating-bar readings — gantt, waterfall, dumbbell — sit on the same category axis and flip with it, as measured above. Their models do not re-sort either. They are worth doing in the same change, with the one extra care that `GanttData.lanes` is a second array that has to move with `points`.

**Not in scope, deliberately:** line and scatter. A Chart.js line flips too (measured: `line, x reverse` → `charlie, bravo, alpha`), but ordering for those two traces is the subject of #1007, and picking an answer for Chart.js alone would settle by accident a question that issue is holding open for a maintainer.

**Not established:** whether the same defect exists for Chart.js's radar, pie or box readings. `reverse` on those is either meaningless or unmeasured, and nothing here should be taken as covering them.

jooyoungseo:
Closes #1015

A Chart.js bar chart with `scales.x.reverse: true` is drawn from the far end, but the adapter emitted the categories in the order they were written. The chart loaded, navigated and announced a complete reading that was the mirror image of what is on screen. #1011 fixed exactly this for the matrix chart by asking the scale instead of the data; the bar builders were not part of that change.

## Measured

Chart.js 4.5.1 in Chromium, `labels: ['alpha', 'bravo', 'charlie']`, reading each bar's own resolved position off `getDatasetMeta(0).data[i]`:

```
vertical, plain            left→right: alpha, bravo, charlie
vertical, x reverse        left→right: charlie, bravo, alpha
horizontal (indexAxis y)   top→bottom: alpha, bravo, charlie
horizontal, y reverse      top→bottom: charlie, bravo, alpha

dodged bar, x reverse      left→right: charlie, bravo, alpha
stacked bar, x reverse     left→right: charlie, bravo, alpha
floating bar, x reverse    left→right: charlie, bravo, alpha
dot (showLine: false)      left→right: charlie, bravo, alpha
```

`chart.data.labels` stays `["alpha","bravo","charlie"]` in every one of those, and `chart.options.scales.<axis>.reverse` resolves readably after construction — `false` when undeclared, `true` when set.

Driving `extractChartData` on the same charts before the fix:

```
bar,        scales.x.reverse: true   →  x: ["alpha","bravo","charlie"]
dodged_bar, scales.x.reverse: true   →  x: ["alpha","bravo","charlie"]
bar (horz), scales.y.reverse: true   →  y: ["alpha","bravo","charlie"]
```

Exactly the reverse of the drawn order in each case.

## The change

One helper, `drawnCategoryPositions(chart, count)`, which asks `categoryAxis(chart)` — the axis the categories are actually on, which `indexAxis: 'y'` moves to y — and hands back the positions to read in. Every bar-family builder iterates it instead of `0..n-1`:

| commit | builders | trace types |
| --- | --- | --- |
| 1 | `singleDatasetToBarLayer`, `extractSegmentedBarLayers` | `bar`, `dot`, `dodged_bar`, `stacked_bar`, `diverging_bar` |
| 2 | `extractGanttLayer`, `extractWaterfallLayer`, `extractDumbbellLayer` | `gantt`, `waterfall`, `dumbbell` |

`extractDotLayers` builds its layers by calling `singleDatasetToBarLayer`, so the dot plot is fixed by the bar fix rather than separately; there is a test pinning that so it stops being an accident.

(An earlier revision of this description also listed `normalized_bar`. That was wrong — `extractSegmentedBarLayers` has one caller, which chooses between `DIVERGING`, `STACKED` and `DODGED`, and the only `NORMALIZED` in this adapter is `NORMALIZED_AREA` on the line path, which this change does not touch.)

## Why reordering the payload is the whole fix

Two things checked rather than assumed, because the same reasoning failed in #1001:

- **The models do not re-sort.** `BarTrace`, `SegmentedTrace`, `GanttTrace`, `WaterfallTrace` and `DumbbellTrace` all take `layer.data` as given — unlike `ScatterTrace`, which sorts by ascending x in its constructor and so cannot be steered from an adapter at all (#1007).
- **Chart.js paints to canvas**, so none of these layers carry `selectors`. No highlight index is keyed to the category order, and none of the both-halves-must-move hazard of #988/#990 or #1000 applies here.

The gantt is the one case with a second array to keep in step: `GanttData.lanes` names the lanes `points` holds, position for position, so both are built from the same order.

Nothing about a waterfall's arithmetic depends on the order it is walked in. `isWaterfallSequence` has already run on the written order, where the chaining that identifies a waterfall lives, and a step's `start`, `end`, `delta` and `kind` are properties of the step rather than of its neighbours. A test pins that a reversed waterfall is still recognised as one.

## Tests

`test/adapters/chartjs/reversedBarOrder.test.ts` (14) and `test/adapters/chartjs/reversedFloatingBarOrder.test.ts` (10).

Falsification — each mutation applied alone, against both files:

| mutation | red |
| --- | --- |
| helper returns `0..n-1` unconditionally (i.e. the pre-fix behaviour) | 10 |
| helper always asks the `x` scale rather than the category axis | 2 — both horizontal cases |
| gantt `lanes` left in written order while `points` are reordered | 2 — including "keeps every lane under its own name" |

The cases that stay green under the first mutation are the ones pinning that an *un*reversed chart is untouched, which is what they are for.

Only the *relative* order is asserted. Which end of a horizontal bar list MAIDR reads first is a separate convention, and this change does not touch the plain case that establishes it.

## Verified

- `npx tsc --noEmit` — clean
- `npx eslint --fix` on all four files — clean
- `npx jest --selectProjects unit` — 333 suites, all passing
- `npm run build` — exit 0

## Not done

- **No E2E.** There is no Chart.js reversed-axis spec, and the existing Chart.js specs do not cover a reversed axis for any trace type. The reviewer on #1014 suggested a cross-producer E2E smoke test for this whole family of fixes rather than one per PR; that suggestion still stands and this PR does not act on it.
- **No screen-reader check.**
- **Line and scatter are deliberately excluded.** A Chart.js line flips too — measured `line, x reverse` → `charlie, bravo, alpha` — but ordering for those two traces is the subject of #1007, and settling it for Chart.js alone would answer by accident a question that issue is holding open for a maintainer. (Measurements posted there since: Chart.js is the one producer with no both-halves hazard, because it paints to canvas and emits no path to keep in step.)
- **Radar, pie and box are not covered.** `reverse` on those is either meaningless or unmeasured, and nothing here should be read as covering them.


jooyoungseo:
Acted on the review's coverage note, and correcting a mistake in my own PR body that chasing it turned up.

**`normalized_bar` is not a reading this adapter emits.** My table listed it under commit 1. `extractSegmentedBarLayers` is called from exactly one place:

```ts
const traceType = isStacked(chart)
  ? (isDivergingSplit(data.datasets) ? TraceType.DIVERGING : TraceType.STACKED)
  : TraceType.DODGED;
```

and the only `NORMALIZED` in the whole Chart.js adapter is `NORMALIZED_AREA`, on the line/area path this change does not touch. So the trace list for commit 1 is `bar`, `dot`, `dodged_bar`, `stacked_bar`, `diverging_bar` — five, not six. PR body corrected.

**`diverging_bar` was a real gap and now has its own case** (`fcacdac`). It is worth covering in its own right rather than inheriting from the dodged and stacked cases, because the type is chosen by `isDivergingSplit` reading the *values* — so a reordering that lost a sign would change what the chart is announced as, not merely what order it is read in. The test stacks two datasets sitting wholly on opposite sides of the baseline and asserts the type survives alongside the order:

```ts
expect(type).toBe('diverging_bar');
expect(groups.map(g => g.map(p => p.x))).toEqual([DRAWN, DRAWN]);
expect(groups[0].map(p => p.y)).toEqual([-3, -2, -1]);
expect(groups[1].map(p => p.y)).toEqual([6, 5, 4]);
```

14 tests now in `reversedBarOrder.test.ts`, 10 in `reversedFloatingBarOrder.test.ts`. `tsc --noEmit` and `eslint --fix` clean.

On the other two notes: the `singleDatasetToBarLayer` JSDoc is left as it is on purpose — the reasoning belongs with `drawnCategoryPositions`, and repeating it at five call sites is five places to drift. The E2E gap stands as described.

---

Issue #1017, opened 2026-08-18
A Recharts chart with a reversed axis is read backwards, and the adapter cannot see that it is


jooyoungseo:
`<XAxis reversed />` is Recharts' own way of drawing a chart right-to-left. The chart draws that way, MAIDR announces it the other way, and nothing errors.

What makes this one different from the reversed-axis fixes already landed (#995/#1000 Highcharts, #1008/#1009 Highcharts heatmap, #1010/#1011 Chart.js matrix, #1015 Chart.js bars) is that the Recharts converter is not merely *failing* to consult the axis. It is never given it.

## Measured

`examples/recharts/examples/BarChartExample.tsx` with `reversed` added to its `<XAxis>`, nothing else changed, rendered by `npm run dev:recharts` in Chromium. Data is `Q1: 4200, Q2: 5800, Q3: 3900, Q4: 7100`.

Reading each `.recharts-bar-rectangle` in **DOM order**, with its own bounding box:

```
DOM index   x     height    → which datum (by height)
    0      731      34        Q1  (4200)
    1      598      46        Q2  (5800)
    2      466      31        Q3  (3900)
    3      333      57        Q4  (7100)
```

Two facts fall out, and they point in opposite directions:

- **The drawing is reversed.** Sorted by x, left→right is `Q4, Q3, Q2, Q1`. `reversed` does what it says.
- **The DOM is not.** The rectangles are emitted in data order — `Q1` is still the first `.recharts-bar-rectangle` in the document, sitting at the far right of the plot.

The converter emits the categories in data order too, so today the payload and the DOM agree with each other and both disagree with the chart. The reader is told `Q1` first while the leftmost bar is `Q4`.

## Why this is not the same fix as the others

Every previous one in this family was "the adapter had the axis and did not ask it". Here the adapter provably does not have it. `MaidrRecharts` hands the converter its **props only**:

```tsx
const maidrData = useMemo(
  () => convertRechartsToMaidr({
    id, title, subtitle, caption, data, chartType, xKey, yKeys, /* … */ selectorOverride,
  }),
  [/* … */],
);

return (
  <Maidr data={maidrData}>
    {subplots ? renderPanelGrid(subplots, columns, children) : children}
  </Maidr>
);
```

[...]

Issue #1004, opened 2026-08-18
Resolving a selector inserts a sibling, so a list of positional selectors is only correct read backwards


jooyoungseo:
`Svg.selectElement` does not just find an element — it inserts a hidden clone immediately after the one it found:

```ts
const clone = this.markOwned(element.cloneNode(true) as T);
clone.setAttribute(Constant.VISIBILITY, Constant.HIDDEN);
element.insertAdjacentElement(Constant.AFTER_END, clone);
return clone;
```

So resolving one selector changes the answer to the next. For a **positional** selector — `:nth-child(N)`, `:nth-of-type(N)` — the sibling count is the whole meaning, and every position after the one just resolved has shifted by one.

## Measured

Three sibling `<path class="pt">` elements, `p1 p2 p3`, addressed as `#c .grp .pt:nth-child(k)` and resolved through `Svg.selectElement`:

```
ascending   1, 2, 3   →   p1   p1   p1
descending  3, 2, 1   →   p3   p2   p1
```

Ascending returns the **first element three times**: each clone lands in front of the next position, so `:nth-child(2)` matches the clone of `p1`, and so on. Descending is correct, because a clone inserted after position `k` cannot disturb any position below `k`.

The failure is silent and total, and it looks exactly like a working highlight — an element is found, it is a real bar of the chart, it is highlighted. It just does not move.

## Why this is not theoretical

`test/model/box.ts` already works around it, in as many words:

```ts
// Phase 1: Collect all original elements without cloning (prevents nth-child() DOM shifts)
```

and `src/adapters/d3/selectors.ts` avoids positional selectors for the same reason:

```
 * `:nth-child(N)`, which any later DOM insertion would shift
```

as do the Vega-Lite and Recharts selector modules. So the hazard is known and has been handled three or four times locally — but `SegmentedTrace.mapGridToSvgElements` and `AbstractBarPlot.mapToSvgElements` both resolve their lists one at a time, cloning as they go.

Nothing is broken *today* by that, as far as I measured, and the reason is luck rather than design: the one adapter emitting positional selectors into those paths (Highcharts, #995) emits them for a reversed axis, so the list comes out in descending order and survives. Emit the same selectors for an unreversed chart — ascending — and every cell of the layer highlights the first bar.

The plotly per-point selectors (#989) are safe for a different accidental reason: their positional part sits on an **ancestor** (`g.point:nth-of-type(k) > path`), so the clone lands inside the group rather than beside it and the count is untouched.

[...]

Issue #1003, opened 2026-08-18
A segmented layer's DOM pairing silently depends on the mark's tag name, and the two branches default oppositely


jooyoungseo:
This is the thing that made my analysis in #1001 wrong, so it is worth writing down whether or not the Highcharts half is fixed. It will make the next person's analysis wrong the same way.

## The asymmetry

`SegmentedTrace.mapToSvgElements` picks its pairing strategy from the tag of the first resolved element, and the two branches disagree about what "nothing declared" means.

Path branch:

```ts
if (isRowMajor || !this.layer.domMapping) {   // ← row-major when nothing is declared
```

Rect branch:

```ts
const isRowMajor = this.layer.domMapping?.order === 'row';
if (isRowMajor) { ... } else { /* column-major (default) */ }
```

Same layer, same DOM order, same absent `domMapping` — **opposite answers**, decided by whether the library drew `<path>` or `<rect>`.

There is also a third case with no branch at all: an element that is neither leaves `svgElements` empty, and the layer silently gets no highlight rather than a wrong one.

## Why this is not academic

Highcharts changed its bar mark tag between majors. Measured, same chart and same adapter, three versions:

| Highcharts | mark | DOM order | branch | default | pairing |
| --- | --- | --- | --- | --- | --- |
| 10.3.3 | `rect` | series-major | rect | column-major | **wrong** |
| 11.4.8 | `path` | series-major | path | row-major | correct |
| 12.6.0 | `path` | series-major | path | row-major | correct |

The DOM order never changed — only the tag. So `convertStackedBar`, `convertDodgedBar` and the normalized layer they share pair correctly on v11+ and incorrectly on v10, with no code change on either side. `convertDivergingBar` declares `domMapping: { order: 'row' }` and is right on both, which reads very much like it was written when v10 was current.

`src/adapters/highcharts/selectors.ts` deliberately supports both tags —

```
 * The element tag varies: `<rect>` for square corners, `<path>` for rounded corners
 * (the default in Highcharts v11+). Matching by class only handles both cases.
```

— so the adapter treats `<rect>` as a supported shape while the pairing quietly treats it as a different contract. `borderRadius: 0` does *not* bring `<rect>` back on v12, measured, so today the rect side is reachable only on older Highcharts.

## Why it is hard to catch

[...]

PR #1019, opened 2026-08-18
fix(victory): read a chart in the order an inverted axis draws it

### LINKED ISSUE #1018: A Victory chart with invertAxis is read backwards, and the adapter already has what it needs to know
`<VictoryAxis invertAxis />` draws the categories from the far end. MAIDR announces them in the order the data was written. Nothing errors; the reading is simply the mirror image of the chart.

This is #1017's defect in a neighbouring adapter, but with the opposite conclusion about what a fix would take. Recharts hands its converter props only, so the axis is genuinely out of reach. **Victory hands its adapter the children**, and the adapter already walks them for axis properties — so nothing new has to be invented here.

## Measured

`examples/victory/examples/BarChartExample.tsx` with `invertAxis` added to its independent `<VictoryAxis>`, nothing else changed, rendered by `npm run dev:victory` in Chromium. Data is `Q1: 4200, Q2: 5800, Q3: 3900, Q4: 7100`.

```
x tick labels, left→right:  Q4, Q3, Q2, Q1

marks in DOM order    centre x    height    → which datum
        0               631        104        Q1  (4200)
        1               530        144        Q2  (5800)
        2               430         97        Q3  (3900)
        3               329        176        Q4  (7100)
```

The axis labels confirm the drawing is inverted. The marks do not follow: `Q1` is still the first `path[role="presentation"]` in the document, sitting at the far right. The payload is emitted in data order too, so the reader is told `Q1` first while the leftmost bar is `Q4`.

## Why this one needs no new affordance

`MaidrVictory` passes `children` straight into the adapter:

```tsx
const maidrData = useVictoryAdapter({ id, title, subtitle, caption, children, layout }, containerRef);
```

and `extractAxisLabels` already walks them, reading a `<VictoryAxis>`'s own props to decide which axis it is:

```ts
Children.forEach(children, (child) => {
  if (!isValidElement(child)) return;
  const name = getVictoryDisplayName(child.type);
  if (name !== 'VictoryAxis') return;

  const props = child.props as Record<string, unknown>;
  ...
  if (props.dependentAxis) {
    result.y = label;
  } else {
    result.x = label;
  }
});
```

`props.invertAxis` sits right beside `props.dependentAxis` on the same element, in a loop that is already running. That is the whole of the "how do we know" half — no prop for the author to remember, and no introspection machinery that does not already exist.

## Both halves still have to move

Victory tags its marks rather than selecting them structurally, but with **one shared attribute**:

```ts
for (const el of matched) {
  el.setAttribute(attrName, '');
  claimed.add(el);
  ...
}

return `${scope}[${attrName}]`;
```

confirmed by the measurement — all four marks carry `data-maidr-victory-0`. One selector for the series, resolved positionally, i.e. in DOM order, which is data order. So reversing the payload alone would announce `Q4` and outline `Q1`.

The fix therefore has to stamp per-index attributes and return a `string[]`, in the order the payload is emitted. `stampOrderedSelectors` in the d3 binder is the same shape and can be read across; because the attributes would be distinct rather than positional, #1004's descending-order hazard does not apply.

## Scope

`tagLayerElements` dispatches to seven taggers — area, stacked area, error bar, polar area, line, candlestick, box plot, and discrete. Only the discrete one is measured here, and only for `VictoryBar`. Which of the others `invertAxis` reaches, and whether a line's single `<path>` has the same trace-order-not-drawn-order property measured for plotly and Recharts on #1007 and #1017, is **not established**.

## Not established

- Every Victory reading other than a plain vertical `VictoryBar`.
- `invertAxis` on the dependent axis.
- Google Charts' `hAxis.direction: -1`, AnyChart, Tableau. (Observable already handles reversed axes in several places and looks to be out of scope.)

jooyoungseo:
Closes #1018

`<VictoryAxis invertAxis />` draws the categories from the far end. Victory keeps rendering its marks in data order, and the adapter emitted them that way too, so the chart was announced as the mirror image of what it draws. Nothing errored.

## Measured

`examples/victory/examples/BarChartExample.tsx` with `invertAxis` added to its independent `<VictoryAxis>`, nothing else changed, rendered by `npm run dev:victory` in Chromium. Data `Q1: 4200, Q2: 5800, Q3: 3900, Q4: 7100`.

```
x tick labels, left→right:  Q4, Q3, Q2, Q1

marks in DOM order    centre x    height    → which datum
        0               631        104        Q1
        1               530        144        Q2
        2               430         97        Q3
        3               329        176        Q4
```

The axis labels say the drawing is inverted. The marks say the DOM is not: `Q1` is still first in the document, at the far right.

## How the adapter learns it

Nothing new. `MaidrVictory` hands the adapter the chart's `children`, and `extractAxisLabels` already walked its `<VictoryAxis>` elements reading their own props:

```ts
if (props.dependentAxis) {
  result.y = label;
} else {
  result.x = label;
}
```

`invertAxis` sits on the same element, so it is read in the same pass — the function becomes `extractAxisInfo` and returns both. One real change of behaviour in that walk: it used to `return` early when an axis carried no `label`, and an inverted axis commonly carries none, so the flag would have been missed. There is a test for exactly that.

This is where Victory and Recharts part company. #1017 is the same defect in the Recharts adapter, which is handed props only and cannot see its `<XAxis>` at all — so that one still needs a decision about what affordance to add, and this PR does not touch it.

## Both halves move together

`tagDiscreteElements` stamped **one shared attribute** on every mark and returned a single selector:

```ts
for (const el of matched) {
  el.setAttribute(attrName, '');
  ...
}
return `${scope}[${attrName}]`;
```

confirmed by the measurement — all four marks carried `data-maidr-victory-0`. The model resolves that positionally, i.e. in DOM order, i.e. data order. So reversing the payload alone would have announced `Q4` and outlined `Q1`: #988/#990 and #1000 again.

A reversed layer now has each mark named individually, by the position it holds in the payload:

```ts
// `matched` is in DOM order, which is data order; the payload was turned
// round, so payload position i is the mark standing at the other end.
return matched.map((_, i) => {
  const own = `${attrName}-${i}`;
  matched[matched.length - 1 - i].setAttribute(own, '');
  return `${scope}[${own}]`;
});
```

Distinct attributes rather than `:nth-child`, so the list is order-independent and #1004's descending-only rule does not apply. An unreversed chart keeps its single selector — there is nothing to re-point, and a list of four would be four things able to drift instead of one.

**Verified in the browser, not only in jsdom.** With the fix in place and the same example:

```
marks in DOM order (centre x, height, maidr attrs)
  cx=631 h=104   data-maidr-victory-0  data-maidr-victory-0-3
  cx=530 h=144   data-maidr-victory-0  data-maidr-victory-0-2
  cx=430  h=97   data-maidr-victory-0  data-maidr-victory-0-1
  cx=329 h=176   data-maidr-victory-0  data-maidr-victory-0-0
```

Selector `-0` — the one the reading now leads with — lands on the mark at `cx=329`, the leftmost, height 176, which is `Q4`. That is the bar a sighted reader meets first.

## Scope

`REVERSIBLE_KINDS` is `bar`, `dot`, `histogram`, `waterfall` — the readings that draw one mark per datum, announce `layer.data` as it arrives, and are tagged by `tagDiscreteElements`. All four are read by `AbstractBarPlot` or `WaterfallTrace`, both of which already document a `string[]` as "one element per point".

Left alone, each for its own reason, all recorded in a docblock beside the set:

| kind | why not |
| --- | --- |
| `line`, `area`, `stackedArea`, `polarArea` | one `<path>`, and `LineTrace` pairs vertex `i` with point `i`. Both are in data order and agree today; reversing the payload alone breaks that agreement rather than fixing anything, and the adapter cannot rewrite the path. #1007's question. |
| `scatter` | `ScatterTrace` sorts by ascending x whatever arrives. (A `VictoryScatter` over *categories* is `dot`, and is in the set.) |
| `segmented`, `diverging` | one mark per datum, but flattened from a single selector under a row/column-major convention that is itself unsettled (#1003). |
| `box`, `candlestick`, `errorBar` | taggers of their own. |
| `pie` | no axis to invert. |

The line, area, scatter and stacked readings were all measured to flip with the axis — the measurements are on #1018. They are excluded because a fix here could not be made to hold, not because they are unaffected.

## Tests

`test/adapters/victory/invertedAxisOrder.test.ts`, 11 cases. Falsification, each mutation applied alone:

| mutation | red |
| --- | --- |
| never set `invertedIndependent` (the pre-fix behaviour) | 5 |
| read `invertAxis` on either axis rather than the independent one | 1 — the inverted-dependent-axis case |
| reverse the payload but keep the single shared selector | 3 |
| name the marks in DOM order rather than payload order | 2 |

## Verified

- `npx tsc --noEmit` — clean
- `npx eslint --fix` on all four files — clean
- `npx jest --selectProjects unit` — 334 suites, 4899 tests, all passing
- Browser verification above

## Not done

- **No E2E.** There is no Victory reversed-axis spec, and the existing Victory specs do not cover one.
- **No screen-reader check.**
- **`invertAxis` on the dependent axis is deliberately ignored**, with a test. It turns the magnitudes over and moves no category, so reading it here would reorder a chart that did not move.
- **`VictoryBoxPlot`, `VictoryCandlestick`, `VictoryErrorBar` and polar are not measured** under `invertAxis`, and are untouched.

---


jooyoungseo:
Acted on all four review notes (`794a746`). Two were worth more than polish.

**1. The redundant shared attribute — taken, and it was not quite harmless.** On a reversed layer the shared `data-maidr-victory-N` names the same marks in the *opposite* order to the reading, so it was a second, wrong answer sitting next to the right one, waiting for something to resolve it. Now skipped when the marks are named individually; `claimed.add(el)` still runs, since claiming is what stops a later layer taking the same marks.

Worth saying how this was caught rather than just made: I removed the guard again and **every test still passed**. A change nothing can falsify is not a change anyone can rely on, so it now has a test — and re-running the same mutation turns exactly that one red.

```ts
expect(svg.querySelectorAll('[data-maidr-victory-0]')).toHaveLength(0);
expect(svg.querySelectorAll('[data-maidr-victory-0-0]')).toHaveLength(1);
```

**2. `readInDrawnOrder` exported for no caller — taken.** It was leftover intent, not design; the test drives it through `extractVictoryLayers`, which is the public surface. Unexported.

**3. Horizontal untested — taken, as tests rather than as a caveat.** `horizontal` swaps which *screen* axis the categories run along but not which `<VictoryAxis>` declares them — the independent one is still the element without `dependentAxis` — so the fix should not depend on orientation. Two cases now say so instead of me asserting it: a horizontal chart with `invertAxis` reads `['Q4','Q3','Q2','Q1']` and keeps `orientation: 'horz'`, and one without is untouched.

**4. The stale `@returns` — taken, and it is mine.** You are right that it predates this PR, but widening the return type to include `string[]` is what made it wrong in a new way, so it belongs here:

> Whatever names this layer's marks: one CSS selector for a series the model can walk positionally, a `string[]` when the marks have to be named one by one, or the structured selector a box or a candle needs.

15 tests now. Falsification re-run after the changes:

| mutation | red |
| --- | --- |
| never set `invertedIndependent` (pre-fix behaviour) | 5 |
| read `invertAxis` on the dependent axis instead | 8 |
| reverse the payload but keep the single shared selector | 3 |
| name the marks in DOM order rather than payload order | 2 |
| keep stamping the shared attribute on a reversed layer | 1 |

`tsc --noEmit` and `eslint --fix` clean; `jest --selectProjects unit` 334 suites / 4903 tests passing.

On your note about not being able to run the suite in the review sandbox — understood, and the CI run on this PR is the check to trust, not my summary.

---

Issue #1021, opened 2026-08-18
An AnyChart bar chart on an inverted ordinal scale is read backwards


jooyoungseo:
`chart.xScale().inverted(...)` draws an AnyChart bar or column chart's categories from the far end. The adapter emits them in data order regardless, so the chart is announced as the mirror image of what it draws.

This is the same defect as #1015 (Chart.js), #1017 (Recharts) and #1018 (Victory), and like Victory the adapter is already holding everything it needs to fix it.

## Measured

AnyChart 8.13.0 served locally, `examples/anychart/bar.html` with one line added and nothing else changed, rendered in Chromium. Data `Sat: 87, Sun: 76, Thu: 62, Fri: 19`, so a bar is identifiable by its size.

**`anychart.bar` (horizontal), `xScale().inverted(false)`**

```
category tick labels, top→bottom:  Fri, Thu, Sun, Sat

marks in DOM order    centre y    width    → which datum
        0               442        535       Sat  (87)
        1               367        467       Sun  (76)
        2               292        381       Thu  (62)
        3               217        116       Fri  (19)

emitted:  [{"x":"Sat"},{"x":"Sun"},{"x":"Thu"},{"x":"Fri"}]
```

**`anychart.column` (vertical), `xScale().inverted(true)`**

```
category tick labels, left→right:  Fri, Thu, Sun, Sat

marks in DOM order    centre x    height   → which datum
        0               675        260       Sat  (87)
        1               521        227       Sun  (76)
        2               367        185       Thu  (62)
        3               213         56       Fri  (19)

emitted:  [{"x":"Sat"},{"x":"Sun"},{"x":"Thu"},{"x":"Fri"}]
```

In both, the reading leads with `Sat` while the chart puts `Sat` at the far end.

## The defaults differ by chart type, which is why nobody has hit this yet

Reading `chart.xScale().inverted()` on a **freshly constructed** chart, before touching anything:

```
anychart.bar(...)    → inverted() === true
anychart.column(...) → inverted() === false
```

`bar` is horizontal, and inverting its ordinal scale is what puts the first category at the *top* — which is the conventional way round, and which happens to agree with the adapter's data order. `column` defaults the other way and also agrees. So the adapter is accidentally correct on every default chart, and wrong only when an author states the opposite. That is exactly the shape that survives a long time unnoticed.

It also means my first attempt at this measurement was a false negative: I set `inverted(true)` on a `bar` chart and nothing changed, because it was already true.

[...]

Issue #1012, opened 2026-08-18
An amCharts heatmap with an inversed x renderer reads its columns backwards


jooyoungseo:
Third result of the #1008 sweep, and the last measurable one. The column half of #981/#982.

## What

`extractHeatmapData` asks the y axis which way it is drawn and never asks the x axis:

```ts
function drawsFirstRowAtTop(series: AmXYSeries): boolean {
  const renderer = settingOf(series.get('yAxis'), 'renderer');
  return settingOf(renderer, 'inversed') === true;
}
...
if (drawsFirstRowAtTop(series)) {
  return { x: xLabels, y: yLabels, points };
}
return { x: xLabels, y: [...yLabels].reverse(), points: [...points].reverse() };
```

`xLabels` is the order the data items arrived in, on both paths.

## Measured

amCharts 5 bundled to an IIFE with esbuild, rendered in Chromium, a 3×2 `ColumnSeries` grid on two `CategoryAxis`, reading each column's own global x:

```
x inversed=false   renderer setting=false   columns left→right: c0, c1, c2   dataItem order: c0, c1, c2
x inversed=true    renderer setting=true    columns left→right: c2, c1, c0   dataItem order: c0, c1, c2
```

The data items arrive in the same order either way. Only the drawing moves, and the adapter follows the data items — so an inversed x renderer is announced left to right while it is drawn right to left.

The unreversed case is right because the two happen to agree, exactly as the y axis' unreversed case is *wrong* for the opposite reason (amCharts counts a category axis from the bottom, which is why #981 had to turn the rows over).

## Severity

The one-sided case, like #976 and #982: `buildHeatmapLayer` emits no `selectors` because amCharts paints to canvas, so there is no highlight index to desynchronise, and nothing is mis-valued — the values are keyed to their labels through `valueMap`. Only the **order** is false, and with it arrowing, panning, the braille row and autoplay.

## Suggested

Symmetric with the row half, and adapter-side:

```ts
function drawsFirstColumnAtLeft(series: AmXYSeries): boolean {
  const renderer = settingOf(series.get('xAxis'), 'renderer');
  return settingOf(renderer, 'inversed') !== true;
}
```

then reverse `x` and each row of `points` when it is false. Note the polarity differs from the y helper — an unreversed **x** is already left-first, while an unreversed **y** is *not* top-first — so the two are not copy-paste of each other. Worth asking both in one place for the reason given on #1008: having the question written down once and applied once is how these column halves keep being missed.

[...]

Issue #955, opened 2026-08-17
Every horizontal Google Charts bar variant is silent: `horz` declared over an unswapped payload


jooyoungseo:
A Google Charts `BarChart` — the plain, standard horizontal bar chart — announces nothing on any bar. So do the stacked, dodged, diverging and funnel variants. The column variants are all fine.

## Reproduction

Running the adapter over a three-row table (`Apples 30 / Bananas 70 / Cherries 50`) and feeding the emitted layer to the real `TraceFactory`:

| declared type | orientation | `data[0]` | `audio.freq.raw` |
|---|---|---|---|
| `BarChart` | `horz` | `{x: "Apples", y: 30}` | **`null`** |
| `StackedBarChart` | `horz` | `{x: "Apples", y: 30, z: "Sales"}` | **`null`** |
| `DodgedBarChart` | `horz` | `{x: "Apples", y: 30, z: "Sales"}` | **`null`** |
| `DivergingBarChart` | `horz` | `{x: "Apples", y: 30, z: "Sales"}` | **`null`** |
| `FunnelChart` | `horz` | `{x: "Apples", y: 30}` | `1` (a ratio, not a count) |
| `ColumnChart` | `vert` | `{x: "Apples", y: 30}` | `30` ✓ |
| `StackedColumnChart` | `vert` | … | `30` ✓ |

Each `horz` row also trips the core's own guard (#950), verbatim:

> `[BarTrace] Layer "…" is horz, so its magnitude is read from `x` — but that field holds "Apples". A horizontal bar puts the magnitude in `x` and the category in `y`; see MaidrLayer.orientation. Every bar of this layer will be silent.`

The chart loads, navigates and highlights. Only the audio and the braille are empty, which is why nothing has flagged it.

## Cause

`buildBarLayer` and `buildSegmentedLayer` in `src/adapters/google-charts/converters.ts` both take an `orientation` parameter, pass it straight through to the emitted layer, and build their points the vertical way regardless:

```ts
const label = formatCellValue(dt, r, 0);
const value = numericValue(dt, r, dataCol);
data.push({ x: label, y: value });   // <- always this way round
…
orientation,                          // <- but this may say horz
```

Per `MaidrLayer.orientation` (#947), `horz` means the magnitude is read from `x`. So the key and the payload contradict each other, and `toBarValue('Apples')` is `NaN` — indistinguishable from a deliberate gap, hence the silence rather than an error.

[...]

PR #1023, opened 2026-08-18
fix(recharts, google-charts): read a chart in the order its reversed axis draws it

### LINKED ISSUE #1017: A Recharts chart with a reversed axis is read backwards, and the adapter cannot see that it is

### LINKED ISSUE #1020: A Google Charts category axis reversed with direction: -1 has no way to reach the adapter
Google Charts reverses an axis with `hAxis: {direction: -1}` (or `vAxis` for a bar chart). That is a **draw option**, and the Google Charts adapter is never given the draw options — so a chart drawn right-to-left is read left-to-right, and there is nothing in the adapter that could know otherwise.

This completes the reversed-axis sweep across producers: #995/#1000 Highcharts, #1008/#1009 Highcharts heatmap, #1010/#1011 Chart.js matrix, #1015/#1016 Chart.js bars, #1017 Recharts, #1018 Victory. Google Charts is the last one, and it is the case the codebase has already met and solved twice — just not for this option.

## The gap, from the source

The adapter's entry point takes the chart, its DataTable, its container and a config:

```ts
export function createMaidrFromGoogleChart(
  chart: GoogleChart,
  dataTable: GoogleDataTable,
  container: HTMLElement,
  options: GoogleChartAdapterOptions,
): Maidr
```

and the caller passes the draw options separately, to Google:

```js
bindMaidr(chart, data, container, { chartType: 'ColumnChart', title: '…' });
chart.draw(data, { legend: { position: 'none' }, width: 600, height: 400 });
```

`types.ts` says this in its own words, about a different option:

> Not a Google class: **the adapter never sees the draw options**, so a percent-stacked area has to be named by the caller.

So the established answer is: **the caller names it in `chartType`**. `'StackedAreaChart'`, `'NormalizedAreaChart'` and `'StackedColumnChart'` all exist for exactly this reason.

## This has already happened once, for the other axis

`'BumpChart'` is the precedent, and its docblock is precisely this problem:

> Not a Google class: the table is any multi-series line chart's, and **the axis direction that gives it away lives in the draw options**. Naming it is what inverts the pitch, so that a competitor climbing the table is heard climbing.

That is `vAxis: {direction: -1}` — the *value* axis — handled by giving the caller a name for the reading it produces. `hAxis: {direction: -1}` on the **category** axis has no equivalent name, and `direction` appears nowhere in `GoogleChartAdapterOptions`.

A bump chart read without its name is heard falling instead of climbing. A column chart drawn with a reversed category axis and read without a name is heard in the opposite order to the one on screen — the same class of wrongness, and the same shape of fix.

## Explicitly not measured

Unlike #1015, #1017 and #1018, **I have not measured this in a browser.** Google Charts loads from `https://www.gstatic.com/charts/loader.js`, and Chromium in this environment cannot reach it (`net::ERR_CONNECTION_RESET`, with and without the egress proxy configured), so there is no way to render a chart here and read where its bars land.

What is verified is the part the fix turns on, and it is verified from source rather than inferred:

- `createMaidrFromGoogleChart` receives no draw options — signature above, and the example in `index.ts` shows `chart.draw(data)` called separately.
- `GoogleChartAdapterOptions` has no `direction` and no reversed-axis name.
- The codebase already treats "this lives in the draw options" as a reason to add a caller-side name, twice.

What is **not** verified here:

- That `hAxis: {direction: -1}` reverses the drawn order. This is Google's documented behaviour for the option, but documented is not measured, and this issue should not be read as claiming otherwise.
- Which order Google emits the bar elements in. That decides whether a fix is one-sided (payload only) or has to move the selectors too, as it did for Recharts (#1017) and Victory (#1018) — where the DOM was measured to stay in **data** order while the drawing reversed. Anyone taking this on should measure that first; it is the half that has caught this work out before.

Somebody with a working Google Charts environment can settle both in about five minutes, and it would be worth doing before writing any code.

jooyoungseo:
Closes #1017
Closes #1020

The same defect in the last two adapters that had it: a reversed category axis draws the bars from the far end while the library goes on emitting its marks in data order, so the chart is announced as its own mirror image. Two commits, one per adapter, because the *reason* each was unfixed differs and so does the fix.

Both were filed as needing a decision. Neither turned out to.

---

# Recharts — #1017

`<XAxis reversed />`. Measured under `npm run dev:recharts` in Chromium, `Q1: 4200, Q2: 5800, Q3: 3900, Q4: 7100`:

```
bars in DOM order    x      height    → which datum
        0           731      153        Q1
        1           598      211        Q2
        2           466      142        Q3
        3           333      258        Q4
```

Left→right is `Q4, Q3, Q2, Q1`. The drawing is reversed; the DOM is not.

## The decision it makes

#1017 left open whether to take a declaration (like d3's `xOrder`) or introspect `children`, and said I would not pick. **Introspection**, because the two precedents are not symmetric:

- **d3** takes a declaration because a d3 scale lives in the author's closure. There is genuinely nothing to consult.
- **Victory** introspects (#1019) because its adapter is handed `children` and already walks `<VictoryAxis>`.

Recharts is Victory's case. The axis element is right there in the subtree this component renders — it just was not being looked at; `reversed` appears in the whole adapter once, in a comment. And this is an accessibility defect: a fix that has to be opted into leaves every chart already written still wrong, with its author no reason to suspect anything.

The brittleness cost is paid by answering **no** whenever the walk is unsure — an axis it cannot find, a subtree shaped in a way it does not recognise — so a chart is never turned round on a guess. Four tests cover that.

## Both halves

Recharts stamps nothing per point and the schema is built before the chart renders, so position is what there is. Measured: the bars are the only children of their own `<g class="recharts-layer">`, so `:nth-child` counts exactly them. `:nth-child` is inserted after the wrapper class, not after the first token, because the base selector carries the `#maidr-article-…` scope.

**The list runs descending, and that is load-bearing.** Payload position 0 is the last bar drawn, so the list is `nth-child(4), (3), (2), (1)` — the direction #1004 requires, since `Svg.selectElement` inserts its clone beside the match mid-resolution and an ascending positional list returns the first element every time.

Verified in the browser, resolving the emitted forms against the live DOM:

```
nth-child(4) → x=333  h=258     ← leftmost bar, Q4
nth-child(1) → x=731  h=153     ← rightmost bar, Q1
```

The walk lives in `src/adapters/recharts/axisDirection.ts` rather than in the component: importing the component to reach it drags the whole UI tree, and with it the ESM-only markdown stack, into any CommonJS test — the clash `.claude/rules/testing.md` documents.

---

# Google Charts — #1020

`hAxis: {direction: -1}`. I filed #1020 saying this could not be measured here and would need a caller-side name. **Both were wrong**, and the issue carries the correction.

## It is measurable

Chromium here cannot reach `gstatic.com`; Node can. Playwright bridges them — intercept the browser's requests, fulfil them from Node's fetch, and the real library runs unmodified:

```js
await page.route('**://*.gstatic.com/**', async (route) => {
  const res = await fetch(route.request().url());
  await route.fulfill({ status: 200, body: Buffer.from(await res.arrayBuffer()), … });
});
```

```
plain                    tick labels L→R: Sat, Sun, Thu, Fri
                         rects in DOM order: 133/213  226/186  318/152  411/46
hAxis: {direction: -1}   tick labels L→R: Fri, Thu, Sun, Sat
                         rects in DOM order: 411/213  318/186  226/152  133/46
```

## It needs nothing from the caller

The draw options really are never handed to the adapter. They do not have to be — the layout interface reports where each row landed:

```
plain                     getXLocation(0) = 162   getXLocation(3) = 439
hAxis: {direction: -1}    getXLocation(0) = 439   getXLocation(3) = 162
```

and `markBarElements` **is already calling that interface** to find the rects. It just never asked what the positions imply. The highlight half was already in place too: every rect is stamped `data-maidr-bar="<series>-<row>"`, so naming the marks in the payload's order needs no new attribute.

Verified against the real library with the fix built in:

```
plain           emitted [Sat, Sun, Thu, Fri]   selectors: one string       ← untouched
direction -1    emitted [Fri, Thu, Sun, Sat]   selectors: list of 4
                selector 0 → x=133 h=46        ← leftmost bar, Fri
                selector 3 → x=411 h=213       ← rightmost bar, Sat
```

---

## Tests

| file | cases |
| --- | --- |
| `test/adapters/recharts/reversedAxisOrder.test.ts` | 13 |
| `test/adapters/google-charts/reversedAxisOrder.test.ts` | 7 |

Falsification, each mutation applied alone:

**Recharts**

| mutation | red |
| --- | --- |
| never reversed (pre-fix) | 4 |
| count `nth-child` upward | 2 |
| walk only the top level of `children` | 3 |
| always ask the `x` axis whatever the orientation | 1 |
| rebuild a caller's own selector | 1 |

**Google Charts**

| mutation | red |
| --- | --- |
| never reversed (pre-fix) | 4 |
| compare the two locations the wrong way round | 6 |
| reverse the payload but keep the single selector | 3 |
| name the marks in row order | 2 |

The Recharts override case needed a second test to become load-bearing: my first one used a selector the builder rejects anyway, so the guard was invisible to the suite. The added case uses an override that *does* name `.recharts-bar-rectangle`, which is what the builder would otherwise renumber.

## Verified

- `npx tsc --noEmit` — clean
- `npx eslint --fix` on every touched file — clean
- `npx jest --selectProjects unit` — 337 suites, 4934 tests, all passing
- `npm run build` — exit 0
- Browser verification of both adapters, above

## Not done

- **No E2E**, and no screen-reader check, for either adapter.
- **Bar only.** Recharts' line, area and scatter flip too (measured on #1017) and are untouched: a line's vertex-to-point pairing is #1007's question, and the segmented builder's shared selector wants its own change. Google Charts' non-rect readings are likewise untouched, and a funnel has no axis to reverse.
- **Recharts multi-panel is not measured.** The panel scope survives by construction, since `:nth-child` is inserted into the base rather than replacing it — but that is reasoning, not a result.
- One unexplained `SyntaxError` appeared in the Google Charts verification page's console. The chart drew and the adapter produced the payload above, so it did not affect the measurement, but I have not chased it and am not claiming it is nothing.


jooyoungseo:
Widened this PR to carry the Google Charts fix (#1020) alongside the Recharts one (#1017), rather than opening a third. Same defect, same session, and both were filed as "needs a decision" — so keeping the reasoning in one place is more useful than splitting it. Title and description updated; the two commits stay separate, one per adapter.

The Google Charts half is worth flagging because **it retracts what I wrote when I filed #1020.** I said the behaviour could not be measured in this environment and that a fix would need a caller-side name. Neither was true:

- Chromium here cannot reach `gstatic.com`, but Node can. Playwright can route the browser's requests through Node, so the real library runs unmodified. That was an obstacle to my first approach, not to measurement.
- The draw options genuinely never reach the adapter — but `markBarElements` **already calls `chart.getChartLayoutInterface()`** to find the rects, and that interface reports where each row landed. It simply never asked what the positions imply.

So #1020 turned out to be the same shape as #1021: the answer was already in hand.

Full measurements and the browser verification of both adapters are in the description.

---

jooyoungseo:
All three findings taken (`373e708`). I checked each against the code before acting rather than taking them on the review's word, and all three held.

## 1. Google Charts asked the wrong axis on a horizontal `BarChart` — confirmed, and measured

You were right that this was dead code for `BarChart`, and right about why. I measured it rather than reasoning from the convention:

```
BarChart, plain                  labels top→bottom: Sat, Sun, Thu, Fri
                                 getYLocation(0) = 108   (3) = 293
BarChart, vAxis: {direction:-1}  labels top→bottom: Fri, Thu, Sun, Sat
                                 getYLocation(0) = 293   (3) = 108
```

and on that same reversed chart `getXLocation` read **116 then 127** — ascending, and unchanged from the plain chart. So the check answered "not reversed" on a chart that plainly was. `drawsCategoriesReversed` now takes the orientation and asks `getYLocation` when the categories are on y. The comparison itself is unchanged: Google numbers y downward, so "the last row lands before the first" is `last < first` on both axes.

This is the third time in this family that axis polarity has caught me — #1009 (Highcharts), #1012 (amCharts), and #1022, where a browser run caught the same class of error. Worth naming as a pattern rather than a coincidence: **whenever a fix asks "which way round is this axis", the answer differs by orientation, and a test that exercises one orientation will pass a rule that is wrong for the other.** My suite covered `ColumnChart` only, exactly as you noted.

## 2. Subplot mode — and your second-order point was the more important one

The missing `categoryAxisReversed` in `merged` is fixed. But your follow-up — that even a correctly-wired global verdict would apply one panel's axis to every other — is the part that would have shipped a subtler bug, so it is fixed properly rather than by propagation:

- `MaidrRecharts` now derives **one verdict per panel**, from that panel's own child element, via `Children.toArray(children).map(...)`.
- `buildSubplotMaidr` counts a flat row-major index and `buildPanelSubplot` looks up its own.

A test covers a mixed grid — `[reversed, not reversed]` — and asserts each panel reads its own way. Propagating a single verdict fails it.

## 3. Composed mode — confirmed, and the two builders now share one guard

`buildComposedLayers` never read the flag. Rather than add a second copy of the condition, both builders now call one `reversedBarSelectorsFor`: they emit the same `bar` reading from the same selector, and a rule living in one of them is a rule the other quietly does not have — which is precisely how this gap existed.

It answers **selectors only**, not the payload. The builders hand it a union of every payload shape, and narrowing that inside the helper made TypeScript apply the rule to the first union member alone; each caller reverses its own typed array instead.

## Falsification of the fixes

| mutation | red |
| --- | --- |
| always ask `getXLocation` (the bug you found) | 1 |
| drop `categoryAxisReversed` from the panel config | 2 |
| use one global verdict for every panel | 1 |
| composed mode does not read the flag | 1 |

One correction to my own test while adding the `BarChart` cases: both horizontal cases failed at first, including the control. That was my helper, not the fix — a horizontal layer carries its category in `y`, not `x` (#955). `categoriesOf` is now orientation-aware.

Recharts 18 cases, Google Charts 9. `tsc --noEmit`, `eslint` and 337 suites / 4941 tests all clean.

On your closing note — you were right not to take the "Verified" section as covering these paths. It did not, and the gaps were exactly where the tests were not.

---

Issue #749, opened 2026-08-06
A subplot with an empty `layers` array throws in `Subplot.activeTrace` and makes the entire figure inert


jooyoungseo:
Found while verifying r-maidr's patchwork output in a browser (xability/r-maidr#63). Reproduced against `main` at `f17d573` (v3.75.1). Sibling of #748 — same user-visible outcome, different path.

## Symptom

A subplot whose `layers` is `[]` throws while the plot context is being initialised, before any user interaction. MAIDR never attaches, so the whole figure is inert — Tab does nothing and no subplot is reachable, including well-formed ones.

```
Uncaught TypeError: Cannot read properties of undefined (reading '0')
    at get activeTrace              (maidr.js:341:87538)
    at get state                    (maidr.js:341:87770)
    at getStateWithFigurePosition   (maidr.js:341:88115)
    at get state                    (maidr.js:341:85572)
    at isFigureLevel                (maidr.js:341:88952)
    at initializePlotContext        (maidr.js:341:89018)
```

## Minimal reproduction

First subplot has no layers, second is well-formed:

```html
<!DOCTYPE html><html lang="en"><head><meta charset="utf-8">
<link href="lib/maidr-3.75.1/maidr.css" rel="stylesheet"/>
<script src="lib/maidr-3.75.1/maidr.js"></script></head><body>
<svg xmlns="http://www.w3.org/2000/svg" width="400" height="200" viewBox="0 0 400 200"
     maidr-data='{"id":"repro","subplots":[[
       {"id":"sp-1","layers":[]},
       {"id":"sp-2","layers":[{"id":"maidr-layer-1","type":"bar","title":"Good layer",
         "axes":{"x":{"label":"x"},"y":{"label":"y"}},
         "data":[{"x":"a","y":1},{"x":"b","y":2},{"x":"c","y":3}],
         "selectors":"#bars > rect"}]}
     ]]}'>
  <g id="bars">
    <rect id="b1" x="20" y="100" width="40" height="80" fill="#444"/>
    <rect id="b2" x="80" y="60" width="40" height="120" fill="#444"/>
    <rect id="b3" x="140" y="30" width="40" height="150" fill="#444"/>
  </g>
</svg></body></html>
```

Observed: the `TypeError` above, nothing announced. Give `sp-1` the same layer as `sp-2` and the identical file announces and navigates normally.

## Root cause

`src/model/plot.ts:405` builds traces from the layer list, so an empty `layers` yields an empty `traces`:

```ts
[...]

PR #1077, opened 2026-08-18
fix(model): stop one malformed subplot from taking down the whole figure

### LINKED ISSUE #1076: fix(model): one subplot without `layers` takes down the entire figure
## What happens

`Subplot`'s constructor reads `layers.length` unguarded (`src/model/plot.ts:388`):

```ts
const layers = subplot.layers;
this.size = layers.length;
this.primaryTitle = layers[0]?.title ?? '';
```

`Figure`'s constructor maps every cell through it eagerly (`src/model/plot.ts:142`):

```ts
this.subplots = subplots.map(row => row.map(subplot => new Subplot(subplot)));
```

So a single cell arriving without a `layers` key throws `TypeError: Cannot read properties of undefined (reading 'length')` during figure construction, the map aborts, and **no** part of the chart initialises — including the cells that were perfectly well-formed.

## Why this is the worst failure shape

The SVG is already in the DOM by then. The page looks finished, the chart is visibly a chart, and the key that should start navigation does nothing at all. There is no message and no degraded mode — a screen-reader user gets a decorative image and no way to know why.

Confirmed in Chromium against a 4.4.0 bundle, comparing a figure with one bare cell against the same figure with none:

| | uncaught error on activation | runtime UI built |
|---|---|---|
| one bare cell | `Cannot read properties of undefined (reading 'length')` | no |
| no bare cells | none | yes |

## Where it came from

Found via py-maidr, which was emitting bare `{}` for the unoccupied positions of a sparse subplot grid (xability/py-maidr#512, now fixed on the producer side). But the producer fix does not close this: `layers` being required in `MaidrSubplot` (`src/type/grammar.ts:262`) is a compile-time claim about TypeScript callers, and the schema arrives at runtime as JSON from Python, R, or a hand-authored `maidr` attribute. Any of them can emit a cell the type says is impossible.

## Suggested shape

Two things, and the first matters more than the second:

1. **Do not let one cell abort the figure.** `layers?.length ?? 0` would be enough to keep construction going, and the surrounding code is already written this way — `maidr.axes?.x?.label ?? DEFAULT_FIGURE_AXIS` on line 138, `layers[0]?.title ?? ''` on the very next line. Line 388 is the odd one out among its own neighbours. The empty-`layers` path is already a supported case: `isFigureLevel()`, `Movable`, and `Plot.getActiveTrace` all document and handle a subplot with no layers.

2. **Say something.** Silently treating a malformed cell as empty trades a total failure for an invisible one. A console warning naming the position (`subplot [1][3] has no layers`) costs nothing and is the only signal a producer author would ever get. There is no schema validation on this path today, so this would be the first.

## Reproduction

Any figure whose `maidr` attribute has a cell without `layers`:

```json
{"subplots": [[{"id": "a", "layers": [...]}, {}]]}
```

jooyoungseo:
Closes #1076.

## What breaks

`Subplot`'s constructor read `subplot.layers.length` unguarded, and `Figure`'s constructor maps every cell through it eagerly:

```ts
this.subplots = subplots.map(row => row.map(subplot => new Subplot(subplot)));
```

So one cell arriving without a `layers` key threw `TypeError: Cannot read properties of undefined (reading 'length')` during figure construction, the map aborted, and **no** part of the chart initialised — every well-formed subplot with it.

Measured in Chromium against a 4.4.0 bundle, one figure with a bare cell versus the same figure without:

| | uncaught error on activation | runtime UI built |
|---|---|---|
| one bare cell | `Cannot read properties of undefined (reading 'length')` | no |
| no bare cells | none | yes |

The SVG is already in the DOM by then, so the page looks finished, the chart is visibly a chart, and the key that starts navigation does nothing at all. No message, no degraded mode — a screen-reader user gets a decorative image and no way to know why.

## Why the type does not prevent it

`layers` being required in `MaidrSubplot` is a compile-time claim about TypeScript callers. The schema arrives at runtime as JSON — from py-maidr, from r-maidr, from a hand-authored `maidr` attribute. Any of them can emit a cell the type says is impossible, and py-maidr did, for every unoccupied position of a sparse subplot grid (xability/py-maidr#512, now fixed on the producer side).

That producer fix does not close this. The next producer to get it wrong takes the same figure down.

## The change

Degrade to the empty subplot rather than throwing. That is already a supported state — `isFigureLevel()`, `MovableGrid` and `getActiveTrace()` all document and handle a subplot with no layers (#749) — so the rest of the figure costs nothing. The surrounding lines were already written defensively; `layers.length` was the odd one out between `maidr.axes?.x?.label ?? DEFAULT_FIGURE_AXIS` and `layers[0]?.title ?? ''`.

Silently reading a malformed cell as empty would trade a loud failure for an invisible one, so `Figure` names the position on the console:

```
[Figure] Subplot [0][1] has no `layers`; reading it as an empty subplot.
The producer of this schema should emit `{ id, layers: [] }` for positions
no chart occupies.
```

The coordinates are the point — without them the warning cannot be acted on in a grid of any size, and they are the only signal a producer author ever gets. There is no schema validation on this path today; this is the first.

`subplot?.selector` picked up the same guard, since a cell can be `null` as easily as `{}`.

## Tests

Six added to `test/model/emptySubplot.test.ts`, next to the #749 cases they are the stricter cousin of — that issue was an empty `layers` *array* failing later, in `activeTrace`; this one is a missing `layers` *key* failing earlier, at construction.

Both halves were mutation-checked independently:

- guard reverted → **5 of the 6 fail** (the sixth is the well-formed control for the warning, which correctly still passes)
- `console.warn` removed, guard kept → **exactly 1 fails**, the one asserting the position is named

```
Test Suites: 355 passed, 355 total     Tests: 5143 passed    # unit + component
Test Suites:  10 passed,  10 total     Tests:  137 passed    # esm
lint:fix — clean       type-check — clean
```

---
_Generated by [Claude Code](https://claude.ai/code/session_01MAVKCcEEugLbot6CcYU2f8)_

### DIFF
--- a/src/model/plot.ts
+++ b/src/model/plot.ts
@@ -79,6 +79,25 @@ export function focusedSubplotTitle(state: PlotState): string {
 }
 
 /**
+ * Whether a subplot cell arrived with a `layers` array we can iterate.
+ *
+ * `layers` is required in `MaidrSubplot`, but that is a claim about
+ * TypeScript callers and the schema arrives at runtime as JSON -- from
+ * py-maidr, from r-maidr, from a hand-authored `maidr` attribute. Any of
+ * them can emit a cell the type says is impossible.
+ *
+ * Checked with `Array.isArray` rather than for truthiness, because the
+ * shapes that break are not only the falsy ones: `{ layers: {} }` passes
+ * `?? []` untouched and then throws `layers.map is not a function`, which
+ * is the same failure as a missing key wearing a different hat.
+ *
+ * One predicate for both callers below, so the two cannot drift apart.
+ */
+function hasLayerArray(subplot: MaidrSubplot | undefined): subplot is MaidrSubplot {
+  return Array.isArray(subplot?.layers);
+}
+
+/**
  * Represents a figure containing one or more subplots
  */
 export class Figure extends AbstractPlot<FigureState> implements Movable, Observable<FigureState>, Disposable {
@@ -141,17 +160,14 @@ export class Figure extends AbstractPlot<FigureState> implements Movable, Observ
     const subplots = maidr.subplots as MaidrSubplot[][];
     this.subplots = subplots.map((row, r) =>
       row.map((subplot, c) => {
-        // `layers` is required in `MaidrSubplot`, but that is a claim about
-        // TypeScript callers and the schema arrives as JSON -- from Python,
-        // from R, from a hand-authored `maidr` attribute. Reporting the
-        // position is the only signal a producer author ever gets, and
-        // treating the cell as empty silently would trade a loud failure
-        // for an invisible one.
-        if (!subplot?.layers) {
+        // Reporting the position is the only signal a producer author ever
+        // gets, and reading the cell as empty in silence would trade a loud
+        // failure for an invisible one.
+        if (!hasLayerArray(subplot)) {
           console.warn(
-            `[Figure] Subplot [${r}][${c}] has no \`layers\`; reading it as `
-            + `an empty subplot. The producer of this schema should emit `
-            + `\`{ id, layers: [] }\` for positions no chart occupies.`,
+            `[Figure] Subplot [${r}][${c}] has no \`layers\` array; reading `
+            + `it as an empty subplot. The producer of this schema should `
+            + `emit \`{ id, layers: [] }\` for positions no chart occupies.`,
           );
         }
         return new Subplot(subplot);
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Good catch, and correct — reproduced it before changing anything:

```
THREW: TypeError: layers.map is not a function
```

from `new Figure({ subplots: [[ <bar>, { layers: {} } ]] })`, propagating out of the constructor exactly as described. The first guard only closed half the door.

Both call sites now go through one `hasLayerArray` predicate, which also settles the second half of your note — the `Figure` check that decides whether to warn and the `Subplot` check that builds the default can no longer drift apart.

Covered across every shape a producer plausibly emits, each asserting both that the figure still builds with its well-formed neighbours intact and that the position is reported:

| cell | previous guard | now |
|---|---|---|
| `{}` | caught | caught |
| `null` | caught | caught |
| `{ layers: {} }` | **threw** | caught |
| `{ layers: 'bar' }` | **threw** | caught |

The two non-array rows fail against the previous guard and pass now; the other two pass either way, which is what makes them the control.

```
Test Suites: 355 passed        Tests: 5151 passed     # unit + component
Test Suites:  10 passed        Tests:  137 passed     # esm
lint:fix — clean      type-check — clean
```

On the sandbox note at the end: understood, and no need to re-run anything on your side — the numbers above are from a full local run, and the mutation checks are reproducible by reverting `hasLayerArray` to `subplot?.layers ?? []`.

---

Issue #1059, opened 2026-08-18
Recharts: a step line has no way to say it is one, so a staircase is read as an interpolated line


jooyoungseo:
`RechartsChartType` has no `'step'`, and `stepDirection` reaches a layer only
from inside `SurvivalCurveConfig`:

```ts
function layerOptions(chartType, config): Partial<MaidrLayer> {
  switch (chartType) {
    …
    case 'survival':
      return { stepDirection: config.survivalConfig?.stepDirection ?? 'hv' };
    default:
      return {};
  }
}
```

So a `<Line type="stepAfter">` or an `<Area type="step">` that is not a survival
curve can only be declared as `'line'` / `'area'`, and comes out as
`TraceType.LINE` / `TraceType.AREA` with no `stepDirection`. Recharts is the
only supported library with a first-class step curve that MAIDR cannot name:
Victory reads it off `interpolation`, Google Charts has `SteppedAreaChart`
(#1055), Chart.js has `stepped`, plotly has `line.shape`.

## What is lost

`StepTrace` is not a cosmetic relabel of `LineTrace`. Read as a line, the chart
loses all three of:

- **navigation by transition** rather than by sample;
- **a description in terms of runs** — `Transitions`, `Longest run`, `Levels`,
  and `Step direction`;
- **the announcement itself** — `plotType: 'step'` in the instruction text,
  the layer-switch cue and the subplot entry cue.

The values announced at each sample are right either way. What a reader is told
about the shape between them is not: a line chart implies the value moved
gradually from one sample to the next, and a step chart is drawn to say it did
not move at all until it jumped.

## Measured

Rendered the real components in jsdom over
`[{ t: 0, v: 10 }, { t: 1, v: 30 }, { t: 2, v: 20 }]` at 320×220 and read the
`.recharts-curve` paths. The samples land at x = 65, 190, 315 and
y = 128.75 (10), 16.25 (30), 72.5 (20):

| Recharts `type` | rendered `d` | riser sits |
| --- | --- | --- |
| `stepAfter` | `M65,128.75 L190,128.75 L190,16.25 L315,16.25 L315,72.5` | at the **next** sample → `hv` |
| `stepBefore` | `M65,128.75 L65,16.25 L190,16.25 L190,72.5 L315,72.5` | at the **current** sample → `vh` |
[...]

PR #1078, opened 2026-08-18
feat(observable): read a step curve as the staircase it draws

### LINKED ISSUE #1073: Observable Plot: a step curve is skipped entirely, though its direction is recoverable from the path
## What happens

`Plot.line(data, { x, y, curve: 'step-after' })` produces **no MAIDR layer at all**. The chart is silent — not misread, unread.

`parsePathVertices` (`src/adapters/observable/converters.ts:1489`) returns `null` when the path's vertex count does not match the datum indices Plot bound to it, and a `null` skips the mark:

```ts
return expected !== null && vertices.length !== expected
  ? null
  : { vertices, pixelError };
```

`docs/observable.md` states the reason, and it was the right reason when written:

> **Lines whose path does not pass through the data.** `curveBasis` and `curveBundle` draw through control points that are not data points, and the step curves (`curveStep`, `curveStepAfter`, `curveStepBefore`) draw a corner between every pair of points. The adapter detects the mismatch … and skips the mark rather than announcing the corners.

Lumping the step curves in with `curveBasis` is what is now wrong. A basis spline genuinely does not pass through its data. **A step curve does** — it passes through every data point and adds corners between them, and the corners are at positions that say which convention drew them.

## Measurement

`@observablehq/plot@0.6.17`, Chromium, `[{t:0,y:1},{t:1,y:3},{t:2,y:2}]`. Data x pixels are 40/330/620, data y pixels 370/20/195:

| curve | path `d` | verts | corners at |
|---|---|---|---|
| `linear` | `M40,370 L330,20 L620,195` | 3 | — |
| `step-after` | `M40,370 L330,370 L330,20 L620,20 L620,195` | 5 | 330, 620 — the **next** sample |
| `step-before` | `M40,370 L40,20 L330,20 L330,195 L620,195` | 5 | 40, 330 — the **current** sample |
| `step` | `M40,370 L185,370 L185,20 L475,20 L475,195 L620,195` | 6 | 185, 475 — the **midpoints** |
| `basis` | `M40,370 L88.333,311.667 C136.667,253.333,…` | cubics | not data |

All three step curves are distinguishable from the rendered path alone, and each gives the data back exactly:

- **`2n - 1` vertices, corners on data x** → `step-after` or `step-before`. Told apart by the first segment: horizontal first is `step-after`, vertical first is `step-before`. Taking every other vertex from index 0 recovers `(40,370) (330,20) (620,195)` — identical to the `linear` path's vertices.
- **`2n` vertices, corners at midpoints** → `curveStep`, the centred riser.
- Anything else → still not interpolating, still skipped.

## Why this is now cheap

`TraceType.STEP` and `MaidrLayer.stepDirection` exist, and `StepTrace` navigates by transition and describes in runs. The mapping is already settled across three adapters:

| Plot curve | `stepDirection` |
|---|---|
| `step-after` | `'hv'` |
| `step-before` | `'vh'` |
| `step` (centred) | **undeclared** — `StepDirection` cannot name a centred riser, and `StepTrace` is written to expect `undefined` |

This adapter is in a better position than the d3 binder, where the convention has to be declared (#1066) because `d3.curveStep*` leaves nothing in the path to tell it from a line whose samples happen to land on a staircase. Here Plot binds the datum indices to the path, so the expected sample count is known — which is exactly what makes the extra vertices identifiable as corners rather than as data.

## Scope

Read the step curves as `STEP` layers with the direction recovered from the corner positions; leave `curveBasis`/`curveBundle` skipped as they are. The centred `curveStep` is read as a step whose direction goes undeclared, rather than being guessed at as `hv`.

An **area** drawn with a step curve is the same shape and keeps `TraceType.AREA` with the direction riding along, as `bindD3Area` does.

### LINKED ISSUE #1075: Recharts: a centred step chart (type="step") is read as an interpolated line, though 'mid' names it
Follow-up to #1059, and a gap I left there.

## What happens

`<Line type="step" />` is read as `TraceType.LINE` with no `stepDirection`. Navigation is by sample rather than by transition, the description carries no runs, and the chart announces itself as a line plot — the same three losses #1059 was filed about, still in place for one of the three step types.

`stepDirectionFor` (`src/adapters/recharts/childProps.ts:34`) names two of the three:

```ts
const STEP_DIRECTIONS: Record<string, StepDirection> = {
  stepAfter: 'hv',
  stepBefore: 'vh',
};
```

`type="step"` is not in the map, so the lookup misses and the layer stays a line.

## Why it was left out, and why that reason was wrong

I wrote #1059 believing a centred riser had no name in the grammar. It has one. `StepDirection` is `'hv' | 'vh' | 'mid'` (`src/type/grammar.ts:1140`), and `'mid'` has been there since the STEP trace landed in #723:

> `mid` — jump at the midpoint of the two x values (matplotlib `steps-mid`).

`StepTrace` handles it end to end, including the highlight mapping, which distinguishes `2N - 1` path vertices for `hv`/`vh` from `2N` for `mid` (`src/model/step.ts:33`). The Vega-Lite adapter already maps `'step': 'mid'` (`src/adapters/vegalite/converters.ts:630`), so this is the odd one out rather than a new capability.

## The mapping, from Recharts' own source

`recharts/lib/shape/Curve.js` resolves `type` against a factory map of d3 curves:

```js
curveStep:        d3.curveStep,
curveStepAfter:   d3.curveStepAfter,
curveStepBefore:  d3.curveStepBefore,
```

so `type="step"` is `d3.curveStep`, the centred riser, and the full mapping is:

| Recharts `type` | d3 curve | `stepDirection` |
|---|---|---|
| `stepAfter` | `curveStepAfter` | `'hv'` |
| `stepBefore` | `curveStepBefore` | `'vh'` |
| `step` | `curveStep` | `'mid'` |

## Scope

Add `step: 'mid'` to `STEP_DIRECTIONS`, and correct `docs/recharts.md` where it says the centred form cannot be declared.

Two related corrections belong with it, both the same mistaken belief written down elsewhere:

- `test/adapters/d3/stepDirection.test.ts` — the test "stays the line it was when nothing declares a convention" carries a comment claiming `d3.curveStep` "is neither convention and which `StepDirection` cannot name". The *behaviour* is right (a caller who declares nothing gets a line, since d3 leaves no trace of the curve in the path), but the stated reason is false: a caller drawing `curveStep` can and should pass `stepDirection: 'mid'`, which `bindD3Line` already forwards.
- `docs/d3.md` repeats the same claim.

No code change is needed on the d3 side — only the two places that tell a reader a capability does not exist.

jooyoungseo:
Closes #1073. Closes #1075.

Two commits, both about staircases, and the second is a correction of my own.

## 1. Observable Plot: a step curve was silent

`Plot.line(data, { curve: 'step-after' })` produced **no layer at all**.
`parsePathVertices` refuses a path whose vertex count does not match the datum
indices Plot bound to it, and `docs/observable.md` grouped the step curves with
`curveBasis` under that rule.

The rule is right; the grouping was wrong. A basis spline genuinely does not
pass through its data. A staircase passes through **every** sample and adds a
corner between each pair — and where the corner sits says which convention drew
it. Measured with `@observablehq/plot@0.6.17`, four samples at x pixels
40/233.333/426.667/620:

| curve | vertices | corners at | read as |
|---|---|---|---|
| `linear` | 4 | — | line |
| `step-after` | 7 | the **next** sample | `hv` |
| `step-before` | 7 | the **current** sample | `vh` |
| `step` | 8 | the **midpoints** | `mid` |
| `basis` | cubics | control points | still refused |

So the whole reading comes off the drawing with nothing declared. That is the
opposite of the d3 binder, and the reason is worth stating: Plot binds the
datum indices to the path, so how many samples there should be is known, and
the surplus vertices can be *identified* as corners rather than mistaken for
data. A binder reading its data from the source has no such handle.

**Every corner is verified, not counted.** A path carrying `2n - 1` vertices
without the shape is refused — reading it as a staircase would announce
transitions the chart never drew.

The centred curve needs one extra step: its interior samples are **not on the
path**, only the midpoints between them, so they come back by
`x[k+1] = 2·m[k] − x[k]` and the last drawn sample checks the result. The error
does not accumulate — a wrong `x` reaches the next negated rather than
amplified. Over 50 samples at deliberately uneven spacing the largest departure
was **0.002 px**, twice the quantum the coordinates were written at.

An **area** keeps its type and carries the convention, as `bindD3Area` does. A
**stacked** one works because the baseline is stepped alongside the top edge, so
the halves stay aligned and the same sample indices cut both.

### The harness gained what it could not represent

Plot leaves the datum indices on the element as `__data__`; captured markup
cannot carry them. Without them **no fixture could reach any check that
compares a path against how many samples it has** — including the pre-existing
`curveBasis` refusal. Fixtures now opt in by stating a `samples` count, so
every existing fixture mounts exactly as before.

## 2. Recharts: `type="step"` was read as a line — my gap from #1059

I left the centred convention out of #1059 believing the grammar had no name
for it. It has: `StepDirection` is `'hv' | 'vh' | 'mid'` and has been since the
STEP trace landed in #723, where `mid` is documented as matplotlib's
`steps-mid`. `StepTrace` handles it end to end, its highlight mapping
distinguishing `2N - 1` vertices from `2N` — the same two shapes the
measurement above found. The Vega-Lite adapter already maps `'step': 'mid'`, so
Recharts was the odd one out rather than this being a new capability.

Recharts resolves `type` against d3's curves in `lib/shape/Curve.js`, so
`type="step"` is `curveStep`. All three now map.

The same mistaken belief was written down in `docs/d3.md` and a d3 test comment.
No code change was needed there — `bindD3Line` always forwarded whatever the
caller declared, and the field is typed `StepDirection` — so only the two places
that told a reader the capability did not exist are corrected, plus a case
pinning it. The d3 binder still asks the caller to declare, and that reason now
stands on its own: it reads `__data__` and never looks at the path.

## Falsification

Nine mutations, one at a time, against the observable suite:

| --- | --- |
| staircase never attempted on a line | 6 |
| corner check loosened to the level only | 1 — flat `step-before`, where only the corner's x separates the conventions |
| centred branch removed | 1 |
| centred riser check removed | 1 |
| midpoint drift check removed | 1 |
| stepped line not retyped `STEP` | 3 |
| stepped **area** retyped `STEP` | 1 |
| area baseline left uncut by the sample indices | 1 — stacked bands |
| `stepDirection` not carried onto the layer | 7 |

Plus one for Recharts: dropping `step: 'mid'` reddens its case.

Two of those needed shapes **no chart Plot draws produces** — all twelve gappy
and non-interpolating variants measured miss both `2n - 1` and `2n` outright —
so they are pinned with hand-written paths, labelled as such, one per check.
Each check alone refuses my first attempt, which is exactly why one example was
not enough: it left the other deletable with the suite still green.

An earlier round left four mutations surviving. The fixtures that kill them —
a flat staircase, a stacked stepped area, a two-sample staircase — came from
that round, and one guard turned out to be **unreachable** (a one-sample step
line draws exactly one vertex, which the caller already handles) and was
removed rather than kept as a line no test could reach.

## Found but not fixed

A **gappy** stepped line can coincidentally match its sample count:
`[1, null, 3, 2]` with `step-after` draws `M40,370 Z M426.667,20 L620,20 L620,195`
— four vertices for four samples — so the existing equality branch accepts it
and announces a corner as a sample. This is pre-existing (the same branch, the
same result, before this change) and belongs to the multi-subpath question
rather than to step curves, so it is left alone here. Filed separately.

## Verification

```
npm run lint:fix     clean
npm run type-check   clean
npm test             356 suites / 5,161 unit, 10 suites / 137 esm — all pass
```

---


jooyoungseo:
Good catch, and fixed in a5361fb8 — all four, plus two more the sweep turned up.

You were right that the worst of them is `childProps.ts`: the JSDoc denying a centred riser has a name sat four lines above the map I had just given `step: 'mid'`. Searching the phrase rather than the files found two further copies in `src/adapters/recharts/types.ts` — the chart-type list at line 49 and the `stepDirection` field at 1139 — so the corrected set is six:

| file | what it said |
| --- | --- |
| `src/adapters/recharts/childProps.ts` | JSDoc over `STEP_DIRECTIONS` |
| `src/adapters/recharts/converters.ts` | the `layerOptions` case for step and area |
| `src/adapters/recharts/types.ts` | the chart-type list |
| `src/adapters/recharts/types.ts` | the `stepDirection` field |
| `src/adapters/d3/types.ts` | the `stepDirection` field |
| `test/adapters/recharts/stepDirection.test.ts` | the file docblock |

Two of them were doing more than describing — they justified a real decision with the now-false reason, so the decision is kept and the reason replaced rather than the sentence deleted:

- **Recharts supplies no default `stepDirection`.** Not because the grammar lacks a name for the centred case, but because a config that declares none *and* has no `<Line>` to read one off has not said which way its risers go. Every curve Recharts draws does have a name, which is exactly why a default here would be substituting one of them for silence.
- **The d3 binder asks the caller to declare.** Not because `'mid'` is unnameable — `bindD3Line` has always forwarded it — but because the binder reads `__data__` and never looks at the path, so which curve was used is the author's knowledge and not something the drawing gives back.

`grep -rniE "cannot name|neither convention|is neither"` over `src/`, `test/` and `docs/` now returns one unrelated hit in an amCharts geography test.

Re-verified after the change: `lint:fix` and `type-check` clean, all 14 builds complete, 356 suites / 5,161 unit and 10 suites / 137 esm passing.

On the verification note — understood that the job could not run the suite. The figures above are from a full local run on this commit.

---

PR #1083, opened 2026-08-18
feat(observable): read a regression line as the fitted trend it is

### LINKED ISSUE #1081: Observable Plot: linearRegressionY / linearRegressionX are unread, though the mark names itself and MAIDR has SMOOTH
`Plot.linearRegressionY(data, { x, y })` produces no layer. MAIDR has `TraceType.SMOOTH`, and unlike the box composite (#1074) this mark needs no heuristic at all — Plot gives it a group of its own with a name nothing else uses.

## Measurement

`@observablehq/plot@0.6.17`, Chromium, `[{a:1,b:2},{a:2,b:4.1},{a:3,b:5.9},{a:4,b:8.2}]` with a `dot` mark and a `linearRegressionY`:

```
g[aria-label="dot"]                4 circles
g[aria-label="linear-regression"]  2 paths
```

`linear-regression` is the whole detection. `convertMark` already dispatches on `aria-label`, and no other mark produces that one — `linearRegressionX` uses the same label, so both variants arrive the same way.

The group's two children are, in order:

```
1  stroke="none"   closed path, ~130 vertices   the confidence band
2  fill="none"     M40,370.565 L620,25.081      the fitted line
```

and the group carries `fill-opacity="0.1"`, which is the band's shading.

Grouped by `stroke`, the children come in the same pairs, one pair per series:

```
path stroke="none"    fill="#4269d0"   band, series 1
path fill="none"      stroke="#4269d0" line, series 1
path stroke="none"    fill="#efb118"   band, series 2
path fill="none"      stroke="#efb118" line, series 2
```

so the colour scale already resolves the series name the way it does for every other mark here, and `fill`/`stroke` being mutually `"none"` tells band from line without depending on child order.

## The shape question, and what the measurement says about it

`SmoothPoint` wants samples along the curve, with the band's bounds on them — the shape settled for r-maidr's `geom_smooth(se = TRUE)` and py-maidr's plotly trendline.

The fitted line has **only two vertices**, its endpoints. That is complete for a straight line but a poor read: two points to arrow through, and no sense of the trend between them.

The band supplies the grid. Its ~130 vertices are the upper edge at 4-pixel steps followed by the lower edge back, so each x on that grid has both bounds, and the fitted y at that x comes from interpolating the line's two endpoints — exact, because the line is straight.

So the reading is available in full. What it needs is a decision about density: the band's own grid is ~65 samples across the frame, which is a lot to navigate but is precisely what the chart draws. Downsampling it would be choosing a number, which is the kind of thing worth agreeing rather than picking.

## Scope

Read `linear-regression` as `TraceType.SMOOTH`, one series per band/line pair, with `yMin`/`yMax` from the band. `linearRegressionX` is the same mark transposed. Both are Plot's own names for the mark, so nothing has to be declared.

Related: #1074 (box composites, which do need a declaration), #1073 (step curves, merged in #1078).

jooyoungseo:
Closes #1081.

`Plot.linearRegressionY` and `linearRegressionX` produced **no layer at all**.
MAIDR has `TraceType.SMOOTH`, and this is the one Plot mark that needs neither a
heuristic nor a declaration: Plot names its group `aria-label="linear-regression"`,
which no other mark produces, and `convertMark` already dispatches on that
attribute. The detection is a switch case.

## What the markup gives

Measured with `@observablehq/plot@0.6.17`:

```
g[aria-label="linear-regression"]
  path  stroke="none"            the confidence band
  path  fill="none"              the fitted line, M40,370 L620,20
```

Split by `stroke`, the group holds four paths in band/line pairs, one pair per
series. They are told apart by `fill="none"` rather than by their order, so a
future Plot that emits them the other way round still reads.

## Two defects the fixtures then pinned

Neither would have surfaced without capturing real markup with the bound datum
indices restored — which is what #1078 taught the harness to do.

**The count check refused every fit.** The fit's `__data__` is the *four
observation indices it was computed from*; its path has *two* vertices. So
`parsePathVertices` — whose whole job is to refuse a curve that does not pass
through its samples — refused it. That check is right for a line and meaningless
here, because the datum indices mean something different for this mark: on a
line they are its samples, on a fit they are the observations behind it. The
geometry reading is now a function of its own and the count check belongs to the
caller that needs one. Putting the old call back reddens **five** tests.

**The series name came back empty.** `strokeOrFill` prefers `fill`, and a fitted
line's fill is `none` by construction — the very attribute that identified it.
The colour is on the stroke, and this mark reads the stroke. The helper is left
alone, since every other mark depends on its current precedence.

## What is deliberately not read

**The confidence band.** `SmoothPoint` is `{ x, y, svg_x, svg_y }` — no bounds —
and `SmoothTrace` announces none, so an interval has nowhere to go on a smooth
layer. I originally proposed carrying it as `yMin`/`yMax` and
[corrected that on the issue](https://github.com/xability/maidr/issues/1081#issuecomment-5333165314)
once I read the type. Giving smooth layers bounds would serve r-maidr's
`geom_smooth(se = TRUE)` and py-maidr's plotly trendline as much as this mark,
and is a decision about the grammar rather than about Plot.

The fit's two vertices are its ends, and both are read along with the pixels
they were drawn at — `svg_x`/`svg_y`, which this adapter has because pixels are
what it inverts from. A straight line is completely described by its ends, so
two of them is the whole trend, not a truncation of it.

## Falsification

| --- | --- |
| the `linear-regression` case removed | 5 |
| the band no longer skipped | 3 |
| the count check put back | 5 |
| `svg_x`/`svg_y` zeroed | 1 |
| series colour read via `strokeOrFill` | 1 |
| only the first series kept | 1 |
| typed `LINE` instead of `SMOOTH` | 2 |

## Verification

```
npm run lint:fix     clean
npm run type-check   clean
npm test             358 suites / 5,187 unit, 10 suites / 137 esm — all pass
```

Run again after merging `main` (which brought in #1082). That merge conflicted
in the fixtures file and in `docs/observable.md`; I resolved the first and
missed the second, and `test/scripts/conflictMarkers.esm-test.ts` caught it —
doing exactly what it is there for. Both are resolved and the full suite is
green on the merge commit.

`docs/observable.md` gains a table row and a "Regression lines" section, and its
box-mark exclusion now says *why* the parts cannot be recognised as one chart
rather than only that they are separate marks.

---


jooyoungseo:
You were right to ask, and it was a real regression. Fixed in 7179e920.

`convertLine` **had** lost `strokeOrFill`. The cause was a careless replace-all: the same call appears in both `convertLine` and `convertRegression`, and my substitution had no count, so it changed both. I had described the change as local to the new function; it was not.

The consequence is worse than a colour lookup falling back. A fitted line reads its colour off the stroke *because* its fill is `none` — that attribute is what identifies it. An area is the exact inverse:

```
area path    <path fill="#4269d0" d="…">      no stroke at all
fit path     <path fill="none" stroke="#4269d0" d="…">
```

So `getAttribute('stroke')` returned null for every area, `valueAtColor` returned null, and every multi-series area lost its band names and its legend — `z` undefined on each point — while the values stayed correct. Silent, and exactly the kind of thing that reads as "this chart has no series" rather than as a failure.

Your instinct about why to double-check was also right: **nothing caught it.** The suite was green across 5,187 tests. Checking why, the area cases assert values and the legend cases are all bars and lines — no test pinned an area's series names, so there was nothing to break.

There is one now, and it is the test I should have had before making the change rather than after:

```
with the regression reintroduced:
  ● a linear regression mark › leaves an area mark reading its colour off the fill
  Tests: 1 failed, 101 passed, 102 total
restored:
  Tests: 102 passed, 102 total
```

`convertRegression` still reads the stroke, which is correct for it and now the only place that does.

Re-verified on the pushed commit: `lint:fix` and `type-check` clean, all 14 builds complete, 358 suites / **5,188** unit and 10 suites / 137 esm passing.

On the other note — a local type alias for `readPathGeometry`'s return — I've left it as the inline shape for now, since the file's other small readers are written the same way and there is still only the one caller of each. Worth doing if a third arrives.

---

PR #1086, opened 2026-08-18
fix(model): give every grid cell a subplot number, not only the drawn ones

### LINKED ISSUE #1085: fix(model): an empty subplot announces "Subplot 1", colliding with the real first panel
## What a reader hears

A figure with an unoccupied grid position announces two different cells as the same one. Driving a 2×2 whose top row is a single spanning panel (so `(0,1)` is empty) in Chromium against a 4.4.0 bundle:

| key | announced |
|---|---|
| `→` | Subplot **1** of 4, Top: This is a bar plot. |
| `→` | Subplot **1** of 4 is empty, nothing to describe. |
| `↓` | Subplot **3** of 4, Right: This is a bar plot. |
| `←` | Subplot **2** of 4, Left: This is a bar plot. |
| `↑` | Subplot **1** of 4, Top: This is a bar plot. |

Two problems, and the first is the one that costs a reader their place:

1. **The empty cell claims index 1**, which the top-left panel also claims. Position announcements are the only way to know where you are in a grid, and here they are ambiguous.
2. **The index and the total count different things.** Occupied panels number 1–3; the total says 4. So "Subplot 3 of 4" is reachable but "Subplot 4 of 4" never is, and a reader has no way to tell whether they have seen everything.

## Root cause

`collectAxesPositions` (`src/util/subplotLayout.ts:156`) only yields a position for a cell whose axes element resolved, and `buildOrderMap` (`:186`) numbers exactly those:

```ts
function buildOrderMap(sorted: AxesPosition[]): Map<string, number> {
  const map = new Map<string, number>();
  for (let i = 0; i < sorted.length; i++)
    map.set(`${sorted[i].row},${sorted[i].col}`, i + 1);
  return map;
}
```

An empty cell has no axes element, so it gets no entry. `Figure.state` (`src/model/plot.ts:287`) then falls through its own guard:

```ts
const currentIndex = this.visualOrderMap.get(key);
if (currentIndex === undefined) {
  console.warn(`[Figure] Visual order map missing key "${key}". Was applyLayout() called?`);
}
...
index: currentIndex ?? 1,
```

The warning does fire — captured from the console on the run above:

```
warning: [Figure] Visual order map missing key "0,1". Was applyLayout() called?
```

It just names the wrong cause. `applyLayout()` *was* called; the key is missing because the cell has nothing to measure, which is a legitimate state (`isFigureLevel`, `MovableGrid` and `getActiveTrace` all handle an empty subplot, and xability/maidr#1076 has since made a malformed one degrade to the same thing). And `?? 1` turns a missing index into a confidently wrong one.

Meanwhile `size` is `this.size`, the total cell count, so the two halves of "N of M" are computed from different sets.

## Not new, and about to be more visible

Reproduces on a plain `subplot2grid((2, 2), (0, 0), colspan=2)` layout that predates any recent change. But xability/py-maidr#517 has just made `seaborn.jointplot` emit a 2×2 with one genuinely empty corner instead of a 2×6 with nine, so figures that reach an empty cell are about to become ordinary rather than pathological:

```
→  Subplot 1 of 4: This is a hist plot.
→  Subplot 1 of 4 is empty, nothing to describe.
↓  Subplot 3 of 4: This is a hist plot.
```

## Suggested shape

Number every cell rather than every *measurable* cell. The empty ones have positions in the grid even when they have none on the screen — `buildOrderMap` could fill them in from their row/column after sorting what it can measure, so the sequence stays visual where measurement is possible and stays complete where it is not.

Whatever the numbering, the index and the total should be drawn from the same set, so that every announced index is reachable and the last one is the total.

The `?? 1` fallback deserves a look too. A missing index is not index 1, and silently claiming it is turns a diagnosable gap into a wrong announcement a reader cannot detect. The console warning should also stop blaming `applyLayout()` for a case that is nobody's mistake.

## Reproduction

```python
import matplotlib; matplotlib.use("Agg")
import matplotlib.pyplot as plt
import maidr

fig = plt.figure()
top = plt.subplot2grid((2, 2), (0, 0), colspan=2, fig=fig); top.set_title("Top")
left = plt.subplot2grid((2, 2), (1, 0), fig=fig); left.set_title("Left")
right = plt.subplot2grid((2, 2), (1, 1), fig=fig); right.set_title("Right")
for ax in (top, left, right):
    ax.bar(["p", "q"], [1, 2])

maidr.save_html(fig, file="gap.html", use_cdn=False)
```

Open it, activate the chart, and press `→` twice.

jooyoungseo:
Closes #1085.

## What a reader hears

A figure with an unoccupied grid position announced two different cells as the same one. Driving a 2×2 whose top row is a single spanning panel (so `(0,1)` holds nothing) in Chromium:

| key | before | after |
|---|---|---|
| `→` | Subplot **1** of 4, Top | Subplot **1** of 4, Top |
| `→` | Subplot **1** of 4 is empty | Subplot **2** of 4 is empty |
| `↓` | Subplot **3** of 4, Right | Subplot **4** of 4, Right |
| `←` | Subplot **2** of 4, Left | Subplot **3** of 4, Left |
| `↑` | Subplot **1** of 4, Top | Subplot **1** of 4, Top |
| console | `[Figure] Visual order map missing key "0,1"` | *(silent)* |

Position announcements are the only way to know where you are in a grid, and two cells answered to the same number. The total meanwhile counted every cell while the index counted only drawn panels, so "Subplot 4 of 4" was never reachable and a reader had no way to tell they had seen everything.

## Root cause

`collectAxesPositions` yields nothing for a cell with no axes element, so `buildOrderMap` left it out of the map entirely. `Figure.state` then fell through its own guard:

```ts
const currentIndex = this.visualOrderMap.get(key);
if (currentIndex === undefined) {
  console.warn(`[Figure] Visual order map missing key "${key}". Was applyLayout() called?`);
}
...
index: currentIndex ?? 1,
```

`applyLayout()` *was* called. The key is missing because the cell has nothing to measure, which is a supported state — `isFigureLevel`, `MovableGrid` and `getActiveTrace` all handle an empty subplot, and #1076 made a malformed one degrade to the same thing. `?? 1` turned a missing index into a confidently wrong one.

## The change

`buildOrderMap` now covers every cell. Only the **direction** of each axis is read from the measurements, not the individual positions, so a cell nothing could be measured for still lands where it belongs — interpolating a coordinate for it would be inventing a measurement rather than using one.

Both halves of "N of M" now come from the same set: the indices run 1..cells, and every one of them is somewhere a reader can stand.

Ragged grids are handled (a shorter row simply has no cell at that column), and the existing `?? 1` fallback becomes unreachable for grid positions — left in place rather than removed, since it still guards a `Figure` used before `applyLayout()`.

## Behaviour change worth naming

For a figure **with** an empty cell, the drawn panels renumber: in the example above Left goes from 2 to 3 and Right from 3 to 4. That is the point — the old numbers were a sequence that skipped positions while the total counted them. A figure with no empty cells is numbered exactly as before, which is what the eight pre-existing tests pin.

## Tests

Four added, and all twelve pass. Against a build with the old measured-only `buildOrderMap`, the four new ones fail and the eight old ones pass.

They go through the `<g id="axes_*">` path **deliberately**, and the test docstring says why: the panel fallback next door demands geometry for every subplot and returns nothing when one is missing, so a grid with an empty cell lands on data-array order there — which already numbers every cell. Three of these tests passed in their first draft without exercising the change at all; only the fourth failed, which is what caught it.

```
Test Suites: 357 passed, 357 total
Tests:       5186 passed, 5186 total
lint:fix — clean      type-check — clean
```

Plus the browser run in the table above, against a build of this branch.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MAVKCcEEugLbot6CcYU2f8)_

### DIFF
--- a/src/util/subplotLayout.ts
+++ b/src/util/subplotLayout.ts
@@ -84,7 +84,7 @@ export function resolveSubplotLayout(subplots: Subplot[][]): SubplotLayout {
   }
 
   const sorted = sortByVisualPosition(positions);
-  const visualOrderMap = buildOrderMap(sorted);
+  const visualOrderMap = buildOrderMap(sorted, subplots);
   const topLeftRow = sorted[0].row;
   const invertVertical = detectInversion(positions, subplots.length);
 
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Took the suggestion — the left-packed assumption is now written down at the `col < subplots[row].length` check rather than left to be re-derived:

> The length is enough to say which, because a row is left-packed — a producer fills every position it emits, and a cell with nothing drawn in it is a `Subplot` carrying `layers: []` rather than a hole in the array. A row missing its *leading* column is not a shape the grammar can express.

On the monotonicity point: agreed it is a new assumption for columns, and worth stating that it is the same one `detectInversion` has always made for rows. It holds because these coordinates come from a gridspec — a row's cells share a y band and a column's share an x band by construction, which is what makes the axes separable at all. A layout where it failed would not be a grid, and the model has no way to represent one.

---

jooyoungseo:
One thing worth recording, since you traced `getSubplotSummaries` while reviewing: it has the same `?? ` fallback one line down, and it was hitting it for the same reason.

```ts
const visualIndex = this.visualOrderMap.get(key) ?? summaries.length + 1;
```

On the 2×2 spanning case that gave `(0,1)` — the empty cell — the value `summaries.length + 1` = **2**, which is the index `(1,0)` legitimately holds. Two summaries with the same index, and the `sort` immediately below is then deciding between them arbitrarily.

That falls out with this change rather than needing its own: the map now has an entry for every cell, which `numbers every cell exactly once, 1 through the cell count` pins directly, so the `??` branch is unreachable there too and the summaries carry the same distinct 1..N.

I did not add a test for it. The property it would assert is the one that test already holds, and `getSubplotSummaries` needs a jsdom environment the model tests do not run in — a test that had to construct one would be asserting the map's contents through two layers of indirection rather than more strongly.

---

Issue #1084, opened 2026-08-18
Observable Plot: hexbin, contour and density are unread — one of the three is recoverable, and the other two share cell's root cause


jooyoungseo:
Three more Plot marks MAIDR has trace types for and the adapter does not read. Measured together because the answers differ, and the two that fail do so for the same reason `cell` already does.

## `hexbin` — the data is exactly recoverable

`Plot.dot(data, Plot.hexbin({ r: 'count' }, { x, y }))` draws one `<path>` per bin inside a `g[aria-label="dot"]`. Measured with `@observablehq/plot@0.6.17`, three bins holding 9, 5 and 2 points:

```
r scale     type 'pow', domain [0, 9], range [0, 10]

bin  transform                             hexagon radius   count
  1  translate(610.5, 294.449)                     10         9
  2  translate(30.5,  363.731)                      7.454     5
  3  translate(350.5,  17.321)                      4.714     2
```

Both halves invert **exactly**, and neither involves a colour:

- the **centre** is the `transform` translate, through the x and y scales like any other mark's position;
- the **count** is the hexagon's radius through the `r` scale — d3's default square-root scale, so `10·√(5/9) = 7.454` and `10·√(2/9) = 4.714`, matching the drawn radii to the last digit.

The hexagon itself is drawn as `M0,11.547 L10,5.774 L10,-5.774 L0,-11.547 L-10,-5.774 L-10,5.774 Z` — a regular hexagon whose vertices sit at `(0, ±r·2/√3)` and `(±r, ±r/√3)`.

**The open question is detection, not extraction.** A hexbin's marks arrive in a group labelled `dot`, exactly like a scatter's, so the label does not separate them. The element tag gets close and not all the way:

```
plain dot                <circle>
dot with symbol:'square' <path d="M-2.659,-2.659h5.317v5.317h-5.317Z">
hexbin                   <path d="M0,11.547L10,5.774L…Z">
```

So a hexbin is a path among paths. Its shape is a genuine signature — six vertices in the `2/√3` ratio — but a dot mark drawn with a hexagonal symbol would match it. The failure that would cause is milder than the box-plot one (#1074): the positions and sizes announced would still be right, and only the chart's *name* and navigation model would be wrong. That is a judgment worth making deliberately rather than by default, which is why it is written down here rather than implemented.

## `contour` and `density` — the level is in the colour

[...]

PR #1087, opened 2026-08-18
feat(observable): read a declared hexbin as the lattice of bins it draws

jooyoungseo:
Addresses the hexbin half of #1084.

`Plot.dot(data, Plot.hexbin({ r: 'count' }, { x, y }))` was read as a scatter,
so its cells were announced at positions that are lattice centres rather than
observations, and the tally each one stands for went unread entirely.

## Both halves invert exactly, and neither goes through a colour

Which is what separates this from the `cell`, `contour` and `density` marks the
adapter declines. Measured with `@observablehq/plot@0.6.17` over clusters of 5,
2 and 9 points:

```
r scale   type 'pow', exponent 0.5, domain [0, 9], range [0, 10]

transform="translate(30.5, 363.731)"      the centre, through x and y
d="M0,8.607L7.454,4.303L…Z"               radius 7.454 = 10·√(5/9) → 5
```

## The author declares it, because nothing in the markup can

The cells arrive in a group labelled `dot` like any scatter's — and Plot's own
`symbol: 'hexagon'` draws the **identical path shape** at a different radius:

```
symbol: 'hexagon'   M0,3.464  L3,1.732  L3,-1.732  L0,-3.464  …
hexbin              M0,11.547 L10,5.774 L10,-5.774 L0,-11.547 …
```

Both are `M0,r·2/√3 L r,r/√3 L r,-r/√3 L 0,-r·2/√3 …`, differing only in `r`.
So this is declared through `markTypes`, the option that already exists for the
same problem on a binned rect. A hexbin is a single mark, so one label maps to
one type and nothing is widened.

**The declaration does not override the geometry.** A cell still has to be six
vertices at the positions a regular hexagon of that width puts them, so a
declaration aimed at a diamond bubble chart is declined rather than read as a
lattice of invented tallies.

## Rounding the tally, but only when the drawing is unambiguous

`HexbinPoint.count` is documented as "how many points fell in it", and a radius
reaches the `d` attribute rounded to three decimals — so a bin of five inverts
to **5.00059**, inside the geometry's own error of an integer. Rounding there
is reading the field's contract.

A hexbin sized by something that is not a count keeps its fractional tally: a
fixture sized by `r: 'mean'` over a weight holds 0.4 and 2.9, and rounding
those would report 0 and 3.

## The harness gained a `pow` scale

It had no branch for one, so it inverted a radius linearly and this chart's
bins came back as **6.71 and 4.24**. Plot reports `exponent: 0.5` alongside the
type; a fixture now records it, as it already records the datum indices from
#1078.

## Falsification

Eight mutations, one at a time:

| --- | --- |
| the declaration gate removed | 5 |
| the six-vertex check removed | 1 |
| the vertex-position check removed | 1 |
| rows ordered top-first | 3 |
| rounding removed | 1 |
| rounding applied unconditionally | 1 |
| radius read off `y` instead of `x` | 5 |
| a missing `r` scale tolerated | 1 |

Three rounds were needed. The first left both shape checks alive, so I captured
a **diamond bubble chart** — paths, in a `dot` group, *with* an `r` scale:
everything a hexbin has except the shape. That killed one check and showed the
other was redundant on it, because each alone refuses every real Plot symbol.
Separating them needs shapes Plot never draws — a six-sided path that is not a
hexagon, and a hexagon's vertices buried among others — so those two are
hand-written and labelled as such, one per check, following #1078's precedent.

One of those hand-written paths was itself wrong at first: with integer
coordinates the tolerance derives to a whole pixel, which admitted it. The
current path's top vertex is far outside any tolerance.

## Verification

```
npm run lint:fix     clean
npm run type-check   clean
npm test             359 suites / 5,198 unit, 10 suites / 137 esm — all pass
```

Run again after merging `main` (which brought in #1083) and resolving the
conflicts in the fixture file, the converter and the docs.

`docs/observable.md` gains a table row and a "Hexbins" section covering the
declaration, the geometry check, and the rounding rule.

---


jooyoungseo:
Both observations were fair and are corrected in 6c21686f. Neither changed behaviour; both were comments describing their arithmetic loosely, which is worth fixing on its own terms — a reader trusting a stated derivation should get the real one.

**On the tolerance doubling** — you are right, and my comment was simply wrong about where the second error came from. The expected vertices are computed from the *measured* radius via `√3`, not read from a second independently-rounded path, so only the drawn path carries the rounding. It now says the doubling is generosity, and that erring loose there can only admit a shape that is nearly a hexagon, never reject one that is.

**On `binCount`'s factor of two** — it was implicit and is now named: it is the square-root scale's, since `r ∝ √tally` makes a relative error in the radius arrive doubled in the tally. You are also right about the direction of the failure, and it is worth being precise about it: with a linear `r` scale the estimate would be *twice* the true error, so the risk is rounding a value that is genuinely not near an integer — the opposite of conservative. `Plot.hexbin` always sizes by the square root, so it is not a live case, but the comment now says which scale the formula belongs to rather than leaving it to be inferred.

Re-verified after the change: `lint:fix` and `type-check` clean, 112 observable tests passing.

---

jooyoungseo:
The row-grouping coverage note was a real hole, not a theoretical one — thank you. Fixed in 56c26b09.

You were right that no fixture had two bins in one row. What I had not appreciated until I went to check is that with singleton rows, *both* halves of the ordering are unobservable: sorting a one-element row by `x` is a no-op, and grouping on `bin.row` gives the same three singleton groups as grouping on anything else unique per bin. The suite could not tell a correct implementation from a broken one.

The new `crowdedHexbin` fixture is real `Plot.plot()` output over five clusters arranged into two rows — counts 3, 7, 2 along the bottom and 5, 4 along the top. What makes it useful is something I did not have to arrange: Plot draws the hexagons largest first, so the DOM order is

```
7 (bottom)  5 (top)  4 (top)  3 (bottom)  2 (bottom)
```

— the two rows interleaved, and neither row in reading order. Recovering `[[3, 7, 2], [5, 4]]` from that needs the grouping *and* the sort.

Falsified mutation by mutation, each applied alone against `test/adapters/observable/hexbin.test.ts`:

| Mutation | Before | After |
| --- | --- | --- |
| drop `.sort((a, b) => a.x - b.x)` within a row | 11 passed | **1 failed** (new test only) |
| key `byRow` on `bin.x` instead of `bin.row` | 11 passed | **1 failed** (new test only) |
| drop `.sort(([a], [b]) => b - a)` on the rows | 3 failed | 3 failed |

The first two are the ones the old fixtures could not see. The third was already pinned, and interestingly the new test does *not* add to it: with the row sort gone the rows fall back to `Map` insertion order, which for this fixture happens to be bottom-row-first anyway. Worth stating plainly rather than claiming the new case covers more than it does.

On your other two notes — the non-bijective vertex matching is real and I agree it is unreachable, since it needs two coincident vertices in a path Plot generates from a fixed hexagon template; I have left it rather than add a check for a shape that cannot arrive. The `pow`-specific error formula and the tolerance comment are as you describe.

Verified after the change: `lint:fix` and `type-check` clean, full unit suite 359 suites / 5,199 tests passing (up one).

---

Issue #1088, opened 2026-08-18
Observable: a box plot whose outliers pull a quartile past its whisker is not recognised, and its medians are announced as a dot plot


jooyoungseo:
Found while building the reading for #1074, in the detection that already exists rather than in anything new. It reproduces on `main` today.

## What happens

`isBoxComposite` in `src/adapters/observable/introspect.ts` requires each whisker to *contain* its box:

```ts
function runsThrough(box: Box, line: Segment): boolean {
  return (spans(box.x, line.x) && contains(line.y, box.y))
    || (spans(box.y, line.y) && contains(line.x, box.x));
}
```

That is true of most box plots and false of a perfectly ordinary one. When outliers sit far enough out, they move a quartile **past** the whisker's end, and the box then extends beyond the whisker rather than inside it.

## Reproduction

One category, six values and two low outliers:

```js
Plot.plot({ marks: [Plot.boxY(
  [40, 42, 44, 46, 48, 50, 10, 4].map(v => ({ g: 'B', v })),
  { x: 'g', y: 'v' },
)] });
```

Measured from what Plot drew (0.6.17), with the `y` scale `[4, 50] → [370, 20]`:

| part | pixels | values |
| --- | --- | --- |
| `bar` (box) | 46.63 … 153.15 | q3 46.5, q1 **32.5** |
| `rule` (whisker) | 20 … 96.09 | max 50, min **40** |
| `tick` (median) | | q2 43 |
| `dot` | | outliers 4 and 10 |

`q1 = 32.5` is below `min = 40`, so `contains([20, 96.09], [46.63, 153.15])` is `false` — the box's lower edge (153.15) falls outside the whisker's span.

Nothing is wrong with the chart. Over `4, 10, 40, 42, 44, 46, 48, 50` the first quartile genuinely is 32.5, the outlier fence is 11.5, and the lower whisker genuinely stops at 40 — the smallest value that is not an outlier. Two low outliers are enough to drag Q1 below it.

## What the reader gets

The composite is not recognised, so the four groups are read as unrelated marks. Measured on a two-category version of the same chart:

```
layer 1  type "dot"   data [{x: 'A', y: 26}, {x: 'B', y: 43}]     ← the medians
layer 2  type "dot"   data [ … the outliers … ]
```

[...]

PR #1089, opened 2026-08-18
feat(observable): read a box plot as a distribution instead of skipping it

### LINKED ISSUE #1074: Observable Plot: boxY / boxX carry a full five-number summary the adapter can read, but nothing marks the four marks as one composite
Raising this rather than implementing it, because the blocker is a design decision rather than a missing extraction.

## The half that works

`Plot.boxY(data, { x: 'g', y: 'v' })` puts every number a `BoxPoint` needs into directly-readable attributes. Measured with `@observablehq/plot@0.6.17` in Chromium, two groups of iris sepal lengths:

```
g[aria-label="rule"]   2 × <line>  x1=x2=centre  y1..y2 = whisker low .. high
   line 0  x1=68  y1=370      y2=218.649
g[aria-label="bar"]    2 × <rect>  y = q3 edge, y+height = q1 edge
   rect 0  x=68  y=247.027  width=248  height=73.311
g[aria-label="tick"]   2 × <line>  y1=y2 = the median
   line 0  x1=68  y1=289.595  x2=316
g[aria-label="dot"]    1 × <circle>  the outliers
   circle  __data__ = 7        (the index of {g:'setosa', v:7.9})
```

Five numbers, all in attributes rather than in a serialised path — the class `docs/observable.md` describes as inverting exactly, not the rounded-path class. Better still, an outlier dot's `__data__` is its **index into the source data**, so the outlier values come from the data itself and never from a pixel.

The marks line up by index the way `BoxTrace` needs: rule `k`, bar `k` and tick `k` share an x centre and belong to category `k`.

## The half that blocks it

A box plot is a composite, and Plot leaves nothing saying so. Compared against a hand-drawn `ruleX` + `barY` + `tickY` that is *not* a box plot:

| | `boxY` | hand-drawn |
|---|---|---|
| rule `<g>` | `aria-label="rule"` `stroke="currentColor"` `transform` | `aria-label="rule"` `stroke="currentColor"` `transform` |
| bar `<g>` | `aria-label="bar"` `fill="#ccc"` | `aria-label="bar"` |
| tick `<g>` | `aria-label="tick"` `stroke="currentColor"` `stroke-width="2"` `transform` | `aria-label="tick"` `stroke="currentColor"` `transform` |

The only differences are `fill="#ccc"` and `stroke-width="2"` — Plot's default styling for a box, which an author can set on a hand-drawn mark and which a `boxY` given `{fill: 'steelblue'}` does not have. They are not a signal.

`convertMark` dispatches on `aria-label`, so under the current model these four groups are four independent marks. Reading them as a box needs a **heuristic** — same element counts, shared x centres, the bar's y-range nested inside the rule's, the tick inside the bar. That heuristic would fire on a range chart with a median line drawn by hand, and announce a five-number summary for a chart that has none. Same class of failure as deriving a violin's quartiles from its KDE (#1068): a reader told "Q1 is 4.8" cannot tell that nothing in the chart said so.

## What it would take

A declaration. The adapter already has `markTypes`, but it maps one `aria-label` to one trace type and cannot say "these four groups are one layer". Options, none obviously right:

1. **Extend `markTypes` to composites** — something like `composites: [{ type: 'box', rule: 0, bar: 0, tick: 0, dot: 0 }]` indexing the marks in draw order. Explicit and unambiguous; verbose, and the indices break when the author adds a mark.
2. **A dedicated option** — `boxPlots: true`, applying the heuristic only when the author has said the chart contains boxes. Cheap to write and to explain, and the heuristic can then be strict (refuse when the groups do not line up) because a false negative is a chart the author knows to fix.
3. **Leave it.** `boxY` is one of a handful of Plot marks MAIDR cannot read, and the docs already say so.

My inclination is (2): it puts the ambiguity where the author can resolve it, keeps the extraction — which measurement shows is exact — and lets the adapter refuse rather than guess when the shapes do not match. But it adds public API for one mark family, so it is worth a maintainer's call before anything is written.

`boxX` is the same chart transposed and would follow whatever is decided here.

Related: `docs/observable.md` § "What it does not read" lists composite marks as an exclusion; this issue is about whether that line should stay.

jooyoungseo:
Closes #1074.

`Plot.boxY` and `Plot.boxX` are read as a `box` trace. Each category is announced with its five-number summary and its outliers, and every number comes back from the pixels the chart was drawn at.

> **Note on reading this PR.** The commit history is not a clean line, and deliberately so: three separate times it shipped a reading that announced fabricated quartiles for a chart that was not a box plot, and each was found by me writing an adversarial chart against my own previous fix rather than by review. The comment thread has the full account. This description reflects where it landed, not how it got there.

## What changed

The adapter already identified a box plot — `boxComposites` walks every `(bar, rule, tick)` combination and requires each box to have a median crossing it and a whisker running past both its ends — and used that to **skip** all four groups so the interquartile box was not announced as a bar. This turns the skip into a reading:

- `boxCompositeGroups` → `boxComposites`, returning the groupings (`{ bar, rule, tick, dot? }`) rather than a set of indices to drop.
- `collectFacetCells` converts each composite, aligning the four groups' facets **by offset** — Plot emits a facet's `dot` group only where that facet has outliers, so the nth dot facet is not the nth box facet.
- `readBoxLayer` sorts the parts along the category axis, pairs them by position, and requires each pair to agree on where its category is.

**No author declaration is involved.** I proposed one on the issue (`markTypes: { bar: TraceType.BOX }`, by analogy with the hexbin in #1087) and withdrew it: asking someone to declare a chart the adapter has already identified makes a correct chart unreadable until a line is typed.

## The hard part: what a box plot is not

The geometry that justifies *skipping* a composite does not justify *reading* one. A false positive costs a skipped chart in the first case and invented statistics in the second — and a floating rect inside a rule with a line across it **is** a box plot's shape. Three ordinary charts share it, all drawn in the same mark order:

| chart | what it really is | read as a box it would announce |
| --- | --- | --- |
| `barRangeAndTarget` | bar + target line + range rule | the bar's height as `q3`, the target as `q2` |
| `bulletChart` | qualitative range, measure, target | the measure as `q3`, the target as `q2` |
| `floatingRangeBar` | candlestick with a marker | open and close as `q1` and `q3` |

No arrangement of the marks separates them, because the difference is what the numbers mean. So the reading asks Plot what it drew: **`Plot.boxY` builds its median from a `tick` with `strokeWidth: 2`**, which a tick mark does not otherwise carry.

Measured against Plot 0.6.17, that survives the box's `fill` **and** `stroke` being overridden, so an author-styled box plot still reads. `fill` does not survive — `{fill: 'steelblue'}` writes `steelblue` — which is why the grey `#ccc` interquartile box is not what is tested. When the mark is faceted Plot writes the styling onto the facet containers instead, so the check looks where `facetGroupsOf` looks.

**Detection is untouched**, so a chart this refuses is skipped exactly as on `main` — this can add a reading, never take one away or introduce a wrong one. What detection *should* do with those three charts, and the false negative it also has, is #1088.

## What measurement corrected

- **The parts of one category do not share a centre.** Plot's half-pixel crispness offset puts `bar`/`tick` at 192 and `rule`/`dot` at 192.5. Exact equality would decline every box plot Plot draws, so pairing tolerates a pixel.
- **A band does not start where its scale's range does.** In `boxHorizontal` the `y` range starts at 20 and the first band sits at 23, so centres are read from the drawn rects.
- **A facet offset is not a measurement.** Plot writes the same translation into every mark's group for a facet, so those are compared exactly — a tolerance there could only make adjacent facets confusable.

## The whisker

`BoxSelector` names `min` and `max` separately; Plot draws one `<line>` spanning both ends, so both fields point at it. `buildViolinBoxSelectors` (`src/adapters/plotly/extractor.ts:3045`) already does this for a plotly violin's inner box over a single path holding whiskers, quartile box *and* median, so Observable is the easier case. Splitting the line would mean inserting elements into the author's chart, which #1004 is the record of not doing.

## Verification

The four existing box fixtures covered the withholding; they now assert the reading, and the two negative cases (`errorBarsWithTarget`, `barWithRule`) are unchanged and still pass. Five fixtures are new, all real `Plot.plot()` output:

- `boxTails` — outliers on both sides, lower tail drawn 14-before-10 so document order is not value order
- `boxesOnBaseline` — counts data where *every* first quartile is genuinely zero
- `barRangeAndTarget`, `bulletChart`, `floatingRangeBar` — the three impostors above

Values are asserted against arithmetic on each fixture's own geometry, never restated from the code. `boxHorizontal` is `boxPlot` transposed and must return the identical five numbers with `horz`, which checks the whole pipeline independently.

Every change falsified by applying each mutation alone:

| Mutation | Tests reddened |
| --- | --- |
| `q1`/`q3` from the rect's top/bottom instead of min/max | 2 |
| ignore the mark group's `translate` | 12 |
| pair facets by index instead of by offset | 2 |
| drop the equal-count requirement | 1 |
| stamp `min` and `max` as separate tokens | 2 |
| one selector token shared by every box | 2 |
| require the parts to agree exactly (tolerance 0) | 12 |
| invert the lower/upper outlier split | 5 |
| report a tail in document order | 1 |
| stamp outliers in document order while values are sorted | 1 |
| drop the median-styling check | 3 |
| read styling from the mark group, never the facet groups | 2 |
| accept `stroke-width: 1` | 6 |
| treat a mixed child set as faceted | 1 |
| never faceted | 5 |

Several of those started as **survivors**, and each was a real gap rather than a redundant guard — no fixture had a lower outlier, none had two outliers in one tail (so nothing pinned that a value stays in step with its own selector), the equal-count guard needed a second decline case to reach, and half the facet rule had never been tested in any of the three places it was written out.

`eslint`, `type-check` and `build` clean with exit codes checked directly; 361 suites / 5,223 tests passing.


jooyoungseo:
Both notes acted on in 7f2a1b1a, and the tolerance one turned out to be more than a maintenance hazard.

**On `PART_TOLERANCE` serving two purposes** — you were right that they are conceptually different, and measuring showed one of them did not want a tolerance at all. A facet's offset is not a measurement of anything: Plot writes the same translation into every mark's group for a given facet, so groups that belong together carry the same *string* and parse to the same number. Across all four marks of `facetedBox`:

```
rule [[0,0],[305,0]]
bar  [[0,0],[305,0]]
tick [[0,0],[305,0]]
dot         [[305,0]]      ← only the facet that has outliers
```

Exact integers, identical. So `facetAt` now compares exactly, and `PART_TOLERANCE` is left to the one thing it was written for — the sub-pixel crispness offset between the parts *inside* a facet. That resolves the coupling by removing the sharing rather than by duplicating the constant, and it is also the safer of the two: a pixel of slack on a facet offset could only ever make two adjacent facets confusable.

Both are still load-bearing after the split, checked separately:

| Mutation | Tests reddened |
| --- | --- |
| `facetAt` returns `facets[0]` (pair by index) | 2 |
| `PART_TOLERANCE = 0` | 14 |

**On the outlier split** — agreed that `>= low` reads as though it is missing a check, and the comment now says what it rests on: a `dot` in this mark is an outlier by construction, since Plot draws it only for observations outside the whiskers. Anything not below the lower end is above the upper one, and testing that separately would create a third branch for a point that cannot occur.

Re-verified: `lint:fix` and `type-check` clean, `build` succeeds, 360 suites / 5,215 tests passing.

On the verification note — understood, and no need to grant Bash approval on my account: the numbers above are from runs in this session, and CI's own Unit Tests, Type Check, Lint and Build jobs are green on the PR independently of either of us.

---

jooyoungseo:
Hold on this — I found a regression in my own change while investigating #1088, and pushed a fix in b221884a. Flagging it rather than quietly amending, because the PR body's central claim ("can add a reading, never introduce a wrong one") was false as written.

## What was wrong

The composite check is satisfied by charts that are not box plots. This one is:

```js
const rows = [
  { g: 'a', v: 10, lo: 0, hi: 12, target: 6 },
  { g: 'b', v: 7, lo: 0, hi: 9, target: 5 },
];
Plot.plot({ marks: [
  Plot.barY(rows, { x: 'g', y: 'v' }),
  Plot.ruleX(rows, { x: 'g', y1: 'lo', y2: 'hi' }),
  Plot.tickY(rows, { x: 'g', y: 'target' }),
] });
```

A bar chart with a target line inside each bar and a range rule that happens to reach the baseline. The tick lies inside the bar, and the rule runs along it and past both ends — so `crosses` and `runsThrough` both pass, and the composite is claimed. Run through the adapter as this PR stood:

```json
{ "z": "a", "lowerOutliers": [], "min": 0, "q1": 0, "q2": 6, "q3": 10, "max": 12, "upperOutliers": [] }
```

`10` is the bar's own height announced as the third quartile, `6` is the target line announced as the median, and none of the five numbers is a quartile of anything. On `main` this chart is claimed and skipped, so it reads as nothing; the change turned silence into confident wrong data, which is worse and is the one direction this was not supposed to move in.

## The fix

`Plot.boxY` emits its parts in one fixed order, because they come from one call rather than from three someone wrote:

| fixture | mark order | box plot? |
| --- | --- | --- |
| `boxPlot` | `rule → bar → tick → dot` | yes |
| `boxHorizontal` | `rule → bar → tick → dot` | yes |
| `ruleThenBox` | `rule → rule → bar → tick → dot` | yes |
| `facetedBox` | `rule → bar → tick → dot` | yes |
| `boxTails` | `rule → bar → tick → dot` | yes |
| `barRangeAndTarget` | `bar → rule → tick` | **no** |
| `errorBarsWithTarget` | `bar → rule → tick` | **no** |

The reading now requires `rule` before `bar` before `tick`. That is a statement about where the marks came from rather than about how they happen to overlap, and it is strictly narrowing — a composite refused there is skipped exactly as before, so this really can only decline more.

`barRangeAndTarget` is a new fixture (real Plot output) and the case is pinned; removing the guard reddens it.

## What I did not change

Detection still claims that chart, so it is still skipped rather than handed back to be read as the bar chart it is. Fixing that means changing what `isBoxComposite` accepts, which is the same decision as the false negative — so I have added this to #1088 rather than doing it here. Both halves of that issue are now measured: the check is wrong in **both** directions, rejecting real box plots whose outliers pull a quartile past a whisker, and accepting bar charts whose parts happen to line up.

Re-verified after the fix: `lint:fix` and `type-check` clean, build succeeds, 360 suites / 5,216 tests passing.

---

jooyoungseo:
The draw-order fix in my previous comment was **not enough**, and I found the hole by trying to break it rather than by waiting for someone else to. Replaced in 31becf3e.

## What it missed

I said the draw-order rule was necessary but not sufficient, so I went looking for a chart that satisfies it. There is an obvious one, and it is a real chart type:

```js
const rows = [
  { g: 'revenue', v: 10, lo: 0, hi: 14, target: 8 },
  { g: 'profit', v: 6, lo: 0, hi: 9, target: 4 },
];
Plot.plot({ marks: [
  Plot.ruleX(rows, { x: 'g', y1: 'lo', y2: 'hi' }),   // qualitative range
  Plot.barY(rows, { x: 'g', y: 'v' }),                // the measure
  Plot.tickY(rows, { x: 'g', y: 'target' }),          // the target
] });
```

A **bullet chart**, written in the order anyone would write one — which is the same order `boxY` emits its parts. It passed the new guard and came out as:

```json
{ "z": "revenue", "min": 0, "q1": 0, "q2": 8, "q3": 10, "max": 14 }
```

`10` is the measure, `8` the target, `14` the top of the range.

## What actually separates them

Where the rect stands. `Plot.barY` draws from zero by definition, so a measure bar rests on the baseline; a box rests on `q1`, which is a quantile of the data and is zero only by coincidence. The check counts boxes on the baseline and declines when every one of them is there, reusing the `baselinePixel` / `countTouching` pair the bar reading already uses for the same distinction.

This **replaces** the draw-order test rather than joining it, for two reasons: it catches strictly more (the bar chart from the previous round as well as the bullet chart), and it is a statement about what the marks mean rather than about the order Plot happens to emit them in — an implementation detail that could change, and would then silently stop every box plot being read.

## Counted, not found

`onBaseline === boxes.length`, not `> 0`. The stricter version passed every test I had, which meant nothing justified the choice — so `boxOnBaseline` is a new fixture: a real box plot whose first category is counts data with enough zeros that its first quartile genuinely is zero, so one box stands on the baseline and the other does not.

| Mutation | Tests reddened |
| --- | --- |
| drop the baseline guard | 2 (`barRangeAndTarget`, `bulletChart`) |
| decline when *any* box touches the baseline | 1 (`boxOnBaseline`) |

Both fixtures are real `Plot.plot()` output, and `boxOnBaseline`'s quartiles are asserted against its own geometry: `q1` of `[0, 13.5]`, `q3` of `[4.5, 20.5]`.

## Standing back

Two rounds of this PR shipped a reading that announced fabricated quartiles for a chart that was not a box plot, and in both cases the geometry check that has been in the repository all along was what let it through. That is the substance of #1088, which I have updated: the composite test is wrong in both directions, and this PR now guards the reading against both of its false positives without touching detection — so a chart it refuses is skipped exactly as on `main`.

Re-verified: `lint:fix` and `type-check` clean, build succeeds, 360 suites / 5,218 tests passing.

---

jooyoungseo:
Third false positive, and this one changed the approach rather than adding another patch. d6df896a.

## What got through the baseline test

A **candlestick on a categorical axis with a marker in its body** — a high-low rule, a floating open-close bar, a tick:

```js
const rows = [
  { d: 'Mon', low: 4, high: 14, open: 7, close: 11, mark: 9 },
  { d: 'Tue', low: 6, high: 18, open: 9, close: 15, mark: 12 },
];
Plot.plot({ marks: [
  Plot.ruleX(rows, { x: 'd', y1: 'low', y2: 'high' }),
  Plot.barY(rows, { x: 'd', y1: 'open', y2: 'close' }),
  Plot.tickY(rows, { x: 'd', y: 'mark' }),
] });
```

Its box stands on nothing, so the baseline test cannot see it:

```json
{ "z": "Mon", "min": 4, "q1": 7, "q2": 9, "q3": 11, "max": 14 }
```

— the open and the close announced as the quartiles.

## The actual conclusion

Three charts in a row walked past a geometric test, and that is the finding rather than three separate bugs. **A floating rect inside a rule with a line across it simply is a box plot's shape.** No arrangement of those three marks distinguishes a box plot from a bullet chart or a candlestick, because the difference is what the numbers mean, and meaning is not in the geometry. Each guard I added was a proxy for provenance, and each one had a chart that satisfied it.

So the reading stops inferring and asks Plot what it drew. `Plot.boxY` builds its median from a `tick` with `strokeWidth: 2`, which a tick mark does not otherwise carry:

| chart | median's `stroke-width` |
| --- | --- |
| every `boxY` / `boxX` fixture | `2` |
| `barRangeAndTarget`, `bulletChart`, `floatingRangeBar` | absent |

Measured on 0.6.17: it survives the box's `fill` **and** `stroke` being overridden, so a chart styled by its author still reads. `fill` does not survive — `Plot.boxY(data, {fill: 'steelblue'})` writes `steelblue` — which is why the grey `#ccc` box is not what is tested. When the mark is faceted Plot writes the styling onto the facet groups instead of the mark, so the check looks where `splitFacets` looks.

## Why this replaced the baseline test rather than joining it

Falsification said the baseline guard had become indistinguishable — dropping it reddened nothing once provenance was read directly. But it was worse than redundant: **`boxesOnBaseline`**, a real box plot of counts data with enough zeros that *every* first quartile is zero, is a chart the baseline test would have refused. A proxy that is no longer needed had turned into a source of false negatives, so it is gone and that chart is now pinned as reading correctly (`q1` of `[0, 0]`, `q3` of `[6.75, 11.25]`).

| Mutation | Tests reddened |
| --- | --- |
| drop the median-styling check | 3 (all three false positives) |
| look at the mark group only, never the facet groups | 2 |
| accept `stroke-width: 1` | 6 |

## On the shape of this pull request

It has now shipped a fabricated-quartiles reading three times, each caught by me rather than by review, and each time the fix was too weak in a way the next measurement exposed. The principle I should have applied from the first commit is the one now written into the code: **the evidence that justifies withholding a composite is not the evidence that justifies reading one.** `boxComposites` was written to decide whether to stay quiet, where a false positive costs a skipped chart; the reading needs a different and stronger question answered, because there a false positive costs invented statistics.

Verified on the pushed tree: `lint:fix` and `type-check` clean, `build` succeeds (`All builds complete in 139.8s`), 360 suites / 5,219 tests passing.

---

jooyoungseo:
CI's Lint Code step failed on d6df896a and I had reported that same tree as lint-clean. Fixed in 4b3a3d0f, but the reason I missed it is worth writing down because it invalidates several verification claims I have made on this PR.

**The error** was `ts/no-use-before-define`: `BOX_MEDIAN_STROKE` declared below the function that reads it. A one-line move.

**Why my run said otherwise.** I was invoking it as

```bash
npm run lint:fix 2>&1 | tail -1 && npm run type-check && ... && echo "ALL VERIFIED"
```

A pipeline's exit status is the *last* command's, so `tail` returning 0 masked eslint returning 1 — and `&&` then happily ran the type-check, the tests and the build, ending in `ALL VERIFIED` on a tree that could not pass CI. Every "lint clean" I have written on this PR came through that pipe and was worth less than it looked.

It is the same mistake in a different costume: earlier in this work `git merge … | tail` masked a conflict exit code the same way. I noticed it then, wrote it down, and repeated it an hour later.

Re-verified without the pipe, checking exit codes directly:

```
npx eslint .        → ESLINT_EXIT=0
npm run type-check  → TYPECHECK_EXIT=0
jest (full)         → 360 suites, 5219 tests, all passing
```

The build was independently confirmed green on the same code before the constant moved (`All builds complete in 139.8s`, exit 0), and a statement reorder cannot change it — but CI is the check that counts here, not my say-so, and it is running now.

---

jooyoungseo:
Acted on the drift note in 7bed72a8, and it turned up something.

You flagged that `drawnAsMedian`'s facet-detection predicate duplicates `splitFacets`'. It was actually a **third** copy — `leafElements` has it too, and those two predate this PR:

```
introspect.ts:384   leafElements
introspect.ts:419   splitFacets
converters.ts:531   drawnAsMedian        ← mine
```

With three callers there is now a reason to have the rule in one place rather than a comment asking future readers to keep them in step, so `facetGroupsOf` holds it and all three ask.

**Pinning it showed half of it was never tested.** The rule has two parts — some child is a group, *and* every last child is one — and mutating the second half away:

```ts
return facets.length > 0 && facets.length === children.length ? facets : [];
// →
return facets;
```

passed all 5,219 tests. No fixture has a mark that draws groups of its own alongside other elements, which is exactly the case the condition exists for (the comment names an axis tick's line and its label). That gap was in the codebase before this PR and would have stayed invisible while the rule was written out three times.

`facetGroupsOf` is exported, so the new tests exercise the rule directly rather than through one of its three uses:

| Mutation | Tests reddened |
| --- | --- |
| treat a mixed child set as faceted | 1 (the new case) |
| never faceted | 5 (the new case, three faceting tests, the faceted box) |

On your `n = 1` note about the previous commit — that edge case is gone rather than narrowed, since the baseline test it belonged to is no longer there. A single-category box plot with `q1 = 0` now reads like any other, and `boxesOnBaseline` covers the multi-category form of it.

Verified with exit codes checked directly, not through a pipe: `eslint` 0, `type-check` 0, 361 suites / 5,223 tests passing.

---

PR #1090, opened 2026-08-18
fix(observable): recognise a box plot by what Plot drew, not by where its parts sit

### LINKED ISSUE #1088: Observable: a box plot whose outliers pull a quartile past its whisker is not recognised, and its medians are announced as a dot plot

jooyoungseo:
Closes #1088.

The composite check was wrong in **both** directions, and one signal fixes both.

## It rejected real box plots

Outliers move a quartile, and enough of them on one side pull it past the whisker's end — so the box sticks out rather than sitting inside it. Over `4, 10, 40, 42, 44, 46, 48, 50` the first quartile is 32.5 while the lower whisker stops at 40, the smallest value that is not an outlier. Nothing is wrong with that chart.

Requiring the whisker to *contain* the box left it unrecognised, and its four marks were then read separately:

```
layer 1  type "dot"   data [{x: 'A', y: 43}, {x: 'B', y: 25}]     ← the medians
layer 2  type "dot"   data [ … the outliers … ]
```

The medians announced as the data points of a dot plot; the box, the quartiles and the whiskers never announced at all.

## It accepted charts that were not box plots

A bar chart with a target line, a bullet chart, and a candlestick with a marker in its body all satisfy the geometry, because a floating rect inside a rule with a line across it simply **is** a box plot's shape. Those were claimed and then skipped, so they went out silent — nothing announced for an ordinary chart.

## One signal for both

`Plot.boxY` builds its median from a `tick` with `strokeWidth: 2`, which a tick mark does not otherwise carry. #1089 established that reading it settles what the arrangement cannot; moving that check into the recognition does both jobs at once.

**The impostors stop being claimed**, so their marks are handed back and read as what they are rather than dropped:

| chart | before | after |
| --- | --- | --- |
| `bulletChart` | nothing announced | bar `[6, 10]` + dot `[8, 4]` |
| `barRangeAndTarget` | nothing announced | bar `[10, 7]` + dot `[6, 5]` |
| `floatingRangeBar` | nothing announced | dot `[9, 12]` |

**And the geometry is free to be loose enough to be true of every box plot.** `runsThrough` (containment) becomes `runsAlong` (overlap), and `boxPastItsWhisker` now reads:

```json
{ "z": "A", "lowerOutliers": [4, 10], "min": 40, "q1": 32.5, "q2": 43, "q3": 46.5, "max": 50, "upperOutliers": [] }
```

The **ordering matters**. Relaxing the geometry on its own is what I said on the issue was worth being afraid of, because it would leave the distinction resting on the median crossing the box — which a bullet chart also satisfies. Behind the provenance check it costs nothing.

## What this removed

- The reading's own copy of the median check, now unreachable: with the strong question answered in detection, the two are no longer different questions. Mutating it away reddens nothing.
- `contains`, which had no callers left.

## A check that had never been tested

Verifying the relaxation, I mutated `runsAlong` to `return true` and **all 135 observable tests still passed**. `ruleThenBox` looks like the test for it — "finds the box even when another rule is drawn first" — but a baseline `ruleY([0])` draws *one* line for *two* boxes and is turned away by the count guard before the geometry is consulted. The whisker's geometry had no coverage, and neither did `runsThrough` before it.

`referenceLinesThenBox` is the case that needs it. `Plot.ruleY([12, 26])` drawn before a two-category box plot gives two rule groups with **two lines each**, so counting cannot choose between them:

```
rule(2)   ← the reference lines, drawn first
rule(2)   ← the whiskers
bar(2)  tick(2)
```

Taking the first announces 12 and 26 as the categories' extremes. The test asserts `min`/`max` of `[10, 20]` and `[20, 30]`.

## Verification

Three new fixtures, all real `Plot.plot()` output: `boxPastItsWhisker`, `referenceLinesThenBox`, and the reading assertions against each one's own geometry.

| Mutation | Tests reddened |
| --- | --- |
| drop the median-styling check from detection | 1 (the bullet chart is claimed and silent again) |
| restore containment in place of overlap | 1 (`boxPastItsWhisker`) |
| whisker geometry always satisfied | 1 (`referenceLinesThenBox`) |
| drop the reading's copy of the median check | 0 — which is why it is gone |

`eslint`, `type-check` and `jest` all exit 0, checked directly rather than through a pipe; 361 suites / 5,226 tests passing.

---


jooyoungseo:
Both points acted on in 95a380a9, and the second one was a sharper observation than "slightly less coverage".

**The stale docstring** — you are right, and it was stale in both of the ways that matter. It claimed every whisker *contains* its box, which is the check this change replaces, and it claimed the arrangement is "conclusive enough to read the composite" — which is precisely the claim the change exists to contradict. `docs/observable.md` had been rewritten and the function had not. It now says the same thing as the docs: two gates, neither sufficient alone, and the geometry is only as strict as every box plot really is *because* provenance answers what it cannot.

**The coverage note.** `not.toContain(TraceType.BOX)` passes whether a chart is skipped or handed back, so those two assertions could not see the difference this change makes at all — the PR's table claimed corrected behaviour they were not watching. All three now assert what they are actually read as:

| fixture | layers |
| --- | --- |
| `bulletChart` | `[bar, dot]` |
| `barRangeAndTarget` | `[bar, dot]` |
| `floatingRangeBar` | `[dot]` |

The third is not an omission: a candlestick's body floats free of the baseline, so the bar reading declines it on its own account — a ranged bar has two ends and a bar point has room for one — and only its marker is read. Worth having written down, since "why does this one differ" is the first question the table raises.

Dropping the median check now reddens **six** tests rather than three.

**One thing the fix itself turned up.** My first version of that table used `it.each([...] as const)`, which type-checks fine as a value and fails as a jest argument — `TYPECHECK=2`, while `eslint` and the full suite both passed. I only saw it because I now capture exit codes rather than piping to `tail`; the pipeline habit that broke #1089's lint would have hidden this one too, and CI would have caught it a second time.

Verified: `eslint` 0, `type-check` 0, `jest` 0 — 361 suites / 5,229 tests.

---

PR #1096, opened 2026-08-19
feat(observable): read a waffle chart as a bar chart of its tallies

### LINKED ISSUE #1093: Observable Plot: a waffle chart is not read at all
Found while sweeping the marks the Observable Plot adapter does not read.

## What happens

`Plot.waffleY` draws a group labelled `waffle`, which `convertMark` does not have a case for, so it falls to `default: return null`. A chart whose only mark is a waffle is left unbound entirely.

A waffle is a bar chart drawn as a count of cells, and the count is in the markup exactly — no colour inversion and no approximation.

## Measured

`@observablehq/plot@0.6.17`, `waffleY([{g:'A',v:12},{g:'B',v:7},{g:'C',v:19}], {x:'g', y:'v'})`:

```
g[aria-label="waffle"]
  <pattern id="plot-pattern-1-0" width="55.263" height="55.263">…</pattern>   ×3
  <path fill="url(#plot-pattern-1-0)" transform="translate(60.105,370)"
        d="M0,0…L165.789,0L165.789,-221.053L0,-221.053…Z"/>                   ×3
```

- The `<pattern>` gives the cell size: `55.263` square.
- The `<path>` is the outline of the filled cells — a rectangle when the count fills whole rows, a staircase when the last row is partial.
- `165.789 / 55.263 = 3` columns; `221.053 / 55.263 = 4` rows; `3 × 4 = 12`, which is A's value exactly. B's outline steps from a full 3-wide block two rows tall to a 1-wide top row: `6 + 1 = 7`. C's is `18 + 1 = 19`.

So `count = polygon area / cell area`, and it comes back as an integer rather than as an inverted pixel. The category is the band the `transform` lands in, the same way a bar's is.

## Scope

The obvious reading is a bar layer of counts, which is what the chart is: `x` the category, `y` the tally. That needs no new trace type and no declaration.

Two things worth knowing before a fix is written:

- **A waffle can be stacked or faceted**, and neither was measured here. A `fill` channel presumably splits one category's cells across several paths in the same band, which is a stacked bar and needs the same treatment the stack reading already has.
- **`Plot.waffleX`** — the horizontal form — was not measured either. If it lays cells out in columns rather than rows, the area arithmetic is unchanged but the category axis moves.
- The pattern is referenced by `fill="url(#…)"`, so the cell size has to be looked up through the id rather than read off the path. A path whose pattern cannot be resolved has no cell size and so no count, and should be declined rather than guessed at.

jooyoungseo:
Closes #1093.

## What was missing

`Plot.waffleY` and `Plot.waffleX` draw a group labelled `waffle`, which `convertMark` had no case for, so both fell to `default: return null` and a chart whose only mark was a waffle was left unbound entirely.

## How the tally is read

A waffle draws each category as one `<path>` around its filled cells and fills it with a `<pattern>` giving the cell's size. So the count is the outline's **area over the cell's** — the shoelace formula, which does not care whether the outline is a rectangle or the staircase a partial last row leaves. Nothing is inverted from a colour and nothing is approximated.

What gets announced is not the tally but the height those cells would have as a solid bar, put back through the scale. The two agree only while a cell is one unit:

```
waffleY(rows, {unit: 5})   values 12, 7, 19   ->   area ÷ cell = 2.4, 1.4, 3.8
```

Announcing that would report fractions of a cell as the data. Going through the scale returns what the axis shows, and never has to learn what `unit` was.

## Measured

`@observablehq/plot@0.6.17`. Recovered against drawn, exactly, in every layout:

| chart | drawn | lanes | recovered |
|---|---|---|---|
| `waffleY` | 12, 7, 19 | 3 | 12, 7, 19 |
| `waffleY`, partial first row | 2, 17 | 3 | 2, 17 |
| `waffleY`, `unit: 5` | 12, 7, 19 | 1 | 12, 7, 19 |
| `waffleY`, `unit: 2` | 12, 7, 19 | 2 | 12, 7, 19 |
| `waffleY`, large | 240, 97 | 13 | 240, 97 |
| `waffleY`, tiny | 1, 2 / 3, 5 | 1 | 1, 2 / 3, 5 |
| `waffleX` | 12, 7, 19 | 1 | 12, 7, 19 |
| `waffleY` + `fill` | A 5+7, B 3+4 | 4 | 5, 7, 3, 4 |
| fractional | 2.5, 4 | 1 | 2.5, 4 |

Three things fell out of the measurement rather than being assumed:

- **The lattice's width has to be taken across the whole mark.** A category holding less than a full row is only as wide as the cells it drew, so a width taken from it divides every tally on the chart by the wrong number. Some category always fills a row, because Plot sizes the lattice to the largest value — checked down to values of 1 and 2, where it picks one column.
- **`waffleX` needs no separate arithmetic.** It lays cells along the band rather than up it, so each outline is a plain rectangle and the same area gives the same count.
- **A zero category is drawn, not omitted** — `M0,0L0,0…Z`, every vertex on the origin, enclosing no area. So it reads as the zero it is. That also made a minimum-vertex guard I had written unnecessary, and it was removed rather than left as an untested branch.

The series colour for a `fill`-split waffle is on the pattern's `<rect>`, not on the path, which carries only `url(#…)` — so the series is named through the colour scale the same way every other mark names one, just from a step further in.

## Tests

`test/adapters/observable/waffle.test.ts`, on seven new fixtures captured verbatim from real `Plot.plot()` output. Each mutation applied alone:

| mutation | tests reddened |
|---|---|
| dispatch case removed | 9 |
| lanes dropped from the divisor | 6 |
| lanes taken from the first path only | 2 |
| cell axes swapped | 1 |
| growth direction inverted | 9 |
| series never resolved through the colour scale | 1 |
| subpath guard dropped | 1 |
| lanes rounded up instead of to nearest | 1 |
| category read off the wrong axis | 9 |
| `url(#…)` regex loosened | 9 |
| pattern lookup falls back to any pattern | 1 |

The last one is worth noting: it reddened **nothing** on the first pass, because the "not a lattice" test used a plain colour fill, which fails the `url(#…)` regex before the id lookup runs. A second case — a reference to an id that resolves to nothing — was added to reach it.

`npx eslint .`, `npm run type-check` and the unit project all exit 0: 364 suites, 5,245 tests.

## Not covered

A **faceted** waffle. The lane count would have to be taken per facet rather than per mark, and I have not measured what Plot draws there.

---


jooyoungseo:
Both observations checked. The first is real in the code and unreachable in practice; the second I'm leaving as is, with a reason.

## The all-zero waffle

You're right about the arithmetic — `widest` would be 0, `lanes` would round to 0, and the mark would be dropped. It cannot be reached, because Plot does not draw that chart at all. Measured on `@observablehq/plot@0.6.17`:

| chart | `g[aria-label="waffle"]` | paths | patterns |
|---|---|---|---|
| `waffleY`, one category, value 0 | **absent** | 0 | 0 |
| `waffleY`, two categories, both 0 | **absent** | 0 | 0 |
| `waffleY`, three categories, all 0 | **absent** | 0 | 0 |
| `waffleX`, all 0 | **absent** | 0 | 0 |

No group means `convertMark` is never handed a `waffle` facet, so the question of what `convertWaffle` would do with it does not arise. A zero *among* non-zero categories is the case that does occur, and it is drawn (`M0,0L0,0…Z`, every vertex on the origin) and read as a zero — that's the `zeroWaffle` fixture.

So the conscious call is: the guard stays, because a lattice of no lanes must not divide a tally, but it is not the all-zero case it looks like. `6e940a19` records that measurement next to the guard so the next reader doesn't have to re-derive it.

## The pattern scan

Leaving it. `patternOf` is O(patterns) per patch, so O(n²) in categories as you say — but the constant is one `getAttribute` per comparison, and a waffle's category count is bounded by what fits on a chart legibly. At 50 categories that's 2,500 attribute reads once per conversion, which is not measurable against the DOM traversal the rest of the adapter already does. Building an id map first would be faster and would add a lifetime to manage for a cost nothing pays.

Worth noting the alternative I did reject on correctness rather than speed: `querySelector('#id')`, which is O(1) but only correct when the id happens to be a valid CSS identifier, and `CSS.escape` is not something this adapter can assume in every document it is handed.

## Verification

CI runs `npm test`, `npm run type-check`, `npm run lint` and the E2E suite on every push, so those are covered where your sandbox couldn't. Locally on this tree: `npx eslint .`, `npm run type-check` and the unit project all exit 0 — 364 suites, 5,245 tests. I check the exit codes directly rather than reading the output, having been caught earlier in this branch's history by `cmd | tail` swallowing a non-zero status.

---

Issue #1092, opened 2026-08-19
Observable Plot: an area given an explicit y1/y2 band announces the band's width as the data


jooyoungseo:
Found while sweeping the marks the Observable Plot adapter does not read. **Retitled and rescoped** after measuring further — see "Correction" below; the first version of this issue named Bollinger charts specifically and proposed a pairing signal that does not exist.

## What happens

`convertLine`'s area reading takes a band's value to be its height: `y = top − bottom`. That is right for a **stack**, where the lower edge is the series below and the band between them is this series' own value — it is the fix #788 was written for.

It is wrong for an area whose lower edge is not a series. Given `y1`/`y2` explicitly, an area is an *interval*: a confidence band, a min/max range, a Bollinger band. Its lower edge is not data below it, and the distance between the edges is not a value the chart is about.

## Measured

`@observablehq/plot@0.6.17`, read through `observablePlotToMaidr`.

**A plain confidence band** — `areaY(rows, {x: 't', y1: 'lo', y2: 'hi', fillOpacity: 0.2})` over `10±2, 13±2, 11±2, 15±2`:

```
area  [{x:0,y:4},{x:1,y:4},{x:2,y:4},{x:3,y:4}]
```

A flat 4 — the width — announced four times. The chart's real values, `8/12`, `11/15`, `9/13`, `13/17`, appear nowhere. A reader arrowing across hears one unchanging tone.

**A Bollinger chart** — `bollingerY(rows, {x: 't', y: 'v', n: 4, k: 1})` over `v = 10…21` — is the same defect with a second mark beside it:

```
area  [{x:3,y:2.582}, … ×9]        <- 2σ, the band's width
line  [{x:3,y:11.5},{x:4,y:12.5}, … {x:11,y:19.5}]   <- the moving average, correct
```

Both series are constant only because these fixtures are linear; on real data the width varies, and it is still the width announced as the data.

## Correction

The first version of this issue treated the Bollinger case as its own and suggested pairing the `area` with the `line` beside it. Measuring the markup shows there is nothing to pair on. `Plot.bollingerY` is `marks(areaY(…), lineY(…))` and leaves no trace of itself:

```
bollingerY        area  {aria-label: "area", fill-opacity: "0.2"}
plainBand+line    area  {aria-label: "area", fill-opacity: "0.2"}     <- hand-written, identical
bollingerY        line  {aria-label: "line", fill: "none", stroke: "currentColor", stroke-width: "1.5", …}
[...]

PR #1099, opened 2026-08-20
feat(observable): read a spike mark as a magnitude standing at a place

### LINKED ISSUE #1098: Observable Plot: `spike` and `vector` marks are not read
The last mark from the Observable Plot sweep that carries data and is not read. `Plot.spike` and `Plot.vector` both draw a group labelled `vector`, which `convertMark` has no case for, so both fall to `default: return null`.

The two are not the same problem, and only one of them is a gap I would fix.

## A spike is exactly readable

A spike map is a magnitude standing at a position, and both halves are in the markup:

```
Plot.spike(pts, {x, y, length: 'm'})    length scale: linear [0, 12] -> [0, 18]

  translate(40,253.333)   d=M-3.5,0L0,-7.5L3.5,0     ->  7.5/18 × 12 =  5
  translate(330,20)       d=M-3.5,0L0,-18L3.5,0      ->   18/18 × 12 = 12
  translate(620,370)      d=M-3.5,0L0,-12L3.5,0      ->   12/18 × 12 =  8
```

Drawn from `m = 5, 12, 8`. The position is the `transform`, inverted through x and y like any other mark's; the magnitude is the drawn length inverted through the `length` scale. Exact, no approximation, no colour.

## A vector has a direction there is nowhere to put

`Plot.vector` draws the same length with an arrowhead, and adds `rotate`:

```
Plot.vector(pts, {x, y, length: 'm', rotate: 'r'})

  translate(40,253.333) rotate(0)  translate(0,3.75)  d=M0,0L0,-7.5M-1.5,-6L0,-7.5L1.5,-6
  translate(330,20)     rotate(45) translate(0,9)     d=M0,0L0,-18M-3.6,-14.4L0,-18L3.6,-14.4
  translate(620,370)    rotate(90) translate(0,6)     d=M0,0L0,-12M-2.4,-9.6L0,-12L2.4,-9.6
```

A flow field is a direction **and** a magnitude at each position, and the direction is the part the chart is usually drawn for. MAIDR has no field for a bearing, so reading only the magnitude would announce half of what is there without saying which half — which is the failure mode #1092 was about, at a chart's scale rather than a series'.

The two are distinguishable in the markup: a spike is one subpath (`M…L…L…`, a triangle), a vector is two (shaft then head), and the rotation is on the `transform` where it can be seen.

## The open question

`ScatterPoint` carries an optional `z?: number`, which the trace passes through and announces, so a spike's magnitude has somewhere plausible to go. I have **not** checked what `z` means in the announcement — whether it reads as a third measurement, a bubble's size, or something else — and putting a spike's height there without knowing would be guessing at the reader's experience rather than the markup.

So the scope is:

- **A spike with no rotation** is worth reading, once it is settled what field its magnitude belongs in and how that is announced.
- **A vector with a rotation** should stay unread and be documented as such, rather than announced as a spike with its direction quietly dropped.
- **A vector whose rotation is zero everywhere** is a spike drawn with a different arrowhead, and could go either way — I have not measured whether Plot omits the `rotate()` entirely in that case or writes `rotate(0)`, which is what would decide whether the two can be told apart at all.

jooyoungseo:
Closes #1098.

## What was missing

`Plot.spike` and `Plot.vector` share the `vector` label, which `convertMark` had no case for, so a spike map went unread entirely.

## How a spike is read

Both halves are written down. The place is the mark's `transform`; the magnitude is how far it reaches, put back through the `length` scale that sized it.

```
Plot.spike(pts, {x, y, length: 'm'})      length scale: linear [0, 12] -> [0, 18]

  translate(40,253.333)  d=M-3.5,0L0,-7.5L3.5,0   ->  7.5 -> 5
  translate(330,20)      d=M-3.5,0L0,-18L3.5,0    ->   18 -> 12
  translate(620,370)     d=M-3.5,0L0,-12L3.5,0    ->   12 -> 8
```

Drawn from `m = 5, 12, 8`. Announced as a scatter carrying `z` — a labelled third measurement the trace speaks alongside the position *and* scales into an audio intensity across the layer's range, so the magnitude is both said and heard.

## The sign the direction carries

The scale sizes a **distance**, so a negative magnitude reaches down rather than up:

```
m = 5, -8, 12
  d=M-3.5,0L0,-7.5L3.5,0     up 7.5
  d=M-3.5,0L0,12L3.5,0       down 12     <- |-8| through the same scale
  d=M-3.5,0L0,-18L3.5,0      up 18
```

My first implementation measured the rise as `-min(y)`, which gives **0** for that middle spike and dropped it silently. A spike map of net migration would have lost every loss. The apex is now taken as the vertex furthest from the base in either direction, and its sign is the magnitude's.

That bug survived the first test pass and was caught by mutation testing: swapping `-min(y)` for `max|y|` reddened nothing, because every fixture I had grew upward.

## What is refused

**A vector that points somewhere.** A flow field's direction is usually what it was drawn for, and the grammar has no field for a bearing, so reading only the magnitude would announce half a chart without saying which half.

The discriminator is measured, not assumed. Plot writes `rotate()` into the transform exactly when a rotation channel was given — **even where every value is zero**:

```
vector(pts, {x, y, length})                translate(40,253.333) translate(0,3.75)
vector(pts, {…, rotate: 'r'})  r = 0,45,90 translate(40,253.333) rotate(0) translate(0,3.75)
vector(pts, {…, rotate: 'r'})  r = 0,0,0   translate(40,253.333) rotate(0) translate(0,3.75)
```

So the question asked is whether a direction was *encoded*, not whether it varied. Worth contrasting with #1092: there a constant floor is read as a base, because the drawing cannot tell a chosen base from a constant bound. Here it can, so a constant direction is still a direction.

**A vector with no `length` channel**, too — every arrow is then the same default height and the only number to hand is the mark's own styling.

## Tests

`test/adapters/observable/spike.test.ts`, on six new fixtures captured verbatim from real `Plot.plot()` output. Each mutation applied alone:

| mutation | tests reddened |
|---|---|
| dispatch case removed | 3 |
| rotation guard dropped | 2 |
| sign of the rise ignored | 1 |
| raw pixels announced as the magnitude | 3 |
| `length` never read into the scale set | 3 |
| apex taken as the minimum `y` | 1 |

One guard was **removed** rather than kept: an early `isContinuous(scales.length)` check reddened nothing, because inverting through a scale that is not there already yields nothing and drops the point. The lengthless case is still pinned by its own fixture, and the reasoning now lives at the point where it actually happens.

Adds `length` to the adapter's scale set (`scales.ts`, `types.ts`) — two lines, and mutation-checked.

`npx eslint .`, `npm run type-check` and the unit project all exit 0: 366 suites, 5,262 tests.

---


jooyoungseo:
The finding is right, and the degenerate path is now measured rather than assumed. `@observablehq/plot@0.6.17`, magnitudes `[5, 0, 12]`, length scale `[0, 12] → [0, 28.8]`:

```
<path transform="translate(40,253.333)" d="M-3.5,0L0,-12L3.5,0">
<path transform="translate(330,20)"     d="M-3.5,0L0,0L3.5,0">
<path transform="translate(620,370)"    d="M-3.5,0L0,-28.8L3.5,0">
```

Exactly the shape you predicted. `Plot.vector` is the same story: `M0,0L0,0M0,0L0,0L0,0`. The place is on the chart, so the reading should have an answer for it.

Fixed in cc68f9b0. `readPathGeometry` already reports an unreadable path as `null`, so `spikeRise` had no reason to invent a second meaning for it — the apex is returned as it stands, and `null` now means only what it says.

The fix carried a second, smaller bug with it. The sign was written `rise < 0 ? size : -size`, so a rise of zero fell to the `-size` branch and produced `-0` — a magnitude with a minus sign in front of it and no side to be on. Written `rise > 0 ? -size : size` the flat case keeps a positive zero, and the two non-zero branches are unchanged. Jest's `toEqual` does distinguish `-0` from `0` (checked), so this is pinned rather than merely intended.

Two tests on a new `zeroSpike` fixture, both falsified:

| mutation | reddens |
|---|---|
| `spikeRise` restored to `apex === 0 ? null : apex` (the reported bug) | both new tests |
| sign restored to `rise < 0 ? size : -size` (yields `-0`) | both new tests |
| sign dropped entirely (`z = size`) | `keeps the sign the direction carries` |

Full suite green: eslint 0, type-check 0, jest 0 — 366 suites, 5,264 tests.

Two things measured on the way that are worth recording, since neither needed a change:

- **Every magnitude zero.** The `length` scale then has domain `[0, 0]` and range `[null, null]`, and Plot emits **no paths at all**. Nothing to read and nothing to mis-read; `points.length === 0` returns `null` as it already did.
- **A null magnitude.** Plot drops the datum itself — three points in, two paths out — so the adapter never sees it.

On the nits: agreed on both, and I've left both alone. The `axes.z` one in particular is a real gap, but it is adapter-wide (`convertDot` has it too), so fixing it here would fix one caller of a shared shortcoming and leave the rest — worth its own issue rather than a rider on this branch.

---

Issue #1094, opened 2026-08-19
Observable Plot: `link` and `arrow` marks — the range half of a dumbbell chart — are not read


jooyoungseo:
Found while sweeping the marks the Observable Plot adapter does not read.

## What happens

`Plot.link` and `Plot.arrow` draw groups labelled `link` and `arrow`. `convertMark` has no case for either, so both fall to `default: return null`.

A `link` given `x1`/`x2` on a shared `y` is a **range** or **dumbbell** chart — the shape #792 added a trace for. Drawn the usual way, with `Plot.dot` at each end, the dots *are* read today, so the chart announces two scatter layers and nothing that says the two ends belong together.

## Measured

`@observablehq/plot@0.6.17`, `link(rows, {y: 'g', x1: 'lo', x2: 'hi'})` with a dot at each end, over `A 3→9`, `B 5→14`, `C 2→6`:

```
g[aria-label="link"]   M88.333,32L378.333,32     M185,55L620,55     M40,78L233.333,78
g[aria-label="dot"]    88.333,32                 185,55             40,78
g[aria-label="dot"]    378.333,32                620,55             233.333,78
```

Two vertices per path, both ends on the same `y`, and the y scale is a `point` scale over the categories. Both ends invert exactly.

`Plot.arrow` draws the same segment plus its head **in the same `d`**:

```
M40,33L356.364,33M349.435,29L356.364,33L349.435,37
```

That is two subpaths, which #1079's guard already refuses on a line — so an arrow could not be read as a two-point line by accident even if the label matched.

## Scope

- The bound datum on a `link` path is a **number** (the datum index), not the array a line carries — so the vertex-count check that guards `parsePathVertices` has nothing to compare against here, and a reading would rest on the mark's own shape instead.
- A `link` is not always a range bar. Given `x1`,`y1`,`x2`,`y2` it connects two arbitrary points, which is an edge in a node-link diagram rather than a category's span. Reading those as ranges would announce a category axis that isn't there. What separates them is whether the two ends share a coordinate on one axis — measured above for the range form; **not** measured for the diagonal form, or for a `link` with a `curve`, which draws more than two vertices.
- An `arrow` additionally has a direction, which a range has no field for. Its head is in the same path, so any reading has to drop the head's subpath rather than treat its vertices as data.

jooyoungseo:
Measured the two forms the issue left open — the diagonal one and the curved one — plus a case neither of us thought of. `@observablehq/plot@0.6.17`.

```
[...]

PR #1101, opened 2026-08-20
feat(observable): read a rule mark as a gantt of the intervals it draws

### LINKED ISSUE #1100: Observable Plot: a `rule` mark carrying an interval is unread, and it takes the whole chart with it
`Plot.ruleX` and `Plot.ruleY` are how Plot draws a high–low chart, a gantt, a range plot, a candlestick's wick. `convertMark` has no `case 'rule'`, so the mark converts to `null`. On a chart whose only mark is a rule, that is not a dropped layer — it is the whole chart:

```
observablePlotToMaidr(element)  →  null
```

Measured in jsdom against the adapter's own fixture harness, on `Plot.ruleY(rows, { y: 'day', x1: 'lo', x2: 'hi' })`. The reader gets a static image of a chart every one of whose numbers is sitting in an attribute.

## What a rule actually writes

`@observablehq/plot@0.6.17`. `rows = [{day: 'Mon', lo: 2, hi: 8}, {day: 'Tue', lo: 3, hi: 11}, {day: 'Wed', lo: 1, hi: 6}]`.

**A gantt** — `Plot.ruleY(rows, { y: 'day', x1: 'lo', x2: 'hi' })`, x linear `[1, 11] → [40, 620]`, y point `[Mon, Tue, Wed] → [20, 90]` step 23:

```html
<g aria-label="rule" stroke="currentColor" transform="translate(0,0.5)">
  <line x1="98"  x2="446" y1="32" y2="32"></line>
  <line x1="156" x2="620" y1="55" y2="55"></line>
  <line x1="40"  x2="330" y1="78" y2="78"></line>
</g>
```

`98 → 2`, `446 → 8`, and so on down. Exact: no path to tokenise, no rounding to undo, no geometry to infer — four attributes, four numbers.

**A high–low chart** — `Plot.ruleX(rows, { x: 'day', y1: 'lo', y2: 'hi' })`, x point `[Mon, Tue, Wed] → [40, 620]`, y linear `[1, 11] → [370, 20]`:

```html
<line x1="137" x2="137" y1="335" y2="125.00000000000001"></line>
<line x1="330" x2="330" y1="300" y2="20"></line>
<line x1="523" x2="523" y1="370" y2="195"></line>
```

`335 → 2`, `125 → 8`. Same story on the other axis.

This is the shape `convertLink` already reads (#1094): a segment flat on one axis, spanning the other. A rule is that chart drawn as a `<line>` rather than a `<path>`, so it is a reading the adapter has already made once and can make again more cheaply.

## What must go on being refused, and how the drawing says so

Three other things wear the same `aria-label="rule"`, and each is turned away by something measured rather than guessed.

**A reference line.** `Plot.ruleY([5])` beside a dot mark:

```html
<line x1="40" x2="620" y1="370" y2="370"></line>
```

One line, and `40`/`620` are exactly the x scale's range endpoints — it spans the frame because Plot handed it the frame, not because a datum said so. Read as an interval it would announce a span from 1 to 3 that no row of the data contains, and it would do so as a second layer beside a scatter that is already correct. That is the #1084 failure: announcing something the chart does not contain.

**A bare positional rule.** `Plot.ruleX(rows, { x: 't' })` — vertical lines at each position, no value:

```html
<line x1="20" x2="20" y1="0" y2="30"></line>
```

This one refuses itself. The plot has **no y scale at all** — `svg.scale('y')` throws — so inverting the span yields nothing and the mark drops out, exactly the way `Plot.vector` with no `length` channel is turned away in #1098. No new guard needed; worth a test that says so.

**A lollipop stem.** `Plot.ruleX(rows, { x: 't', y: 'hi' })`, usually paired with a `dot` at the tip:

```html
<line x1="40"  x2="40"  y1="370" y2="115.45454545454545"></line>
<line x1="330" x2="330" y1="370" y2="20"></line>
<line x1="620" x2="620" y1="370" y2="179.0909090909091"></line>
```

Every stem starts at `y1 = 370`, the baseline — value 0 on a domain of `[0, 11]`. Read as intervals these announce "0 to 8" where the chart means "8", and the `dot` mark beside them is already read as a scatter carrying that 8. Two layers, one number, two names for it.

The drawing does distinguish this: every line shares an endpoint, so that end is the frame's baseline rather than anything a row said. That is the same whole-mark question `floorIsLevel` asks of an area in #1095, and the same "ask it of the mark, not the path" rule #1094 settled.

## Proposed reading

Read a `rule` mark as a set of spans when, across the whole mark, **neither end is shared by every line** and the span axis has a scale to invert through. Emit what `convertLink` emits, reusing `readSpan`.

Refuse otherwise:

- every line sharing one end → a baseline, so the span is a stem and not a measurement;
- every line sharing both ends, or a single line at the frame's extent → a reference line;
- no scale on the span axis → nothing to invert, and the point drops on its own.

## Not proposed

- **A genuine interval chart whose rows all start at the same value.** A gantt where every task begins at day 0 is drawn exactly like a stem, and the markup cannot separate them. Refusing costs that chart its reading; reading it would mis-announce every lollipop, which is far commoner and, unlike the gantt, is *not* silent today — its `dot` already carries the value. The trade goes the other way from #1095's constant floor for that reason, and it should be written into `docs/observable.md` rather than left implicit.
- **`Plot.ruleY([5])` as an annotation layer.** Announcing "there is a reference line at 5" is a real thing a reader would want, but there is no field in the grammar for it, and inventing one is a separate change from reading a mark that already has a shape.

jooyoungseo:
Closes #1100.

`Plot.ruleX` and `Plot.ruleY` are how Plot draws a high–low chart, a range plot and a gantt. `convertMark` had no case for them, so a chart whose only mark was a rule converted to `null` — not a dropped layer but the whole chart, every number of which was sitting in an attribute.

## The reading

A `<line>` carries both of its ends as attributes, so this is the cheapest exact reading in the adapter — no `d` to tokenise and nothing rounded to undo. Measured on `@observablehq/plot@0.6.17`, `Plot.ruleY(rows, {y: 'day', x1: 'lo', x2: 'hi'})` over Mon 2–8, Tue 3–11, Wed 1–6:

```html
<g aria-label="rule" stroke="currentColor" transform="translate(0,0.5)">
  <line x1="98"  x2="446" y1="32" y2="32"></line>
  <line x1="156" x2="620" y1="55" y2="55"></line>
  <line x1="40"  x2="330" y1="78" y2="78"></line>
</g>
```

`98 → 2`, `446 → 8` on a scale of `[1, 11] → [40, 620]`, and so on down. It is the same span a `link` draws, so it is the same gantt — `convertLink`'s body is now shared between the two rather than written twice.

`pixelError` is `0`, which is what routes the value through `cleanNumber` rather than through a geometric quantum: an attribute has not been through a serialiser, so there is no rounding to undo, only float noise to clean. `y2="125.00000000000001"` comes back as `8`.

## What is refused, and what says so

Three other things wear this label, and each gives itself away the same way — every line ends where every other one does:

| what | drawn | why it is not an interval |
|---|---|---|
| `Plot.ruleY([5])` | `x1="40" x2="620" y1="370" y2="370"` | 40 and 620 are the x range's endpoints — Plot handed it the frame |
| `Plot.ruleX(rows, {x})` | `y1="20" y2="370"` on every line | the frame's full height, three times over |
| `Plot.ruleX(rows, {x, y})` | `y1="370"` on every line | the baseline, 0 on a domain of `[0, 11]` — these are a lollipop's stems |

Asked of the whole mark, as it is for a link (#1094): if every line agrees on an end, that end is the frame or the baseline rather than anything a row of the data said. **Both** ends are asked, because which one a constant lands in is the caller's spelling — `y1: 'v', y2: 0` and `y1: 0, y2: 'v'` are the same lollipop, and `sharedEndRules` is the fixture for the second.

A reference line is also refused a second way, and this is the one that survives when there are two of them at different heights: it lies *across* the lanes of the chart it annotates rather than along one, so `lineSpan` finds its ends disagreeing on the lane axis and gives it back.

## One existing expectation moves

`test/adapters/observable/box.test.ts` asserted that `floatingRangeBar` — a candlestick with a marker in its body, declined as a box plot by #1088's stroke-width gate — reads as `[DOT]`. It now reads as `[GANTT, DOT]`, because its wicks are a rule mark carrying a high and a low.

That follows the design the test itself states: *"declining to claim them is not the same as skipping them: their marks go back to the readings they belong to."* `[DOT]` was the state of the world while a rule had no reading; now one of the chart's marks has the reading it belongs to. The wicks' numbers are the period's low and high, which is true and was previously absent.

The other two fixtures in that case — `bulletChart` and `barRangeAndTarget` — are unchanged, and not by accident: their rules stand on the baseline, so the new reading declines them for its own reason. A real `Plot.boxY` is claimed as a composite and never reaches any of this.

## Verification

`eslint 0 / type-check 0 / jest 0` — 367 suites, 5,273 tests.

Every guard falsified by applying each mutation alone, across `rule.test.ts`, `box.test.ts` and `link.test.ts` (39 tests):

| mutation | result |
|---|---|
| `case 'rule'` removed | 4 failed |
| agreement guard removed entirely | 6 failed |
| only the `from` half of it | 1 failed (`sharedEndRules`) |
| only the `to` half of it | 4 failed |
| `lineSpan` drops its lane-agreement check | 1 failed (`crossingRules`) |
| spans read on the wrong axis | 4 failed |

Two mutations did **not** redden, and neither is a guard I am adding:

- **`pixelError: 0 → 1`.** `cleanToGeometry` short-circuits to `cleanNumber` when `pixelError <= 0`, which is what "this came from an attribute" means; at a span of 10 over 350 pixels a one-pixel error happens to round to the same digits. Reddening it would need a chart contrived to sit on the rounding boundary, which is testing the test.
- **the `drawn` predicate widened to `() => true`.** Plot's rule and link groups hold only lines and only paths respectively, so no fixture separates `isLine`/`isPath` from "every child". The *check* it feeds — every element or none — is live, via `mixedLinks`. The predicate was already there on the link side; I kept it rather than widen a contract on the strength of what Plot happens to emit today.

Two false passes were found and fixed during the sweep rather than shipped: the first `referenceRule` and `stemRules` fixtures were on two continuous axes, so `barOrientation` refused them before the new guard ran at all. Both were re-captured on a categorical axis, where the guard is what does the work.

## Not doing

- **A gantt whose rows genuinely all begin — or all end — together.** Drawn exactly as a lollipop is, and the markup cannot separate them. The trade goes the other way from the constant floor in #1095 because there refusing cost a chart its only reading, while here what is refused is either not data at all or already announced by the mark beside it. Written into `docs/observable.md` rather than left implicit.
- **Announcing `Plot.ruleY([5])` as an annotation.** A reader would want to know a reference line is there, but there is no field in the grammar for one, and inventing it is a separate change.
- **`src/adapters/observable/index.ts`'s "marks it reads" list** is stale — it predates `tick`, `linear-regression`, `waffle`, `link` and `vector` as well as `rule`. Left alone here rather than half-corrected; `docs/observable.md` is the maintained reference and does have the new section.

---


jooyoungseo:
Good catch on the stray comment — it was exactly what you suspected, and it was mine. I inserted `convertRule` immediately above `interface Span`, and the anchor I matched on was the `interface Span {` line rather than the doc comment sitting on top of it, so the comment stayed put and the new function slid in underneath it. `Span` was left undocumented and the line was left dangling over `convertRule`'s own JSDoc.

Put back in 8f0fc4f7, along with one thing that became inaccurate through this change rather than through the move: `Span.pixelError` was documented as "the quantum the path's coordinates were written at", which is no longer the whole story now that a span can come off a `<line>`. It now says what a `0` means and why a line has one.

On the other two observations, both correct and neither changed:

- **`barOrientation` computed twice.** It is — once in `convertRule` to read the spans it needs to ask the agreement question of, once in `convertSpans`. Threading it through would couple `convertRule` to the shared body's internals to save a comparison of two scale types, so I left it. (Small correction to the reading: there is no early return on it in `convertRule` — a null orientation yields `alongX === false`, the spans then fail to invert, and `convertSpans` refuses the mark. `bareRules` is that path.)
- **`from`/`to` are in draw order, not sorted.** Right, and deliberately the same as `readSpan`'s. `GanttPoint` does not claim an ordering, and a rule *can* be drawn either way round — `y1: 'hi', y2: 0` is the `sharedEndRules` fixture and reaches `from > to`. Sorting here would quietly assert something the mark does not.

`eslint 0 / type-check 0 / jest 0` after the fix — 367 suites, 5,273 tests.

---

PR #1110, opened 2026-08-21
feat(adapters): read a Plot text mark as the labelled scatter it draws

### LINKED ISSUE #1106: Coverage sweep: `Plot.text` is a labelled scatter the adapter drops, and a tree announces its layout coordinates without its node names
Ran eighteen Observable Plot marks through a browser and looked at what each one leaves in the DOM, now that the adapter reads twelve labels plus the box composite (#1105). Plot 0.6.17, captured with Playwright.

### `Plot.text` → a scatter carrying each point's name

```
<g aria-label="text" transform="translate(0.5,0.5)">
  <text y="0.32em" transform="translate(40,253.33)">alpha</text>
  <text y="0.32em" transform="translate(330,20)">beta</text>
  <text y="0.32em" transform="translate(620,370)">gamma</text>
</g>
```

Both halves are exact. The **position** is the mark's `transform`, the same place `dot`, `tick` and `vector` take theirs from. The **name** is the element's text content, verbatim — nothing inverted, nothing rounded, no colour involved.

This is not a decorative mark. `Plot.text` is how a labelled scatter is drawn — country names on a GDP-against-life-expectancy chart is Plot's own canonical example — and today the whole layer is handed back.

**What blocks it is one field.** `ScatterPoint` carries `x`, `y`, `z`, `xLabel` and `yLabel`, and the last two name *the category a coordinate is a position for* (#927), which is a different statement from "this point is Norway". The field that says the latter exists already, on `VolcanoPoint`, and its own documentation makes the case better than I can:

> **What the point *is* — a gene, a SNP, a probe.** Identity is the payload on these charts, not the coordinates. A reader told "x is 2.3, y is 14.1" has been given the two numbers they can see the shape of already and withheld the one thing they came for.

Reading `text` as a plain scatter would do exactly that. Promoting `label` from `VolcanoPoint` to `ScatterPoint` is the change that makes the mark readable, and it costs `VolcanoPoint` nothing — it inherits the field it already declares.

### A tree announces its layout coordinates, and drops every node name

`Plot.tree(['a', 'a/b', 'a/c', 'a/b/d', 'a/b/e'])` draws three groups: `link`, `dot`, and two `text`. The dots are read — they are an ordinary `dot` mark — so the chart is bound and navigable. What a reader gets is this:

```
<circle cx="0"     cy="266" r="3"><title>/a</title></circle>
<circle cx="319.5" cy="133" r="3"><title>/a/b</title></circle>
<circle cx="639"   cy="0"   r="3"><title>/a/b/d</title></circle>
```

The announced `x` and `y` are **d3 tree layout coordinates** — depth and sibling offset, on scales of domain `[0, 2]` and `[-0.5, 1]` that the chart draws with `axis: null` precisely because they mean nothing to a reader. The node's path is sitting in the `<title>` of the very element being announced, and in a `text` mark beside it, and both are dropped.

So the reading is not incomplete, it is misleading: a hierarchy is announced as a five-point scatter of coordinates nobody plotted. Worth deciding separately from `text` — the honest options are a real hierarchy reading (MAIDR has network and treemap traces), or declining a `dot` mark whose axes are suppressed.

## Correctly unread, and worth recording as such

| Mark | What it leaves | Why it stays unread |
|---|---|---|
| `cell` | `<rect>` per cell | magnitude is an 8-bit fill; already documented |
| `raster` | one `<image>` | magnitude is pixels of that image |
| `contour` | `<path>` per band | drawn as filled bands, and the level is in the fill — the same wall #1084 hit |
| `density` | `<path>` | likewise |
| `geo` | `<path>` per feature | value in the fill; the centroid a `ChoroplethPoint` wants is not in the DOM |
| `image` | `<image>` per row | positions are readable, but the payload is a picture |
| `hull`, `voronoi`, `delaunay link` | `<path>` | geometry *over* points rather than data of their own; the points are their own mark and read already |
| `crosshair`, `frame` | decoration | nothing to announce |

## Already handled correctly, verified rather than assumed

- **`bollingerY`** draws an `area` and a `line`. The band's floor follows the data, so the area is handed back as an interval without a centre (the rule #1047 records) and the moving average reads as the line it is. Correct as it stands.
- **A tree's `link` mark is not claimed as a gantt.** Its four paths run `M0,266C…,319.5,133` — both ends differ on both axes — so #1094's whole-mark span test refuses them. Worth stating because a node-link diagram is precisely the case that test exists for, and this is the first one in the wild.

## Not read, and needs a grammar decision rather than an adapter one

**`differenceY`** draws three groups: `positive difference`, `negative difference`, and `line`. The line is read, so today the chart announces **one of the two series and nothing about the comparison** — which is the entire chart. The shape it wants is two series and the signed gap between them, which `DumbbellData` says for a category and `SmoothPoint` says for a band, and neither is this. Flagging rather than proposing.

## Suggested order

`Plot.text` first: it is a whole chart type, the data needs no interpretation, and the only new thing it asks of the grammar is a field that already exists one interface down. Then the tree question, which is a defect rather than a gap. `differenceY` last, since it needs a shape decided before an adapter can emit one.

jooyoungseo:
Closes #1106.

## What it does

`Plot.text` is how a labelled scatter is drawn — country names against GDP is Observable's own canonical example — and the whole layer was handed back. It is also how another mark's labels are drawn, which is the same markup put to a different use, and `Plot.tree` is built from it.

Measured in Chromium against the built bundle, Plot 0.6.17:

| chart | before | after |
|---|---|---|
| `Plot.text(data, {x, y, text: 'name'})` | no layer at all | 1 scatter — `{x: 3.1, y: 14.2, label: 'alpha'}`, … |
| `Plot.dot` + `Plot.text` over the same data | 1 unnamed scatter + 1 nameless duplicate | **1** named scatter |
| `Plot.tree(['a', 'a/b', 'a/c', 'a/b/d', 'a/b/e'])` | dots announcing layout coordinates, no names | **1** layer, nodes named `/a`, `/a/b`, `/a/c`, … |

## Which of the two uses a `text` mark is

Position, and it separates them exactly. Plot writes a label's position from the same channel it writes the labelled point's, so a label lands on what it labels. Measured with `dy: -8`, which is how you would offset a label so it does not sit on its dot:

```
GROUP "dot"    circle cx=40 cy=238.37606837606842
GROUP "text"   text   translate(40,238.37606837606842)  "alpha"
```

The offset is applied when the text is laid out, not to its `transform`. So a `text` mark with another mark's points beneath it is that mark's names — they move onto it and the group is not read again — and one with nothing beneath it is a series.

This is what keeps the tree to one layer. A tree is four marks, of which two are `text` sitting exactly on the `dot`:

```
GROUP "link"  n=4
GROUP "dot"   n=5   circle cx=330 cy=136.667  <title>/a/b</title>
GROUP "text"  n=3   text translate(330,370)   "c"  <title>/a/c</title>
GROUP "text"  n=2   text translate(40,253.333) "a" <title>/a</title>
```

The `<title>` is preferred over the glyph where a mark has both — a tree's glyph shows only the leaf, and reading `textContent` off that element would give `"c/a/c"`, since a `<title>` child's text is part of its parent's `textContent`.

The tree's `x` and `y` stay what the chart drew: d3 tree layout depth and sibling offset, on scales `Plot.tree` renders with `axis: null` because they mean nothing to a reader. Naming the nodes does not make the coordinates meaningful — it stops them being all a reader gets, which is what #1106 asks for.

Only on two continuous scales. A `text` on a categorical axis is a value written over a bar, and the bar mark already carries that value; reading it here would announce every count twice.

## The grammar change

`label` moves from `VolcanoPoint` to `ScatterPoint`, which is the change #1106 identifies as the one that makes the mark readable. The field's own documentation already made the general case — identity is the payload on some charts, not the coordinates — and that is as true of a country as of a gene. `VolcanoPoint` inherits it unchanged.

It stays distinct from `xLabel` / `yLabel`, which name *the category a coordinate is a position for*. "This slot on the x axis is called Norway" and "this point is Norway" are different statements, and a chart can make either: a strip plot makes the first, a labelled scatter the second.

`ScatterTrace` announces it as an aside — `main` carries a value on an axis, formatted as that axis formats, and a name is not one — and only where the cursor is on **exactly one point**. That is the rule `VolcanoTrace` already stated: naming one of a stack would name the wrong one. A stack of exactly one has no wrong one to pick, which is the ordinary case for a labelled scatter, where every point has its own x.

That last part changed an existing volcano test. `test('says nothing extra where the cursor is not on one point')` asserted `asides` was undefined in column mode — but every gene in its fixture has a distinct x, so its cursor *was* on exactly one point, and the test was pinning a name withheld from a cursor that identified it. It is now two tests: one for the column that holds one point, and one for a column holding two, which is the case the rule was always about.

## Tests

Three fixtures captured from real charts in Chromium and added to `test/adapters/observable/fixtures.ts`, plus `test/adapters/observable/text.test.ts`.

Falsification, each change reverted alone:

| mutation | result |
|---|---|
| drop the `'text'` dispatch case | standalone-scatter test **red** |
| never claim an overlay | both overlay tests **red** (`dot`+`text`, and the tree) |
| `markName` reads `textContent` instead of the `<title>` | tree test **red** — it gets `"b/a/b"` |
| `ScatterTrace` stops announcing the name | volcano naming test **red** |

The `markList` guard caught that `text` was dispatched but undocumented, which is why `src/adapters/observable/index.ts` and `docs/observable.md` both gained it — a table row, and a `## Text marks` section covering the two uses and the categorical-axis refusal.

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all clean.

---


jooyoungseo:
The facet finding was real, and measuring it turned up that the consequence is worse than "regresses to the pre-fix behaviour". Fixed in a540bde.

**Your diagnosis of the mechanism was exactly right.** `centresOf` read `group.children`, which for a faceted mark are the per-facet wrappers, so `dotCentre` fell back to each wrapper's own `transform` — the panel's offset.

**What that produces is not a nameless duplicate.** Measured in Chromium against the built bundle, a faceted `Plot.dot` + `Plot.text`:

```
subplot 0,0   1 layer   [{x:1,y:10}, {x:2,y:20}]     ← no labels
subplot 0,1   1 layer   [{x:3,y:30}, {x:4,y:40}]     ← no labels
```

Two panels, one layer each, and **no names at all**. The pairing did not fail — it *succeeded* on a comparison that meant nothing. Two marks faceted alike have identical panel offsets, so every "label" found a "point", the `text` group was claimed, and `markName` returned null for a wrapper whose text lives one level down. So the names were lost rather than duplicated, which is strictly worse than the reading this PR set out to replace, and it would have shipped.

The fix is the one you named: split both marks with `splitFacets` and pair within each panel, which is what every other reader in the file already does. After it, the same chart gives `[['a','b'], ['c','d']]` across the two subplots.

`facetedDotWithLabels` is a new fixture, captured from that chart, and `test('pairs the labels within each panel rather than across panels')` covers it. Falsified: replacing both `splitFacets` calls with `Array.from(group.children)` — the shape the old code had — turns exactly that one test red and nothing else, which is the "not currently caught by CI" you predicted.

One incidental note on the fixture: it needed `positions: [0, 305]` on its `fx` band scale, read off the facet wrappers' own transforms, or the harness places both panels in one cell. Without it the test would have been measuring the rebuilt scale rather than the adapter.

`npm run type-check`, `npm run lint:fix` and `npm test` clean; 193 tests in the Observable suite.

---

PR #1111, opened 2026-08-21
fix(focus): give a framed chart a way back to the page that embeds it

jooyoungseo:
# Pull Request

## Description

A chart on a Quarto `revealjs` slide swallows every key and gives the reader no way back to the deck. Tab reaches the chart; from inside it, Space does not advance the slide, the arrows do not go back, and Shift + Tab leaves the document altogether for the browser's own UI.

Keyboard events do not cross a frame boundary, and every binding embeds charts in an iframe (py-maidr through `srcdoc`, r-maidr through a `data:` URL). While the reader is inside the chart the page around it hears nothing — which is *right* while they are reading, since MAIDR binds the arrows, Space, Backspace, Page Up and Page Down itself. That leaves Shift + Tab as the way back.

Usually the browser handles that on its own: Shift + Tab at the frame's first tab stop moves to whatever precedes the frame in the host. It cannot when nothing reachable precedes it — and a Quarto deck is exactly that case, because it renders no controls of its own, so a chart is the first thing on the page.

## Related Issues

The reveal.js half of what was observed — content on non-current slides staying in the tab order — is upstream and already filed as [hakimel/reveal.js#1587](https://github.com/hakimel/reveal.js/issues/1587) (open since 2016). It reproduces in a stock Quarto `revealjs` deck containing nothing but links, with no MAIDR involved, and is not addressed here.

## Changes Made

`FrameFocusService` (`src/service/frameFocus.ts`), constructed and disposed by `src/controller.ts` alongside the other services, so its listener exists exactly while a chart holds focus.

- Watches for **Shift + Tab at the frame's first tab stop**, and steps in **only when the browser would strand the reader** — where the host has somewhere real to go, the native tab order is left exactly as it was. Tab in the forward direction is never touched.
- When it does step in, focus goes to the element holding the frame, which on a slide deck is the slide's section element: the reader hears which slide they landed on, and the deck's own shortcuts work again because the keydown now reaches the host document.
- Reachability is **checked, not assumed**. reveal.js leaves the slides on either side of the current one rendered, so a chart on the previous slide is a tab stop in document order even though it is marked hidden — sending the reader there would land them on an off-screen chart instead of the slide.
- A same-origin host is served directly through `frameElement`, in the same event. A host on another origin — r-maidr's `data:` frames have an opaque origin — is asked by `postMessage` instead (`maidr:frame-focus-escape`), which `xability/r-maidr` answers on the parent side; a handoff already in flight is not asked again, so a held Shift + Tab cannot push its deadline out.
- **Nothing can trap a reader.** If the handoff cannot be completed, the chart blurs so the next Shift + Tab does what the browser would have done. Outside a frame the service adds no listener at all.

Nothing was needed in `xability/py-maidr`: its `srcdoc` frames are same-origin, so the engine reaches the host document directly.

Not visual. A Chromium (Playwright) trace of the reader's path through a real three-slide Quarto deck, charts drawn by matplotlib through py-maidr:

```
Tab        → chart          role=application, "This is a maidr plot of…"
→ →        → chart          "X is a, Y is 3" / "X is b, Y is 5"
Shift+Tab  → section#slide-1-bar        ← back on the page
Space      → slide-2                    ← the deck responds again
Tab        → chart on slide 2
Shift+Tab  → section#slide-2-line
```

Before this change every step after the first left the reader stuck in the chart, and Shift + Tab put focus in the browser UI.

## Checklist

## Additional Notes

**Verification.** `npm run lint:fix`, `npm run type-check` and `npm run build` all clean; `npm test` green at 5286 tests across 369 suites, including eight new cases in `test/service/frameFocus.test.ts`. Both halves of the decision are covered — the fixing case and the leave-it-alone cases, since getting the latter wrong would rewrite the tab order of every notebook and article that embeds a chart. Checked in a browser against Quarto 1.10.18 decks built from both R (ggplot2 via r-maidr) and Python (matplotlib via py-maidr), on every slide, plus a plain Quarto HTML document to confirm its tab order is untouched.

**Reaching users.** Both bindings load maidr 4.4.0 today, from a CDN or a vendored copy, so this arrives with the next engine release.

**Related.** [`xability/r-maidr#208`](https://github.com/xability/r-maidr/pull/208) carries the parent-side half for its cross-origin frames.

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -14,6 +14,7 @@ import { CommandPaletteService } from '@service/commandPalette';
 import { DescriptionService } from '@service/description';
 import { DisplayService } from '@service/display';
 import { FormatterService } from '@service/formatter';
+import { FrameFocusService } from '@service/frameFocus';
 import { GoToExtremaService } from '@service/goToExtrema';
 import { HelpService } from '@service/help';
 import { HighContrastService } from '@service/highContrast';
@@ -55,6 +56,7 @@ export class Controller implements Disposable {
   private readonly notificationService: NotificationService;
   private readonly settingsService: SettingsService;
   private readonly formatterService: FormatterService;
+  private readonly frameFocusService: FrameFocusService;
 
   private readonly audioService: AudioService;
   private readonly brailleService: BrailleService;
@@ -104,6 +106,7 @@ export class Controller implements Disposable {
 
     this.notificationService = new NotificationService();
     this.formatterService = new FormatterService(maidr);
+    this.frameFocusService = new FrameFocusService();
     this.textService = new TextService(this.notificationService, this.formatterService);
     this.displayService = new DisplayService(this.context, plot, this.textService);
     this.settingsService = new SettingsService(
@@ -502,6 +505,7 @@ export class Controller implements Disposable {
     this.brailleService.dispose();
     this.audioService.dispose();
     this.formatterService.dispose();
+    this.frameFocusService.dispose();
 
     this.settingsService.dispose();
     this.notificationService.dispose();
--- a/src/service/frameFocus.ts
+++ b/src/service/frameFocus.ts
@@ -209,21 +209,30 @@ export class FrameFocusService implements Disposable {
   }
 
   /**
-   * Asks the host to take focus, and takes it for the host if it does not.
+   * Moves focus into the host, by whichever channel that host allows.
    */
   private requestHostFocus(): void {
-    // A cooperating host knows its own page and can pick a better landing
-    // place than this frame can, so it gets the first say. It is also the only
-    // channel to a host on another origin, where the fallback cannot reach.
+    if (FrameFocusService.hostFrameElement() !== null) {
+      // The host's DOM is reachable, which is both faster and surer than
+      // asking: focus moves in this same event, with nothing to wait for.
+      this.focusHostContainer();
+      return;
+    }
+
+    // A host on another origin can only be asked. Once a handoff is in flight
+    // there is nothing to add by asking again — which is what a held Shift +
+    // Tab would otherwise do on every auto-repeat, pushing the deadline out
+    // for as long as the key is down.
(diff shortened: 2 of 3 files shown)


jooyoungseo:
Thanks — acted on two of the three, pushed in `2b3e832`.

**The 50 ms tax.** Worth more than "likely imperceptible": the same-origin case *is* the reported scenario, since py-maidr's `srcdoc` frames are what a Quarto `revealjs` deck embeds. Reaching that host's DOM is both faster and surer than asking it, so `requestHostFocus()` now returns straight after `focusHostContainer()` when `frameElement` is non-null — focus lands in the same event, no message, no deadline. The message stays exactly what it always was, the only channel to a host on another origin, and now goes out only where it is the only channel.

**Auto-repeat.** Fixed rather than tolerated, because it costs one line: while a handoff is in flight the service now returns instead of re-arming, so the first ask stands until it resolves. Only the cross-origin path can reach that code at all now, and a held key no longer pushes the deadline out for as long as it is down.

**The layering deviation.** Fair point, and the note is now in the `focusHostContainer()` doc comment. The reason it cannot follow the rule is specific rather than general: the rule's remedy is to fire an event and let a ViewModel render the result, but the element focus has to land on belongs to *another document* — no ViewModel of ours renders it and no React tree of ours contains it. Writing to it is the whole operation, so there is nothing left to delegate.

Left alone: the `indexOf(frame) < 0` fallback, for the reason you gave — favouring the native tab order over intervening is the direction this should fail in.

**Re-verified after the change**, since it moves behaviour on the primary path: `type-check` and `build` clean, `npm test` green at 5286 (one new case pinning the ask-once behaviour, and the same-origin case now asserts focus moves with no message posted and no timer armed). In Chromium against real Quarto 1.10.18 decks, on every slide of both — Python/matplotlib for the same-origin path, R/ggplot2 for the cross-origin one — Shift + Tab still lands on the slide and Space still advances the deck.

---

PR #1121, opened 2026-08-21
fix(focus): check that the host actually took focus before calling it done

jooyoungseo:
# Pull Request

## Description

`FrameFocusService` (#1111) assumed that calling `focus()` on a host element moved focus there. It does not always: an element with no rendered box refuses silently, and `display: contents` is the case that bites.

Shiny wraps **every** output in a `display: contents` div, and a chart's frame sits directly inside one with no sectioning element above it. So the handoff picked that wrapper, called `focus()`, got nothing, and returned as though it had succeeded — leaving the reader inside the chart with nothing to show for the keypress. That is the very failure the service exists to prevent, reproduced in a different host.

It was also slightly worse than doing nothing there: the old native behaviour at least moved focus out of the page, whereas the intercepted keypress moved it nowhere at all.

## Related Issues

Follow-up to #1111, which introduced the handoff. The Shiny case was not among the hosts checked before that merged — this is the gap.

## Changes Made

- `focusHostContainer()` now reads the outcome back (`ownerDocument.activeElement === element`) instead of trusting the call, and walks up the ancestor chain until one actually takes focus. The sectioning element is still tried first; the whole chain follows it rather than only `parentElement`.
- A `tabindex` the handoff added is **removed again** when the element refuses, so a wrapper that cannot hold focus is not left carrying an attribute claiming it can. That also shrinks the change's footprint on host pages generally.
- `isFocusable()` folded into the new `takeFocus()`, which is where the duck-typing note now lives; `hostCandidates()` builds the list.

Not visual. A real py-shiny app driven in Chromium, sampling the host document after one Shift + Tab out of the chart:

```
before                                    after
t=  50ms  top=iframe#…  frameHasFocus=1   t=  50ms  top=div.container-fluid  frameHasFocus=0
t= 300ms  top=iframe#…  frameHasFocus=1   t= 300ms  top=div.container-fluid  frameHasFocus=0
t=2000ms  top=iframe#…  frameHasFocus=1   t=2000ms  top=div.container-fluid  frameHasFocus=0

tabindex written on: div.shiny-html-output    tabindex written on: div.container-fluid
                     (display:contents,                            (the one that took it)
                      never took focus)
```

Walking the ancestor chain in that page, measured rather than assumed:

```
div#first_chart.shiny-html-output   display=contents   takesFocus=false
div.container-fluid                 display=block      takesFocus=true
body                                display=block      takesFocus=true
```

## Checklist

## Additional Notes

**The regression test earns its place.** jsdom has no layout, so it happily focuses an element a browser refuses — a first attempt at this test passed against the unfixed code and pinned nothing. The committed case reproduces Shiny's actual shape (a `display: contents` wrapper with no sectioning ancestor) and stubs `focus()` to the no-op a browser performs; it fails against `main` and passes here, which was checked both ways.

**Verification.** `npm run lint:fix`, `npm run type-check` and `npm run build` clean; `npm test` green at 5389 tests across 384 suites. Then driven in Chromium against four hosts, since the point of this fix is that one host had been missed:

| host | before | after |
| --- | --- | --- |
| py-shiny, chart first | stuck in the chart | focus on `div.container-fluid` |
| Quarto `revealjs` (R and Python) | lands on the slide, Space advances | unchanged |
| JupyterLab, live notebook | native tab order kept, nothing written | unchanged |
| nbconvert HTML export | native tab order kept, nothing written | unchanged |
| bare page, chart first | focus on `body` | unchanged |

Both notebook hosts give every cell wrapper a `tabindex`, so something focusable always precedes the frame and the service correctly leaves them alone — confirmed by checking that no ancestor of the chart's frame gains a `tabindex` from us (the two that carry one in JupyterLab already carry it before any chart is touched).

**Related.** [`xability/r-maidr#213`](https://github.com/xability/r-maidr/pull/213) carries the same fix for the parent-side copies of the listener.

### DIFF
--- a/src/service/frameFocus.ts
+++ b/src/service/frameFocus.ts
@@ -259,37 +259,80 @@ export class FrameFocusService implements Disposable {
    */
   private focusHostContainer(): void {
     const frame = FrameFocusService.hostFrameElement();
-    const container = frame === null
-      ? null
-      : frame.closest(HOST_CONTAINER_SELECTOR) ?? frame.parentElement ?? frame.ownerDocument.body;
-
-    if (container === null || !FrameFocusService.isFocusable(container)) {
-      // Nothing more this side can do. Blurring means the next Shift + Tab
-      // behaves as it would have without this service, so the reader is
-      // left no worse off than before — never trapped.
-      (document.activeElement as HTMLElement | null)?.blur();
-      return;
+    if (frame !== null) {
+      for (const candidate of FrameFocusService.hostCandidates(frame)) {
+        if (FrameFocusService.takeFocus(candidate)) {
+          return;
+        }
+      }
     }
 
-    // `tabindex="-1"` makes the container focusable without adding a tab stop
-    // to the host's own order.
-    if (!container.hasAttribute('tabindex')) {
-      container.setAttribute('tabindex', '-1');
-    }
-    container.focus();
+    // Nothing in the host would take it. Blurring means the next Shift + Tab
+    // behaves as it would have without this service, so the reader is left no
+    // worse off than before — never trapped.
+    (document.activeElement as HTMLElement | null)?.blur();
   }
 
   /**
-   * Whether a host element can be told to take focus.
+   * The host elements worth handing focus to, nearest meaningful first.
    *
-   * Duck-typed on purpose: everything reached through `frameElement` lives in
-   * the host's realm, where `instanceof HTMLElement` — which resolves to *this*
-   * frame's constructor — is false for every element the host owns.
-   * @param {Element} element - An element belonging to the host document
-   * @returns {boolean} True when the element exposes `focus()`
+   * The whole ancestor chain follows the sectioning element, because the
+   * nearest wrapper is not always able to hold focus: Shiny wraps every output
+   * in a `display: contents` div, which has no box for focus to land in, and
+   * the chart's frame sits directly inside one.
+   * @param {Element} frame - This document's frame, in the host's DOM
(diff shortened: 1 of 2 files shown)


jooyoungseo:
On the `tabindex` note — I tried to close it, then backed the change out, and the reason is worth recording because it says something about the code rather than about the change.

The idea was to ask a candidate as it stands first and only add `tabindex="-1"` if it refused, so a natively tabbable ancestor would never be taken out of the host's tab order. Writing the test for it is what showed the case cannot be reached: an ancestor precedes its own descendant in document order, so a natively tabbable ancestor **is** a tab stop before the frame, `hasTabStopBefore()` returns true, and `handleKeyDown()` returns before `takeFocus()` is ever called. The test asserting the fix fired came back with `defaultPrevented === false` — the service had correctly left the tab order alone.

A second attempt aimed at the reachable half — an ancestor the host already gave `tabindex="-1"`, which is not a tab stop and so does not stop the service — passed against both the changed and the unchanged code, because the existing `!element.hasAttribute('tabindex')` guard already declines to write over what the host set.

So the change was defensive against something I could not construct, and unpinnable by a test. `rules/testing.md` asks that new behaviour ship with a test that would have failed before it, and this had none to offer, so it is out. Reverted; the diff here is what you reviewed.

Worth keeping in mind if `TABBABLE_SELECTOR` ever grows to cover an element that is natively focusable but currently missing from it — `<summary>` is the one I can name — since that is the gap through which the case would become reachable.

---

PR #1123, opened 2026-08-21
feat(vegalite): read a rule spanning two bounds as the range chart it draws

### LINKED ISSUE #1122: A Vega-Lite range chart reads or does not depending only on whether it was drawn with `bar` or `rule`
Found by sweeping the Vega-Lite mark vocabulary against `resolveTraceType` in `src/adapters/vegalite/converters.ts`, the way the Observable Plot marks were swept for #1104.

## Measured

Two specs, same data, same channels, differing only in the mark name. Run through `vegaLiteToMaidr` on this branch:

```js
// mark: 'bar'
{ data: SCHEDULE, mark: 'bar', encoding: {
    y: { field: 'task', type: 'ordinal' },
    x: { field: 'start', type: 'quantitative' },
    x2: { field: 'end' } } }
// -> [ 'gantt' ]

// mark: 'rule'  — everything else identical
// -> []
```

`[]` is a figure with no layers: the chart is unread. The interval each row draws is in the spec, on channels the adapter already reads, and it is dropped because of the mark's name.

`rule` **is** how Vega-Lite's own gallery draws a range chart. `resolveTraceType`'s `switch` has cases for `bar`, `line`, `trail`, `area`, `point`, `circle`, `square`, `tick`, `rect`, `boxplot`, `arc`, `errorbar`, `errorband` and `geoshape`; there is no `rule`, so it falls to `default: return null`. The adapter already says so out loud in a comment, at `convertPairedLayers`:

> Read layer by layer, the segment is dropped — **a bare `rule` resolves to no trace type**

## What must keep resolving to nothing

Three other charts wear the `rule` mark, and the spec separates all three cleanly — unlike the Observable side (#1100), where the same question had to be asked of the drawn geometry:

| spec | today | must stay |
| --- | --- | --- |
| `rule` + `y` only — a reference line | `[]` | `[]` — one bound is not an interval |
| `rule` + `x` ordinal + `y` — lollipop stems | `[]` | `[]` — the stem runs from the baseline; the dots beside it carry the value, and `convertPairedLayers` already collapses that pair |
| `rule` + `x`/`x2` under a dot layer — a ranged dot plot | `dumbbell` | `dumbbell` — `convertPairedLayers` consumes both layers before per-layer conversion |
| `rule` + `x`/`x2` + ordinal `y` — a range chart | **`[]`** | **`gantt`** |

Each of those was measured, not reasoned about.

## Proposed

Route `rule` through `resolveRangedBarType`, the helper `bar` already uses: two positional bounds on one axis, with the opposite axis categorical, is an interval per row whichever mark drew it. Nothing else needs to move — `extractGanttData` reads the spec rather than the geometry, and the selector machinery already knows this mark (`markToCssClass` gives `mark-rule`, and `markToChildElement` already returns `line` for it, since the `errorbar` whip is a rule).

One wrinkle: that helper warns, when both bounds are present but the opposite axis' type was left to Vega-Lite's inference, that it is "reading it as an ordinary bar". For a `rule` that sentence is wrong — an unresolved rule reads as nothing at all — so the message needs to name what actually happens to the mark it is talking about.

## Also turned up by the same sweep, not proposed here

- **`text`** — no case. The Observable adapter reads a standalone `Plot.text` as a labelled scatter (#1106); the Vega-Lite `text` mark with `x`, `y` and `text` is the same chart and is unread. Worth its own issue, with the same reading decision to make.
- **`image`** — no case, and correctly so: an image mark places pictures at positions and carries no magnitude to announce. Same conclusion py-maidr reached for `ax.imshow` of an RGB array (xability/py-maidr#564).

## Related

- #1100 — the same reading on the Observable side, where the shapes had to be told apart by geometry rather than by spec
- #1094 — the `link` mark read as a gantt of its spans

jooyoungseo:
Closes #1122.

## What was wrong

Two specs, same data, same channels, differing only in the mark's name:

```js
{ data: SCHEDULE, mark: 'bar', encoding: {
    y: { field: 'task', type: 'ordinal' },
    x: { field: 'start', type: 'quantitative' },
    x2: { field: 'end' } } }
// -> [ 'gantt' ]

// mark: 'rule', everything else identical
// -> []
```

`[]` is a figure with no layers: the chart is unread. `resolveTraceType` had a case for `bar`, `line`, `trail`, `area`, `point`, `circle`, `square`, `tick`, `rect`, `boxplot`, `arc`, `errorbar`, `errorband` and `geoshape`, and none for `rule`, so it fell to `default: return null`. The adapter already said so out loud, in a comment at `convertPairedLayers`: *"a bare `rule` resolves to no trace type"*.

Which of the two an author writes is a choice about how thick the span is painted, not about what it means — the same two bounds on the same two channels — and `rule` is what Vega-Lite's own gallery reaches for to draw a range chart.

## What changed

`rule` goes through `resolveRangedBarType`, the helper `bar` already uses. Nothing else had to move: `extractGanttData` reads the spec rather than the geometry, and the selector machinery already knew this mark — `markToCssClass` gives `mark-rule`, and `markToChildElement` already returned `line` for it, since the `errorbar` whip is a rule.

The three other charts that wear this mark keep resolving to nothing, and the spec separates them **without measuring anything**:

| spec | reads as |
| --- | --- |
| `rule` + `x`/`x2` + ordinal `y` | `gantt` — was `[]` |
| `rule` + `y`/`y2` + ordinal `x` | `gantt` — was `[]` |
| `rule` + `y` only (a reference line) | `[]` |
| `rule` + `x` ordinal + `y` (lollipop stems) | `[]` |
| `rule` + `x`/`x2` where `x2` is a `datum` constant | `[]` |
| `rule` + `x`/`x2` under a dot layer (ranged dot plot) | `dumbbell`, unchanged — `convertPairedLayers` consumes both layers first |

A reference line and a lollipop's stem each carry **one** positional field, so neither states an interval any row of the data contains. That is what makes this cheaper than the same reading on the Observable side (#1100), where the three shapes had to be told apart by measuring the drawn geometry and asking whether every line shared an end.

## One thing the shared helper could not keep saying

It warns, when both bounds are present but the opposite axis' type was left to Vega-Lite's inference, that it is *"reading it as an ordinary bar"*. True of a bar — whose magnitude then becomes the lower bound by itself, a silent misread — and false of a rule, which has no other reading and is simply left out of the figure. The sentence now names the mark it is about and what actually became of it. Both halves are asserted.

## Tests

`test/adapters/vegalite/rangedRule.test.ts`, 9 cases.

The one carrying the most weight is **"reads it exactly as the same chart written with a bar"**: the two specs differ in one string, and differed in whether the chart was read at all, so type, data, axes and orientation are compared field by field between them rather than both being asserted against a literal — they cannot drift apart while each still matches something written down.

**"highlights through the element Vega draws a rule as"** is there because a `rule` renders as `<line>` where a `bar` renders as `<path>`, and a selector naming the wrong tag matches nothing — costing the layer its highlighting without costing it anything else, which no other assertion in the file would notice.

Falsified — each mutation applied alone against the fixed baseline:

| mutation | result |
| --- | --- |
| `case 'rule'` removed | 5 failures |
| `rule` returns `GANTT` unconditionally | 4 failures — the reference line, the stem, the constant bound, and the untyped lane axis |
| the warning always says "ordinary bar" | "leaves a rule unread when the lane axis has no declared type" fails |
| `markToChildElement` returns `path` for `rule` | the selector test fails |

`npm run lint:fix && npm run type-check && npm run build && npm test`: all clean, `385 passed` suites / `5397 passed` tests plus `10` / `137` on the ESM project. `lint:fix` changed nothing.

## Recorded in the issue, not fixed here

The same sweep found **`text`** has no case either. The Observable adapter reads a standalone `Plot.text` as a labelled scatter (#1106); the Vega-Lite `text` mark with `x`, `y` and `text` is the same chart and is unread. It needs its own reading decision. **`image`** has no case and correctly so — it places pictures at positions and carries no magnitude to announce.

---

### DIFF
--- a/src/adapters/vegalite/converters.ts
+++ b/src/adapters/vegalite/converters.ts
@@ -2887,9 +2887,14 @@ function warnCompositeDeclaration(spec: VegaLiteSpec): void {
  *
  * Vega-Lite has no lollipop mark and no dumbbell mark: both are authored as
  * a `layer:` of a segment and its dots. Read layer by layer, the segment is
- * dropped — a bare `rule` resolves to no trace type — and the dots are
- * announced as a scatter, so the chart loses the thing it is drawn to show:
- * the stem's magnitude, or the gap between the pair.
+ * dropped — a `rule` with one positional field resolves to no trace type,
+ * which is every stem this pairs — and the dots are announced as a scatter,
+ * so the chart loses the thing it is drawn to show: the stem's magnitude, or
+ * the gap between the pair.
+ *
+ * This pass runs *before* per-layer conversion, which is what keeps a genuine
+ * ranged dot plot reading as a dumbbell now that a `rule` spanning two bounds
+ * has a reading of its own (#1122).
  *
  * The two are told apart by the data. Two values at every category, told
  * apart by a grouping field, is a dumbbell; one value at each is a
--- a/src/adapters/vegalite/converters.ts
+++ b/src/adapters/vegalite/converters.ts
@@ -699,8 +699,10 @@ function getStepDirection(spec: VegaLiteSpec): StepDirection | undefined {
  * Some marks produce different MAIDR types depending on encoding:
  *   - `bar` with `bin` on x/y          → HISTOGRAM
  *   - `bar` with color/fill + stack    → STACKED / NORMALIZED / DODGED
- *   - `bar` spanning x–x2 / y–y2       → GANTT, or WATERFALL when the
+ *   - `bar` or `rule` spanning x–x2 / y–y2 → GANTT, or WATERFALL when the
  *     bounds are consecutive running totals
+ *   - `rule` with one positional field → nothing: a reference line or a
+ *     lollipop's stem states no interval of its own
  *   - `rect`                           → HEATMAP
  *   - `point` / `circle` / `square` / `tick` → SCATTER, or DOT when one
  *     positional channel is a category
@@ -913,18 +915,26 @@ function resolveFoldedAxes(
 }
 
 /**
- * Classify a bar that spans a range on one axis instead of standing on a
+ * Classify a mark that spans a range on one axis instead of standing on a
  * baseline.
  *
- * Both shapes draw a floating bar from a positional channel to its `2`
+ * Both shapes draw a floating span from a positional channel to its `2`
  * counterpart, with the remaining axis naming the row: a gantt lane, a
  * waterfall step. They are told apart by {@link hasRunningSumTransform}.
  *
+ * `bar` and `rule` both reach this. Which of the two an author writes is a
+ * choice about how thick the span is drawn, not about what it means — the
+ * same two bounds on the same two channels — and reading only one of them
+ * left an ordinary range chart unread for the sake of its mark's name
+ * (#1122).
+ *
+ * @param mark - The mark being classified, named only for the warning
  * @param encoding - The layer's encoding, merged with any parent's
  * @param transform - The transforms in scope for the layer
- * @returns The trace type, or `null` when the bar is an ordinary one
+ * @returns The trace type, or `null` when the mark spans no range
  */
 function resolveRangedBarType(
+  mark: string,
   encoding?: VegaLiteEncoding,
   transform?: VegaLiteTransform[],
 ): TraceType | null {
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Both observations were right and both are in 4b8cb5f.

**The stale comment.** `convertPairedLayers`' "a bare `rule` resolves to no trace type" was a fair description of the whole mark before this branch and is now only true of the stems that pass actually pairs — the ones carrying a single positional field. Reworded to say that, and the ordering guarantee you traced is written down there rather than left to be re-derived: this pass runs before per-layer conversion, which is what keeps a genuine ranged dot plot reading as a dumbbell now that a two-bound `rule` has a reading of its own.

**The untested `WATERFALL` path.** Pinned. Your reasoning for why it is "very likely fine" is exactly why it was worth a test: the mark's name reaches only the warning string today, so a future change that gave `rule` a branch of its own would take that path silently. `reads a rule over a running total as the waterfall a bar would be`, and the case says in its comment that nobody draws a waterfall with a rule and why it is pinned anyway.

Falsified by passing `undefined` for the transforms in the new `rule` case, which fails that one test and nothing else.

`npm run lint:fix && npm run type-check && npm test`: clean, `385 passed` suites / `5399 passed` tests, plus `10` / `137` on the ESM project.

---

PR #1125, opened 2026-08-21
feat(vegalite): read a text mark as the labelled scatter it draws

### LINKED ISSUE #1124: A Vega-Lite `text` mark is unread, so a labelled scatter announces nothing at all
The second half of the mark sweep that produced #1122.

## Measured

Run through `vegaLiteToMaidr` on `main`:

| spec | today |
| --- | --- |
| `text` + `x`/`y` quantitative + `text` field | `[]` |
| `text` + `x`/`y` quantitative, no `text` field | `[]` |
| `text` + `x`/`y` nominal + `text` field | `[]` |
| `image` + `x`/`y` | `[]` |

`[]` is a figure with no layers. The first row is a **labelled scatter** — the canonical country-names-against-GDP chart — and it is the one shape where the mark carries something no other mark can: `resolveTraceType` has no `text` case, so it falls to `default: return null`.

The Observable adapter already reads its equivalent as a labelled scatter (#1106), and the grammar has the field to carry it. `ScatterPoint.label`'s own doc says why:

> A reader told "x is 2.3, y is 14.1" has been given the two numbers they can see the shape of already and withheld the one thing they came for — which is why a volcano plot, a Manhattan plot and Observable's canonical country-names-against-GDP scatter are all drawn this way, and why `Plot.text` exists as a mark at all.

## Proposed

Read a `text` mark as a `SCATTER` whose points carry `label`, under the same rule #1106 settled for `Plot.text`, and for the same reason:

> Only on two continuous scales, which is what a labelled scatter is. A `text` on a categorical axis is a value written over a bar, and the bar mark already carries that value; reading it here would announce every count twice, once as a bar and once as a point standing at the same place.

Three shapes therefore keep resolving to nothing, and the spec separates all three without measuring geometry:

- **a categorical positional channel** — bar labels, heatmap cell values. The mark beside them already announces the number.
- **no `text` channel** — nothing is drawn to read.
- **a `text` layer inside a `layer:` that shares both positional fields with another layer** — the label overlay on a scatter. Vega-Lite's way of labelling points is `layer: [{mark: 'point'}, {mark: 'text', …}]` on the same channels, and reading both would give the figure two scatters over one set of points. Observable's adapter solves this with `labelOverlays`, which claims the `text` group and moves its names onto the points it labels; the Vega-Lite side can start by *declining* the overlay, which leaves the layered chart exactly as it reads today while the standalone case gains its reading.

Moving those names onto the points is the natural follow-up and is deliberately not in this — it is the difference between "a labelled scatter is read" and "every scatter that has labels beside it is named", and only the first is a chart that reads as nothing today.

## Not proposed

**`image`.** It places pictures at positions and carries no magnitude to announce, so `null` is the right answer and it should stay one. The same conclusion py-maidr reached for `ax.imshow` of an RGB array (xability/py-maidr#564).

## Related

- #1106 — the same reading on the Observable side, and the rule this follows
- #927 — `ScatterPoint.label`, the field this fills
- #1122 — the `rule` half of the same sweep

jooyoungseo:
Closes #1124. The second half of the mark sweep whose first half was #1122.

## What was wrong

`resolveTraceType` had no `text` case, so the mark fell to `default: return null` and a figure drawn entirely of it came back with **no layers at all**:

```js
{ data: CITIES, mark: 'text', encoding: {
    x: { field: 'area', type: 'quantitative' },
    y: { field: 'pop', type: 'quantitative' },
    text: { field: 'city', type: 'nominal' } } }
// -> []
```

That is the country-names-against-GDP chart, and the one shape where the mark carries something no other mark can. The grammar already had the field for it — `ScatterPoint.label`'s own documentation names `Plot.text` as the reason it exists:

> A reader told "x is 2.3, y is 14.1" has been given the two numbers they can see the shape of already and withheld the one thing they came for.

## The rule

#1106's, settled for the Observable side and holding here for the same reason: **only on two continuous scales**. A `text` on a categorical axis is a value written over a bar or into a heatmap cell, and the mark beside it already announces that number — reading it would say every count twice, once as a bar and once as a point standing at the same place.

| spec | before | after |
| --- | --- | --- |
| `text` + `x`/`y` quantitative + `text` field | `[]` | `scatter`, each point carrying `label` |
| `text` + `x`/`y` quantitative, no `text` field | `[]` | `[]` — empty strings carry no name |
| `text` + one categorical positional channel | `[]` | `[]` — a bar's own count |
| `text` + both categorical | `[]` | `[]` — a heatmap cell's value |
| `layer: [point, text]` on the same channels | `scatter` | `scatter` — unchanged |
| `layer: [point, text]` on *different* channels | `scatter` | `scatter`, `scatter` |
| `image` | `[]` | `[]` |

## The layered spelling needed a second rule

Vega-Lite labels a scatter with `layer: [{mark: 'point'}, {mark: 'text'}]` over the same two channels, and **both halves pass the test above** — so such a figure would have gained a second scatter announcing the same three points, once with names and once without.

`labelsAnotherLayer` declines the overlay. That leaves the layered chart reading exactly as it did before, while the standalone case gains the reading it never had. It is told by the **fields, not the order**: a label layer may be written either side of the mark it labels, and a `text` over different channels is a chart of its own rather than an annotation — both asserted.

Moving those names *onto* the points they label — what the Observable adapter's `labelOverlays` does — is the natural follow-up and is deliberately not here. It is the difference between "a labelled scatter is read" and "every scatter with labels beside it is named", and only the first reads as nothing today.

## Tests

`test/adapters/vegalite/labelledText.test.ts`, 11 cases, covering all seven rows of the table above plus the axis titles and the selector.

The one doing quiet work is **"leaves an ordinary scatter without labels it never had"**: `extractScatterData` is shared with `point`, `circle`, `square` and `tick`, and it now reads a `text` channel, so a mark that never had one must come back exactly as before.

Falsified — each mutation applied alone against the fixed baseline:

| mutation | result |
| --- | --- |
| `case 'text'` removed | 4 failures |
| `text` always `SCATTER` (categorical axes accepted) | the bar-label and heatmap-label cases fail |
| the label overlay is not declined | both layered cases fail |
| any sibling counts as the labelled mark | "keeps a text layer that charts something of its own" fails |
| the label is never read onto the point | the labelled-scatter case fails |

`npm run lint:fix && npm run type-check && npm run build && npm test`: all clean, `386 passed` suites / `5410 passed` tests plus `10` / `137` on the ESM project. `lint:fix` changed nothing.

## Not changed, and pinned so

`image` places pictures at positions and carries no magnitude to announce, so `null` is the right answer — the same conclusion py-maidr reached for `ax.imshow` of an RGB array (xability/py-maidr#564). Asserted here so the sweep cannot be read later as having missed it.

---

### DIFF
--- a/src/adapters/vegalite/converters.ts
+++ b/src/adapters/vegalite/converters.ts
@@ -2940,6 +2940,12 @@ function warnCompositeDeclaration(spec: VegaLiteSpec): void {
  * or after the mark it labels, and a `text` layer over *different* channels
  * is a chart of its own rather than an annotation.
  *
+ * Any non-`text` sibling counts, not only one that would itself read as a
+ * scatter. A `line` with per-point annotations over the same two channels is
+ * the same double-announcement -- the coordinates are the line's, and the
+ * text is written on top of them -- so what matters is that another mark
+ * already draws those positions, not what that mark resolved to.
+ *
  * @param specs - The layered spec's children
  * @param index - Position of the candidate label layer
  * @param parentEncoding - Encoding hoisted onto the layered parent
--- a/src/adapters/vegalite/converters.ts
+++ b/src/adapters/vegalite/converters.ts
@@ -164,6 +164,12 @@ export function vegaLiteToMaidr(
         i += 2;
         continue;
       }
+      // A `text` layer written over the mark it labels is not a second
+      // chart -- see `labelsAnotherLayer`.
+      if (labelsAnotherLayer(layerSpecs, i, spec.encoding)) {
+        i += 1;
+        continue;
+      }
       const layer = convertLayerSpec(
         layerSpecs[i],
         i,
@@ -703,6 +709,9 @@ function getStepDirection(spec: VegaLiteSpec): StepDirection | undefined {
  *     bounds are consecutive running totals
  *   - `rule` with one positional field → nothing: a reference line or a
  *     lollipop's stem states no interval of its own
+ *   - `text` on two continuous axes    → SCATTER whose points carry `label`
+ *   - `text` on a categorical axis, or with no `text` channel → nothing:
+ *     a value written over a bar, which the bar already announces
  *   - `rect`                           → HEATMAP
  *   - `point` / `circle` / `square` / `tick` → SCATTER, or DOT when one
  *     positional channel is a category
@@ -1086,6 +1095,28 @@ function resolveTraceType(
     // dumbbell it draws before this runs, by `convertPairedLayers`.
     case 'rule':
       return resolveRangedBarType(mark, encoding, transform);
+    // A `text` mark writes a name at a position, and on two continuous axes
+    // that is a labelled scatter -- the country-names-against-GDP chart
+    // `ScatterPoint.label` exists for. It resolved to nothing, so such a
+    // figure came back with no layers at all (#1124).
+    //
+    // Only on two continuous scales, which is the rule #1106 settled for the
+    // Observable side and settles here for the same reason: a `text` on a
+    // categorical axis is a value written over a bar or into a heatmap cell,
+    // and the mark beside it already announces that number. Reading it here
+    // would announce every count twice, once as a bar and once as a point
+    // standing at the same place.
+    //
+    // A `text` with no `text` channel draws nothing to read, and a `text`
+    // layer that labels another layer is declined before this runs -- see
+    // `labelsAnotherLayer`.
+    case 'text':
+      if (!hasField(encoding?.text))
+        return null;
+      return hasField(encoding?.x) && hasField(encoding?.y)
+        && !isCategorical(encoding?.x) && !isCategorical(encoding?.y)
+        ? TraceType.SCATTER
+        : null;
     case 'rect':
       return TraceType.HEATMAP;
     case 'boxplot':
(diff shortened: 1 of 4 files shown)


jooyoungseo:
The 🔴 finding is real and it is fixed in 60c63577. Thank you — that is a genuine defect I shipped, and the reasoning you gave is exactly right.

I verified the DOM claim against the codebase rather than taking it: `test/adapters/vegalite/facets.test.ts:661` builds `<g class="mark-text role-title-text"><text>…</text></g>`, and `facets.ts:433` reads a facet header out of `g.mark-text.role-title-text text`. A `mark-text` group holds `<text>`, so `g.mark-text.role-mark.marks path` matched nothing in a real chart: the labelled scatter read correctly through audio, text and braille while nothing lit up under the point being read.

`markToChildElement` now returns `text` for that mark, and its comment records that `text` reached the function for the first time in #1124 — the default branch's `path` was a guess nothing had exercised.

**The test is the part worth naming.** `expect(layer.selectors).toContain('mark-text')` is true whichever tag follows it, which is precisely the failure `markToChildElement`'s own doc predicts: selecting the wrong tag costs the layer its highlighting and nothing else, so it is invisible from every other channel. I wrote that same assertion for `rule` in #1123 *because* of that hazard and then wrote the weaker version of it here. It now resolves the selector against a rendered group, as `facets.test.ts` does:

```ts
const dom = new JSDOM('<svg><g class="mark-text role-mark marks">'
  + '<text>Aa</text><text>Bb</text><text>Cc</text></g></svg>');
expect(Array.from(dom.window.document.querySelectorAll(layer.selectors as string))
  .map(node => node.textContent)).toEqual(['Aa', 'Bb', 'Cc']);
```

Falsified: reverting `markToChildElement` to `path` fails that test alone.

**On `labelsAnotherLayer`'s breadth** — you read the intent correctly, and it was only in my head, so it is now in the comment. Declining for *any* non-`text` sibling over the same channels is deliberate: a `line` with per-point annotations is the same double-announcement, since the coordinates are the line's and the text is written on top of them. What matters is that another mark already draws those positions, not what that mark resolved to.

`npm run lint:fix && npm run type-check && npm test && npm run build`: clean, `386 passed` suites / `5410 passed` tests plus `10` / `137` on the ESM project.

---

PR #1127, opened 2026-08-21
fix(vegalite): give a layered spec's layers the data written above them

### LINKED ISSUE #1126: A layered Vega-Lite spec's inline data never reaches its layers, so every spec-only layered chart resolves to no rows
Found while starting the `text`-overlay follow-up recorded in #1124 — the first thing that work needs is a layered spec that actually carries data, and it turns out none of them do.

## Measured

`vegaLiteToMaidr(spec)` on `main` at `1b4b25da`, no compiled view supplied:

| spec | types | rows per layer |
| --- | --- | --- |
| single `point`, `data` at top level | `["point"]` | `[2]` |
| **layered** `point`, `data` on the parent | `["point"]` | **`[0]`** |
| **layered** `bar`, `data` on the parent | `["bar"]` | **`[0]`** |
| **layered** `line`, `data` on the parent | `["line"]` | **`[1]`** (one empty series) |
| **layered** `point` + `line`, encoding hoisted too | `["point","line"]` | **`[0, 1]`** |
| layered `point`, `data` repeated on the child | `["point"]` | `[2]` |

Only the last one works, and it is the form nobody writes: Vega-Lite's own examples put `data` once at the top level and `layer:` beneath it.

## Cause

`resolveData`'s inline fallback reads the **layer's own** spec:

```ts
// Inline data fallback.
if (spec.data && typeof spec.data === 'object') {
  const d = spec.data as { values?: Record<string, unknown>[] };
  if (Array.isArray(d.values))
    return d.values;
}
```

`convertLayerSpec` is handed the child spec. The parent's encoding and transforms are both threaded down to it (`parentEncoding`, `parentTransform`) — its `data` is not, so the fallback has nothing to find and returns `[]` after a `console.warn`.

## How much this bites

Stating the limit honestly rather than overclaiming: **`bindVegaLite` always passes a compiled view**, and the view path runs first, so an ordinary browser chart gets its rows from the view and never reaches this fallback. What this does affect:

1. **The fallback stops being a fallback for layered specs.** When the view path misses — `rowsCarryFields` rejects every candidate dataset, or the enumeration finds only internal datasets — a single-view spec still recovers from its inline data and a layered one cannot. It returns no rows and warns.

2. **Every spec-only test of a layered chart is asserting on empty data**, which is why this went unnoticed. `test/adapters/vegalite/labelledText.test.ts`'s overlay cases assert only `layer.type`; they pass identically whether the data is right, wrong, or absent. The same shape appears elsewhere in the layered tests.

The second is the reason to fix it now: #1124's follow-up is about the *values* a layered chart announces, and there is currently no way to test that spec-only.

## Proposed

Thread the parent's `data` down beside the encoding and transforms it already threads, and use it in the inline fallback when the layer has none of its own. A layer's own `data` keeps precedence, which is Vega-Lite's own rule.

Then the overlay tests can assert on rows, and #1124's follow-up — moving a `text` layer's names onto the points it labels — becomes testable.

## Related

- #1124 — the `text` overlay follow-up this was found under
- #1125 — the PR that added the overlay decline
- #927 — `ScatterPoint.label`, what the follow-up fills

jooyoungseo:
Closes #1126.

Found while starting #1124's recorded follow-up — the first thing that work needs is a layered spec that carries data, and it turned out none of them do.

## Measured

`vegaLiteToMaidr(spec)` on `main` at `1b4b25da`, no compiled view:

| spec | types | rows per layer |
| --- | --- | --- |
| single `point`, `data` at top | `["point"]` | `[2]` |
| **layered** `point`, `data` on parent | `["point"]` | **`[0]`** |
| **layered** `bar`, `data` on parent | `["bar"]` | **`[0]`** |
| **layered** `line`, `data` on parent | `["line"]` | **`[1]`** (one empty series) |
| **layered** `point`+`line`, encoding hoisted | `["point","line"]` | **`[0, 1]`** |
| layered `point`, `data` repeated per child | `["point"]` | `[2]` |

Only the last works, and it is the form nobody writes.

## Cause

`resolveData`'s inline fallback reads the spec it was handed, and `convertLayerSpec` gets the **child**:

```ts
if (spec.data && typeof spec.data === 'object') { … }
```

`parentEncoding` and `parentTransform` are both threaded down. `data` is not.

## Fix

A new `inheritData` gives each child the parent's `data` where it declares none — Vega-Lite's own precedence, so a layer that names a dataset keeps it. Applied to the specs rather than threaded as a tenth parameter, because `resolveData` is reached from several places (per-layer conversion, and the paired lollipop/dumbbell pass either side of it) and each would otherwise need the parent handed to it separately.

**The facet path already did exactly this** — `{ ...layerSpec, data: childSpec.data ?? spec.data }` — so this is following an idiom already in the file, not inventing one. The plain layered and concat paths had simply missed it.

## Scope, stated plainly

`bindVegaLite` always passes a compiled view, and the view path runs first, so an ordinary browser chart takes its rows from the view and never reaches this fallback. I'm not claiming every layered chart in the wild was broken. What changes:

1. **The fallback becomes a fallback again for layered specs.** When the view path misses — `rowsCarryFields` rejects every candidate, or enumeration finds only internal datasets — a single-view spec recovers from its inline data and a layered one could not.
2. **Spec-only tests of layered charts were asserting against no data.** That is why this went unnoticed: the existing layered tests check `layer.type` only and pass whether the data is right, wrong, or absent. #1124's follow-up is about the *values* a layered chart announces, and there was no way to test that.

## Verification

`npm run lint:fix`, `npm run type-check`, `npm run build` all clean. Full suite `5410 passed` + `137 passed`, no regressions — this touches shared data resolution, so that was the real risk.

Three mutations, each applied alone, each hitting exactly its own tests:

| mutation | fails |
| --- | --- |
| revert the layered inheritance | 3 (incl. the own-data-wins case, which needs the code path to run at all) |
| revert the concat inheritance | 1 (the concat test only) |
| let the parent override a layer's own data | 1 (the precedence test only) |

A sixth test pins that a layered spec with no data anywhere still gets an empty payload rather than anything conjured.

---

### DIFF
--- a/src/adapters/vegalite/converters.ts
+++ b/src/adapters/vegalite/converters.ts
@@ -3583,10 +3583,10 @@ function buildConcatMaidr(
     if (childSpec.layer) {
       const layers = inheritData(childSpec.layer, childSpec.data)
         .map((layerSpec, j) => {
-        // Vega names this child's mark groups `concat_<i>_layer_<j>_marks`
-        // (local layer index `j`, not the global data index), so drive the
-        // selector off those while keeping `globalLayerIndex` for the data
-        // lookup, which follows Vega's sequential dataset numbering.
+          // Vega names this child's mark groups `concat_<i>_layer_<j>_marks`
+          // (local layer index `j`, not the global data index), so drive the
+          // selector off those while keeping `globalLayerIndex` for the data
+          // lookup, which follows Vega's sequential dataset numbering.
           const layer = convertLayerSpec(
             layerSpec,
             globalLayerIndex,
@@ -4023,7 +4023,14 @@ function buildFacetMaidr(
     warnCompositeDeclaration(spec.spec);
   }
   const isLayered = childSpec.layer != null && childSpec.layer.length > 0;
-  const layerSpecs = isLayered ? childSpec.layer! : [childSpec];
+  // Both branches, not only the layered one: a facet's single child is
+  // itself the spec whose data sits on the enclosing `spec`, so wrapping
+  // only `childSpec.layer` left the unlayered case with none -- measured,
+  // panel enumeration then saw one cell where the chart has three.
+  const layerSpecs = inheritData(
+    isLayered ? childSpec.layer! : [childSpec],
+    childSpec.data ?? spec.data,
+  );
   const facetFields = [
     descriptor.rowChannel?.field,
     descriptor.columnChannel?.field,
@@ -4045,10 +4052,7 @@ function buildFacetMaidr(
   const layerRows = layerSpecs.map((layerSpec, j) => {
     if (perLayerDatasets)
       return perLayerDatasets[j];
-    const specForData: VegaLiteSpec = layerSpec.data != null
-      ? layerSpec
-      : { ...layerSpec, data: childSpec.data ?? spec.data };
-    let rows = resolveData(specForData, j, view);
+    let rows = resolveData(layerSpec, j, view);
     if (rows.length > 0 && !facetFields.every(field => field in rows[0])) {
       const source = resolveSourceRows(spec, view);
       if (source.length > 0 && facetFields.every(field => field in source[0])) {
@@ -4268,15 +4272,15 @@ function buildRepeatMaidr(
     }
     const childName = repeatChildName(cell.mapping);
     const isLayered = cellSpec.layer != null && cellSpec.layer.length > 0;
-    const layerSpecs = isLayered ? cellSpec.layer! : [cellSpec];
+    const layerSpecs = inheritData(
+      isLayered ? cellSpec.layer! : [cellSpec],
+      cellSpec.data,
+    );
     const parentEncoding = isLayered ? cellSpec.encoding : undefined;
 
     const rawLayers = layerSpecs.map((layerSpec, j) => {
-      const specForData: VegaLiteSpec = layerSpec.data != null
-        ? layerSpec
-        : { ...layerSpec, data: cellSpec.data };
       const layer = convertLayerSpec(
-        specForData,
+        layerSpec,
         globalLayerIndex,
         view,
         parentEncoding,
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Pushed `78eb71c6` — both findings acted on, and #1 was worth more than it looked.

## #1 — folding facet and repeat into `inheritData`

Right, and the point lands: the whole rationale for the helper was one copy of the rule, and I'd left three. `buildFacetMaidr` and `buildRepeatMaidr` now both call it.

Doing it exposed a trap I would not have found by reasoning — **the existing facet tests caught it**. Those inline copies applied to the *unlayered* branch too:

```ts
const layerSpecs = isLayered ? childSpec.layer! : [childSpec];
// …then, for both:
const specForData = layerSpec.data != null ? layerSpec : { ...layerSpec, data: childSpec.data ?? spec.data };
```

A facet's single child **is** the spec whose data sits on the enclosing `spec`. My first pass wrapped only `childSpec.layer`, so the unlayered case lost its data entirely and panel enumeration returned one cell for a three-panel chart — 12 facet tests red:

```
● converts a column facet into a 1xN subplot grid with per-cell data
    Expected length: 3
    Received length: 1
```

The inheritance now covers both branches, which is what the code it replaced did. Worth saying plainly since it is the second time in this PR that the *facet* path knew something the others didn't.

## #2 — concat comment indentation

Fixed. Confirming your suspicion: `npm run lint:fix` does **not** catch it — I ran it before pushing the first commit and it left those four lines alone, so eslint isn't checking comment indentation inside a moved callback. Corrected by hand.

## #3 and the verification note

Understood on not being able to run the commands — for what it's worth, re-run after this change: `lint:fix` and `type-check` clean, `npm run build` complete, full suite `5416 passed` (up 6 from the new tests) + `137 passed`.

---

PR #1128, opened 2026-08-21
feat(vegalite): give a label layer's names to the points it labels

### LINKED ISSUE #1124: A Vega-Lite `text` mark is unread, so a labelled scatter announces nothing at all

jooyoungseo:
Closes #1124 — the follow-up that issue recorded and #1125 deliberately left out.

## What was missing

Vega-Lite labels a scatter with a layer pair on the same two channels:

```json
"layer": [
  {"mark": "point", "encoding": {"x": {"field": "area"}, "y": {"field": "pop"}}},
  {"mark": "text",  "encoding": {"x": {"field": "area"}, "y": {"field": "pop"},
                                 "text": {"field": "city"}}}
]
```

#1125 declined the label layer, which stopped the figure announcing two scatters over one set of points. But declining is not reading: it left the chart exactly as it had always been — a scatter whose points are nameless — and those names are the whole reason the second layer was written. As `ScatterPoint.label`'s own doc puts it, a reader told "x is 2.3, y is 14.1" has been given the two numbers they can already sense the shape of and withheld the one thing they came for.

Measured before: `[{x: 10, y: 3}, {x: 20, y: 5}]`. After: `[{x: 10, y: 3, label: 'Aa'}, {x: 20, y: 5, label: 'Bb'}]`.

## How

`labelsAnotherLayer` becomes `labelOverlays`, reporting not only which `text` layers to skip but **which layer each one labels**, so that layer is converted carrying the `text` channel.

Named after the Observable adapter's function for the same job. That one pairs labels to points by DOM geometry because Plot hands it no data; here the spec says outright which rows both layers draw, so they pair by field — no tolerance, no measuring.

From there nothing new was needed: `extractScatterData` already fills `ScatterPoint.label` from `encoding.text`, which is how a standalone `text` mark gets its names. The overlay now reaches the same field by the same route.

Order-independent, like the decline it replaces — a spec that draws labels behind their marks reads identically.

## What still declines

| shape | behaviour |
| --- | --- |
| `bar` + `text` on a categorical axis | declined, bar payload unchanged |
| `line` + `text` | declined, line unchanged |
| `text` overlay with no `text` field | declined, points gain no empty name |
| standalone labelled scatter | unchanged (#1125) |

A bar's `text` writes the bar's own value, which the bar already announces — the reason #1124 gave for keeping that one declined.

## A guard I wrote and removed

I first guarded the move on `resolveTraceType(...) === TraceType.SCATTER`, so names only travelled to a layer that could hold them. Mutation testing showed **no test could distinguish it from its absence** — `encoding.text` has exactly two readers (`resolveTraceType`'s `text` case, unreachable for a non-`text` mark, and `extractScatterData`), so a `bar` or `line` handed the channel simply ignores it.

Rather than ship a branch nothing exercises, I removed it and let the bar and line tests pin the *outcome*. If another extractor ever starts reading `encoding.text`, those are the tests that go red. That's noted in the code so the next reader doesn't re-add it.

## Verification

`lint:fix`, `type-check`, `build` clean. Full suite `5421 passed` + `137 passed`.

Mutation: replacing `names.set(at, labels.text)` with a no-op — i.e. reverting to #1125's decline-only behaviour — fails exactly the two tests that assert the names arrive, and nothing else.

---

### DIFF
--- a/src/adapters/vegalite/converters.ts
+++ b/src/adapters/vegalite/converters.ts
@@ -1132,8 +1132,8 @@ function resolveTraceType(
     // standing at the same place.
     //
     // A `text` with no `text` channel draws nothing to read, and a `text`
-    // layer that labels another layer is declined before this runs -- see
-    // `labelsAnotherLayer`.
+    // layer that labels another layer is claimed before this runs, its names
+    // handed to the layer it labels -- see `labelOverlays`.
     case 'text':
       if (!hasField(encoding?.text))
         return null;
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Pushed `b02c8542`. Both doc findings were mine and both are fixed; the third turned into a test.

**#1 and #2 — stale after the rename.** Correct on both counts. `resolveTraceType`'s `text` case pointed at `labelsAnotherLayer`, and the function's own JSDoc still described the signature it had before this change — an `index` parameter it no longer takes, a boolean it no longer returns. `grep -rn labelsAnotherLayer src/ test/` is now empty. The JSDoc is rewritten to describe what it actually does and returns, including that the channel is handed over whatever the labelled mark is, with the reason at the assignment rather than duplicated in two places.

**The two-`text`-layers observation.** I measured it rather than take it as read, and it behaves exactly as you described:

```
layer: [point, text(n1), text(n2)]
→ types=["point"]  data=[{x: 1, y: 2, label: "second"}]
```

Both label layers are claimed, so the figure is one layer rather than three — no double-announcement — and `names` being keyed by the *labelled* layer means the later one wins. Now pinned by a test and stated in the JSDoc. Not reconciling it: "which of two names is the name" has no right answer, and silently concatenating or picking the first would each be inventing one. Better to have the behaviour known than tidy.

`lint:fix` / `type-check` clean, full suite `5422 passed` + `137 passed`.

On the verification note — understood, and it's worth saying that your static trace caught two things my green test suite could not, which is exactly the split that makes both worth having.

---

PR #1133, opened 2026-08-23
feat(grammar): give a band with bounds and no estimate a shape to be read as

### LINKED ISSUE #1047: A band with bounds and no estimate has no shape in the grammar (Highcharts arearange)
Found while measuring the Highcharts area family for #1046. Three series types
emit no layers at all; `streamgraph` was a missing bucket entry and is fixed
there. `arearange` is not — it is a shape the grammar cannot express.

## Measured

Real Highcharts 11 plus `highcharts-more.js` in Chromium, driven through
`highchartsToMaidr`:

| chart | `series.type` | layers emitted |
|---|---|---|
| `arearange`, data `[[5,15],[30,50],[12,28],[22,38]]` | `arearange` | **none** |
| `areasplinerange`, same data | `areasplinerange` | **none** |
| `polygon` | `polygon` | **none** |

`polygon` is a shape mark with no statistical reading, and declining it is the
right answer. The two range types are not.

## Why a bucket entry will not do

An `arearange` point is `[low, high]` — a band with two bounds and **nothing
between them**. Every existing interval shape in the grammar carries an
estimate:

```ts
export interface ErrorBarPoint {
  x: number | string;
  /** The estimate itself: a mean, a median, a fitted value. */
  y: number;
  yMin?: number;
  yMax?: number;
  z?: string;
}
```

`y` is required, and the three obvious ways to satisfy it are all wrong:

- **Midpoint** — `(low + high) / 2` is a number the chart never draws. A
  reader who hears "10" at a column drawn from 5 to 15 has been told something
  false, which is worse than being told nothing.
- **`low` or `high`** — picks one bound and announces it as the estimate,
  losing the other and implying a point reading the chart does not make.
- **Read it as a `gantt`** — the point shape nearly fits (an interval per
  position), but the *reading* does not: a gantt announces "this lane runs from
  a to b", and an `arearange` is one continuous band over a continuous axis,
  not a schedule of lanes.

This is the same class of question #942 settled for grouped error bars — a
shape the data has and the grammar did not — and it wants the same treatment
rather than a coercion at the adapter.

## The recommendation, and why it is not just done here

Relax `ErrorBarPoint.y` to optional and have `ErrorBarTrace` announce the
interval alone when there is no estimate.

The reason this is a maintainer's call rather than a change to make on its own:
**`ForestPoint extends ErrorBarPoint`**, and a forest plot's estimate is not
optional — the whole reading is whether the interval crosses the null
*relative to the estimate*. Relaxing the parent silently removes a guarantee
the child depends on, so the change is either

- a separate point shape for a band with no estimate, or
- an optional `y` on the parent with `ForestPoint` re-declaring it required,

and which of those is right is a statement about the grammar rather than about
Highcharts. Thirteen files reference `ErrorBarPoint` today.

There is also a sonification question worth deciding at the same time: with no
estimate there is no obvious pitch. Sweeping the band, or pitching its midpoint
while announcing only the bounds, are both defensible — a pitch is a positional
cue rather than a claimed number — but that is a design choice, not something
to infer.

## Who else draws this

Not only Highcharts. `plotly` has no native band type, but the same shape
arrives from `Plot.areaY` with `y1`/`y2` in Observable, from `geom_ribbon` in
ggplot2, and from `fill_between` without a centre line in matplotlib — all of
which currently either drop the band or attach it to a line that happens to be
in the same layer.

jooyoungseo:
Closes #1047.

Highcharts' `arearange` and `areasplinerange` emitted **no layer at all** — a chart drawn entirely as a band was silent — because every interval shape in the grammar required an estimate, and a band has none.

**This is the decision #1047 reserved, taken as a proposal rather than settled.** The issue set out two candidate shapes and I have implemented the second; if you prefer a separate point type, say so and I will redo it that way. What follows is why this one held up under the compiler and the tests.

## The grammar

`ErrorBarPoint.y` is optional; **`ForestPoint` re-declares it required.**

That is exactly the half the issue named as the reason not to just do it — relaxing the parent would silently remove a guarantee a forest plot depends on, since its whole reading is whether the interval crosses the null *relative to the estimate*. Re-declaring puts the requirement on the shape that has it, and the guarantee then stops being a convention and becomes something the compiler checks.

It immediately earned that: `tsc` found **every** site relying on the parent's promise — three, all building forest rows in the amCharts and Recharts adapters. The issue's "thirteen files reference `ErrorBarPoint`" worry resolved to those three, precisely, rather than by inspection.

None was a runtime gap. Both producers already guarantee an estimate — amCharts `continue`s past a sample without one, Recharts reads it with a total function — so the fix was to *say* so: `EstimatedPoint` names the guarantee at the point where it is enforced, instead of every consumer asserting it.

## The trace

One line, and it is the line the bounds already lived by:

```diff
-    sections = SECTIONS.filter(s => s === 'value' || <was it drawn?>)
+    sections = SECTIONS.filter(s => <was it drawn?>)
```

The estimate now earns its row the same way a bound does. `magnitudeOf` returns `NaN` for an absent estimate exactly as it already did for an absent bound, and every downstream guard (`isMeasured`, the per-row ranges, `valueRow`'s `Math.max(0, …)`) already handled that — the trace was shaped for this before the shape allowed it.

A band walks `lower` then `upper`. The `value` row of `NaN` — a third of every interval announcing nothing, with the pointer entering on it — is gone.

**That answers the sonification question the issue left open by not needing an answer.** With no `value` row there is no invented midpoint to pitch, and the two bounds are themselves positions the chart drew, on the shared scale the trace already builds. Asserted: a band spanning 5–15 and 30–50 pitches from 5 to 50.

## The adapter

`convertRangeSeries` reads `low`/`high` into `yMin`/`yMax` and no `y`.

Deliberately **not** `convertErrorBarSeries`' midpoint fallback. That is defensible for an unlinked whip, which *is* drawn about a centre, so the midpoint is where the chart put the estimate visually. A band is drawn about nothing, and `(low + high) / 2` would be a number the chart never shows — a reader told "10" at a region spanning 5 to 15 has been told something false, which is worse than being told only the two numbers that are true. Asserted directly (`expect(point.y).toBeUndefined()`).

`polygon`, the third silent type #1047 measured, stays declined — a shape mark with no statistical reading.

## Verification

`npm run lint:fix`, `npm run type-check`, `npm run build` all clean. **5429 + 137 tests pass**, up from 5422 + 137.

7 new tests: the band's two rows and their announcements, that there is no third row, that it pitches by its own bounds, both range series types through the real adapter path, the no-invented-estimate assertion, and a one-sided sample carried rather than dropped.

The band tests sit directly beside `omits a bound the data never carries`, which is the same rule read the other way round — the estimate and the bounds now differ in nothing but name.

## Not addressed

The issue's closing note — that `geom_ribbon`, `Plot.areaY` with `y1`/`y2` and `fill_between` without a centre line are the same shape from other producers — is a binding-side change in each of the three repositories. The grammar can now express what they draw; making them emit it is separate work.

---

### DIFF
--- a/e2e_tests/specs/errorbar.spec.ts
+++ b/e2e_tests/specs/errorbar.spec.ts
@@ -73,10 +73,15 @@ test.describe('Error Bar Plot', () => {
 
       expect(points).toHaveLength(3);
       for (const point of points) {
+        // The estimate is optional on the shape since #1047, because a band
+        // may draw only bounds. This chart is not one -- it draws all three,
+        // and asserting so is what makes the two comparisons below mean
+        // something rather than compare against `undefined`.
+        expect(typeof point.y).toBe('number');
         expect(typeof point.yMin).toBe('number');
         expect(typeof point.yMax).toBe('number');
-        expect(point.yMin!).toBeLessThan(point.y);
-        expect(point.y).toBeLessThan(point.yMax!);
+        expect(point.yMin!).toBeLessThan(point.y!);
+        expect(point.y!).toBeLessThan(point.yMax!);
       }
     });
 
--- a/src/adapters/amcharts/declaration.ts
+++ b/src/adapters/amcharts/declaration.ts
@@ -39,7 +39,7 @@ import type {
   VolcanoDeclaration,
 } from '@type/declaration';
 import type {
-  ErrorBarPoint,
+  EstimatedPoint,
   ForestPoint,
   ScatterPoint,
   SurvivalPoint,
@@ -892,7 +892,7 @@ function readBounds(
  */
 export function extractErrorBarSamples(
   declared: AmDeclaredLayer,
-): DeclaredSamples<ErrorBarPoint> {
+): DeclaredSamples<EstimatedPoint> {
   const declaration = declared.declaration as ErrorBarDeclaration;
   const horizontal = isDeclaredHorizontal(declared);
   const misses = explicitRefs(declaration, ['yMin', 'yMax', 'error']);
@@ -901,7 +901,10 @@ export function extractErrorBarSamples(
     : null;
 
   const { value } = valueFields(horizontal);
-  const data: ErrorBarPoint[] = [];
+  // `EstimatedPoint`, because the loop below drops any sample whose estimate
+  // is missing -- so this promises what it already enforces, and a forest row
+  // can be built from it without an assertion (#1047).
+  const data: EstimatedPoint[] = [];
   const items: AmDataItem[] = [];
 
   for (const item of declared.series.dataItems) {
--- a/src/adapters/highcharts/adapter.ts
+++ b/src/adapters/highcharts/adapter.ts
@@ -1579,6 +1579,14 @@ function convertSeries(
       return convertWaterfallSeries(series, containerId);
     case 'errorbar':
       return convertErrorBarSeries(series, chart, containerId);
+    // A band with two bounds and nothing between them. Same `low`/`high`
+    // shape as an error bar and the same trace, minus the estimate -- which
+    // is why these emitted no layer at all until `ErrorBarPoint.y` became
+    // optional (#1047). `areasplinerange` is `arearange` with a smoothed
+    // outline; the curve between the samples is drawing, not data.
+    case 'arearange':
+    case 'areasplinerange':
+      return convertRangeSeries(series, chart, containerId);
     case 'dumbbell':
       return convertDumbbellSeries(series, containerId, options);
     // A gantt series is an xrange with dates and lanes — `GanttSeries` extends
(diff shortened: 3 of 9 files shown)


jooyoungseo:
The review is right, and I reproduced it before fixing rather than taking it on the diff. On the `BAND` fixture, `getExtremaTargets()` returned:

```
Max value at feb = 30
Min value at jan = 5
```

30 is feb's **lower** bound. The chart's highest drawn point is 50, which `this.max` already reports in the description's stats — so "go to max" and the description contradicted each other, and the menu called a bound "value".

Fixed in 69784ece.

**What changed.** The bounds are still not ranked alongside an estimate, for the reason the original docstring gives: widest interval is a different question from largest value. But a band has no estimate for that rule to be *about*, so a new `extremaRowsOf(group)` ranks the estimate row alone when there is an estimate, and every section row together when there is not — the highest and lowest positions the chart actually drew, which is what `this.min`/`this.max` already span.

Each target now carries the section it was found in, so three things follow rather than being patched separately:

- the label names it — `Max upper bound at feb`, and still exactly `Max value at c` where there is an estimate, since `SECTION_LABEL.value` is `'value'`;
- `segment` is the real section instead of an unconditional `'value'`;
- `navigateToExtrema` resolves the row from `target.segment` rather than from the fallback, so the cursor lands on the upper bound it named.

**One thing the review didn't name, which the fix required.** The same-sample check was `lowest.index !== highest.index || lowest.groupIndex !== highest.groupIndex`. A one-sample band's two bounds share both a column and a group while being the two ends of the very spread being reported, so that check would have dropped the minimum and announced a band with no spread at all. It now compares the section too.

**Verification.** Three tests added, and each part of the fix mutation-tested alone against a restored baseline — all five mutations killed:

| mutation | test that failed |
|---|---|
| band branch never taken | ranks a band across both its bounds *(+2)* |
| dedup ignores the section | offers both ends of a one-sample band |
| navigation pins to `valueRow` | lands a band's maximum on its upper bound |
| `segment` always `'value'` | ranks a band *(+1)* |
| label always says "value" | ranks a band across both its bounds |

`npm run lint:fix`, `npm run type-check` and `npm run build` clean; `npm test` → 5436 passed (389 suites) + 137 passed (10 ESM suites).

The stale `valueRow` docstring the review flagged as minor is updated in the same commit: it now says it falls back to `lower` on a band and points at `hasEstimate` for anything that means the estimate specifically.

---

PR #1135, opened 2026-08-23
fix(segmented): stop the mark's tag deciding how a layer pairs with the DOM

jooyoungseo:
Refs #1003. This is **piece 1**, and it is a proposal — the decision the issue reserved, implemented the way the issue itself recommends, with the cost stated so you can redirect it.

Pieces 2 and 3 are already in (`test/model/segmentedDomBranch.test.ts`, and #1115's per-cell grid for Highcharts).

## The change

The two branches were the same code apart from their default, so they collapse into one:

```ts
const isColumnMajor = this.layer.domMapping?.order === 'column';
```

One walk, one default, and **the tag decides nothing**. Three consequences:

1. A library that changes its mark tag between majors — as Highcharts did between v10 and v11, with no change to its DOM order — no longer flips how its cells pair.
2. A mark that is **neither** `<path>` nor `<rect>` gets a pairing instead of falling through both branches and leaving the layer with no highlight at all. Nothing in the tree draws such a mark; a d3 user could, since their DOM is their own.
3. ~80 lines of duplicated walk logic go, which is what made the two defaults drift apart unnoticed in the first place.

## What it retires, stated plainly

Row-major as the single default is what the issue recommends ("Row-major is the better default on the evidence"), and my earlier sweep confirmed it is a no-op for every adapter in the tree. But it is **not** merely aligning two branches that drifted: it retires the default's claim, documented on `groupsRunForward`, that an undeclared producer draws each category's segments bottom-up.

That convention is not lost — it is what `order: 'column'` now says, and `groupsRunForward` still decides which end of such a category the elements start from. What changes is that a producer has to *say* it rather than get it by drawing the right tag. **A producer added tomorrow that draws bottom-up and declares nothing would be right under the old default and wrong under this one.** That is the trade, and it is the reason this was reserved. The one-line revert is `!== 'row'` in place of `=== 'column'`.

## What moved in the suite

Exactly what the issue predicted — three cases, two files, **no adapter test** — plus the neither-tag case, which now pairs rather than reporting nothing:

| case | what it did | what it does now |
|---|---|---|
| `walks a rect-marked layer category by category…` | pinned the asymmetry | replaced by `answers the same whichever tag the marks are` |
| `pairs nothing at all with a mark that is neither` | pinned the silent no-highlight | `pairs a mark that is neither tag like any other` |
| `barGaps`: gap stands aside | inherited the rect default | declares `order: 'column'`; assertion unchanged |
| `diverging`: reads the sides in declared order | inherited the rect default | declares `order: 'column'`; assertions unchanged |

The last two declare rather than move because their subjects are elsewhere. The bar-gap case is about a gap standing aside instead of consuming a real element; the walk is scaffolding. The diverging fixture genuinely draws each band's two bars adjacent — and **I deliberately did not move that onto `DivergingTrace`**, because diverging producers do not agree on a layout: Victory draws series-major and declares nothing, so a trace-level override would have broken it. It stays a fact about the fixture.

## Verification

Applied alone against a restored baseline; all four killed:

| mutation | tests that failed |
|---|---|
| column-major when nothing declared (the old rect default) | 3 |
| a declared `column` never honoured | 4 |
| `groupDirection` ignored within a category | 3 |
| a skipped cell consumes an element anyway | 1 |

`npm run lint:fix`, `npm run type-check` and `npm run build` clean. `npm test` → 5436 passed (389 suites) + 137 passed (10 ESM suites).

---


jooyoungseo:
E2E caught something that changes the evidence in this PR, and I want to correct it plainly rather than leave the original claim standing.

**The claim was wrong.** I said on #1003, and repeated in the PR body, that no producer in the tree relies on the rect branch's column-major default. That was measured **across adapters**, and it holds there. It is not true of the repo as a whole: **authored specs are producers too, and two of this repo's own examples relied on it.**

```
examples/pyramid.html             diverging_bar            <rect>, category-major, nothing declared
examples/normalized-barplot.html  stacked_normalized_bar   <rect>, category-major, nothing declared
```

`pyramid.spec.ts › should light up the bar it just announced, not the one across the axis` failed — the highlight landed at `x=221.8` where the test expects `x=400`, which is a Men bar lit up for a Women announcement. That is exactly the failure that spec exists to catch.

**The second one is the more useful finding.** I did not stop at the failing spec; I swept every inline `maidr` attribute under `examples/`. `normalized-barplot.html` is in the same position — `<rect>`, category-major, bottom-up within each category, nothing declared — and it has **no highlight assertion of its own**, so it would have shipped pairing every cell with somebody else's bar with all 518 e2e tests green. That is precisely the blind spot #814 names: audio, text and braille read from the unaffected side and stay correct.

## What I changed (809e91db)

Both examples now declare `"domMapping": { "order": "column" }` — one line each, and a true statement about the layout they draw.

`test/model/authoredSegmentedDomMapping.test.ts` asks for the declaration outright on every authored segmented layer that hands over a flat selector, rather than inferring each page's DOM order from its SVG. A page whose DOM really is series-major loses nothing by saying so. Falsified by removing one declaration: the test names the offending page and type. Layers handing over a grid of selectors are exempt — a grid names every cell already.

## What this does to the proposal

It does not withdraw it, but it sharpens the trade, and you should weigh the corrected version rather than the one in the PR body:

- **There is no free unification.** Row-major as the single default costs these two authored examples. Column-major as the single default would cost plotly, recharts and victory, which draw `<path>` series-major and declare nothing. Whichever way it goes, somebody declares. Row-major is the cheaper side — two authored pages against three shipping adapters — but it is not zero, and I previously said it was.
- **The two dependents are the hand-authored chart types**, a population pyramid and a 100% stacked bar. That is evidence for reading the default as "a sensible guess for producers who will not declare" rather than "a legacy shape to phase out" — which is the framing you reserved this decision on. It cuts against the change, and I would rather say so than bury it.
- **The tag-dispatch defect is independent of all that.** A mark's tag deciding the strategy, and a mark of any other tag getting no highlight at all, are wrong under either default. If you want the other default, the same collapse works with `!== 'row'` on that one line and these two examples need no declaration.

Verification after the fix: full e2e run locally, **518 passed**; `npm test` → 5438 passed (390 suites) + 137 passed (10 ESM suites); lint, type-check and build clean.

---

jooyoungseo:
Answering the review's one non-blocking question — the `barValues.length === 0` path returning `null` where a `<path>` row-major call used to return `[]`.

**Deliberate**, as part of consolidating a guard that was previously in one branch and not the other: the rect branch had `if (!this.barValues || this.barValues.length === 0) return null;` and the path branch had nothing. With one walk there is one guard, and `null` is the honest answer — `[]` says "resolved a grid, it has no rows", which is not what happened.

**Inert at both consumers**, checked rather than assumed:

- `bar.ts:415` — `!this.highlightValues || this.highlightValues.length === 0` treats them identically.
- `segmented.ts:281` — `this.highlightValues === null || this.row === this.barValues.length - 1`. With `[]`, `barValues.length - 1` is `-1`, `this.row` starts at `0`, so that check passes and the next line reads `[][0]` → `undefined` → the `!rowElements` guard → `outOfBoundsState`. With `null` it returns `outOfBoundsState` one line earlier. Same state, either way.

**And the input is unreachable for this class.** `mapToSvgElements` runs inside `super()` at `bar.ts:150`; `SegmentedTrace`'s constructor then calls `createSummaryLevel()`, which indexes `this.barValues[0].length` at `segmented.ts:37`. An empty `barValues` therefore throws one line after the branch in question, so no `SegmentedTrace` with empty `barValues` is ever successfully constructed. The guard is a cheap statement of an invariant, not a live path — which is also why I dropped the two per-branch guards the review identified as dead (`!this.barValues[r]` and `!this.barValues[0]`): `barValues` is dense by construction at `bar.ts:145`, and `totalExpected`'s `row.length` would throw before either could fire.

No change made for this. If you would rather the guard return `[]` to preserve the older path-branch answer exactly, it is one line — but on the above it would be preserving a value nothing can produce and nothing distinguishes.

---

PR #1144, opened 2026-08-23
feat(keybinding): walk a layer's lowest and highest values with the brackets

jooyoungseo:
# Pull Request

## Description

Adds two keybindings that move the cursor to a layer's extreme values:

| Key | Action |
| --- | --- |
| `[` (open bracket) | Go to the **minimum** value of the current layer |
| `]` (close bracket) | Go to the **maximum** value of the current layer |

Reaching an extreme previously meant opening the extrema dialog with `g`, arrowing through its list and pressing Enter. The brackets are the one-press shortcut for the two entries people reach for most.

A value can be shared by several points. Each press steps to the next point holding that value and wraps at the end, and announces where it landed among them — `A, 90, 1 of 3`. Without this a reader has no way to learn that a high was reached three times rather than once.

## Related Issues

None filed — this came in as a direct request.

## Changes Made

**`src/service/keybinding.ts`** — binds `GO_TO_MIN_VALUE` (`[`) and `GO_TO_MAX_VALUE` (`]`) in the `TRACE`, `BRAILLE` and `CANDLESTICK_DELTA` keymaps, each with a description and a `helpKey` that spells the bracket out. The help menu and the command palette are both generated from the keymaps, so there is no second list to update.

Braille scope is included even though the extrema dialog (`g`) is not: a modal would fight the braille text area for focus, but a jump that opens nothing has no such problem, and braille readers benefit most from reaching an extreme in one press. `Ctrl+↑` / `Ctrl+↓` already do the same kind of model-level jump from that scope.

**`src/service/goToExtrema.ts`** — new `goToExtremeValue(type)`. It collects every target sharing the extreme value, steps to the next one, and navigates through the same `navigateToExtrema` the dialog uses, so the move is sonified, brailled and highlighted by the usual observer chain. Three decisions worth a look:

- **Position is read back from the trace, not remembered.** A stored index would go stale the moment the reader moved by any other means, or the data changed under a live chart. `isCursorAt` mirrors what `navigateToExtrema` does with a target — a group target addresses row and column, a point target a column — so the two cannot drift apart.
- **Entry is the one position not read back.** Before the reader has moved, the cursor is parked on the first point without their having heard it, so it is not a position to advance from; treating it as one would skip the first tie when a chart opens on it. Guarded by `isInitialEntry`, and covered by a test.
- **`intersection` targets are excluded.** They are neither a minimum nor a maximum, and their value would otherwise compete for the extreme.

**`src/command/goTo.ts`** — `GoToMinValueCommand` and `GoToMaxValueCommand` over a small shared base, which owns the announcement:

- The position is appended to the point text rather than sent on its own. The live region holds one message at a time and a notification *replaces* the navigation text, so announcing `1 of 3` alone would cost the reader the point they pressed for.
- A value held by one point announces nothing extra — `1 of 1` is noise, and the move already announced itself.
- **Terse keeps the position; off suppresses it.** Terse shortens the point description, it does not drop navigational position — `Layer n of m` is announced in either mode — and the count is the one thing no point text carries in any mode. Off is a deliberate choice to navigate by tone and braille, and a notification announces regardless of the text setting, so that guard is the only thing standing between the position and a reader who turned text off.
- When there is nowhere to jump at all it announces so, rather than falling silent: the keys are bound across the whole scope, so they fire on the figure lobby and on trace types without extrema support too, and a silent press reads as a broken shortcut.

**`src/model/bar.ts`** — bar traces reported only `indexOf`'s first match, so two bars of equal height looked like one to the dialog as well as to the brackets. They now list every tied index, as line traces already did.

**`src/command/command.ts`, `src/controller.ts`, `src/command/factory.ts`** — `goToExtremaService` joins `CommandContext` and the factory gains the two cases. The jump touches no Redux state, so the commands take the service directly rather than going through `GoToExtremaViewModel`.

**`e2e_tests/`** — bracket key constants, `goToMinimumValue()` / `goToMaximumValue()` on `BasePage` alongside the existing jump helpers, a tied-extremes fixture page, and a spec that presses the keys across trace scope, braille scope and both text modes.

### Known limitation

Walking ties needs the trace to report them. `line` already did and `bar` now does. `heatmap`, `segmented`, `candlestick`, `errorBar`, `dumbbell` and `waterfall` still report one target per extreme, so a walk there is one long and announces no position — unchanged behaviour rather than a regression. Extending them is a per-trace change against six different data shapes and is left out of this PR deliberately.

Not applicable — the change is keyboard-only and produces no new visual surface. The two new rows appear in the generated help menu and command palette.

## Checklist

- [x] I have tested the changes locally following `ManualTestingProcess.md`, and all tests related to this pull request pass. — the repo has no `ManualTestingProcess.md`; the keys were driven in a real browser (see below) and every automated suite passes. No screen-reader pass was run.
- [x] I have updated the documentation, if applicable. — the help menu is generated from the keymap descriptions, and the docs carry no keybinding table.

## Additional Notes

**Verification on the current head, run locally:**

- `npm run lint:fix` — clean
- `npm run build` — all 14 targets built
- `npm test` — 394 suites / 5483 tests, plus 10 suites / 137 tests, all passing
- chromium e2e (the suite CI runs on every PR) — **530 passed**

**Tests**

`test/command/goToExtremeValue.test.ts` (22 cases) covers the selection and walk against a real `BarTrace` — stepping and wrapping, the two directions walking independently, restarting when the reader moved away by other means, ties reported rather than collapsed, intersection targets skipped, the announcement composed with and without a position, and silence under text-off.

`e2e_tests/specs/goToExtremeValue.spec.ts` (12 cases) presses the keys in a browser. Five use the bar example for the single-extreme cases including braille scope; seven use a new `e2e_tests/fixtures/tied-extremes.html` for the walk, the position announcement, and the two text modes. That fixture exists because **none of the 207 charts in `examples/` has a tied high or low** — I scanned them — so nothing there can exercise a press that has somewhere further to go.

Every behaviour claim here was checked by mutation, **per direction** rather than per feature — breaking one thing must fail exactly the cases that depend on it. That distinction is not academic: a coarser check on the braille pair let a vacuous assertion through earlier in this PR, which the review caught.

| Mutation | Result |
| --- | --- |
| `]` rebound to an unused chord | the two `]` cases fail, `[` stays green |
| braille-scope `]` dropped | the braille case fails, trace-scope cases stay green |
| walk never advances (`next = 0`) | the four tie-walking cases and three unit tests fail, the single-extreme cases stay green |
| `isOff()` guard replaced with `false` | the text-off case fails alone |

**Key codes** — `hotkeys-js` maps `[` to 219 and `]` to 221 in its own key map, so the bindings resolve correctly rather than falling through to `charCodeAt`.

**Review history** — seven automated passes, ending with no blocking issues and every finding closed. No correctness bug was ever found in `src/`; all five changes the reviews prompted were to the tests.

| Finding | Outcome |
| --- | --- |
| No test exercised the `CommandFactory`/keymap wiring | Fixed — `473d46f` |
| Overlapping guard in `goToExtremeValue` | Deliberate, kept ([why](https://github.com/xability/maidr/pull/1144#issuecomment-5388486442)) |
| `BRAILLE` bindings — keep or drop? | Kept; reviewer agreed |
| No browser-level check of the `BRAILLE` claim | Fixed — `dabe3d5` |
| Per-trace-type e2e coverage | Not added; the jump is scope-level ([why](https://github.com/xability/maidr/pull/1144#issuecomment-5388598130)) |
| Braille test's max assertion was vacuous | Fixed — `49207fe` ([why it slipped](https://github.com/xability/maidr/pull/1144#issuecomment-5388650825)) |
| Should terse suppress the position too? | No — pinned in a browser, `8a96436` ([why](https://github.com/xability/maidr/pull/1144#issuecomment-5389057325)) |


jooyoungseo:
Thanks for the trace through the full path — pushed `473d46f` for note 2, and here is where the other two landed.

**Note 2 — factory/keymap wiring is now covered.** You are right that this was the one real gap: a keymap name and its `CommandFactory` case are matched by string alone, and `create` throws `Invalid command name` only when the key is pressed. Nothing in the suite presses a key, so a rename on either side would have shipped green. `test/command/goToExtremeValue.test.ts` now pins both directions — the three scopes bind `[`/`]`, the factory resolves each name to its command, and both carry a help description so the generated menu lists them. The factory call needs an `as Keys` cast: `Keys` is `keyof Keymap[Scope]`, the intersection of every scope's keys, and `SETTINGS_KEYMAP` is empty, so it narrows to `never` — `KeybindingService` casts at the same boundary when it walks `SCOPED_KEYMAP`, and the test mirrors that rather than working around a quirk it did not introduce.

**Note 1 — leaving the overlapping guard as is.** Your reading is right that it is intentional: `context.state` is the broader `PlotState` union here, so the `state.type !== 'trace'` check is what narrows it, and `isExtremaNavigable` separately proves the active plot is an `AbstractTrace` that opted in. Neither implies the other — a populated trace state does not guarantee the active plot implements extrema — so removing either would widen what reaches `getExtremaTargets()`.

**Note 3 — keeping the `BRAILLE` bindings.** Agreed, and thanks for confirming the `hotkeys.filter` path; that was the part I could not verify without a browser.

**On what you could not run:** all four are now green locally on the pushed head — `npm run lint:fix`, `npm run type-check`, `npm test` (394 suites / 5475 tests, plus 10 / 137), and `npm run build`. I also ran the chromium e2e suite that CI runs on every PR: **518 passed**. That took working around a sandbox quirk rather than a repo one — the pinned Playwright wants browser build 1217 and this image ships 1194, so every spec died at launch until I pointed `executablePath` at the installed binary through a scratch config. That config was never committed; the diff is unchanged by it.

---

jooyoungseo:
Took the braille suggestion — pushed `dabe3d5`.

You were pointing at the one claim in this PR that rested on a reading rather than a run. The braille binding only holds if two things are true at once: `hotkeys.filter` lets the key through to a focused text area, **and** `preventDefault` keeps the bracket character out of the field. You confirmed both statically; nothing had actually pressed the key there.

The new case turns braille mode on, presses `]` then `[`, and checks the braille cursor lands on the extreme cell (`selectionStart` against the bar's index, read from the example's data) while the field's contents stay byte-for-byte unchanged. Both assertions matter: the cursor alone would pass if the character were also typed, and the contents alone would pass if the key did nothing at all.

I checked it discriminates rather than just passing: dropping the two bindings from `BRAILLE_KEYMAP` fails this case alone and leaves the four trace-scope cases green. (The same check on the earlier commit: rebinding `]` to an unused chord failed exactly the two `]` cases.)

On your note about trace-type coverage — I agree the jump is scope-level rather than per-trace, so I have not added a spec per trace type. What the bar example exercises is the scope wiring; the per-trace part is `getExtremaTargets()`, which each trace already owns and which `test/command/goToExtremeValue.test.ts` covers at the selection layer.

Current head, all run locally: `lint:fix`, `type-check` and `build` clean; `npm test` 394 suites / 5475 tests plus 10 / 137; chromium e2e **523 passed**.

---

jooyoungseo:
Good catch, and correct — fixed in `49207fe`.

I confirmed it against the primary sources rather than taking the reading: `examples/barplot.html` is `Sat: 87, Sun: 76, Thur: 62, Fri: 19`, and `activateMaidr()` leaves the cursor on the first cell, so `max.index` is 0 and the braille test was asserting a position the cursor already held. A bracket bound to nothing would have passed that half.

Worth being precise about why my teeth check missed it, since that is the part I got wrong rather than just the test. I ran the mutation on the *whole* braille pair — I dropped both bindings from `BRAILLE_KEYMAP` and saw the case go red, and stopped there. That red came entirely from the minimum assertion further down; the maximum assertion was passing vacuously the whole time and the coarse mutation hid it. A per-direction mutation would have caught it, and that is what I ran this time: breaking **only** the braille-scope `]` now fails this case, where before the fix it passed.

The fix goes to the minimum first, which has to travel from index 0 to index 3, then asserts the maximum from there, which has to travel back. I also pinned the two assumptions the ordering rests on — that the cursor starts at 0 and that `min.index` is not 0 — so the test states its reasoning instead of depending on it silently, and would fail loudly if the example's data or the entry behaviour changed rather than quietly going vacuous again.

I left the trace-scope `]` case as it is. It presses from the same index-0 entry point, but the mutation shows it discriminating: before the first move the info region holds the instruction text rather than a data point, so an unbound key leaves nothing to match. That is teeth from the entry state rather than from the cursor having to travel, and `should reach an extreme from anywhere, not just from the entry point` is the case that covers the travel by design.

Current head: `lint:fix`, `type-check` clean; `npm test` 394 suites / 5475 tests plus 10 / 137; chromium e2e **523 passed**.

---

jooyoungseo:
On the terse question — **keeping it**, and pinned in a browser in `8a96436`.

Terse shortens the point *description*; it doesn't drop navigational position. I checked rather than assumed: `formatLayerSwitchAnnouncement` builds `Layer n of m` with no mode gate at all, so layer position is announced in either mode. The one place terse does drop a counter is the subplot lobby, and the comment there gives the reason — an authored title already identifies the panel, so `Subplot n of m` is redundant next to it. Nothing plays that role here: no point text in either mode says how many points share the value, or whether the walk has come back round to the first. In terse the reader hears `A, 90` / `C, 90` / `E, 90`, which distinguishes the points but never says there are three. So terse is if anything where the suffix earns its place more, not less.

Since the decision was only asserted against a mocked `TextService`, both modes now run for real:

- **terse** — presses still announce `1 of 3`, `2 of 3`. The case asserts the mode instead of trusting the cycle: `t` runs verbose → terse → off, and a reordering there would leave it passing from verbose and proving nothing.
- **off** — the position is silent. This is the guard you traced, and it was the only one of the two never exercised end to end; a notification bypasses `TextViewModel`'s enabled check, so nothing else stands between the announcement and a reader who turned text off. Replacing `isOff()` with `false` fails that case alone.

Thanks also for confirming the `isInitialEntry` short-circuit is load-bearing — that one was found by a test rather than by design, so it's good to have it independently traced.

Current head: `lint:fix`, `type-check`, `build` clean; `npm test` 394 suites / 5483 tests plus 10 / 137; chromium e2e **530 passed**.

---

Issue #1138, opened 2026-08-23
Sixteen Highcharts series types emit no layer, and five of them already have a home in the grammar


jooyoungseo:
The Highcharts adapter's `convertSeries` dispatches 36 series types and drops everything else. #814's trace roadmap is closed — the grammar has 48 authorable types — so the question is no longer "what can MAIDR express" but "which of them does this adapter reach".

## Measured

Every candidate driven through `highchartsToMaidr` with a minimal series, counting the layers it produced. `spline` and `areaspline` are the controls, and they show the harness works:

```
bubble           — no layer      columnpyramid    — no layer
packedbubble     — no layer      variwide         — no layer
columnrange      — no layer      bellcurve        — no layer
pareto           — no layer      timeline         — no layer
item             — no layer      organization     — no layer
venn             — no layer      polygon          — no layer
vector           — no layer      windbarb         — no layer
flags            — no layer      pictorial        — no layer

spline             line                     ← control
areaspline         area                     ← control
```

`convertSeries`'s `default:` does `console.warn` and returns `null`, so the series contributes nothing and the chart reads as though it were not there. On a single-series chart that is silence.

## Five already have an exact home

Not new trace types — existing ones this adapter does not reach:

| series | reads as | why it fits |
|---|---|---|
| **`bubble`** | `scatter` with `z` | `ScatterPoint.z` already exists and already drives `zIntensityFor()`. This is exactly what #826 did for Chart.js, so the trace side is done. |
| **`columnrange`** | `error_bar` band | A bar from `low` to `high` per category with no estimate — the shape #1047 just made expressible when `ErrorBarPoint.y` became optional. |
| **`variwide`** | `mosaic` | A column whose **width** is a second magnitude, which is what `MOSAIC` was added for (#804): "the bar's width is the second magnitude". |
| **`columnpyramid`** | `bar` | A column with a tapered outline. Same data, different rendering. |
| **`pictorial`** | `bar` | A column drawn with a repeated icon. Same data, different rendering. |

The last two are the cheapest thing on this list — the series carry ordinary `y` per category and differ from `column` only in how Highcharts paints them.

## Four more with a plausible home, needing a decision

- **`bellcurve`** → `smooth`. It fits a normal curve and evaluates it at renderer-chosen points, so there are no observations to announce and the sample count is a drawing parameter — the same reasoning `stat_function` gets in r-maidr (xability/r-maidr#202).
[...]

PR #1150, opened 2026-08-24
feat(highcharts): read a Pareto curve as the cumulative line it draws

jooyoungseo:
The second of #1138's four judgement cases, after `bellcurve` in #1149. The judgement is [on the issue](https://github.com/xability/maidr/issues/1138#issuecomment-5389814865).

`buildSubplot` sorts series into buckets by type and `pareto` is in none of them, so it reached `convertSeries` as an unsupported type and was declined. The columns beneath it still read, so the chart came out as **half of itself**: the bars without the cumulative curve that is the whole reason a Pareto chart is drawn rather than a bar chart.

## The part worth pinning: the numbers are percentages

Measured on Highcharts 11.4.8 plus `modules/pareto.js` in Chromium, over a base whose total is **not** 100 so that the two candidate readings differ:

```
base column counts   80, 60, 40, 20     (total 200)
pareto series.data   40, 70, 90, 100
a running total would be   80, 140, 180, 200
```

It is the cumulative **percentage**. So nothing here may convert the values back into counts — the chart does not draw counts, and its axis has a maximum of 100. That axis is the secondary one the author titled, conventionally "Cumulative %", which `getAxisLabel` reads off `series.yAxis` without being told anything.

Verified again through the example page's own configuration: base counts of 80, 60, 40, 20, 10 over a total of 210 give 38.1, 66.7, 85.7, 95.2, 100.

## The rest

- **A second layer, not a replacement.** `series.linkedParent` is null and the columns are reached through `series.baseSeries`, which the adapter reads on its own terms. A chart drawing both gets both — columns first, then the curve, which is the order they are drawn in and the order `buildSubplot` already sorts its buckets into.
- **The graph path is the handle**, which is what every line-family layer takes and what `LineTrace` parses for its vertices.

## A claim I had to correct

I first wrote that the curve draws "no point marks", carried over from an earlier sweep. Re-measuring for the example page contradicted it, and the earlier reading turned out to be a **query-scoping artifact**: the markers live in a `highcharts-markers` group that is a *sibling* of the series group, so a query scoped inside the series group finds none. Measured properly at four, five and twenty points, a pareto curve always draws one marker per step. It changes nothing about the reading — they are the same decoration an ordinary `line` draws, and `convertLineSeries` does not address them either — but the comment and the test now say what is true.

(The same sweep's `bellcurve` claim was re-checked with a container-scoped query and stands: that curve genuinely draws no markers, its points being dense enough for Highcharts to suppress them.)

## Testing

- 7 tests in `test/adapters/highcharts/pareto.test.ts`, including an explicit assertion that the values are **not** the running total.
- 5 mutations killed: pareto declined again, read as a `smooth`, named by the wrong axis, `bellcurve` read as a `line`, and a valueless step kept.
- Full gate green: `eslint`, `tsc --noEmit`, `npm run build`, `npm test` — 398 suites / 5503 tests, plus the ESM project at 10 / 137.
- `docs/highcharts.md` gains the row and a note; `examples/highcharts-pareto.html` is a working page whose exact configuration was verified in Chromium.

Two of #1138's four judgement cases remain: `timeline` and `organization`.

---


jooyoungseo:
The reversed-axis point was right, and it is not only a hypothetical — I went and measured it rather than reasoning about it, and it is a live defect. Fixed in d68de407.

Highcharts 11.4.8 in Chromium, the same four causes, once plain and once with `xAxis.reversed`, read through the adapter itself:

```
plain      bar layer  A B C D  (80, 60, 40, 20)   curve  A B C D  (40, 70, 90, 100)
reversed   bar layer  D C B A  (20, 40, 60, 80)   curve  A B C D  (40, 70, 90, 100)
```

Highcharts still accumulates in declared order, and the graph path's vertices come out in that order too (`M 406.875 … L 58.125 …`, i.e. right to left). So on a reversed axis one chart's two layers announced its categories in opposite orders: a reader moving forward through the columns was moving backwards through the cumulative percentage drawn over them. `convertParetoSeries` now calls `drawsSeriesReversed` and emits `domMapping.pointOrder` the way `convertLineSeries` does.

Three tests added (10 in the file now), and three mutations against the new branch all die:

```
baseline                             17 passed
never re-paired                       3 failed
always re-paired                      4 failed
payload reversed, path not told       1 failed
```

Full gate re-run green: eslint clean, `tsc --noEmit` clean, `npm run build` complete, `npm test` → 398 suites / 5506 tests, plus the ESM project at 10 / 137.

On the sibling you noted: `convertBellCurveSeries` really does have the same gap, and so does `convertScatterSeries`. Measured on a `bellcurve` over a reversed axis, the payload comes out 2.46 → 4.18 while the chart draws it 4.18 → 2.46. That is already-merged code rather than this PR's, so I am filing it separately rather than widening this one.

---

PR #1152, opened 2026-08-24
feat(highcharts): read a timeline as the labelled scatter it draws

jooyoungseo:
A Highcharts `timeline` series is in none of `buildSubplot`'s type buckets, so it reached `convertSeries` as an unsupported type and was declined — and the whole chart came out unread. A timeline's entire payload is *which events happened, in what order*, which is exactly what `ScatterPoint.label` is for.

Closes part of #1138. Third of that issue's four judgement cases; `organization` is the one still open.

## `y` is a constant

This is the fact the reading is built around, and it was measured rather than assumed. Highcharts 11.4.8 in Chromium, four events and three, dated and undated alike:

```
point.y     1, 1, 1, 1
point.x     0, 1, 2, 3        no x declared, x axis hidden
point.x     1579046400000 …   with xAxis.type: 'datetime'
```

So the layer carries the 1 the chart drew. **Sonifying it plays one pitch for the whole row**, which is what a chart with no magnitude sounds like. The two alternatives are worse: pitching by `x` would announce a magnitude the chart does not draw, and refusing the layer leaves the reader with nothing. What they actually get is the *sequence* — `ScatterTrace` pans by distinct x value, so moving through the events sweeps the stereo field — and the *identity* of each event.

`x` is whatever the author declared. Nothing converts a bare index into a date: on the undated timeline the dates live in the event's own text, which is where the chart draws them.

## Each event's box draws two strings

Measured on an event declaring both:

```
data label text   "● First artificial satellite" ⟨zero-width space⟩ "1957 Sputnik 1"
```

`name` on the first line, `label` on the second. Neither alone is the box, and on the ordinary timeline — x axis hidden, `x` a bare 0, 1, 2, 3 — `label` is the only place the *date* appears at all. So both are joined into the one string the grammar has room for, and a repetition is said once.

**`description` is dropped.** It is the tooltip's text rather than anything drawn, and the grammar has one string per point.

## The markers needed no new selector

Measured: a timeline's markers carry `highcharts-point` inside a `highcharts-markers` group that is a *sibling* of the series group but carries the same `highcharts-series-N` class. So the existing `scatterSelector` resolves to exactly one element per point, in point order, each identical to that point's own `graphic` — verified on the four-event and three-event charts and again on the example page's own five.

## Falsification

Six mutations against the new code, all killed:

```
baseline                                9 passed
timeline declined again                 9 failed
only the headline named                 2 failed
a repeated string said twice            1 failed
the position pitched as a magnitude     1 failed
a valueless event kept                  1 failed
axes swapped                            1 failed
```

## Gate

`npx eslint src test` clean, `tsc --noEmit` clean, `npm run build` complete, `npm test` → **399 suites / 5515 tests**, plus the ESM project at 10 / 137.

## Also here

- `docs/highcharts.md`: a table row and a note carrying the constant-`y` reasoning.
- `examples/highcharts-timeline.html`: a working example, its exact configuration run through the adapter in Chromium (five events, five selector hits, labels matching the drawn boxes).
- `HighchartsPoint.label` added to the adapter's types, documented as the timeline's second line.

---


jooyoungseo:
Thanks — all five are notes rather than requests, and I agree with each. Two are worth answering on the record.

**On `HighchartsPoint.label` being a reused key.** You are right that this is the one assumption here that could age badly, so I made the JSDoc say so at the declaration rather than leaving it implicit. It is also narrower than it looks in practice: the field is read in exactly one place, `timelineEventLabel`, which is reached only from `case 'timeline'`. A future series type that starts declaring `point.label` for something else would not be affected unless it were also routed through this converter, which would be a deliberate act. If that day comes, the fix is a type-scoped read rather than a wider one.

**On the comma-joined label.** It is not only a grammar constraint — it is what the chart draws. Measured on an event declaring both strings, the rendered data-label box reads

```
"● First artificial satellite" ⟨zero-width space⟩ "1957 Sputnik 1"
```

so `name` and `label` are two lines of *one* box, and either alone is half of it. The case that decides it is the ordinary undated timeline, where `xAxis` is hidden and `x` is a bare 0, 1, 2, 3: there `label` is the only place the **date** appears anywhere in the payload. Joining keeps that whole and costs a dated timeline one repetition — which is itself deduplicated when the two strings are identical.

The `O(n²)` dedup is over a fixed two-element array, as you note; I kept `indexOf` because a `Set` there reads as though the length were unbounded.

E2E is the last check still running. I will merge once it is green.

---

Issue #1153, opened 2026-08-24
A tree with no magnitude cannot be expressed: TreemapTrace announces a value every node lacks


jooyoungseo:
`TraceType.TREEMAP` is the grammar's only hierarchy, and `TreemapPoint.y` is documented as optional — "omitted for an interior node whose value is the sum of its children, which is the ordinary case". But a tree in which *no* node has a value is not expressible: `TreemapTrace` announces a value clause and computes shares from it either way, and both possible spellings announce something the chart does not draw.

Found while working #1138's `organization` case. A Highcharts organization chart is a pure hierarchy — measured on a six-node chart, every node's `value` and `weight` came back `null` and Highcharts' own internal `sum` was `1` for every node alike, because it assigns one unit per link for layout. There is no magnitude anywhere in the declaration.

## Measured

Both spellings, run through `TraceFactory.create` on a six-node org tree (Ada → Bo, Cy; Bo → Engineering, Design; Cy → Finance).

**Omitting `y` on every node:**

```
audio.freq   { min: 0, max: 0, raw: 0 }
text.cross   { label: 'Reports', value: 0 }
text.asides  [ { label: 'Children', value: '2' } ]
```

Every node announces `0` — a number the chart does not draw — and `min === max === 0` leaves no pitch range, so the sonification is flat across the whole tree.

**Declaring `y: 1` on every node** (Highcharts' own per-link unit):

```
audio.freq   { min: 1, max: 1, raw: 1 }
text.cross   { label: 'Reports', value: 1 }
text.asides  [ { label: 'Share of Ada', value: '100.0%' },
               { label: 'Children', value: '2' } ]
```

Strictly worse. Still no pitch range, still a false number, and now **Bo is announced as 100% of Ada — and so is Cy**. Two siblings each reported as the whole of their parent. The share clause is derived from `node.value / parent.value`, and with a declared 1 the parent's derived total is overridden by its own declared 1 rather than summing to 2.

The `Share of ...` clause is already guarded against a zero parent (`parent !== null && parent.value !== 0`), which is why the first spelling suppresses it. Nothing guards the `cross` clause, which is unconditional:

```ts
return {
  main: { label: this.xAxis, value: node.name },
  cross: { label: this.yAxis, value: node.value },
  ...
};
```

## Why it matters

[...]

PR #1154, opened 2026-08-24
docs(highcharts): write down the five types read and the five declined

jooyoungseo:
The Highcharts guide's type coverage had drifted behind the adapter **in both directions**, and a sweep looking for what is left to do would have re-derived both halves.

Closes the last of #1138's four judgement cases — `organization` — by documenting the decline and the reason, rather than leaving it silent.

## Five types are read and were not listed

| Type | Reads as | Added by |
| --- | --- | --- |
| `columnpyramid`, `pictorial` | Bar | #1139 |
| `streamgraph` | Stacked Area | #1139 |
| `bubble` | Scatter | #1139 |
| `arearange`, `areasplinerange`, `columnrange` | Error Bar | #1047 |

All four rows were extended rather than added, since each type is a spelling of a reading already in the table, and a short note explains the two that involve a judgement: `columnpyramid` and `pictorial` are a column drawn differently and carry the same `point.y` per category; `bubble` reads as a scatter rather than a dot plot, because a dot plot is one value per category and a bubble has two, and the category name is not lost either way — it travels as `xLabel`.

## Five types are declined and the reasons lived only in issue threads

A new **Series types that are deliberately not read** section, so the next sweep finds the reasons rather than re-deriving them:

- `venn` — magnitudes are overlap areas, positions are a layout, no axis to announce against.
- `polygon` — an arbitrary closed shape with no statistical reading.
- `vector`, `windbarb`, `flags` — direction and annotation rather than a measured series.
- `packedbubble` — it has a magnitude, but the positions are a packing; announcing x and y would report where the layout put it.
- `organization` — blocked on the core rather than the adapter.

## Why `organization` is the interesting one

I set out to implement it. The judgement recorded on #1138 was that "the tree is expressible without inventing a value", and **that turned out to be wrong** — I only found out by driving a valueless tree through `TreemapTrace` instead of reasoning about the grammar.

The hierarchy itself is exact and readable. Measured in Chromium on a six-node chart: `series.nodes` carries each node's `id`, `name`, `column` and `linksTo`; `series.data` is the links rather than the nodes; and each node draws its own `path.highcharts-node.highcharts-point` inside the series group in `series.nodes` order, so a selector is straightforward and needs no index stamping.

What is missing is a magnitude. Every node's `value` and `weight` came back `null`, and Highcharts' own internal `sum` was `1` for every node alike, because it assigns one unit per link for layout. There is no number anywhere in the declaration.

`TraceType.TREEMAP` cannot express that. Both spellings, through `TraceFactory.create`:

```
omitting y      freq { min: 0, max: 0, raw: 0 }
                cross { label: 'Reports', value: 0 }

declaring y: 1  freq { min: 1, max: 1, raw: 1 }
                cross { label: 'Reports', value: 1 }
                asides [ { label: 'Share of Ada', value: '100.0%' }, … ]
```

The second is strictly worse: **Bo is announced as 100% of Ada — and so is Cy**, two siblings each reported as the whole of their parent, because a declared parent value overrides the derived total that would have summed to 2. Either way every node states a magnitude the chart does not draw, which is the objection this guide already makes about emitting `count` for a variwide.

So this is declined pending #1153 — giving the tree a way to say it has no magnitude — rather than pending anything in the adapter. Worth saying plainly in the note, and it is: an org chart read as a tree would be a real improvement on the nothing it gets today, since navigation, the ancestry breadcrumb and the child count are all correct and useful. One further guard is recorded for whoever picks it up: `TreemapPoint.path` is *one* ancestry, so a node with two parents cannot be expressed.

## Gate

`npm run docs` builds clean. No source change, so nothing else is touched.

---


jooyoungseo:
You were right, and I should not have written that line without checking. Fixed in 9e8c4cf3.

I checked against the packaged files rather than the documentation — `npm pack highcharts@11`, untar, then grep each bundle for the registered type name:

```
highcharts.js          columnpyramid  0
highcharts-more.js     columnpyramid  1
modules/pyramid3d.js   columnpyramid  0
modules/pictorial.js   columnpyramid  0
modules/funnel.js      columnpyramid  0
```

So `highcharts-more.js` alone, exactly as you suspected. `pyramid3d` is the unrelated 3D series and never mentions it.

Since one unchecked claim in this PR means the others deserve the same treatment, I ran all five the same way:

| type | registered in |
| --- | --- |
| `pictorial` | `modules/pictorial.js` |
| `streamgraph` | `modules/streamgraph.js` |
| `bubble` | `highcharts-more.js` |
| `arearange`, `areasplinerange` | `highcharts-more.js` |
| `columnrange` | `highcharts-more.js` |

All hold as written. `columnrange` also appears once in `highcharts.js`, but as a reference rather than a registration — the type is defined in `highcharts-more.js`, which is what the row says.

---

PR #1157, opened 2026-08-24
feat(highcharts): read an organization chart as the hierarchy it draws

jooyoungseo:
Closes the `organization` case of #1138, which #1153 unblocked.

`buildSubplot` sorts series into buckets by type and `organization` was in none of them, so it reached `convertSeries` as an unsupported type and was declined. On a chart that is only the org tree, that is silence.

## Why it was declined, and what changed

The decline was deliberate. An organization chart is a **pure hierarchy** — measured on a six-node chart in Highcharts 11 plus `modules/sankey.js` and `modules/organization.js`:

| what was checked | what came back |
| --- | --- |
| `node.value` | absent on every node |
| Highcharts' internal `node.sum` | `1` on every node alike — one unit per link, for layout |

There is no magnitude anywhere in the declaration, and both available spellings of that were wrong:

- **omitting `y`** announced `0` on every node, over a `freq { min: 0, max: 0 }` that leaves no pitch range;
- **declaring the layout's `1`** was worse — Bo announced as 100% of Ada, and Cy announced as 100% of Ada as well.

#1156 taught `TreemapTrace` to recognise a tree that declares no magnitude anywhere. This is the adapter half: the payload declares no `y`, names no second axis, and the reader gets what the chart actually has — moving up to a manager, down to a report, across to a peer, the ancestry, and how many people report to whoever the cursor is on.

## Measured

The reading, end to end in Chromium against a real chart:

```
type    treemap
axes    { x: { label: 'Node' } }          ← no y

x                    path
Ada Lovelace, CEO    []
Bo Turing, CTO       ['Ada Lovelace, CEO']
Cy, CFO              ['Ada Lovelace, CEO']
Engineering          ['Ada Lovelace, CEO', 'Bo Turing, CTO']
Design               ['Ada Lovelace, CEO', 'Bo Turing, CTO']
Finance              ['Ada Lovelace, CEO', 'Cy, CFO']
```

and each of the six selectors matched **exactly one `path.highcharts-node`** — the drawn box.

## Reading decisions

**The structure comes from `series.nodes`, not from the links.** An organization series is declared as `from`/`to` pairs; Highcharts resolves them into node objects, and that is the only place `linksTo`, the display `name`, the `title` drawn under it, and the rendered box can all be reached.

**A node's label joins its name and its title,** because that is what the box says, and either alone loses half of what a sighted reader gets. A name that matches the title is not repeated, and a node declaring neither falls back to the id, which is what Highcharts names it.

**A node with more than one parent declines the whole series.** `TreemapPoint.path` is one ancestry, so a tree cannot say that someone reports to two managers, and reading it as one would drop a link the chart plainly draws. A silently missing edge is worse than the fallback, which at least says what it is — the same line `qqline` is held to in xability/r-maidr#251.

**A cycle is cut with a warning**, as it already is for a treemap: nothing stops an organization link pointing back up.

## Verification

- 13 tests in `test/adapters/highcharts/organization.test.ts` — the reading, the absent `y`, the absent second axis, the label rules, the paths, the selectors, the stamping, and the three declines (multi-parent, cycle, no nodes).
- Mutation sweep over the dispatch case, the multi-parent guard, the parent walk, the label join, the dedupe, the cycle seed, the axes, the stamping, the empty guard, and a `y: 1` reintroduction: **10 of 10 killed.** One would-be survivor was resolved by adding the case it guarded (a name equal to a title) rather than by deleting the guard.
- `npx eslint src test` clean · `npx tsc --noEmit` clean · `npm run build` succeeds · `npm run docs` succeeds · `npm test`: **401 suites / 5543 tests, plus the ESM project's 10 / 137**, all passing.

`docs/highcharts.md` moves `organization` out of "Series types that are deliberately not read" and into the supported table, and the note there is rewritten from "why it is blocked" to what the reading does and what it declines.

---

### DIFF
--- a/docs/highcharts.md
+++ b/docs/highcharts.md
@@ -119,6 +119,7 @@ import { createHighchartsSync, highchartsToMaidr } from 'maidr/highcharts';
 | Network | `networkgraph` (requires `modules/networkgraph.js`) | [highcharts-network.html](examples/highcharts-network.html) |
 | Treemap | `treemap` (requires `modules/treemap.js`) | [highcharts-treemap.html](examples/highcharts-treemap.html) |
 | Sunburst | `sunburst` (requires `modules/sunburst.js`, which ships treemap) | [highcharts-sunburst.html](examples/highcharts-sunburst.html) |
+| Organization | `organization` (requires `modules/sankey.js` and `modules/organization.js`), read as a tree with no magnitude | — |
 | Gauge | `gauge`, `solidgauge` (require `highcharts-more.js`; solid gauge also `modules/solid-gauge.js`), `bullet` (requires `modules/bullet.js`) | [highcharts-gauge.html](examples/highcharts-gauge.html) |
 | Waterfall | `waterfall` (requires `highcharts-more.js`) | [highcharts-waterfall.html](examples/highcharts-waterfall.html) |
 | Error Bar | `errorbar` (requires `highcharts-more.js`), reading its estimates from the series it is `linkedTo`; `arearange`, `areasplinerange`, `columnrange` (require `highcharts-more.js`), each read as the band of intervals it draws | [highcharts-errorbar.html](examples/highcharts-errorbar.html) |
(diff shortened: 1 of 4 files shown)


jooyoungseo:
Thanks — two of the three are now fixed on the branch, and the third I'm declining with a reason.

**Finding 3 (numeric node ids) was real, and it reproduces.** Measured in Chromium against a real chart declared as `data: [[1, 2], [1, 3], [2, 4]]`: Highcharts builds nodes with numeric ids and draws the tree correctly, while the reading came back

```
'Ada' | path []
'Bo'  | path []
'Cy'  | path []
'Eng' | path []
```

— four roots, the whole hierarchy silently gone. Ids are now keyed and compared as `String(id)` on both sides, and `HighchartsNode.id` is typed `string | number`, which is what the library actually hands over. Two mutations pin it: dropping the coercion in the walk and dropping it in the map both fail the new test.

**Finding 2 (the cast) is fixed.** `stampPointIndices` now takes `{ graphic?: { element: SVGElement } }[][]` — the one field it touches — so both callers pass their real types and the `as unknown as` is gone. You were right that nothing was keeping the two interfaces in step.

**Finding 1 (the duplicated walk) I'm leaving as it is,** and I want to say why rather than just not doing it. The two walks line up on the shape but differ on five things: the parent field, the label function, whether the id can be absent, the return type (`(string | number)[]` against `string[]`), and the wording of the cycle warning. A shared helper would need a parent accessor, a label function, an id accessor and the series name — four parameters and a generic — and both call sites would still need a small adapter each. That trades thirty lines of two things each readable on its own for one thing that is harder to read than either, which is the wrong side of `CLAUDE.md`'s "readable beats clever". If a third tree type arrives the calculation changes, and it should be revisited then.

Full gate re-run after the fixes: eslint clean, `tsc --noEmit` clean, `npm run build` succeeds, `npm test` at 401 suites / 5544 tests plus the ESM project's 10 / 137.

---

Issue #1162, opened 2026-08-24
A Vega-Lite regression or loess curve is announced as a line, where every other adapter calls a fitted curve a smooth


jooyoungseo:
The adapter sweep continued to Vega-Lite. Its primitive marks are all covered — `bar`, `line`, `trail`, `area`, `point`, `circle`, `square`, `tick`, `rule`, `text`, `rect`, `arc`, `geoshape`, plus the composites `boxplot`, `errorbar`, `errorband`. The gap is not a mark. It is the three **statistical transforms**, which change what a mark means and are read as though they had not.

## Measured

Driven through `vega-lite@5.23` → `vega@5.33` with a **real `View`**, which is what `bindVegaLite` passes, then through `vegaLiteToMaidr`. Twenty-four points, y alternating low/high so a smoother is visible:

```
raw y            [0,11,2,13,4,15,6,17]  n=24

plain line       type=line  n=24  y=[0,11,2,13,4,15,6,17]
loess            type=line  n=24  y=[3.04,4.82,6.62,7.84,9.12,9.87,11.13,11.87]
regression       type=line  n=2   y=[4.4,28.6]
density          type=area  n=201 y=[0.01,0.01,0.01,…]
```

**The data is right in every case.** The loess layer carries the smoothed curve, not the observations; the regression layer carries the fitted line's two endpoints; the density layer carries the 201-point estimate. Nothing is being mis-extracted.

What is wrong is the **name**.

## Why `smooth` is the name, and this is not a new opinion

`TraceType.SMOOTH` exists for a fitted curve, and every other place in the tree that reads one uses it:

| producer | fitted curve | read as |
|---|---|---|
| Observable Plot | `linearRegressionY` / `linearRegressionX` | **smooth** |
| Highcharts | `bellcurve` | **smooth** (settled on #1138) |
| plotly (py-maidr) | `trendline` | **smooth** |
| ggplot2 (r-maidr) | `geom_smooth`, `stat_smooth` | **smooth** |
| **Vega-Lite** | `regression`, `loess` | **line** |

`src/adapters/observable/index.ts` says it outright: "`linearRegressionX` / `linearRegressionY` (the fitted line, as a smooth)". Four producers agree and one does not, so this is the cross-adapter inconsistency the project treats as a defect — the same argument that "the two adapters describe one chart the same way" carries elsewhere.

[...]

PR #1163, opened 2026-08-24
feat(vegalite): name a regression or loess curve a smooth, not a line

jooyoungseo:
Closes the first half of #1162.

## The reading that was wrong

A Vega-Lite `line` or `trail` mark whose rows have been replaced by a
`regression` or `loess` transform was announced as a **line**. It is a fitted
curve: the points on the page are not observations, they are a model evaluated
at a grid. `TraceType.SMOOTH` is the name every other binding already gives
that — py-maidr emits it for `seaborn.regplot`/`lmplot`, and the grammar
describes it as a fitted curve rather than a series of readings.

Announcing it as a line tells the reader the marks are the data. They are not,
and the count alone gives it away: a `regression` fit over 24 rows comes back
as **2** points.

## Measured, through a real compiled view

```
raw y        [0,11,2,13,4,15,6,17]  n=24
plain line   type=line   n=24  y=[0,11,2,13,4,15,6,17]
loess        type=smooth n=24  y=[3.04,4.82,6.62,7.84,9.12,9.87,11.13,11.87]
regression   type=smooth n=2   y=[4.4,28.6]
density      type=area   n=201 y=[0.01,…]
```

`loess` keeps the row count and replaces every value; `regression` collapses to
the two endpoints of the fitted line. Either way the rows on the page are a
fit. The `line` control is unchanged, and `density` — which is an `area` mark —
is untouched by this change.

### A false lead, recorded so it is not chased again

My first pass called `vegaLiteToMaidr(spec, 'probe')` — passing a string where
the **`View`** goes. With no view the converter falls back to the spec's
`data.values`, so every transformed chart appeared to announce its raw
observations, which would have been a much more serious bug than this one.
**It is not real**: with a live view, `view.data(…)` returns the post-transform
rows and every reading above is correct. This is written up in #1162 as well.

## The change

- `hasFittedCurveTransform()` in `src/adapters/vegalite/converters.ts`, next to
  the existing `hasRunningSumTransform()` it is modelled on. It checks for a
  `regression` or a `loess` entry, and deliberately not `density` — a density
  is its own distribution estimate drawn as an `area`, not a fit through
  observed points, and it is a question the issue leaves to the maintainer.
- The check runs in the `line`/`trail` branch **after** the three existing ones
  (they are about *what was plotted*; this one is about the rows having been
  replaced by a fit).
- `TraceType.SMOOTH` added to the extraction dispatch's line-family case group.
  Leaving it out was measured to make the layer **vanish** (`type=undefined
  n=0`) — the default branch returns null — so this is not cosmetic and a test
  pins it.
- `docs/vegalite.md` gains the row for it.

## Tests

`test/adapters/vegalite/fittedCurve.test.ts`, 7 cases, driven through
`makeView({ data_0: rows })` so the rows arrive the way a compiled view
delivers them. One asserts the smooth layer **still carries the fitted rows and
does not vanish**, which is the failure the dispatch omission produced.

Five mutations, all killed:

```
KILLED   regression dropped from the check
KILLED   loess dropped from the check
KILLED   density folded in
KILLED   SMOOTH dropped from the dispatch
KILLED   check never fires
```

## Verification

```
npx eslint src test    clean
npx tsc --noEmit       clean
npm run build          All builds complete in 105.6s
npm run docs           Done!
npm test               406 suites / 5591 tests, ESM 10 / 137
```

---


jooyoungseo:
Both review points were right; both are fixed in `7869041c`.

**The displaced JSDoc was a real bug.** Inserting `hasFittedCurveTransform()` immediately before `hasRunningSumTransform()` put it between that function and its own doc block, so the running-sum explanation ended up stacked above the new function and `hasRunningSumTransform` was left undocumented. The block moves back down onto its function; nothing else in the file changes.

**The table row now matches the format.** It reads `| Smooth | — |`, four columns like every other row. The explanation moved into the paragraph below the table that already covers the charts recognised from the transform that built them — bump, parallel coordinates, ridgeline, diverging bar — which this is the fifth of, so "Four more charts" is now "Five more". That paragraph is also where the `density` exclusion is now stated for a reader of the guide rather than only for a reader of the source.

Re-verified after the fixes:

```
npx eslint src test    clean
npx tsc --noEmit       clean
npm run build          All builds complete in 114.1s
npm run docs           Done!
npm test               406 suites / 5591 tests, ESM 10 / 137
```

---

Issue #1140, opened 2026-08-23
Three amCharts hierarchy series are not read, though the adapter already reads their data shape


jooyoungseo:
> **Corrected.** This issue was first filed as "four amCharts series classes are announced as bar charts". **That was wrong**, and the original text is kept at the bottom. I measured `classifySeriesKind` in isolation and generalised from it without checking whether the paths that reach it are reachable. They are not: both entry points filter these classes out first, so nothing is announced as a bar chart. The real gap is smaller and different, and is what follows.

The same sweep as #1138, run against amCharts.

## Measured, end to end this time

Each class driven through `fromAmCharts` by both routes a series can arrive by — pushed straight into a container (the am5hierarchy pattern), and inside an `XYChart`'s series list. `Treemap` is the control:

```
                    standalone container         inside an XYChart
Tree                throws "no XYChart or Pie"    throws "no supported series"
Pack                throws "no XYChart or Pie"    throws "no supported series"
Venn                throws "no XYChart or Pie"    throws "no supported series"
LinkedHierarchy     throws "no XYChart or Pie"    throws "no supported series"

Treemap             treemap                       treemap        ← control
```

Both throws are the binder's documented contract (`@throws If no supported chart is found …`), so this is the adapter declining, not misreading. **No wrong announcement occurs.**

## The gap that is real

**`Tree`, `Pack` and `LinkedHierarchy` are hierarchies the adapter already knows how to read.** They carry the same nested `children` shape as `Treemap`, `Partition` and `Sunburst`, and `extractHierarchyNodes` / `extractHierarchyPoints` read exactly that shape. The dispatch shows how little separates them:

```ts
case 'treemap':
case 'icicle':
case 'sunburst': {
  const data = extractHierarchyPoints(series);
```

One branch, one converter. What keeps the three out is `STANDALONE_SERIES_CLASSES`, derived from `STANDALONE_KINDS`, which gates discovery: a class not in it is never wrapped as a panel, so it never reaches the converter that could read it.

So this is the same shape as #1138's Highcharts finding — a chart MAIDR already reads, under a name the adapter does not know — reached through a discovery gate rather than a switch.

[...]

PR #1165, opened 2026-08-24
feat(highcharts): read a treegraph and a packed bubble as the hierarchies they are

### LINKED ISSUE #1164: Two Highcharts hierarchy series emit no layer, and both now have a name in the grammar
The same shape as #1140, on the Highcharts side. `treemap`, `sunburst` and `organization` are read; `treegraph` and `packedbubble` are not, and both are hierarchies MAIDR has since gained a name for.

## Measured

Highcharts 13.0.1 in jsdom, with `modules/treegraph.js` and `highcharts-more.js`, `treemap` and `bubble` as controls:

```
treegraph      [MAIDR Highcharts] Unsupported series type: "treegraph"; skipping.     layers: (none)
packedbubble   [MAIDR Highcharts] Unsupported series type: "packedbubble"; skipping.  layers: (none)
treemap        layers: treemap n=5                                                    <- control
bubble         layers: point n=2                                                      <- control
```

Both fall through the series-type switch. That is the adapter declining, not misreading — the same shape #1140 had on amCharts, reached through a switch rather than a discovery gate.

## The names exist now

- `TraceType.TREE` (#1158) — a node-link hierarchy, already what `organization` emits.
- `TraceType.PACK` (#1159) — circle packing, added for amCharts' `Pack`.

The naming argument is the one those made: the trace type is what the reader is *told* is on the page. A node-link diagram announced as a treemap names a painting nobody drew, and so does a cluster of circles.

## `treegraph` — same declaration, different magnitude

A treegraph declares exactly what a treemap does — `id`, `parent`, `name` — so the walk up the parent chain is the treemap's.

What differs is where the magnitude comes from. A treegraph's layout sizes nothing by value, and Highcharts fills the field in anyway:

```
treegraph, no values      point.value 0      point.options.value undefined
treegraph, values 1..5    point.value 1..5   point.options.value 1..5
```

Reading `point.value` would emit `y: 0` on every node and announce a magnitude of zero for a chart that has none — #1153's defect, and the reason `convertOrganizationSeries` exists rather than reusing the treemap path. The author's declaration separates the two.

**A treemap must keep reading the computed field.** There it is not an artefact: an interior node with no declared value comes back carrying the sum of its children (measured, 12 and 8 on a leaves-only tree), which is a real total the rectangles are sized by. And a treemap with no values anywhere renders **zero** nodes, so the case this guards against cannot arise there.

One more thing worth recording: Highcharts classes a treegraph's **link paths** `highcharts-point` as well as its nodes — measured, 9 elements for 5 nodes. A selector list indexed by document position would address the links. The existing `data-maidr-node-index` stamp goes on each point's own `graphic`, so it is unaffected, but it is load-bearing here in a way it is not for a treemap.

## `packedbubble` — was deliberately declined, and the reason is now answered

It is currently listed in `docs/highcharts.md` under "Series types that are deliberately not read":

> `packedbubble` — It has a magnitude, but the positions are a packing rather than data — announcing a bubble's x and y would report where the layout put it.

That was right about the chart. It is answered by the grammar as it now stands: `pack` is a trace whose points carry a magnitude and **no coordinate at all**, so nothing announces where the layout put a bubble. `test/util/orientation.test.ts` already files `PACK` under the unoriented types for the same reason — "a node sits inside its parent rather than along an axis".

Its shape:

```
series "Europe"  points  Germany value=12   France value=8    parent undefined
series "Asia"    points  Japan   value=15   India  value=20   parent undefined
```

Nothing declares a `parent`; the grouping is *which series* a bubble is in, and the adapter already gives each series its own layer named after it. So the paths come out empty and each layer is the flat pack the chart draws.

(Its legend swatches also carry `highcharts-point`, but they sit outside `.highcharts-series-group`, which the selectors are scoped to.)

## Not in scope

`venn`, `polygon`, `vector`, `windbarb` and `flags` stay declined, with the reasons already recorded in the docs. Those are about charts with no per-mark magnitude to walk, which is a different argument from the one `packedbubble` was declined under.

jooyoungseo:
Closes #1164. The same shape as #1140, on the Highcharts side.

## Measured

Highcharts 13.0.1 with `modules/treegraph.js` and `highcharts-more.js`, `treemap` and `bubble` as controls:

```
treegraph      [MAIDR Highcharts] Unsupported series type: "treegraph"; skipping.     layers: (none)
packedbubble   [MAIDR Highcharts] Unsupported series type: "packedbubble"; skipping.  layers: (none)
treemap        layers: treemap n=5                                                    <- control
bubble         layers: point n=2                                                      <- control
```

Both fell through the series-type switch. That is the adapter declining, not misreading.

## The names exist now

`TraceType.TREE` (#1158) and `TraceType.PACK` (#1159). The naming argument is the one those made: the trace type is what the reader is *told* is on the page. A node-link diagram announced as a treemap names a painting nobody drew, and so does a cluster of circles.

## `treegraph` — same declaration, different magnitude

It declares exactly what a treemap does (`id`, `parent`, `name`), so the walk up the parent chain is the treemap's and this delegates to it. What differs is where the magnitude comes from:

```
treegraph, no values      point.value 0      point.options.value undefined
treegraph, values 1..5    point.value 1..5   point.options.value 1..5
```

A treegraph's layout sizes nothing by value and Highcharts fills the field in regardless. Reading `point.value` would emit `y: 0` on every node and announce a magnitude of zero for a chart that has none — #1153's defect, and the reason `convertOrganizationSeries` exists rather than reusing the treemap path. The author's declaration is read instead; with nothing declared anywhere, no node carries a `y` and the layer names no value axis either.

**A treemap keeps reading the computed field** and is untouched by this. There it is not an artefact: an interior node with no declared value comes back carrying the sum of its children (measured, 12 and 8 on a leaves-only tree), which is a real total, and a treemap with no values anywhere renders **zero** nodes — so the case this guards against cannot arise there. A test pins that separately.

One more measured fact worth recording: Highcharts classes a treegraph's **link paths** `highcharts-point` as well as its nodes — 9 elements for 5 nodes. The existing `data-maidr-node-index` stamp goes on each point's own `graphic` so it is unaffected, but it is load-bearing here in a way it is not for a treemap, and a test says so.

## `packedbubble` — a recorded decline, answered rather than overruled

`docs/highcharts.md` listed it under "Series types that are deliberately not read":

> It has a magnitude, but the positions are a packing rather than data — announcing a bubble's x and y would report where the layout put it.

That was right about the chart, and the grammar has since answered it: `pack` is a trace whose points carry a magnitude and **no coordinate at all**, so nothing announces where the layout put a bubble. (`test/util/orientation.test.ts` already files `PACK` under the unoriented types for the same reason.) The docs now carry the reversal and why, rather than dropping the row.

Its shape:

```
series "Europe"  Germany value=12   France value=8    parent undefined
series "Asia"    Japan   value=15   India  value=20   parent undefined
```

Nothing declares a `parent` — the grouping is which *series* a bubble is in, and the adapter already gives each series its own layer named after it. So the paths come out empty and each layer is the flat pack the chart draws, which is why it shares the treemap converter rather than getting one of its own.

Its legend swatches also carry `highcharts-point`, but they sit outside `.highcharts-series-group`, which the selectors are scoped to.

## Reading, after

```
treegraph (no values)     type=tree  axes={"x":{"label":"Node"}}                       y=(none) on all five
treegraph (all values)    type=tree  axes={...,"y":{"label":"Value"}}                  y=1..5
treegraph (leaves only)   type=tree  axes={...,"y":{"label":"Value"}}                  y on leaves, none on interiors
packedbubble              type=pack  titles ["Europe","Asia"]                          12/8 and 15/20
treemap (control)         type=treemap                                                 unchanged
treemap leaves only       type=treemap                                                 interior totals 12 and 8 kept
sunburst (control)        type=sunburst                                                unchanged
```

## Mutations

Nine, all killed:

```
KILLED   treegraph branch removed
KILLED   packedbubble branch removed
KILLED   treegraph named TREEMAP
KILLED   packedbubble named TREEMAP
KILLED   treegraph reads the computed value
KILLED   value axis always named
KILLED   value axis never named
KILLED   undeclared node keeps the computed zero
KILLED   declared value never applied
```

## Verification

```
npx eslint src test    clean
npx tsc --noEmit       clean
npm run build          All builds complete in 134.2s
npm run docs           Done!
npm test               407 suites / 5598 tests, ESM 10 / 137
```

(406/5591 before, so the new suite is the +1/+7.)

---

### DIFF
--- a/docs/highcharts.md
+++ b/docs/highcharts.md
@@ -120,6 +120,8 @@ import { createHighchartsSync, highchartsToMaidr } from 'maidr/highcharts';
 | Treemap | `treemap` (requires `modules/treemap.js`) | [highcharts-treemap.html](examples/highcharts-treemap.html) |
 | Sunburst | `sunburst` (requires `modules/sunburst.js`, which ships treemap) | [highcharts-sunburst.html](examples/highcharts-sunburst.html) |
 | Tree | `organization` (requires `modules/sankey.js` and `modules/organization.js`), read as a hierarchy with no magnitude | — |
+| Tree | `treegraph` (requires `modules/treegraph.js`, which ships treemap) | — |
+| Pack | `packedbubble` (requires `highcharts-more.js`) | — |
 | Gauge | `gauge`, `solidgauge` (require `highcharts-more.js`; solid gauge also `modules/solid-gauge.js`), `bullet` (requires `modules/bullet.js`) | [highcharts-gauge.html](examples/highcharts-gauge.html) |
 | Waterfall | `waterfall` (requires `highcharts-more.js`) | [highcharts-waterfall.html](examples/highcharts-waterfall.html) |
 | Error Bar | `errorbar` (requires `highcharts-more.js`), reading its estimates from the series it is `linkedTo`; `arearange`, `areasplinerange`, `columnrange` (require `highcharts-more.js`), each read as the band of intervals it draws | [highcharts-errorbar.html](examples/highcharts-errorbar.html) |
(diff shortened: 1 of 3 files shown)


jooyoungseo:
Added the explicit-zero case in `d4d8f971`. It earns its place — I checked rather than assuming.

Mutating the guard from `typeof value === 'number'` to plain truthiness, **with** the new case:

```
✕ keeps a declared zero, which is not the layout's zero
Tests: 1 failed, 7 passed, 8 total
```

**Without** it, that mutation survived every assertion in the file. So the distinction the code draws was untested, exactly as you said. The case declares zero on all five nodes and asserts both halves: every node keeps `y: 0`, and the value axis *is* named — where the valueless chart, whose `point.value` fields are identical, keeps neither.

On the other two points:

- **The discarded `y`** — agreed it is wasted work per node, and agreed it is not worth restructuring. `convertTreeSeries` owns the parent-chain walk and the stamping, and pulling the magnitude out into a parameter would spread one decision across two functions to save an object spread. Left as is, deliberately.
- **The missing example links** — matching the `organization` row was the intent, so no change.

Re-verified after the addition: `npm test -- test/adapters/highcharts/hierarchyPaintings.test.ts` → 8 passed.

---

PR #1183, opened 2026-08-25
fix(parallel): scale a value by the axis its point names, not by its column

### LINKED ISSUE #1182: A parallel coordinates observation with a gap is pitched against the wrong axis, and the dialog reports a range with two units in it
`ParallelTrace` computes each axis's extent **by column position** — `lineValues.map(row => row[column])`. That is right only while every observation carries a value for every axis. When one does not, every value after the gap slides one column left and is scaled against the axis before it.

The payload already carries the answer: every `LinePoint` names its own axis in `x`. Nothing has to be inferred.

## Measured

Three cars over three axes, with car B missing an `hp` reading — the shape the adapters already emit, because both of them drop a null point rather than holding the column open:

```ts
const GAPPY: LinePoint[][] = [
  [{ x: 'mpg', y: 33 }, { x: 'hp', y: 65 }, { x: 'weight', y: 1800 }],
  [{ x: 'mpg', y: 21 },                     { x: 'weight', y: 2600 }],
  [{ x: 'mpg', y: 15 }, { x: 'hp', y: 230 }, { x: 'weight', y: 3200 }],
];
```

Reading the trace's own state at column 1:

```
car A, col 1 -> freq { min: 65, max: 2600, raw: 65 }
car B, col 1 -> freq { min: 65, max: 2600, raw: 2600 }

description stats:
  Axes, in order  mpg, hp, weight
  mpg             15 to 33
  hp              65 to 2600      <-- a horsepower range with a car's weight in it
  weight          1800 to 3200
```

Three separate things are wrong, and they are the three modalities:

- **Pitch.** Car A's 65 hp is the *lowest* horsepower in the chart, so it should sound at the bottom of the hp axis — 0 of 1. Against `65..2600` it also lands at 0, by coincidence; car C's 230 hp should sound at the top and instead sits at 0.065. The axis the chart is drawn to compare cars on has collapsed to nearly silent.
- **Announcement.** Car B at column 1 announces `x` as **weight** — correctly, the point says so — while the pitch it is played at is scaled to the hp axis. The reader is told one variable and hears another.
- **The dialog.** `Axes, in order` still lists three axes and gives `hp` the range `65 to 2600`. A reader who opens the dialog to learn what the axes are is given a range that is a horsepower figure at one end and a weight in pounds at the other.

Braille goes the same way as the pitch: `normalized` is built from the same positional extents, so the grid encodes each cell against whichever axis happened to line up.

## This is reachable from the adapters today, not hypothetical

Both emitters produce it, and each has a comment saying so:

- `extractParallelLayer` (plotly) — `if (dimension.values?.[row] == null || !Number.isFinite(value)) continue;`, with *"A gap on one axis is a break in the line rather than a value, and the per-axis extents this trace scales its pitch by must not see it."* The intent is right; `continue` is what makes the columns slide.
- `convertParallelSeries` (Highcharts) — `series.data.filter(p => p.y !== null)`.

`ParallelTrace.axisNames()` already knows rows can run to different lengths — *"Read from the longest observation rather than the first, so a chart whose rows run to different lengths still names every column a cursor can reach"* — but it reads the names off the widest row, which is the same positional assumption.

## The fix

Group the per-axis extents by the axis a point **names**, not by where it sits:

```ts
const extents = new Map<string, number[]>();
for (const row of this.lineValues) …            // keyed by String(point.x)
```

and have `positionOnAxis` take the point rather than the column index. `axisNames()` becomes the ordered set of names seen across every observation rather than the widest row's, so an axis only some observations reach is still named and still gets its own extent.

That fixes all three modalities at once, in one place, and it fixes both adapters without either of them changing — the payload was already telling the truth.

## Why now

Measured while reading `chartjs-chart-pcp` for the `pcp` type #1178 leaves open. That plugin makes a gappy row trivially easy to author — one dataset is one axis, so a single `null` in one dataset's array is one observation missing one axis — and a reading that emitted it would be shipping this bug rather than finding it.

Related: #794 (the parallel trace), #1178.

jooyoungseo:
## The bug

`ParallelTrace` took each axis's extent **by column position** — `lineValues.map(row => row[column])`. That holds only while every observation carries a reading on every axis. One that does not is a point shorter, so from the gap onwards its values sit a column left of the axis they belong to, and the extent that column is scaled against ends up holding two units at once.

Measured on three cars over mpg, hp and weight, with car B never measured for power:

```
car A, col 1 -> freq { min: 65, max: 2600, raw: 65 }
car B, col 1 -> freq { min: 65, max: 2600, raw: 2600 }

Axes, in order   mpg, hp, weight
mpg              15 to 33
hp               65 to 2600      <- a horsepower range with a car's weight in it
weight           1800 to 3200
```

2600 is a weight in pounds sitting at the top of a range in horsepower — the exact mixing of units this class exists to prevent.

## All three modalities go with it

- **Pitch collapses.** Car C's 230 hp is the highest power in the chart and should sound at the top of its axis. Against `65..2600` it sits at 0.065, and the axis the chart was drawn to compare cars on is nearly silent.
- **The announcement and the sound name different variables.** Car B's second point announces `weight` — correctly, the point says so — while being played against the hp axis.
- **The dialog is where a reader goes to learn what the axes are**, and it reports that range as `hp`.

Braille follows the pitch: `normalized` is built from the same extents, so the grid encodes each cell against whichever axis happened to line up.

## Reachable from the adapters today

Both emitters produce this shape, and each has a comment about it:

- `extractParallelLayer` (plotly) — `continue` on a null dimension value, under *"A gap on one axis is a break in the line rather than a value, and the per-axis extents this trace scales its pitch by must not see it."* The intent is right; `continue` is what slides the columns.
- `convertParallelSeries` (Highcharts) — `series.data.filter(p => p.y !== null)`.

## The fix

The payload already carried the answer: every `LinePoint` names its own axis in `x`. So the extents are keyed by that name rather than by position, and `positionOnAxis` takes the name. The axis order is the widest observation's — the chart's axes left to right whenever any observation reaches all of them — with any name it lacks appended rather than guessed into place.

Nothing about the adapters changes. Both are correct again exactly as they stand, and a future one that legitimately emits a short row is correct too.

`axisNames()` is gone; the order is computed once at construction alongside the extents, which is where it belongs — the dialog was recomputing it on every read.

## Tests

11 new cases in `test/model/parallel.test.ts`, in one `describe` for the interior gap plus the sharpenings the sweep asked for:

- a value is scaled by the axis its point names, not by its column
- the axis a reader is told they are on is the one they hear
- the shortened row does not widen an axis it never reached
- the dialog reports each axis in its own units
- braille encodes each cell against the axis that cell names
- an axis no full-width observation reaches is still named
- an axis every observation gapped is named without a range
- one observation gapping an axis does not silence it for the rest (the `Math.min`-of-`NaN` trap `LineTrace` filters for)
- an axis only a short observation reaches, and only gaps, is still named
- the drawn order comes from the widest row, not from what was seen first
- an axis with no spread puts every value at the midpoint, not at an end

The existing trailing-gap suite (`observations that do not all reach every axis`) is unchanged and still passes — it was the case the positional reading handled by luck.

An 11-mutation sweep killed 10. The survivor is `extentOf`'s fallback for a name the layer never used: unreachable through either caller, since both ask with a name taken from a point of this layer and every point registered its own name. Documented as such rather than tested.

## Verification

```
npx eslint src test   0
npx tsc --noEmit      0
npm run build         0
npm run docs          0
npm test              0   415 suites / 5736 unit, 10 suites / 137 ESM
```

Unit tests up from 5729 to 5736 on the first pass and 5740 after the sharpenings.

Closes #1182.

---

### DIFF
--- a/src/model/parallel.ts
+++ b/src/model/parallel.ts
@@ -1,7 +1,38 @@
 import type { LinePoint, MaidrLayer } from '@type/grammar';
 import type { AudioState, BrailleState, DescriptionState, TraceState } from '@type/state';
 import { MathUtil } from '@util/math';
+import { isMeasured } from './bar';
 import { LineTrace } from './line';
+
+/**
+ * The axis names, in the order the chart draws them.
+ *
+ * The widest observation is the spine: whenever any observation carries a
+ * reading on every variable, its points are the chart's axes left to right.
+ * A name that observation lacks is appended rather than guessed into place --
+ * an axis only gappy observations reach has no neighbour the payload names,
+ * and putting it somewhere would assert a drawing order nobody wrote down.
+ *
+ * @param points - The observations the layer carries
+ * @param seen - Every axis name the layer used, in first-seen order
+ * @returns One name per axis
+ */
+function axisOrder(points: LinePoint[][], seen: Iterable<string>): string[] {
+  const widest = points.reduce(
+    (best: LinePoint[], row) => (row.length > best.length ? row : best),
+    [] as LinePoint[],
+  );
+
+  const order = widest.map(point => String(point.x));
+  const placed = new Set(order);
+  for (const axis of seen) {
+    if (!placed.has(axis)) {
+      order.push(axis);
+      placed.add(axis);
+    }
+  }
+  return order;
+}
 
 /**
  * Trace implementation for parallel coordinates plots.
@@ -30,9 +61,29 @@ import { LineTrace } from './line';
  * inaudible.
  */
 export class ParallelTrace extends LineTrace {
-  /** Per-axis extent, indexed by column rather than by row. */
-  private readonly axisMin: number[];
-  private readonly axisMax: number[];
+  /**
+   * Each axis's extent, keyed by the name its points carry.
+   *
+   * Keyed rather than indexed because a column position is not an axis. An
+   * observation with no reading on one variable is one point shorter, so from
+   * that gap onwards its values sit a column to the left of the axis they
+   * belong to -- and an extent taken by position would then be a horsepower
+   * range with a car's weight at the top of it, which is the exact mixing of
+   * units this class exists to prevent (#1182). Every point already names its
+   * own axis in `x`, so nothing has to be inferred from where it sits.
+   */
+  private readonly axisExtent: Map<string, { min: number; max: number }>;
+
+  /**
+   * The axis names, in the order the chart draws them.
+   *
+   * Taken from the widest observation, which is the chart's axes in order
+   * whenever any observation reaches all of them. A name no widest-row
+   * observation carries is appended: the payload says the axis exists but
+   * gives nothing that would place it between two others, and inventing a
+   * position would be a claim about the drawing that was never made.
+   */
+  private readonly axisOrder: string[];
 
   /**
    * Each value's position on its own axis, from 0 to 1.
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Both findings were real. Measured each before acting, and both are fixed in `d5663cb` rather than deferred — they are the same defect in places the first commit did not reach, so scoping them out would have shipped a fix that was incomplete on its own terms.

**1. Panning.** Reproduced exactly as described, on four axes with the second observation missing `hp`:

```
torque, full observation   {"x":2,"y":0,"rows":2,"cols":4}
torque, short observation  {"x":1,"y":1,"rows":2,"cols":3}
```

One axis, two places in the stereo field depending on which observation a reader is walking. `panningFor` is now overridden to resolve the position from the point's own name against `axisOrder`, the same way `axisAtCursor`/`extentOf` do — and `cols` is the chart's axis count rather than the row's length, so a gap is audible as a place nothing is played from instead of silently closing up. Three tests cover it: the axis after the gap, the axes either side of it, and the width of the field.

**2. `Infinity to -Infinity`.** Confirmed — `MathUtil.safeMin([])` is `Infinity`, and `spanned` renders the pair literally. You are right that my change makes it more reachable, since registering an all-gap axis is deliberate: a cursor reaches it, so it has to be named. It now reads `no readings`, and the test asserts the stat value rather than only the order. I also folded the old near-duplicate case into that block so there is one test making each claim.

On the note about the verification gate — understood, and it is worth having the numbers here rather than only in the description. Run locally on `d5663cb`:

```
npx eslint src test   0
npx tsc --noEmit      0
npm run build         0   (467s, 14 examples)
npm run docs          0
npm test              0   415 suites / 5745 unit, 10 suites / 137 ESM
```

A mutation sweep over the first commit killed 10 of 11; the survivor was `extentOf`'s fallback for a name the layer never used, which you independently confirmed is unreachable through either caller, and which is documented as such rather than tested.

---

Issue #1195, opened 2026-08-25
Apache ECharts has no adapter at all — 17 series types, 15 of which MAIDR already has a trace for


jooyoungseo:
`src/adapters/` holds fourteen libraries. **ECharts is not one of them**, and it is the largest charting library with no MAIDR support: no adapter, no docs page, no entry point. A chart drawn with it is silent.

Measured against `echarts@6.1.0`, installed and driven here rather than read from documentation.

## Every series type renders and exposes its model

Rendered headlessly (`echarts.init(null, null, {renderer: 'svg', ssr: true})`) and introspected through `chart.getModel()`. All seventeen work:

| ECharts series | drawn | MAIDR trace it maps onto |
|---|---|---|
| `bar` | ✅ | `BAR` / `STACKED` / `DODGED` |
| `line` | ✅ | `LINE`, `AREA` (with `areaStyle`), `STEP` (with `step`) |
| `scatter` | ✅ | `SCATTER` (`symbolSize` → `ScatterPoint.z`, as #826 did for Chart.js bubbles) |
| `pie` | ✅ | `PIE` |
| `boxplot` | ✅ | `BOX` |
| `candlestick` | ✅ | `CANDLESTICK` |
| `heatmap` | ✅ | `HEATMAP` |
| `radar` | ✅ | `RADAR` |
| `funnel` | ✅ | `FUNNEL` |
| `gauge` | ✅ | `GAUGE` |
| `treemap` | ✅ | `TREEMAP` |
| `sunburst` | ✅ | `SUNBURST` |
| `sankey` | ✅ | `SANKEY` |
| `graph` | ✅ | `NETWORK` |
| `parallel` | ✅ | `PARALLEL` |
| `themeRiver` | ✅ | `STACKED_AREA` or `RIDGELINE` — needs deciding |
| `pictorialBar` | ✅ | `BAR`; the pictograph is what the mark looks like |

Fifteen of the seventeen map onto a trace that already exists and is already navigable. Nothing here needs a new trace type — which is why this is worth doing.

## The model hands over what a layer needs

```
dims:   [ 'x', 'y' ]
count:  3
values: [ {name: 'A', x: 0, y: 1}, {name: 'B', x: 1, y: 2}, {name: 'C', x: 2, y: 3} ]
series name: 'Series one'
components: [["xAxis","Category"],["yAxis","Count"],["title","A title"]]
```

[...]

PR #1196, opened 2026-08-25
feat(echarts): add an adapter for cartesian ECharts charts

### LINKED ISSUE #1197: a scatter mark placed by a matrix transform is grouped at its path's own origin
## What

`ScatterTrace.groupSvgElements` recovers a mark's position from the DOM
rather than from the data's order, and it tries four things in turn: the
`x`/`y` attributes, `cx`/`cy`, a `translate(…)` transform, and finally the
`M x y` that opens a `d`.

A library that places its symbols with `transform="matrix(…)"` falls past
the first three and lands on the fourth — and if its symbol path is a unit
shape about the origin, as one drawn by a scaling transform is, then *every*
mark on the chart reports the same coordinate. They all group into one
column, and the highlighting is wrong for every point after the first.

## Measured

ECharts 6.1.0, `renderer: 'svg'`, a three-point scatter, in Chromium:

```
tag        path
fill       #5070dd
d          M1 0A1 1 0 1 1 1 -0.0001
transform  matrix(2.937,0,0,2.937,240,235)
           matrix(2.937,0,0,2.937,390,150)
           matrix(2.937,0,0,2.937,540,65)
```

The `d` is identical on all three; the position is entirely in the matrix.
`M1 0` parses, so the `d` fallback answers `(1, 0)` three times rather than
failing and leaving the mark ungrouped.

`matrix(a, b, c, d, e, f)` translates by `(e, f)` whatever the other four
are, so the mark's centre is the last pair — 240/390/540 above, which is
where the three points sit.

## Why it matters beyond one library

Nothing about this is ECharts-specific: it is what any renderer does when it
draws one symbol path and scales it into place instead of emitting a fresh
path per point. The four existing strategies were each added for the library
that needed it (Plotly's `translate`, Google Charts' `cx`/`cy`, Highcharts'
`d`), and this is the fifth of the same kind.

## Fix

Read `matrix(a, b, c, d, e, f)` as a position, **before** the `d` fallback —
after it is no use, because the `d` answers first and answers wrongly.

Landing with the ECharts adapter in #1195, since a scatter there cannot be
highlighted at all without it.

jooyoungseo:
Tier 1 of #1195. Also fixes #1197.

## What

ECharts had no adapter while fourteen other chart libraries did, so an ECharts figure was unreachable to MAIDR entirely — not degraded, not partially read, simply absent. This adds `src/adapters/echarts/` covering the cartesian series:

| ECharts | read as |
| --- | --- |
| `bar` (single) | `bar` |
| `bar` with `stack` | `stacked` |
| several `bar` series, no `stack` | `dodged` |
| `bar` on a category `yAxis` | the same, `orientation: 'horz'` |
| `line` | `line` |
| `line` with `areaStyle` | `area` |
| `line` with `step` | `line` carrying the step direction |
| `scatter` | `scatter`, with `symbolSize` read into `z` |

A `pie` chart — anything outside the set — is refused rather than half-read.

## The addressability problem, and how this resolves it

Every other adapter addresses its marks by selector because its library
puts something addressable on them. ECharts' SVG renderer does not: the
tree is flat, and no element carries an id, a class, or a `data-*`
attribute that ties it to a datum.

What the renderer does guarantee is that data marks are separable from
chart furniture by paint — a bar or a symbol is filled, a line is stroked
and unfilled, and the furniture (axis lines, splitlines, the background,
the label backgrounds) is neither, or is painted in the small set of
furniture colours. It also draws marks in data order.

So the adapter counts what it expects to find, compares that against what
the paint filter selected, and stamps its own attributes onto the elements
only when the two agree. On a mismatch it warns and emits the layer with
no selectors, which reads and sonifies but does not highlight. Pointing
at the wrong mark is worse than pointing at none, so a mismatch fails
that way deliberately.

The consequence worth stating plainly: **a series painted pure black is
read but not highlighted**, because pure black is how ECharts paints
several pieces of furniture and the filter cannot tell them apart. That
is the conservative failure, and `docs/echarts.md` says so.

## Each layer gets the selector shape its trace can resolve

A selector that finds the right element is only half of it. Each trace class
accepts a different **shape**, and a layer whose selectors are individually
right and collectively the wrong shape resolves nothing at all — silently, with
no warning and nothing wrong in the payload:

| layer | shape | what reads it |
| --- | --- | --- |
| `bar` | one selector per bar | `AbstractBarPlot`'s array branch |
| `stacked`, `dodged` | a row of selectors per series | `SegmentedTrace.mapGridToSvgElements` |
| `scatter` | **one** selector naming the whole series | `ScatterTrace`, which casts `layer.selectors` to `string` |
| `line`, `area` | one selector per series | `LineTrace.mapToSvgElements` |

So marks are stamped twice in one pass — once per mark, once per series — and
each layer takes the address its own trace can use. Two of the four had this
wrong on the first commit and highlighted nothing; the fix is in `07f7284`,
pinned by tests that build the real trace rather than stopping at the string.

**#1197 is the model half of the scatter case.** ECharts does not *move* its
symbols into place, it *scales* them there: `matrix(s, 0, 0, s, e, f)` over a
unit shape whose `d` always begins `M1 0`. `ScatterTrace.groupSvgElements`
recovers a mark's position from the DOM, and its `d` fallback answered `(1, 0)`
for every point on the chart, grouping the whole scatter into one column. It
reads the matrix now, before that fallback. Nothing about it is
ECharts-specific — it is what any renderer does that draws one symbol path and
scales it per point.

## Two bugs the live run caught

Both were invisible to the unit tests as first written, and both are now
pinned by tests:

1. **The per-datum and per-series stamps shared one attribute.** Stamping
   the line series unstamped every bar, so on a mixed bar + line chart no
   bar could be highlighted while every line could.
2. **An axis name draws an `rgb(0, 0, 0)` background rect.** The paint
   filter recognised `#000000` as furniture but not the `rgb()` spelling
   of it, so it counted the rect as a mark, every count check failed, and
   every chart with an axis name lost its highlighting. Paint is now
   normalised before it is compared.

## Verification

- **Live, against echarts 6.1.0 in Chromium**: twelve figures — `bar`,
  unnamed `bar`, horizontal `bar`, `bar` with gaps, stacked, dodged,
  `line`, area, step, `line` with a gap, `scatter`, sized `scatter` —
  with **zero selector failures**, and `pie` correctly refused. Payloads
  spot-checked against what was drawn, including the horizontal swap
  (`{x: 1, y: 'A'}`) and the gap (`{x: 'B', y: null}`).
- **Mutation sweep: 22/22 caught.**
- Full gate: `eslint`, `tsc --noEmit`, `npm test` (427 suites / 5862
  tests, plus the ESM project's 10 / 138), `npm run build` (15/15
  bundles, up from 14), `npm run docs`.

One note for anyone re-running the tests: the ESM project has to be run
through plain `npm test` rather than `--selectProjects esm`, which fails
with `SyntaxError: Unexpected token 'export'` on the jest-resolve
cross-project memoisation bug tracked in #703. That is unrelated to this
change.

## Not in this change

Tiers 2–4 of #1195: the single-series shapes (`pie`, `gauge`, `funnel`,
`radar`, `boxplot`, `candlestick`, `heatmap`), the hierarchies and graphs
(`tree`, `treemap`, `sunburst`, `sankey`, `graph`), and
`themeRiver` / `pictorialBar`.

### DIFF
--- a/docs/echarts.md
+++ b/docs/echarts.md
@@ -101,3 +101,28 @@ One consequence worth knowing: **a series painted pure black loses its
 highlighting.** Its marks are indistinguishable from the furniture, so they are
 excluded, the count check fails, and the chart reads without an outline. That
 is the conservative failure, not a silent wrong one.
+
+### Which shape each layer's selectors take
+
+A selector that resolves the right element is only half of it: each trace
+class accepts a different **shape**, and a layer whose selectors are
+individually right and collectively the wrong shape resolves nothing at all,
+silently.
+
+| layer | shape | what reads it |
+|---|---|---|
+| `bar` | one selector per bar | `AbstractBarPlot`'s array branch |
+| `stacked`, `dodged` | a row of selectors per series | `SegmentedTrace.mapGridToSvgElements` |
+| `scatter` | **one** selector naming the whole series | `ScatterTrace`, which casts `layer.selectors` to `string` |
+| `line`, `area` | one selector per series | `LineTrace.mapToSvgElements` |
+
+So the marks are stamped twice in one pass — once per mark and once per
+series — and each layer takes the address its own trace can use.
+
+A scatter needs one more thing from the model. ECharts does not move its
+symbols into place; it scales them there, with
+`transform="matrix(s, 0, 0, s, e, f)"` over a unit shape whose `d` always
+begins `M1 0`. `ScatterTrace` recovers a mark's position from the DOM rather
+than from the data's order, and it now reads that matrix — without it every
+symbol reported the same coordinate and the whole scatter grouped into one
+column (#1197).
--- a/docs/echarts.md
+++ b/docs/echarts.md
@@ -1,0 +1,103 @@
+# Apache ECharts
+
+MAIDR reads a rendered [Apache ECharts](https://echarts.apache.org/) chart and
+makes it navigable with audio sonification, text descriptions, braille output
+and keyboard navigation.
+
+Measured against **echarts 6.1.0**.
+
+## Quick start
+
+```html
+<script src="https://cdn.jsdelivr.net/npm/echarts@6/dist/echarts.min.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/maidr/dist/maidr.js"></script>
+<script src="https://cdn.jsdelivr.net/npm/maidr/dist/echarts.js"></script>
+<div id="chart" style="width: 600px; height: 400px"></div>
+<script>
+  const container = document.querySelector('#chart');
+  const chart = echarts.init(container, null, { renderer: 'svg' });
+
+  chart.setOption({
+    title: { text: 'Daily Visitors' },
+    xAxis: { type: 'category', data: ['Mon', 'Tue', 'Wed'], name: 'Day' },
+    yAxis: { name: 'Visitors' },
+    series: [{ type: 'bar', name: 'Visitors', data: [120, 240, 180] }],
+  });
+
+  chart.on('finished', function once() {
+    chart.off('finished', once);
+    const maidr = maidrECharts.createMaidrFromEChart(chart, container);
+    container.setAttribute('maidr', JSON.stringify(maidr));
+  });
+</script>
+```
+
+Two things this example does on purpose:
+
+- **`renderer: 'svg'`.** ECharts defaults to canvas, which draws no elements to
+  point at. A canvas chart still reads — audio, text and braille all come from
+  the chart's model — but it highlights nothing.
+- **Calling after `finished`.** The adapter locates marks in the drawn SVG, so
+  it has to run after ECharts has drawn.
+
+## Supported series types
+
+| ECharts `series.type` | Read as | Notes |
+|---|---|---|
+| `bar` | `bar` | One series. `yAxis: {type: 'category'}` makes it horizontal |
+| `bar` ×N sharing a `stack` | `stacked_bar` | Each point carries its series name |
+| `bar` ×N without a `stack` | `dodged_bar` | |
+| `line` | `line` | |
+| `line` + `areaStyle` | `area` | The fill is what makes it an area |
+| `line` + `step` | `line` + `stepDirection` | `'start'` → `vh`, `'end'`/`'middle'` → `hv` |
+| `scatter` | `point` | A `symbolSize` reading a third column becomes `ScatterPoint.z`, which is audible |
+
+**Not yet supported:** `pie`, `boxplot`, `candlestick`, `heatmap`, `radar`,
+`funnel`, `gauge`, `treemap`, `sunburst`, `sankey`, `graph`, `parallel`,
+`themeRiver`, `pictorialBar`. Each of these is *refused by name* rather than
+mapped onto whichever trace is closest — most have a MAIDR trace waiting for
+them, and each wants its own measured layout first. See
+[#1195](https://github.com/xability/maidr/issues/1195) for the tiers.
+
+> **Orientation note:** ECharts has no "horizontal" option. A bar chart is
+> turned on its side by making the **y** axis the categorical one, and that is
+> the only thing that says so — so the adapter reads the orientation from which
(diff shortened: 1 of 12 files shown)


jooyoungseo:
Both bugs were real, and the diagnosis of *why* they were invisible was the
most useful part of the review: my live run and my unit tests both stopped at
"the selector string finds its element" and never built the trace that would
consume it. Fixed in `07f7284`, along with both minor points. Details, and one
thing the review's fix suggestion would not have been enough for:

### 1. Scatter — confirmed, and it needed a model change too

The shape was wrong exactly as described, and one shared selector string is the
right fix. `markPerDatum` now stamps a second attribute saying which *series* a
mark belongs to, so a scatter layer carries `#id svg [data-maidr-echart-group="2"]`
and a bar layer keeps its per-mark list.

But that alone still highlighted nothing useful, because `ScatterTrace` does not
map elements to points by order — `groupSvgElements` recovers each mark's
position from the DOM. Measured on echarts 6.1.0 in Chromium:

```
d          M1 0A1 1 0 1 1 1 -0.0001      ← identical on every point
transform  matrix(2.937,0,0,2.937,240,235)
           matrix(2.937,0,0,2.937,390,150)
           matrix(2.937,0,0,2.937,540,65)
```

ECharts does not *move* its symbols into place, it *scales* them there. No
`x`/`y`, no `cx`/`cy`, and `translate` does not match — so the chain fell
through to the `d` fallback, which parses `M1 0` and answers `(1, 0)` for every
point on the chart. All three grouped into one column.

`groupSvgElements` now reads `matrix(a, b, c, d, e, f)` as a position, before
the `d` fallback that would otherwise answer first and answer wrongly. Filed
separately as #1197 — nothing about it is ECharts-specific, it is what any
renderer does that draws one symbol path and scales it per point.

### 2. Stacked / dodged — confirmed, one correction to the mechanism

The outcome is exactly as described: no ECharts stacked or dodged bar was ever
highlighted. The path is slightly different from the one named, and worth
recording since it changes what the fix has to be. `SegmentedTrace` **does**
override `mapToSvgElements` (`src/model/segmented.ts:397`) and routes an array
to `mapGridToSvgElements`, so `AbstractBarPlot`'s `this.points.length !== 1`
branch is never reached. `mapGridToSvgElements` then declines the flat list on
`grid.length !== this.barValues.length` — same silence, same result.

So the fix is the second of the two options offered: a genuine per-series
`(string | null)[][]` grid, which is what `markPerDatum` already produced before
I flattened it. `domMapping` is not needed — a grid names every cell outright,
which is what its own comment says makes the inference moot.

### 3. Minor points — both taken

- `nextId` now comes from `src/adapters/shared/selectorUtil`.
- `drawnCount` and `scatterLayer` now ask the same question, via a shared
  `drewMark` / `placed` pair. It was reachable, not just theoretical: a scatter
  datum with a finite magnitude and a non-numeric position would have stamped a
  mark no point stood for.

### On the test-coverage gap

Agreed, and it was the whole cause. `test/adapters/echarts/highlightBinding.test.ts`
is new: it builds the real `BarTrace` / `SegmentedTrace` / `ScatterTrace` /
`LineTrace` from what the adapter emits, navigates to a cell, and asserts the id
of the mark that resolves. All four bugs above fail it before the fix. It runs in
the jsdom environment rather than the node one its siblings use, because building
a `LineTrace` reaches `Svg.createCircleElement` and that needs a `window`.

Two more cases went into `cartesian.test.ts` and `highlightBinding.test.ts` to
close mutation-sweep survivors: two scatter series on one chart (a single-series
chart cannot tell a per-series stamp from a constant one), and a scatter datum
with only one coordinate.

### Re-verified

- Live, echarts 6.1.0 in Chromium, twelve figures: every selector resolves, and
  the shapes are now `list` for `bar`, `grid` for `stacked`/`dodged`, and
  `string` for `scatter` and `line` — with the scatter's one string matching all
  three of its marks.
- Mutation sweep **22/22** (was 16/16 before these six new mutants).
- Full gate green: eslint, `tsc --noEmit`, 427 suites / 5862 tests plus the ESM
  project's 10 / 138, 15/15 bundles, docs.

---

PR #1198, opened 2026-08-25
feat(echarts): read a pie, a funnel and a gauge

jooyoungseo:
Tier 2a of #1195, on top of the adapter merged in #1196.

## What

Three series types the adapter refused by name now read.

| ECharts | read as | highlighted |
| --- | --- | --- |
| `pie` | `pie` | yes — one selector naming every slice |
| `funnel` | `funnel` | yes — one selector per stage |
| `gauge` | `gauge` | **no**, deliberately — see below |

They share a shape the cartesian series do not: no axes, one series to a
chart, and a datum that is a name and a number rather than a position and a
magnitude. Measured on echarts 6.1.0, all three report `data.dimensions` as
`['value']` with the label on `getName(i)`, so the whole reading is that
pair — which is why they are one module rather than three.

## Each layer gets the selector shape its trace can resolve

Checked by building the real trace and navigating it, not by resolving the
string. That is the lesson tier 1 paid for: two of its four layers carried
selectors that were individually correct and collectively the wrong shape,
and resolved nothing at all with no warning.

- **pie** — `PieTrace` reads `layer.selectors as string`, so the slices are
  named together in one selector. A list there would match nothing.
- **funnel** — `FunnelTrace` extends `BarTrace` and takes its array branch,
  so each stage is named separately.
- **gauge** — none.

## Why a gauge is read without an outline

It draws **two** filled marks for its one datum: the track (`#e8ebf0`) and
the progress arc (the series colour). The count check every other reading in
this adapter relies on has nothing to check — there is no arrangement in
which one datum and two marks agree — and naming either mark would be a
guess about which one the reader should be shown. So the gauge reads and
sonifies, and does not highlight. `docs/echarts.md` says so.

## Measured, not assumed

- `min` and `max` come off the series rather than being defaulted here.
  `getModel()` resolves ECharts' own `0` and `100` when the author wrote
  neither — confirmed live — so the dial the reader is told about is the
  dial that was drawn, and a caller who wrote `min: 10, max: 200` gets
  theirs.
- A pie draws one filled path per slice **and** one unfilled stroked guide
  line out to each label. The paint filter already declines the guides; the
  fixture draws both so a filter that stopped doing so would fail here.
- A datum with no number is left out rather than read as a slice of nothing.
  ECharts draws nothing for it, and counting it would make the mark check
  expect three where two were drawn — costing the two real slices their
  outline as well. A chart with nothing left to read emits no layer rather
  than an empty one a reader could land on and hear nothing from.

## Verification

- **Live, echarts 6.1.0 in Chromium**: pie (one string matching all three
  slices), funnel (three selectors, all resolving), gauge with a written
  dial and gauge with the default one. The tier-1 cases re-run unchanged,
  and `treemap` is still correctly refused.
- **Mutation sweep: 28/28** (was 22/22 before the six new mutants).
- Full gate: eslint, `tsc --noEmit`, 428 suites / 5874 tests plus the ESM
  project's 10 / 138, 15/15 bundles, docs.

## Not in this change

The rest of tier 2 — `boxplot`, `candlestick`, `heatmap`, `radar` — plus the
hierarchies and graphs of tier 3 and `themeRiver` / `pictorialBar` in tier 4.
Every one of them has a trace waiting and a model that carries what it needs
(the survey is in #1195); what each still wants is its own selector shape
established against the trace that reads it, which is the part this PR and
#1196 show is not safe to assume.

---

### DIFF
--- a/docs/echarts.md
+++ b/docs/echarts.md
@@ -71,6 +71,28 @@ them, and each wants its own measured layout first. See
 > samples either side stay in their places rather than the series closing over
 > the hole.
 
+## Single-value charts
+
+A pie, a funnel and a gauge sit on no grid and own the whole chart, and all
+three report `data.dimensions` as `['value']` with the label on `getName(i)` —
+so the reading is that pair.
+
+| ECharts | read as | highlighted |
+|---|---|---|
+| `pie` | `pie` | yes — one selector naming every slice |
+| `funnel` | `funnel` | yes — one selector per stage |
+| `gauge` | `gauge` | **no** — see below |
+
+A gauge draws **two** filled marks for its one datum: the track and the
+progress arc, measured as `#e8ebf0` and the series colour. The count check
+every other reading here relies on has nothing to check, and naming either
+mark would be a guess about which one the reader should be shown, so the
+gauge is read without an outline.
+
+`min` and `max` come off the series rather than being defaulted here.
+`getModel()` resolves ECharts' own `0` and `100` when the author wrote
+neither, so the dial the reader is told about is the dial that was drawn.
+
 ## Highlighting
 
 ECharts gives a reading almost nothing to address by. Measured on 6.1.0: the
--- a/docs/echarts.md
+++ b/docs/echarts.md
@@ -52,9 +52,9 @@ Two things this example does on purpose:
 | `line` + `step` | `line` + `stepDirection` | `'start'` → `vh`, `'end'`/`'middle'` → `hv` |
 | `scatter` | `point` | A `symbolSize` reading a third column becomes `ScatterPoint.z`, which is audible |
 
-**Not yet supported:** `pie`, `boxplot`, `candlestick`, `heatmap`, `radar`,
-`funnel`, `gauge`, `treemap`, `sunburst`, `sankey`, `graph`, `parallel`,
-`themeRiver`, `pictorialBar`. Each of these is *refused by name* rather than
+**Not yet supported:** `boxplot`, `candlestick`, `heatmap`, `radar`,
+`treemap`, `sunburst`, `sankey`, `graph`, `parallel`, `themeRiver`,
+`pictorialBar`. Each of these is *refused by name* rather than
 mapped onto whichever trace is closest — most have a MAIDR trace waiting for
 them, and each wants its own measured layout first. See
 [#1195](https://github.com/xability/maidr/issues/1195) for the tiers.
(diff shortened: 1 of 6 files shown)


jooyoungseo:
All five taken, in `d718bbf`. The two "minor" ones turned out to be the
interesting half.

### 1–3. The stale documentation — all correct, all fixed

Three statements, all false in the same direction, and all of them saying
that pie, funnel and gauge are refused by name:

- `docs/echarts.md`'s "Not yet supported" list — pie, funnel and gauge
  removed; the rest of the list stands.
- the JSDoc above `CARTESIAN` — it was the comment that used to sit above
  `READ`, so it claimed to describe "the series types this tier reads" while
  naming only three of six. `CARTESIAN` now describes itself as the
  grid-based subset, and the "refused by name" paragraph moved down to
  `READ`, which is what actually decides.
- `EChartsSeriesType` — this one mattered most, since it is re-exported from
  the adapter's barrel and so is part of the public surface. Now the six,
  with a note that it is kept in step with `READ`.

A module whose whole claim is "measured, not assumed" cannot carry three
statements that contradict its own code. Thanks for catching all three.

### 4. Multi-pointer gauges — measured, and it was a real drop

Measured on echarts 6.1.0 in Chromium rather than assumed either way:

```js
chart.setOption({ series: [{ type: 'gauge', data: [
  { name: 'A', value: 30 }, { name: 'B', value: 70 },
]}]});

// count: 2
// rows:  [{ name: 'A', value: 30 }, { name: 'B', value: 70 }]
// filled marks: #e8ebf0 (the track) + two series-coloured needles
```

So the series really does carry two readings and really does draw two
pointers, and `read[0]` really was dropping the second in silence.

`GaugePoint` is a single reading, so two needles are now **two layers**. What
that loses is that they share a dial — nothing in the grammar carries it —
but both come out with the same `min` and `max`, which is as close as the
shape gets, and it beats announcing one of two.

### 5. The mixed chart — now says what it dropped

You were right that the comment explained the design and not the behaviour.
A pie has no grid to share with a bar, so this is not a chart ECharts is
drawn to make; it is not invalid either, and a silent drop is exactly the
failure mode the rest of this adapter is built to avoid. It now warns and
names what it ignored, alongside the mark-count and unsupported-type paths.

### Coverage

Two new cases, both of which fail before the fix: the two-needle gauge
(asserting both readings, with the shared dial on each), and the mixed chart
(asserting the pie layer survives and the warning names `bar`).

Mutation sweep **30/30**, up from 28/28 — the two new mutants are "a gauge
with two needles is read as having one" and "a chart that mixes families
drops half of it in silence", and both were live before this commit.

Full gate re-run green: eslint, `tsc --noEmit`, 428 suites / 5876 tests plus
the ESM project's 10 / 138, 15/15 bundles, docs.

---

Issue #1191, opened 2026-08-25
A sparse heat grid is filled with zeros, so a cell the chart never drew is announced as 0


jooyoungseo:
`HeatmapData.points` is `number[][]`, so every cell of the rectangle must carry a number. An adapter handed data that does not fill the rectangle has no way to say so, and three of them fill the holes with `0` — which is a value, and a finding, that the chart never made.

One adapter already refuses to do this and says why, in `src/adapters/tableau/extractor.ts`:

> `HeatmapData` is a rectangle: `points[r].length` must equal `x.length` for every row. A view whose row count does not match the product has holes, and **a hole filled with a zero is a value the viz never drew.**

`isCompleteGrid()` declines the heat reading on exactly that test. The other three do the opposite.

## Measured

Driven through `highchartsToMaidr` on Highcharts 13.0.1 in jsdom. Two charts over the same 3×2 grid — the first omits `[1, 1]` entirely, the second states it as `0`:

```
== one cell absent
   points  = [[2,0,4],[5,7,9]]
   pointsN = 6
   drawn series points = 5

== that cell zero
   points  = [[2,0,4],[5,7,9]]
   pointsN = 6
   drawn series points = 6
```

The payloads are **identical**. The first chart draws five cells and MAIDR announces six; the sixth is "Tue, PM: 0", read off a blank the chart left blank on purpose. A reader has no way to tell "no record" from "recorded zero", and on a chart where zero is meaningful — a count, a rainfall total, a defect rate — the two are opposite findings.

## Where it happens

| Adapter | Line | What it does |
|---|---|---|
| `highcharts/adapter.ts` | `points[yIdx][xIdx] = cellValue ?? 0`, over a grid `Array.from(… () => 0)` | fills |
| `vegalite/converters.ts` | `const points = yLabels.map(() => xLabels.map(() => 0))` | fills |
| `anychart/converters.ts` | `Array.from<number>({length: xLabels.length}).fill(0)` | fills |
| `tableau/extractor.ts` | `isCompleteGrid()` | declines |

All three fillers build their axis labels from the **union** of what the rows mention, so any (x, y) pair the data skips is a hole by construction — not an edge case reachable only through a malformed chart.

## Why it is worth fixing rather than declining everywhere

[...]

PR #1199, opened 2026-08-25
feat(echarts): read a heat grid and a price chart

jooyoungseo:
Tier 2b of #1195. The two ECharts series that hand over a summary already computed.

A heat grid and a price chart sit on the same cartesian axes the bar family uses and differ from it the same way: a datum is a set of magnitudes rather than one, and ECharts has already worked them out. Measured on **echarts 6.1.0** the model names them outright:

| ECharts | `data.dimensions` | read as | highlighted |
|---|---|---|---|
| `heatmap` | `['x', 'y', 'value']` | `heat` | yes — a selector per cell |
| `candlestick` | `['base', 'open', 'close', 'lowest', 'highest']` | `candlestick` | yes — a selector per candle |

so nothing is recovered from the drawing. Both draw exactly one filled mark per datum — a cell, a candle body with its wick — which is the shape `markPerDatum` already counts and stamps for a bar, so each gets a per-mark outline through the same pass.

## Three things measurement decided

**A heatmap numbers a cell by axis index, and a category y axis runs bottom-up.** A 2×2 grid reports its cells as `[0,0]`, `[0,1]`, `[1,0]`, `[1,1]` with `y = 0` along the bottom. `HeatmapData` is top-first, so the rows are turned over on the way out — and the selector grid is built by *the same walk* that places the values, because `Heatmap` reverses again on construction and indexes selectors by its own bottom-first row. Getting that backwards outlines the top row while announcing the bottom one, which is why the test navigates a real `Heatmap` rather than resolving the strings.

**A `visualMap` legend paints its bands with a `url(#…)` gradient**, which is neither `none` nor a furniture colour. Measured beside a real heat grid: four cells in flat `rgb(…)` and two `url(#zr6-g0)` bands. Counted, the legend would put eight marks where the model says six and cost the grid its highlighting, so `isFilledMark` now declines a gradient fill.

**`Candlestick.mapToSvgElements` reads a bare array as the *legacy* shape** — one selector for the whole chart — and takes `selectors[0]` for every candle. A per-candle list would have outlined the first candle wherever the reader stood. The selectors go in `CandlestickSelector.body`; only `body` is filled, because ECharts draws a candle as one path holding the body and its wick together.

This is the tier-1 lesson (#1196) applied before it could bite: a selector that resolves the right element is only half of it, and each trace class accepts a different *shape*. Every selector assertion here is made by constructing the real trace and navigating it.

## Edge cases, all reachable and all measured

- **A cell with no value stays empty, not zero.** ECharts keeps an empty cell in its data and reports `null`, whether the author wrote `null` or `'-'` — so the count is four and the drawing holds two. The reading counts what was drawn (#1191).
- **A cell placed off the axes is dropped.** A datum at `[5, 5]` on a 2×2 grid is kept in the data and drawn nowhere. Writing it in would index a row that does not exist and take the whole chart down with a `TypeError`.
- **A cell placed between two rows reads without an outline.** A datum at `[0.5, 0.5]` *is* drawn, but it sits on no row and no column, so there is nowhere in a `HeatmapData` grid to put it. It is left out, the count then disagrees with the drawing, and the chart loses its highlighting — the conservative failure rather than a wrong outline.
- **A period missing any of its four prices draws no candle**, so it is left out of the reading and of the count.

## Deferred, with the reason recorded

`boxplot` and `radar` are measured and left for a later tier, both for how they are drawn rather than what they carry. A box plot hands over its five-number summary outright, but `BoxSelector` wants a selector per part and ECharts draws the box and both whiskers as **one path** painted `#fff`. A radar's count does not fit: a two-series radar draws six vertex symbols and also fills its alternating ring backgrounds, one of which (`rgb(234,237,245)`) is neither furniture nor white, so seven marks are found where six are expected. Both notes are in `grid.ts`'s footer.

## Verification

- `test/adapters/echarts/gridValue.test.ts` — 12 cases, jsdom, every selector checked by navigating the real trace
- Live-verified against real echarts 6.1.0 in Chromium via Playwright:
  ```
  heat:    shape=grid sel=6 res=6 nulls=0 match=6 axes=['Col','Row']
           data={"x":["c0","c1","c2"],"y":["r1","r0"],"points":[[10,11,12],[0,1,2]]}
  heatGap: shape=grid sel=4 res=2 nulls=2 match=2
  candle:  sel=3 res=3 match=3 name=ACME  axes=[None,'Price']
  ```
- **Mutation sweep: 46/46 caught**, including all sixteen written for this tier
- Full gate green: `eslint`, `tsc --noEmit`, `npm test` (all projects), `npm run build`, `npm run docs`

---

### DIFF
--- a/docs/echarts.md
+++ b/docs/echarts.md
@@ -52,11 +52,10 @@ Two things this example does on purpose:
 | `line` + `step` | `line` + `stepDirection` | `'start'` → `vh`, `'end'`/`'middle'` → `hv` |
 | `scatter` | `point` | A `symbolSize` reading a third column becomes `ScatterPoint.z`, which is audible |
 
-**Not yet supported:** `boxplot`, `candlestick`, `heatmap`, `radar`,
-`treemap`, `sunburst`, `sankey`, `graph`, `parallel`, `themeRiver`,
-`pictorialBar`. Each of these is *refused by name* rather than
-mapped onto whichever trace is closest — most have a MAIDR trace waiting for
-them, and each wants its own measured layout first. See
+**Not yet supported:** `boxplot`, `radar`, `treemap`, `sunburst`, `sankey`,
+`graph`, `parallel`, `themeRiver`, `pictorialBar`. Each of these is *refused
+by name* rather than mapped onto whichever trace is closest — most have a MAIDR
+trace waiting for them, and each wants its own measured layout first. See
 [#1195](https://github.com/xability/maidr/issues/1195) for the tiers.
 
 > **Orientation note:** ECharts has no "horizontal" option. A bar chart is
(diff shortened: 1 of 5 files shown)


jooyoungseo:
Acting on the duplicated-predicate point, since it is the same argument the code's own comment makes one line above the duplication.

`heatmapLayer` and `drawnGridCount` were two independent walks asking the same question — which cells reached the page. They agreed, but the failure mode if they ever stopped agreeing is nasty and quiet: the count check would pass with a selector list one longer than the marks it addresses, and every cell after the first divergence would be outlined one place off, in silence. That is the same class of bug tier 1 shipped (#1196).

Both now go through one `placedCells()` walk, which returns the drawn cells in data order; `drawnGridCount` is its length and `heatmapLayer` iterates it. The repo's "no abstraction before a second caller exists" rule is satisfied — there are exactly two, and they are the two that must not disagree.

The mutation sweep gained a case for it (`the count and the layer stop asking the same question`, mutating `placedCells(...).length` back to `data.count()`).

On the two other observations, noted and deliberately left as they are:

- **`whole()` vs `converters.ts`'s `measured()`** — they narrow different things (`whole` is "an axis index", `measured` is "a finite number"), and the version in each file is three lines. Unifying would cost an import across a module boundary for no reader's benefit.
- **`isFilledMark`'s `url(` check** — agreed the theoretical false negative is a fill whose literal text begins `url(`, which is not a colour. The real case it was measured against is in the comment.

Re-verified after the refactor: `tsc --noEmit`, `eslint`, and the full unit suite (429 suites / 5888 tests) all pass locally; pushing now so CI re-runs the whole gate.

---

PR #1208, opened 2026-08-27
docs(schema): split the declarable trace types by what they promise

jooyoungseo:
## What

`TraceType` defined **15** authorable types when the chart type coverage roadmap (#814) was filed. It defines **52** now. The 37 that arrived with the roadmap are not on the same footing as the 15 that predate it — most landed inside about two weeks, and **none has been through a user study** — but `docs/SCHEMA.md` listed all 52 in one alphabetical run, so a producer reading it had no way to tell a settled reading from a prototype.

This adds a **Trace type stability** section that splits them and says what the experimental half does not promise.

## The boundary is measured, not editorial

It is the diff of `TraceType` against `84d9003`, the last commit on `main` before #814 was filed:

```bash
git show 84d9003:src/type/grammar.ts | grep -oE "= '[a-z_0-9]+'"
```

That gives 16 values, of which `candlestick_delta` is runtime-only, so **15 stable + 37 experimental = 52 declarable**, which is exactly what the existing list names. The section carries the command so the next reader can re-derive it rather than trust it.

The stable set ends at `violin_kde` and `violin_box`, which matches where the pre-roadmap work actually stopped.

## What the experimental set is told not to promise

- **Under active development** — changed as they are used, not maintained against a settled specification.
- **Unstable** — point shapes, field names, announcement wording and navigation semantics may change without a deprecation period, including in a patch release.
- **Not validated** — each was measured against the drawing it reads, which is what the issues and tests record. Measuring that a reading is *faithful to the chart* is a different claim from establishing that it is *useful to a reader*, and only the first has been done. Nobody has asked a blind or low-vision reader whether navigating a sunburst by depth, or hearing each parallel-coordinates axis on its own scale, is the right way to read one.
- **Not a support commitment.**

## The drift guard

`test/scripts/schemaStability.test.ts` fails if a declarable type lands in **neither** list or in **both**. At 37 additions in two weeks, a new type reaching the enum and not the split is the expected outcome rather than an unlikely one, so the classification has to be a deliberate step.

It pins the claim as well as the lists — the lists alone would read as a changelog, and softening the wording without revisiting the split should fail. Prose assertions normalise whitespace so a reflow is not what breaks them.

The existing `schemaTraceTypes.test.ts` is untouched and still passes: the flat declarable-types line stays as it was, and this section is a second view of the same set rather than a replacement.

## Verified

```
npx eslint src test scripts   exit 0
npx tsc --noEmit              exit 0
npm test                      434 suites / 5949 tests passed, plus 10 / 138
npm run build                 exit 0
npm run docs                  exit 0
```

The docs build also confirmed the rendered anchor is `id="trace-type-stability"` at `https://maidr.ai/docs/SCHEMA.html`, which is what the sibling binding pages link to.

## Companions

The same split, same boundary, same reasoning, in both bindings: xability/py-maidr#687 and xability/r-maidr#287.

### DIFF
--- a/docs/SCHEMA.md
+++ b/docs/SCHEMA.md
@@ -121,7 +121,7 @@ Or multiple plots:
 
 Use the following to define the object properties:
 
-- `type`: the type of plot. The declarable types are `alluvial`, `area`, `bar`, `box`, `boxen`, `bump`, `candlestick`, `chord`, `choropleth`, `contour`, `diverging_bar`, `dodged_bar`, `dot`, `dumbbell`, `error_bar`, `forest`, `funnel`, `gantt`, `gauge`, `heat`, `hexbin`, `hist`, `icicle`, `line`, `lollipop`, `manhattan`, `mosaic`, `network`, `pack`, `parallel_coordinates`, `pie`, `point`, `polar_area`, `radar`, `ridgeline`, `sankey`, `smooth`, `stacked_area`, `stacked_bar`, `stacked_normalized_area`, `stacked_normalized_bar`, `step`, `sunburst`, `sunflower`, `survival`, `tree`, `treemap`, `violin_box`, `violin_kde`, `volcano`, `waterfall`, `word_cloud`. `TraceType` in `src/type/grammar.ts` is the source of truth; `candlestick_delta` appears there but is built at runtime from a candlestick and a reference line, so it is not something a page declares.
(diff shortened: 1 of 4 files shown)


jooyoungseo:
Both observations taken — pushed in `a2f0217`.

**1. Duplicated `declarableTypes()`.** You quoted the rule I should have applied: the second caller now exists, so `test/scripts/schemaTypes.ts` holds it, along with `SCHEMA` and a shared `typesInBackticks()`.

**2. The regex divergence.** Worth more than a nit, and the shared helper is what settles it. Both copies were correct today and exactly one of them would have been wrong the first time a type carried a digit — silently, since the failure mode is a type quietly dropping out of a list that then still looks complete. Written into the helper's docblock so the next reader knows why it accepts digits.

## One thing you didn't raise that your review implies

Your description of the partition check is what made me look again: it is satisfied by **any** split that covers the enum, including a wrong one. Only two types were pinned to the actual boundary, so a contributor could misfile a type and leave both lists internally consistent with nothing to notice.

The stable set is now written out in full (`STABLE_AT_84D9003`) and the experimental set asserted as its complement. Verified by making the mistake:

```
# `area` moved from Experimental into Stable
Tests: 2 failed, 4 passed, 6 total
# restored
Tests: 6 passed, 6 total
```

The two that fail are the new ones. The partition check passes throughout — which is the gap, demonstrated rather than argued.

## On the boundary you couldn't verify

Fair, and worth recording rather than leaving as trust: `git show 84d9003:src/type/grammar.ts` gives 16 `TraceType` values, of which `candlestick_delta` is runtime-only, so 15 stable + 37 experimental = the 52 declarable types. The full stable list is now in the test, so that derivation is checked on every run rather than only in the PR description.

Re-verified after the change: `npm test` 434 suites / **5950** tests, eslint and tsc clean.

---

PR #1209, opened 2026-08-28
fix(adapters): read a chart drawn sideways as one, across every oriented trace type

jooyoungseo:
# Pull Request

## Description

A chart drawn on its side was announced as an upright one, and read as one, in
six adapters and across most of the trace types that have an orientation at
all. Every number was on the chart and every number was against the other
one's axis, which is the failure `MaidrLayer.orientation` exists to prevent:
nothing throws and nothing is missing, the reader is simply told a different
chart.

The bar family has been getting this right since #997. What that fix did not
reach was everything else: histograms, box plots, violins, dot plots,
lollipops, dumbbells and error bars — and, in AnyChart, the bar chart itself.

Found by driving the built adapters in headless Chromium and reading what MAIDR
announced, one oriented trace type at a time, upright and sideways.

Measured, before → after, on the first mark:

| chart | before | after |
|---|---|---|
| Plotly `{y: samples, type: 'histogram'}` | `vertical hist` — "Count is 0.5 through 1.5, Petal Length is 3" | `horizontal hist` — "Petal Length is 0.5 through 1.5, Count is 3" |
| Plotly `{x: samples, type: 'box'}` | `vertical box` — "Petal Length is Setosa, no Lower outlier(s) for Species" | `horizontal box` — "Species is Setosa, no Lower outlier(s) for Petal Length" |
| Vega-Lite `y: {bin: true}` | `vertical hist` — "Count is 0 through 1, …" (bins lost entirely) | `horizontal hist` — "Petal Length is 6 through 7, Count is 2" |
| Highcharts `histogram` + `inverted` | `vertical hist` | `horizontal hist` |
| Highcharts `boxplot` / `scatter` / `lollipop` / `dumbbell` + `inverted` | `vertical …`, no orientation at all | `horizontal …`, read against the axes as drawn |
| Highcharts `errorbar` + `inverted` | `horizontal error_bar` — "Value is alpha, value Group is 10" | "Group is alpha, value Value is 10" |
| Chart.js `boxplot` / `violin` + `indexAxis: 'y'` | `vertical box` / `violin_box` + `violin_kde` | all three `horizontal` |
| ECharts `boxplot` + `yAxis: {type: 'category'}` | `vertical box` | `horizontal box` |
| AnyChart `anychart.bar()` (and its `marker` / `stick` / `range-bar` marks) | identical to `anychart.column()` — `vertical bar` | `horizontal …`, magnitude in `x`, titles swapped |
| Victory `<VictoryBoxPlot horizontal>`, `<VictoryErrorBar>` in a horizontal chart | no orientation, labels swapped without it | `horizontal`, labels and reading agreeing |

## Related Issues

None filed; the bug was reported directly, first for bar charts and histograms
and then for "violins and elsewhere too", which is what the sweep above
answers. It is the same class as #997 (a Highcharts bar on an inverted chart)
and the one `test/model/histogramOrientation.test.ts` pins for the core model.

## Changes Made

Nine commits, one per adapter and concern, each with its own tests and a line
of documentation. What moves in each is decided by the table on
`MaidrLayer.orientation`, not by taste: **the bar family exchanges `x` and
`y`; nothing else does**, and a type that exchanges nothing still needs the key
to be read and announced correctly.

- **Plotly** — the histogram now resolves `orientation` from `_fullData`
  (plotly sets it even when the author never wrote one), transposes the payload
  and reads the bin size from `ybins`; the box plot declares the orientation
  plotly already resolved.
- **Vega-Lite** — `binX` and `binY` compile to the same `bar` mark, and the
  compiled bin columns were looked up under the x field's name for both. For a
  `binY` spec none were found, so every bin fell back to the count's own extent
  — "0 through 1" for all of them. The binned channel now decides, not
  `isHorizontalEncoding`: both of a histogram's channels are quantitative.
- **Highcharts** — `chart.inverted` now reaches the histogram, the box plot,
  the dot plot, the lollipop and the dumbbell, and the error bar and forest
  plot swap their axis titles with the axes they name. Highcharts calls the
  category axis `xAxis` whichever way it draws the chart, which is why the swap
  is needed there and not in libraries that title the axes as drawn.
- **Chart.js** — `indexAxis: 'y'` now reaches the box plot and both violin
  layers.
- **ECharts** — a category **y** axis, which the bar family already reads, now
  reaches the box plot.
- **AnyChart** — `anychart.bar()` is the sideways chart and `anychart.column()`
  the upright one; the adapter treated them as the same. Every mark inside a
  bar chart follows, and `barmekko` deliberately does not: despite the name it
  draws its categories across the page.
- **Victory** — `horizontal` reached only the bar family, so a box plot and an
  error bar had their axis labels swapped without the key that tells the trace
  the labels moved.

Not applicable — the change is in what is announced, not in what is drawn. The
announcements are quoted in the table above.

## Checklist

## Additional Notes

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all
pass (6011 unit + 138 component/ESM tests, 443 suites). Each new test file
fails on `main`: 32 of the 61 new cases fail without the fixes.

Three things found on the way and deliberately left alone, each a decision
rather than an oversight:

- **The funnel family's orientation is inconsistent across adapters, and the
  convention itself is unsettled.** A classic funnel — stages stacked down the
  page, each stage's width the magnitude — is `horz` by the grammar's own
  definition, which is what the Plotly adapter emits. The ECharts, Highcharts
  and amCharts adapters call the same shape `vert`. Settling that means
  choosing what a reader should be told a funnel is, and it changes navigation
  and audio for every funnel already in use; it deserves its own issue rather
  than a quiet change here.
- **amCharts does not detect a horizontal histogram or dumbbell at all.** Its
  histogram detection requires `openValueXField`, so a chart binned up y falls
  through to `bar`; a horizontal dumbbell matches the gantt signature exactly.
  Those are missing readings rather than wrong ones.
- **The D3 binders cannot declare an orientation for a histogram, a mosaic or a
  candlestick**, though `bindD3Bar` takes one. That is a new option on a public
  config rather than a fix, so it is left for a follow-up.

One observation worth a maintainer's eye: a horizontal histogram's bins are
announced in different orders by different adapters. Plotly and Highcharts emit
them ascending (bottom-to-top as drawn); Vega-Lite's `binY` rows arrive
descending, which is top-to-bottom as drawn and matches what
`sortSimpleBarsByVisualOrder` does for horizontal bars. Both are defensible and
neither is this bug.


jooyoungseo:
Both review nits are addressed in `0d41a06`:

1. **Dead default parameter** — dropped. `extractHistogramData`'s `isHorizontal` is now required; the one call site already has to work the answer out for itself to set `orientation`, so the default was a second place the same question could be answered.
2. **Test coverage gap** — the Plotly and Vega-Lite histogram tests now assert `axes` in both orientations. That neither adapter swaps the labels is a fact about plotly's and Vega-Lite's own axis naming rather than an accident, and it is worth pinning where the Highcharts test pins the opposite.

Two notes on the state of the branch, since the review picked up a snapshot of it:

- **The PR body is current.** It was rewritten before the second review job started, and now covers all of the box plot, violin, dot, lollipop, dumbbell and bar work as well as the histogram — the branch grew after the first review because the reporter asked for the same sweep across every trace type that has an orientation.
- **The `Build Project` failure on `87de73d` was infrastructure, not this diff.** It died in `actions/setup-node@v4` with `socket hang up` while resolving the Node LTS manifest, seven seconds in and before any build step ran. The push above starts a fresh run rather than re-running that one. Locally `npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all pass on this head (6013 unit + 138 component/ESM tests, 443 suites).

---

PR #1211, opened 2026-08-29
fix(funnel): read a funnel by the direction its magnitude runs

jooyoungseo:
# Pull Request

## Description

#1209 left one question open: the funnel family's orientation was inconsistent
across adapters, and I could not say which reading was right. This settles it
against the published conventions and against measured geometry, and closes the
inconsistency.

**The convention.** Orientation names the direction the **magnitude** runs, not
which axis happens to be called `x`. Every definition of a horizontal bar chart
says so — bars running left to right, categories down the y axis — and so does
every library that offers the switch:

| library | API | what it says |
|---|---|---|
| Chart.js | `indexAxis: 'y'` | "horizontal bars"; y holds the categories, x the values |
| Highcharts | `chart.inverted` | "invert the axes so that the x axis is vertical and y axis is horizontal" |
| Matplotlib 3.10 | `orientation='horizontal'` (replacing `vert=False`) | "plots the boxes horizontally" |
| Plotly | `orientation='h'` | "the value of each bar spans along the horizontal" |

That is the rule MAIDR's `MaidrLayer.orientation` was already written against,
so **nothing merged in #1209 or #1210 needs reverting** — bars, histograms, box
plots, violins, dot plots, lollipops, dumbbells and error bars all follow it.

**Where two libraries name the other direction.** ECharts' funnel `orient` and
amCharts' `orientation` name the way the *stages progress*. Measured in
Chromium for values 100/60/20:

| chart | uniform | proportional to the value | magnitude runs |
|---|---|---|---|
| ECharts `orient: 'vertical'` (default) | heights 80, 80, 80 | widths 360, 216, 72 | horizontally |
| ECharts `orient: 'horizontal'` | widths 120, 120, 120 | heights 240, 144, 48 | vertically |
| amCharts `orientation: 'vertical'` (default) | heights 100, 100, 100 | widths 500, 300, 100 | horizontally |
| amCharts `orientation: 'horizontal'` | widths 160, 160, 160 | heights 320, 192, 64 | vertically |

So both defaults draw the value as a band's **width** and were announced as
upright funnels — while the identical figure from Plotly was announced, right,
as a horizontal one. The same chart now reads the same way whichever library
drew it:

```
before  ECharts: "vertical funnel"    "Stage is Visits, Y is 100"
after   ECharts: "horizontal funnel"  "Stage is Visits, Count is 100"
        Plotly:  "horizontal funnel"  "Stage is Visits, Count is 100"
```

## Related Issues

Follow-up to #1209, which flagged this as needing its own decision, and to
#1210. No issue filed; the question came from the maintainer directly.

## Changes Made

- **`src/adapters/echarts/single.ts`** — read `series.orient`; the default
  emits `horz` with the count in `x`, `orient: 'horizontal'` stays upright.
- **`src/adapters/amcharts/adapter.ts`** — same for `FunnelSeries`, keyed on
  the series class so its siblings are not swept in.
- **`src/adapters/plotly/extractor.ts`** — name a funnel's two dimensions when
  the layout names neither, which is always: a funnel's panel carries no axis
  titles, so the count was announced as "X".
- **`src/type/grammar.ts`** — write the convention down where `orientation` is
  defined, with the four library statements, so the next adapter reads its
  chart rather than its API's vocabulary.
- Tests for each, and the two docs pages that describe these funnels.

**Deliberately left alone, having measured them:**

- **Highcharts** — "Each element of the funnel has a height that relates to the
  value", and measured 166/100/33 for 100/60/20. Height-encoded, so its upright
  reading was already right.
- **AnyChart** — measured 172/103/34 against widths 343/198/147: height-encoded
  too.
- **An amCharts pyramid** — *neither* dimension is proportional to the value
  (measured 373/471/500 and 239/63/18), because a pyramid encodes it as an
  **area**. It has no magnitude axis to name, so it keeps none rather than
  being given a direction it does not have. Its pictorial-stack sibling goes
  with it.

Not applicable — the change is in what is announced, and the announcements are
quoted above.

## Checklist

## Additional Notes

`npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` all
pass (6020 unit + 142 component/ESM tests, 455 suites).

Three existing expectations changed with the behaviour rather than being
worked around — the amCharts funnel payload, its axis-label override, and the
ECharts stage order — each now carrying the measurement that decides it.

One limit: the amCharts geometry was read through the series' own API
(`slice.width()` / `slice.height()`) rather than from the DOM, because amCharts
5 draws to canvas. The ECharts, Highcharts and AnyChart figures were measured
from their rendered SVG.

---
_Generated by [Claude Code](https://claude.ai/code/session_0175KUdwfNYKRoqirPMiyALD)_

### DIFF
--- a/docs/amcharts.md
+++ b/docs/amcharts.md
@@ -152,6 +152,8 @@ A `PieSeries` lives in amCharts' separate `percent.js` module and is bound to no
 
 A `FunnelSeries` lives in the same `percent.js` module, inside a `SlicedChart` rather than a `PieChart`, and is likewise bound to no axis — its dimensions default to `Stage` and `Value`. Its pyramid and pictorial-stack siblings carry the same ordered `category`/`value` stages and are read the same way.
 
+**A default funnel is announced as horizontal, and that is not a contradiction of amCharts' own `orientation`.** The two settings name different directions: amCharts' names the direction the stages progress, MAIDR's names the direction the magnitude runs — the convention by which a chart with its bars running left to right is a *horizontal* bar chart. Measured through the series' own API for values 100/60/20, a default `FunnelSeries` draws uniform band heights of 100 against widths of 500/300/100: the value is a horizontal extent, so MAIDR reads it with the count in `x` and the stage in `y`. Set `orientation: "horizontal"` and the two flip, giving the layer MAIDR calls upright. A `PyramidSeries` is neither — its bands encode the value as an *area*, with no dimension proportional to it — so it is left with no orientation at all.
+
 `RadarLineSeries` and `RadarColumnSeries` need `radar.js` on top of `xy.js`; a `RadarChart` extends `XYChart`, so the binder finds it with the rest.
 
 The **declared** rows are declared rather than detected — see [Declaring What a Chart Means](#declaring-what-a-chart-means) below.
--- a/docs/echarts.md
+++ b/docs/echarts.md
@@ -80,8 +80,18 @@ so the reading is that pair.
 | ECharts | read as | highlighted |
 |---|---|---|
 | `pie` | `pie` | yes — one selector naming every slice |
-| `funnel` | `funnel` | yes — one selector per stage |
+| `funnel` | `funnel` | yes — one selector per stage. `orient` is read; see below |
 | `gauge` | `gauge` | **no** — see below |
+
+**A funnel's `orient` and MAIDR's `orientation` name opposite directions, on
+purpose.** ECharts names the direction its stages progress; MAIDR names the
+direction the magnitude runs — the convention by which a chart with its bars
+running left to right is a *horizontal* bar chart. Measured in Chromium for
+values 100/60/20, `orient: 'vertical'` (the default) draws uniform band heights
+of 80 against widths of 360/216/72, so the value is a horizontal extent and the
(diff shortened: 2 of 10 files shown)


jooyoungseo:
Fixed in 264e4cf — you were right, and I checked all four rather than just the one you named:

| adapter | funnel axis fallback |
|---|---|
| Highcharts | `Stage` / `Count` (`src/adapters/highcharts/adapter.ts:2925-2926`) |
| ECharts | `Stage` / `Count` (`src/adapters/echarts/single.ts:147-148`) |
| AnyChart | `Stage` / `Count` (`src/adapters/anychart/converters.ts:6088`) |
| amCharts | `Stage` / **`Value`** (`src/adapters/amcharts/adapter.ts:1251-1252`) |

So the pair this PR picks matches three of them, not all of them, and my comment named the wrong three. Both comments now say which is which, and say that amCharts is left as it is: renaming what a chart announces to a reader is a separate change from reading it the right way round, and it is not this fix's to make. Happy to open the follow-up if you'd like `Value` aligned to `Count`.

`npx eslint`, `npx tsc --noEmit` and `npx jest test/adapters/plotly` (326 passed) are clean on the pushed head.

---

PR #1213, opened 2026-09-04
docs: add npm, jsDelivr, CI, license and docs badges

jooyoungseo:
## Description

The README opened with the logo and went straight into prose. Nothing on it said the package is on npm, what version it is, whether CI is green, or where the docs live. For a library at 4.6.0 with 194 releases, that is a lot of trust signal left on the floor.

Six badges, added inside the existing centred block under the logo. Every value below was parsed out of the returned SVG rather than assumed:

| badge | renders |
|---|---|
| `img.shields.io/npm/v/maidr` | `npm \| v4.6.0` |
| `img.shields.io/npm/dt/maidr` | `downloads \| 25k` |
| `img.shields.io/jsdelivr/npm/hm/maidr` | `jsdelivr \| 1.2k/month` |
| `actions/workflows/ci.yml/badge.svg` | `CI \| passing` |
| `img.shields.io/npm/l/maidr` | `license \| GPL-3.0-or-later` |
| docs badge | static link to maidr.ai |

### Why jsDelivr as well as npm

`py-maidr` and `r-maidr` both load `maidr.js` from `cdn.jsdelivr.net` rather than installing from npm, so the npm download count misses most of the real consumption. The two numbers side by side describe this package's usage far better than either does alone — and it is the one badge here that is specific to how maidr is actually delivered.

### Lifetime rather than monthly downloads

`npm/dt` (25k) over `npm/dm` (2.1k/month), matching the py-maidr and r-maidr READMEs. A lifetime count does not visually regress on a quiet month.

## Deliberately not added

A bundle-size badge. Bundlephobia has been unreliable for a while and a badge that fails to resolve reads as neglect rather than as a missing integration — the same reason the r-maidr PR skipped a codecov badge that renders "unknown".

## Type of Change

- [x] Documentation update

## Checklist

- [x] Self-reviewed
- [x] No new warnings
- [x] Every badge URL verified to return a real value, not a placeholder or an error
- [x] Meaningful `alt` text on each badge — this is an accessibility project, and an empty `[]` would be the wrong thing to ship here
- [x] `node scripts/conflictMarkers.js` clean (1348 tracked files)

README-only; no `src/`, no `package.json`, no generated instruction files, so `sync:copilot:check`, `type-check`, eslint and the test suite are untouched by this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,14 @@
   <img src="docs/logo.svg" width="350px" alt="A stylized MAIDR logo, with curved characters for M A, a hand pointing for an I, the D character, and R represented in braille."/>
   <hr style="color:transparent" />
   <br />
+
+[![npm](https://img.shields.io/npm/v/maidr.svg)](https://www.npmjs.com/package/maidr)
+[![npm downloads](https://img.shields.io/npm/dt/maidr.svg)](https://www.npmjs.com/package/maidr)
+[![jsDelivr hits](https://img.shields.io/jsdelivr/npm/hm/maidr.svg)](https://www.jsdelivr.com/package/npm/maidr)
+[![CI](https://github.com/xability/maidr/actions/workflows/ci.yml/badge.svg)](https://github.com/xability/maidr/actions/workflows/ci.yml)
+[![License](https://img.shields.io/npm/l/maidr.svg)](LICENSE)
+[![Docs](https://img.shields.io/badge/docs-maidr.ai-1f6feb.svg)](https://maidr.ai)
+
 </div>
 
 # MAIDR: Multimodal Access and Interactive Data Representation
--- a/README.md
+++ b/README.md
@@ -3,12 +3,12 @@
   <hr style="color:transparent" />
   <br />
 
-[![npm](https://img.shields.io/npm/v/maidr.svg)](https://www.npmjs.com/package/maidr)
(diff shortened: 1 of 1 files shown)


jooyoungseo:
Alt text tightened in ef98997c — the nit was correct, and worth spelling out why, because there is a detail that makes it sharper than it first looks.

shields.io *does* embed the live value in the SVG, as `<title>CI - passing</title>`. That title is only reachable when the SVG is inlined; an SVG loaded through `` exposes its `alt` attribute and nothing inside the file. So for a badge in a README the alt attribute is not merely the better channel, it is the only one — a sighted reader gets `v4.6.0` and `passing` from the image, and everyone else got the bare word `npm`.

Where I went a step further than the suggestion: the alt now names **what each badge reports** rather than asserting a value.

| before | after |
|---|---|
| `npm` | `npm version` |
| `npm downloads` | `npm total downloads` |
| `jsDelivr hits` | `jsDelivr monthly hits` |
| `CI` | `CI build status` |
| `License` | `Package license` |
| `Docs` | `Documentation site` |

`CI build status` rather than `CI passing` is the deliberate part. Badge values are dynamic and alt text is static, so any alt that states a value is correct only until the value changes — `CI passing` would keep saying "passing" on a red build, which is worse than the terse version I started with. Naming the measurement stays true in both states and still tells the reader what the badge is for.

`npm total downloads` and `jsDelivr monthly hits` also carry the distinction the PR body argues for: the two numbers measure different things, and a reader who cannot see them should not have to guess which is which.

PR #1215, opened 2026-09-06
docs: add per-page SEO metadata to the API reference and unify the site's entity graph

jooyoungseo:
# Pull Request

## Description

SEO and generative-engine (Google AI Overviews / AI Mode) optimization of https://maidr.ai, following Google's guide [Optimizing your website for generative AI features on Google Search](https://developers.google.com/search/docs/fundamentals/ai-optimization-guide). Google's position is that AI features need nothing beyond standard SEO: indexed, snippet-eligible pages, structured data that mirrors visible text, and no special AI files (llms.txt neither helps nor hurts). The same pass is applied to py.maidr.ai (xability/py-maidr#768) and r.maidr.ai (xability/r-maidr#299) so the three sites describe one set of entities.

maidr.ai was already in decent shape (canonicals, descriptions, JSON-LD). The gaps were the 1,587 TypeDoc pages (no canonical, a generic "Documentation for MAIDR Documentation" description, version-suffixed titles that churn every release, no structured data), a home page titled "Home - MAIDR", `docs/echarts.md` built twice with the sitemap pointing at the copy nothing links to, sitemap `lastmod` values that were just the build date, an Organization node that disagreed with the sibling sites, and a BreadcrumbList with no visible breadcrumb.

## Related Issues

None filed; companion PRs xability/py-maidr#768 and xability/r-maidr#299.

## Changes Made

- **TypeDoc (`typedoc.json`, `scripts/typedoc-seo-plugin.mjs`, `scripts/add-navbar-to-typedoc.js`):** `hostedBaseUrl` gives an `api/sitemap.xml` and the index canonical; the project is named "MAIDR JavaScript API" without the version so titles stay stable across releases. A small local plugin (`head.end` hook) adds to every API page a self-referential canonical, a description taken from the symbol's doc comment (so it matches the visible text), Open Graph tags, and TechArticle plus BreadcrumbList JSON-LD with stubs of the site-wide nodes so every `@id` resolves on-page. The post-processor strips TypeDoc's hard-coded description so exactly one remains, and rewrites the API sitemap's index URL and `lastmod` (last commit touching `src/`). `docs/robots.txt` lists both sitemaps.
- **`scripts/build-site.js`:** one `INTEGRATION_PAGES` list drives building, exclusion from the generic docs loop and the sitemap (ECharts is built once and listed once); an optional SEO title per page (home: "MAIDR: Accessible Data Visualization with Sonification, Braille and Text"; integrations: "<Library> Accessibility Integration - MAIDR"); a missing description is now a build failure; `<lastmod>` and TechArticle dates come from git commit dates (both docs workflows check out full history); `changefreq`/`priority` dropped because Google ignores them; a visible "Last updated" line on docs pages mirrors the TechArticle date.
- **`docs/template.html`:** JSON-LD rewritten to the entity graph shared with the sibling sites (the (x)Ability Design Lab with its illinois.edu URL and parent university, JooYoung Seo, a dual-typed software node with license, runtime, `alternateName`, `maintainer` and, on the home page, `citation` to the two MAIDR papers by DOI, plus a WebPage with `relatedLink` to py.maidr.ai and r.maidr.ai). A visible breadcrumb nav (`aria-label="Breadcrumb"`, AA contrast) now backs the BreadcrumbList, the primary nav is labelled "Main", and a footer line names the lab and PI so the site-wide nodes have visible text on every page.
- **Examples gallery (`scripts/examplesGallery.js`):** links are real `href="examples/..."` anchors (keyboard and crawler friendly) with the same click handler; the 260 demo pages under `examples/` get `<meta name="robots" content="noindex">` at build time since they have no titles or descriptions. `examples.html` itself stays indexable.
- **README.md:** a definition paragraph under the H1 (what MAIDR is, which libraries it works with, who builds it, its relation to py-maidr and maidr for R), "Binders" renamed to "Related projects" with descriptive cross-links, and plain-text citations with DOI links before each BibTeX block. `docs/llms.txt` now lists every integration guide.
- **Workflows:** `docs.yml` also runs after the "Release" workflow (its commit carries `[skip ci]`, so the site never rebuilt at release time) and both docs jobs use `fetch-depth: 0`. `scripts/check-page-sizes.js` runs last in `npm run docs` and fails if any indexable page exceeds 1,900,000 bytes (Googlebot reads only the first 2 MB of an HTML file); the largest indexable page is 861 KB.

Not applicable.

## Checklist

## Additional Notes

- Verified locally: `npm run lint`, `npm run type-check`, `npm run build`, `npm run docs` (all 1,587 API pages carry exactly one description and a canonical; every JSON-LD `@id` on the home, docs and API pages resolves on the same page), and the `test/scripts` suites (gallery and site-anchor tests updated).
- The TypeDoc hierarchy theme inlines the whole navigation tree (~600 KB) into every API page. That stays under the 2 MB limit but dilutes each page's unique text; switching to the default theme would cut API pages to a few KB each. Left for a separate decision.
- The `api/index.html` entry was removed from the root sitemap because the generated `api/sitemap.xml` lists it.

https://claude.ai/code/session_01JqUywJzoCqyMGLFp8De1mn

---
_Generated by [Claude Code](https://claude.ai/code/session_01JqUywJzoCqyMGLFp8De1mn)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -309,6 +309,9 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          # Full history, as in docs.yml: the site build reads git commit dates.
+          fetch-depth: 0
 
       - name: Set up Node.js
         uses: actions/setup-node@v4
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -14,9 +14,23 @@ on:
       - 'examples/**'
       - 'media/**'
       - 'scripts/build-site.js'
+      - 'scripts/markdown.js'
+      - 'scripts/examplesGallery.js'
+      - 'scripts/typedoc-seo-plugin.mjs'
+      - 'scripts/add-navbar-to-typedoc.js'
+      - 'scripts/check-page-sizes.js'
       - '.github/workflows/docs.yml'
 
   workflow_dispatch:
+
+  # The semantic-release commit is "chore(release): X [skip ci]", which GitHub
+  # never runs `on: push` workflows for, so the site did not pick up a new
+  # version until the next push that touched a docs path. Rebuild when the
+  # Release workflow finishes instead; the job below skips a failed release.
+  workflow_run:
+    workflows: ["Release"]
+    types: [completed]
+    branches: [main]
 
 permissions:
   contents: read
@@ -30,9 +44,14 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-latest
+    if: ${{ github.event_name != 'workflow_run' || github.event.workflow_run.conclusion == 'success' }}
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          # Full history: scripts/build-site.js derives sitemap <lastmod> and
+          # TechArticle dates from git commit dates.
+          fetch-depth: 0
 
       - name: Setup Node
         uses: actions/setup-node@v4
--- a/.github/workflows/pr-preview-build.yml
+++ b/.github/workflows/pr-preview-build.yml
@@ -39,6 +39,9 @@ jobs:
           # Untrusted PR code is checked out only to be BUILT; don't leave the
           # job token in .git/config where a build script could read it.
           persist-credentials: false
+          # Full history: build-site.js dates each page from its last commit,
+          # so a shallow clone would stamp the preview with the build date.
+          fetch-depth: 0
 
       - name: Set up Node.js
         uses: actions/setup-node@v4
--- a/README.md
+++ b/README.md
@@ -14,10 +14,11 @@
 
 # MAIDR: Multimodal Access and Interactive Data Representation
 
-- **Note:** `maidr` package has been completely rewritten in TypeScript for better architecture and performance. The previous version is now archived at [xability/maidr-legacy](https://github.com/xability/maidr-legacy).
+MAIDR (Multimodal Access and Interactive Data Representation, pronounced "mader") is an open-source JavaScript/TypeScript library that makes statistical charts accessible to blind and low-vision people. It adds keyboard navigation, sonification, text descriptions, braille output and AI-generated descriptions to charts drawn with React, Recharts, Plotly, Vega-Lite, D3, Chart.js, Apache ECharts, Highcharts and other libraries, or described directly with its JSON data schema. MAIDR is developed by the [(x)Ability Design Lab](https://xabilitylab.ischool.illinois.edu/) at the University of Illinois Urbana-Champaign and is the engine behind [py-maidr](https://py.maidr.ai/) for Python and [maidr for R](https://r.maidr.ai/).
 
-`maidr` (pronounced as 'mader') is a system for non-visual access and control of statistical plots.
-It aims to provide an inclusive experience for users with visual impairments by offering multiple modes of interaction:
+- **Note:** The `maidr` package has been completely rewritten in TypeScript for better architecture and performance. The previous version is now archived at [xability/maidr-legacy](https://github.com/xability/maidr-legacy).
+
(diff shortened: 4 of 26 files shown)


jooyoungseo:
Thanks for the review. All four findings are addressed in `c0461dc`.

**1. No tests for the new logic.** Added `test/scripts/siteSeo.esm-test.ts` (12 cases), following the `examplesGallery` / `siteAnchors` pattern. To make the logic testable without running the build, three pieces moved into their own modules with hand-written declarations beside them, as `examplesGallery.js` does:

- `scripts/gitDates.js`: dates are checked against a fixture repository built in the test with known commit dates, including a rename, plus the fallbacks for a path git has never seen and for a directory outside any checkout.
- `scripts/pageSizes.js`: the limits, the exemption, and that each offender reports the limit it broke.
- `scripts/typedocSeo.js`: truncation at a word boundary within the limit, and the no-comment fallback naming kind, symbol and module.

`generatePage` throwing on a missing description is covered by the build itself: the docs build fails outright, which is what the `throw` is for.

**2. `api/` fully exempt from the size guard.** Agreed, and fixed rather than argued: those pages are indexable, so they are now held to Googlebot's cutoff itself (2,097,152 bytes) instead of being skipped, with the 1.9 MB headroom limit kept for ordinary pages. Only the noindex example demos stay exempt. The guard now prints which limit each offender broke.

**3. Integration guides had no TechArticle or "Last updated".** Fixed: the `INTEGRATION_PAGES` loop now threads both, exactly as the docs loop does, and a missing description there is a build failure too. Verified in the rebuild: `react.html` carries a TechArticle and `Last updated 2026-08-23`, with all four `@id` references resolving on the page.

**4. `firstCommitDate` without `--follow`.** Fixed; a renamed page keeps its real `datePublished` instead of silently falling back to its last commit date. The fixture repository in the new test renames a file specifically to cover this.

Local run after the changes: `npm run lint`, `npm run type-check`, `npm run build` and `npm run docs` all clean, `test/scripts` at 306 tests across both projects. The size guard reports the largest indexable page at 861 KB, with the three single-file example bundles (2.05 MB, 1.84 MB, 1.05 MB) exempt and noindex.

On the sandbox note: CI here ran the full suite on this PR and every check is green, including Docs Build, E2E and Unit Tests.

---

jooyoungseo:
Both non-blocking items are taken, in `c51ecfc`.

**1. Inconsistent JSON-LD escaping.** Agreed, and worth closing precisely because the two paths disagreed on a safety measure rather than because either is exploitable today. `scripts/jsonLd.js` now holds the escape and both emitters use it, so `buildBreadcrumbSchema`, `buildTechArticleSchema` and `graphNode` in `build-site.js` escape `<` exactly as the TypeDoc plugin already did. The plugin is the path that will read less predictable text first, since its descriptions come from doc comments, so it makes sense for the shared helper to carry the stricter behaviour. A test asserts a value containing `</script>` cannot close the block and still round-trips through `JSON.parse`.

**2. `fixSitemap()` duplicating the date lookup.** Fixed: it imports `lastCommitDate` from `scripts/gitDates.js` rather than repeating the `git log -1 --format=%cs` call and its fallback. `execFileSync` is no longer imported there.

Verified after the change: `npm run lint`, `npm run type-check` and `npm run docs` all exit 0, and `test/scripts` is at 308 tests across both projects. The rebuild reports every indexable page within its limit, and re-parsing the emitted structured data gives no dangling `@id` on any sampled page:

```
_site/index.html                                  -> 1 block,  7 ids, dangling: none
_site/react.html                                  -> 3 blocks, 4 ids, dangling: none
_site/docs/SCHEMA.html                            -> 3 blocks, 4 ids, dangling: none
_site/api/classes/controller.Controller.html      -> 2 blocks, 3 ids, dangling: none
```

On the sandbox note in your review: CI has run the full suite on this PR, and every check was green on the previous head, including Docs Build, Unit Tests and E2E.

---

PR #1221, opened 2026-09-06
fix(service): sound the data tones a suspended context swallowed, and four more service defects

jooyoungseo:
# Pull Request

## Description

Seven defects in the audio, braille, text and format services, found in a codebase audit. The first is the one a reader would notice immediately: an `AudioContext` created while the page has programmatic focus is born suspended, and only the cue paths ever call `resume()`. Every data tone was scheduled against a frozen clock and simply never heard, so navigation was silent until the reader happened to hit a boundary or open a dialog.

Each fix has a test that failed before the change.

## Related Issues

None.

## Changes Made

- **audio**: a data tone defers behind `resume()` when the context is not running, the way the cue tones already do. The sonification body moved into a private method so the whole path can be deferred as a unit; it re-checks the mode and the context state after the async gap. A volume-0 update still neither resumes nor claims the shared cue slot.
- **audio**: turning sound off cancels the chord chain and its queued echoes, and the chain's own timer re-checks the mode. A 30-point scatter column used to keep playing for a second and a half over the "Sound is off" announcement.
- **braille**: the caret cannot be forwarded to a row's newline sentinel, which maps one past the last data column. End, a click past the last cell or a display's cursor-routing key used to hand the model a column that does not exist. The guard is deliberately narrow: only a newline that is the row's last cell entry and is this index, so a mid-row wrap and a box row terminator still map back to their real cells and End still lands on the last cell of the line.
- **braille**: the box encoder never squeezes a summary section out of existence. On a narrow display a box with many outliers drove min, max, q1 and q3 to zero cells: they rendered nothing while still claiming a cursor cell, so the minimum's cell became the quartile's glyph and later cells pointed past the end of the row, onto the next box's line. Overflow now comes out of the outlier runs instead.
- **text**: a layer-level out-of-bounds cue no longer overwrites the stored state. Page Up on a single-layer chart left `getCoordinateText()` null, so the AI chat's "current position" went blank on a keypress that did nothing.
- **format**: a chart-supplied `format.function` body is called inside a try/catch and falls back to the default format. A body that threw for one value on the axis took the rest of the observer chain with it, so the reader got the tone but no text, no highlight and no review update.
- **format**: the `Intl` formatters are built once per axis instead of once per announced value. They are constructed on first use rather than eagerly, so a currency code or locale the chart got wrong still fails where the new guard can turn it into a fallback.

N/A

## Checklist

## Additional Notes

Four existing braille tests placed the cursor on the sentinel and asserted the model was moved there; they now sit on the row's last real column and assert the same round trip, as the commit body explains.

`npm run lint`, `npm run type-check`, `npm test` (6578 + 158) and `npm run build` pass locally.

---

### DIFF
--- a/src/service/audio.ts
+++ b/src/service/audio.ts
@@ -316,6 +316,43 @@ export class AudioService implements Observer<PlotState>, Disposable {
       return;
     }
 
+    // A suspended context reports currentTime === 0, so a data tone scheduled
+    // against it starts and stops at the same frozen instant and is simply
+    // never heard — the context is created from a timeout after focus-in, and
+    // when that focus is programmatic (a screen reader's virtual cursor,
+    // `autofocus`, a host page calling `.focus()`) it is born suspended and
+    // stays that way. The cue paths defer behind resume() one level down; the
+    // data tones defer here, so the first point of a chart is heard rather
+    // than dropped. Keyboard navigation is the user gesture resume() needs.
+    if (this.audioContext.state !== 'running') {
+      // At volume 0 the update is silent whichever branch it takes, so it must
+      // neither trigger resume() nor claim the shared deferred-cue slot —
+      // mirroring the outer guards in playEmptyTone and playMenuTone.
+      if (this.volume > 0) {
+        this.scheduleWhenRunning(() => this.playState(state));
+      }
+      return;
+    }
+
+    this.playState(state);
+  }
+
+  /**
+   * Sounds one plot state on a running {@link AudioContext}.
+   *
+   * Split out of {@link update} so the whole path — cue tones and data tones
+   * alike — can be deferred behind {@link AudioContext.resume} as a unit.
+   * Re-checks mode and context state because it can run after that async gap,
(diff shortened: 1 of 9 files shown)


jooyoungseo:
The **Unit Tests** failure on `aac5637` is not this pull request's.

The failing suite is `test/command/announce-position.test.ts`, two cases in `AnnouncePositionCommand on a multi-violin KDE`:

```
Expected: "Violin 2 of 3, Position is 3 of 5"
Received: "Violin 3 of 5, Position is 2 of 3"
```

Both fail identically on `main` at `d77521a` with nothing from this branch present, so it is a semantic conflict between two changes that had already merged: #1219 added the announcement, reading the position out of `audio.panning`, and #1220 then re-oriented a vertical violin's pan so the stereo image follows the on-screen x. This branch touches the audio, braille, text and format services and nothing in that path.

Fixed in #1223, on `main` directly. Once that lands I will merge `main` into this branch so CI re-runs against a green base. Holding off on pushing it here as well, since the merge is needed anyway and would carry the same change.

---

PR #1222, opened 2026-09-06
docs(seo): give reference managers real bibliographic metadata

jooyoungseo:
## Description

Saving any page of maidr.ai to Zotero produced a Web Page item with **no author and no date**. The site emitted only Open Graph and Twitter tags, and Zotero's Embedded Metadata translator looks for Highwire Press `citation_*` first, then Dublin Core, then Open Graph — finding neither of the first two, it fell back to the `<title>` alone.

## Changes Made

**`scripts/dublinCore.js`** (new, with hand-written `.d.ts` matching the `jsonLd.js` / `gitDates.js` convention) builds the block; `docs/template.html` gains a `{{DUBLIN_CORE}}` slot, and `scripts/typedoc-seo-plugin.mjs` emits the same pairs as JSX so the API reference is covered too.

Each page now carries title, creator, publisher, description, identifier, type, format, language, rights and its own git commit date.

- **Creators are surname-first** (`Seo, JooYoung`) so Zotero splits them into first and last names instead of storing one opaque string.
- **`DC.type` is per page**: the home page is `Software` because it stands for the library; the guides and the API reference are `Text`. Claiming `Software` site-wide would tell a reference manager that a guide page is the program.
- **`DC.title` drops the site name**, which `<title>`, `og:title` and `twitter:title` keep. `stripSiteName` removes only an exact trailing `<separator><siteName>`, so `React Accessibility Integration - MAIDR` is filed as `React Accessibility Integration` while a title that merely contains a separator keeps all of it. This matches the sibling py-maidr and r-maidr sites, whose generators append their own suffixes.
- **The home page gains a date** from `README.md`'s commit, which it did not have before — it has no breadcrumb to hang a visible "Last updated" line under, so the date existed nowhere.
- The plugin resolves the `src` commit date **once**, not once per page; the hook runs for ~1,600 API pages.

**`CITATION.cff`** gives GitHub its "Cite this repository" button, naming the CHI 2024 paper as the preferred citation and the EuroVis 2024 paper as a reference.

## Deliberately no `citation_*` tags

Google Scholar's [inclusion guidelines](https://scholar.google.com/intl/en/scholar/inclusion.html) reserve those tags for scholarly articles, require that a site consist "primarily of scholarly articles", and say explicitly that `citation_title` must not carry the name of a repository. Putting them on software documentation misrepresents the page and risks the site being dropped from Scholar, without gaining indexing — Scholar does not index documentation sites. The two MAIDR papers stay cited by DOI in the JSON-LD and the visible citation section, and reach Scholar through ACM and Eurographics.

The reasoning is recorded in the module docstring so it is not quietly undone later, and a test asserts no emitted tag name falls outside `DC.`.

Not applicable.

## Checklist

## Verification

- `npm run lint:fix`, `npm run type-check` and `npm run docs` clean.
- `npm test`: the `esm` project's 12 suites and 170 tests pass, which is where every test this PR adds lives. **The `unit` project is red on `main` and red here for the same reason** — see below.
- Built site checked: home emits `DC.type` `Software` dated from `README.md`; `react.html` emits `Text` dated `2026-08-23` from its own source; an API symbol page is filed as `HighlightOverlay` rather than `HighlightOverlay | MAIDR JavaScript API`.
- 12 cases in `test/scripts/siteSeo.esm-test.ts` cover the type split, the per-author tags, the surname-first shape, attribute escaping, omitting an empty date, the no-`citation_*` invariant, and the suffix stripping including the cases that must *not* strip.

### Correction to an earlier version of this description

This section previously claimed `npm test` was clean at "166 passed, 12 suites". That figure is the **`esm` project alone**. `scripts/test.js` runs one process per project and returns the first non-zero code; I read only the tail of its output and reported the last project's summary as the whole run. The real exit code was 1.

### The `Unit Tests` failure is not this PR's

`test/command/announce-position.test.ts` fails two multi-violin KDE cases:

```
Expected: "Violin 2 of 3, Position is 3 of 5"
Received: "Violin 3 of 5, Position is 2 of 3"
```

I checked out `origin/main` with none of this branch applied and it fails identically there (2 failed / 6627 passed). This branch touches `scripts/`, `docs/template.html`, `CITATION.cff` and one file under `test/scripts/`; it does not touch `src/command`, `src/model`, or anything the announcement reads.

#1223 fixes it against the same base commit, so this PR goes green when that merges. I have not ported it: a `src/command` change does not belong in a documentation PR, and both of us editing that file gains nothing.

https://claude.ai/code/session_01JqUywJzoCqyMGLFp8De1mn

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,75 @@
+cff-version: 1.2.0
+title: "MAIDR: Multimodal Access and Interactive Data Representation"
+message: >-
+  If you use this software, please cite both the software itself and the
+  MAIDR paper below.
+type: software
+authors:
+  - family-names: Seo
+    given-names: JooYoung
+    email: jseo1005@illinois.edu
+    affiliation: University of Illinois Urbana-Champaign
+repository-code: "https://github.com/xability/maidr"
+url: "https://maidr.ai/"
+abstract: >-
+  A TypeScript library that makes any SVG chart explorable by keyboard,
+  sonification, braille, text and AI description, and the engine behind
+  py-maidr and maidr for R.
+keywords:
+  - accessibility
+  - data visualization
+  - sonification
+  - braille
+  - screen reader
+  - blind
+  - low vision
+license: GPL-3.0-or-later
+preferred-citation:
+  type: conference-paper
+  title: >-
+    MAIDR: Making Statistical Visualizations Accessible with Multimodal Data
+    Representation
+  authors:
+    - family-names: Seo
+      given-names: JooYoung
+    - family-names: Xia
+      given-names: Yilin
+    - family-names: Lee
+      given-names: Bongshin
+    - family-names: Mccurry
+      given-names: Sean
+    - family-names: Yam
+      given-names: "Yu Jun"
+  collection-title: >-
+    Proceedings of the CHI Conference on Human Factors in Computing Systems
+    (CHI '24)
+  year: 2024
+  publisher:
+    name: Association for Computing Machinery
+  doi: 10.1145/3613904.3642730
+  isbn: "9798400703300"
+references:
+  - type: conference-paper
+    title: >-
+      Designing Born-Accessible Courses in Data Science and Visualization:
+      Challenges and Opportunities of a Remote Curriculum Taught by Blind
+      Instructors to Blind Students
+    authors:
+      - family-names: Seo
+        given-names: JooYoung
+      - family-names: O'Modhrain
+        given-names: Sile
+      - family-names: Xia
+        given-names: Yilin
+      - family-names: Kamath
+        given-names: Sanchita
+      - family-names: Lee
+        given-names: Bongshin
+      - family-names: Coughlan
+        given-names: "James M."
+    collection-title: "EuroVis 2024 - Education Papers"
+    year: 2024
+    publisher:
+      name: The Eurographics Association
+    doi: 10.2312/eved.20241053
+    isbn: "978-3-03868-257-8"
--- a/docs/template.html
+++ b/docs/template.html
@@ -21,6 +21,7 @@
   <meta name="twitter:description" content="{{DESCRIPTION}}" />
   <meta name="twitter:image" content="https://maidr.ai/media/logo.jpg" />
   <meta name="twitter:image:alt" content="MAIDR - Multimodal Access and Interactive Data Representation logo" />
+  {{DUBLIN_CORE}}
   <title>{{SEO_TITLE}}</title>
   <link rel="icon" type="image/svg+xml" href="{{BASE_PATH}}media/logo.svg" />
   <script type="application/ld+json">
--- a/scripts/build-site.js
+++ b/scripts/build-site.js
@@ -309,6 +309,7 @@ function generatePage({ title, content, activePage, basePath = '', slug = '', og
     // documentation about it, so only the home page is typed as Software.
     .replace(/\{\{DUBLIN_CORE\}\}/g, () => dublinCoreTags({
       title: seoTitle,
+      siteName: 'MAIDR',
       description,
       identifier: canonicalUrl,
       date: dateModified,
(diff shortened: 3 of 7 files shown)


jooyoungseo:
Both findings addressed, and the first one was a real defect rather than a doc slip.

**1 (title suffix) — fixed in `bfea065`, by changing the behavior rather than the docstring.**

You framed this as "the docstring should be updated to match what's passed", and self-consistency with `og:title`/`twitter:title` is a fair argument. But the deciding fact is outside this repo: the sibling py-maidr and r-maidr sites, whose PRs went out alongside this one, both strip the suffix their generators append (` – py-maidr`, ` • maidr`). These three sites are meant to describe one set of entities, so leaving maidr.ai as the one that files pages under a different convention would have made the record wrong across the family, not just inconsistent here.

The docstring was the honest half of the mismatch. `dublinCorePairs` now takes a `siteName` and strips it:

| page | before | after |
|---|---|---|
| `react.html` | `React Accessibility Integration - MAIDR` | `React Accessibility Integration` |
| an API symbol | `HighlightOverlay \| MAIDR JavaScript API` | `HighlightOverlay` |
| home | unchanged | unchanged |

`<title>`, `og:title` and `twitter:title` keep the suffix — that is what a browser tab and a link preview want, and a bibliographic record is neither.

Your point about a future editor "fixing" the call sites is why the stripping is now pinned by tests instead of left to a comment. Four new cases cover it, including the two edge cases the naive version would get wrong: a title that merely *contains* a separator keeps all of it (the home page's own title, and `Braille - Text - MAIDR` → `Braille - Text`), and a title that is nothing but the site name is left alone rather than emptied.

**2 (single-author `CREATORS`) — deliberately left to a maintainer.**

py-maidr and r-maidr each took their authors from a declared source (`authors` in `pyproject.toml`, `Authors@R` in `DESCRIPTION`). This repo has no equivalent: `package.json` declares no `author` or `contributors`, so the only sourced name is the `maintainer` in the existing JSON-LD. I could have read `git shortlog` and guessed at "significant contributors", but authorship is a credit decision, and guessing it wrong in a citation record is worse than listing fewer people. If you tell me who belongs there, or add an `author`/`contributors` field to `package.json`, both `CREATORS` and `CITATION.cff` should follow it.

**On the checks you could not run:** understood, and the constraint is worth stating rather than glossing. On this end `npm run lint:fix`, `npm run type-check`, `npm test` (170 passed, 12 suites) and `npm run docs` are all clean, and the table above is read from the rebuilt `_site/`, not predicted.

---

jooyoungseo:
**Unit Tests is red on this PR, and the failure is not this PR's.**

`test/command/announce-position.test.ts` fails two multi-violin KDE cases:

```
Expected: "Violin 2 of 3, Position is 3 of 5"
Received: "Violin 3 of 5, Position is 2 of 3"
```

I checked out `origin/main` with none of this branch's changes applied and ran the same file: **it fails identically there**, 2 failed / 6627 passed. This branch touches `scripts/`, `docs/template.html`, `CITATION.cff` and one test file under `test/scripts/`; it does not touch `src/command`, `src/model` or anything the announcement reads.

**A fix already exists and I have not ported it:** #1223, "fix(command): read the violin position from the navigation frame, not the pan", opened against the same base commit. It attributes the red to the interaction of #1219 and #1220 — one added the announcement reading position out of `audio.panning`, the other re-oriented a vertical violin's panning so the violin index moved into `panning.x` — and fixes it in `navigationPosition`. Porting a `src/command` change into a documentation PR would put both of us editing the same file with no benefit, since #1223 lands on the same base and this PR goes green the moment it merges. If it stalls, I'll port it rather than leave this red.

**Worth correcting in my own PR description while I'm here.** It claims `npm test` clean, "166 passed, 12 suites". That number is the `esm` project alone. `scripts/test.js` runs each project in its own process and returns the first non-zero code, and I read only the tail of its output, so I reported the last project's summary as the whole run. The real exit code was 1, from `unit`, for the failure above. The `esm` figure was accurate; calling it the full suite was not.

What is actually verified on this branch: `npm run lint:fix`, `npm run type-check` and `npm run docs` clean, and the `esm` project's 170 tests pass — which is where every test this PR adds lives.

---

PR #1226, opened 2026-09-06
fix(scatter): announce a streamed point where the trace actually keeps it

jooyoungseo:
# Pull Request

## Description

Two defects in the live-data append path, found in a codebase audit. Both were reproduced with a failing test first.

Streaming a point into a scatter layer announced the wrong point, or nothing at all. `appendData` used the raw insertion index as a column, but a scatter trace keeps its points sorted, so the two disagree the moment a point does not arrive in order. With the fix neutered, the reproduction throws `Cannot read properties of undefined (reading 'y')` and the controller swallows the announcement entirely.

A box layer had a related fault: the appended coordinates ignored the trace's section and position layout. A vertical box announced the section "Lower outlier(s)" with an empty value, and a horizontal box with eight boxes announced no section at all.

## Related Issues

None.

## Changes Made

- **scatter**: a streamed point is announced at the position the trace sorted it to, not at the index it arrived at. New public `ScatterTrace.positionOfDataIndex(index)` answers that question.
- **box**: appended coordinates follow the box trace's own section and position layout, so a streamed box is announced on a section it actually has. `BOX_SECTIONS` is exported for it; `BoxTrace` copies the array, since `dispose()` truncates its own.
- **liveData**: `AppendedPointInfo` gains `colShift`, how far a `maxWidth` trim moves the cursor along the column axis. The controller reads that instead of `trimmed`. New exported helper `appendedPointPosition(trace, appended)`.
- **docs**: `docs/LIVE_DATA.md` promised that a trim always carries the cursor onto the same data point. That is no longer true for a scatter or a horizontal box, and the page now says so, along with which reading of a multi-section item monitor mode announces.

N/A

## Checklist

## Additional Notes

Two limits are left in place on purpose, and the code comments claim only what is true:

- Violin box layers have the same class of bug, but their section list depends on the layer options, so they still get the flat default. That is its own change.
- A horizontal box's *row* cursor still drifts by one per trimmed box, because the reversal prepends the new box. There is only `activeColShift` and no row equivalent, and inventing one is beyond this fix.

A third limit was listed here and was wrong: it said a scatter in grid mode still announces the focused grid cell. `ScatterTrace.getStateAt` suspends grid and grid-cell mode along with point and intersection mode, so that limitation no longer exists. It was fixed in #1220 and the note was stale. It never reached the code or the docs.

One existing test changed: the two `AppendedPointInfo` literals in `test/service/liveData.test.ts` carry the new field. Their intent is unchanged.

`npm run lint:fix`, `npm run type-check` and `npm test` (6640 unit across 485 suites, 158 ESM across 12) all pass on this branch with `main` merged in.

### DIFF
--- a/docs/LIVE_DATA.md
+++ b/docs/LIVE_DATA.md
@@ -253,6 +253,7 @@ Notes for real feeds:
 On live charts (`live: true`), pressing **M** toggles **monitor mode**. While monitoring is on:
 
 - Every data point appended to the **focused layer** is **automatically sonified** (a tone at the point's pitch) and **announced** to screen readers (e.g. "Tick is 42, Reading is 3.14"). On multi-layer charts, appends to other layers stay silent — switch layers (PageUp/PageDown) to change which metric you monitor.
+- A trace that navigates sections rather than a plain list of points is announced at the one reading that stands for the new item: a **candlestick** at the new candle's close, a **box** at the new box's median.
 - The user's current position **does not move** — they can keep exploring historical data while hearing new points arrive, then jump to the live edge with `Ctrl/Cmd + Right Arrow`.
 - Toggling announces "Monitoring on" / "Monitoring off". On non-live charts, pressing M explains that monitoring is only available for live charts.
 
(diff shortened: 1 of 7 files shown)


jooyoungseo:
You are right about the grid-mode note, and thank you for checking it against the code rather than taking the description at its word.

I verified `ScatterTrace.getStateAt` on this branch: it saves and clears `isInGridMode` and `isInGridCellMode` alongside `isInPointMode` and `isInIntersectionMode`, and restores all four in the `finally`. So the limitation I listed does not exist. It was fixed in #1220 and my note was stale.

I checked whether the same claim had reached anywhere durable. It had not: the only mention of grid mode anywhere in this diff is a test fixture docblock saying a scatter with no axis ranges cannot enter grid mode, which is accurate. So the correction is confined to the pull request description, which I have updated to say what was wrong and why, rather than quietly deleting the bullet.

On the two minor points, both stand as they are:

The `indexOf` scan in the `ROW` branch runs once per live append, next to a `getStateAt` and an announcement, so an index map would buy nothing measurable and would be a third structure to keep in step with the constructor's grouping. Worth revisiting only if it gains a caller on the navigation path.

The test helper mirroring the controller's two calls rather than driving the real `Controller` matches the existing candlestick test. Changing only this one would leave the pair inconsistent, so if that wiring deserves direct coverage it should be a change that covers both.

---

PR #1229, opened 2026-09-06
fix(highlight): read a line's stroke width from the element that has one

jooyoungseo:
# Pull Request

## Description

Six defects in the highlight, high-contrast and pattern services, found in a codebase audit. Two are visible to a sighted reader; the rest are cost or dead code.

A line's highlight was **always 2px**, never the original width plus two. The width was read from the clone, which is detached from the document and therefore computes nothing from a stylesheet. A chart whose lines are styled in CSS got a highlight thinner than the line it was meant to mark.

The second one leaves the chart wrong after the fact: `setFigure()` during a live update captured the **already-applied** high-contrast colours as the "original" ones. Turning high contrast off then restored the chart to the high-contrast palette, and the original colours were gone for the rest of the session.

## Related Issues

None.

## Changes Made

- **highlight**: the clone's stroke width comes from the computed style already taken from the original element, which is the one attached to the document.
- **high contrast**: `setFigure()` restores before it re-captures. This needed a new private `highContrastApplied` flag, because the setting can be on while the effect is suspended on blur, so the setting alone cannot answer whether the DOM currently shows our colours.
- **high contrast**: restoring writes back the body's own inline declarations rather than its computed colours, so a page that set none is left with none.
- **high contrast**: the colour ramp and the text-ancestor walk are computed once per apply instead of once per element.
- **pattern**: one canvas context is reused for colour parsing instead of a fresh canvas and 2D context per call, twice per element.
- **pattern**: `removePattern` is removed. It had no callers anywhere in `src/`, `test/` or `e2e_tests/`, and would not have undone a style-based fill in any case.

N/A

## Checklist

## Additional Notes

**A behaviour change worth a reviewer's eye.** Restoring the body now writes back its inline declarations rather than its computed colours, so a page that set no inline body colour gets `removeProperty` on restore instead of a pinned computed value. That is the point of the fix, but it means the body is no longer left at a fixed colour after a high-contrast cycle. Nothing in this repository relies on that.

Hoisting the colour ramp also required a smaller change that is easy to miss: `findClosestColor` spliced the background out of the array it was handed, which is why the caller was copying the array every time. That splice is now a filtered copy, and a test pins that a complex path still loses the background without the marks after it losing it too.

Removing `removePattern` is the one change with no failing test first, deliberately: unreachable code cannot be pinned by a test, so the reproduction was a grep across the tree plus the type checker. Say the word if you would rather it were repaired and kept.

`npm run lint:fix` and `npm run type-check` pass. Tests were run as `npx jest --selectProjects unit --coverage=false` (503 suites, 6732 tests) and the esm project under `--experimental-vm-modules` (12 suites, 172 tests), both clean. That split rather than `npm test` because collecting coverage over the whole tree needs about 10 GB on the build machine and does not fit; CI has the headroom and runs it as usual.

---

### DIFF
--- a/src/service/highContrast.ts
+++ b/src/service/highContrast.ts
@@ -95,6 +95,11 @@ export class HighContrastService implements Disposable {
   // Track previous high contrast mode state to detect changes
   private previousHighContrastMode: boolean = false;
 
+  // Whether high contrast colors are currently painted onto the DOM. Distinct
+  // from the setting: the effect is suspended on blur while the setting stays
+  // on, so only this answers "is the DOM showing our colors right now?".
+  private highContrastApplied: boolean = false;
+
   // Shared canvas context for color parsing (reused to avoid GC pressure)
   private sharedCanvasCtx: CanvasRenderingContext2D | null = null;
 
@@ -184,6 +189,15 @@ export class HighContrastService implements Disposable {
    * @param figure - The replacement figure
    */
   public setFigure(figure: Figure): void {
+    // A live update can arrive while high contrast is painted, and the capture
+    // below reads the colors the DOM is showing. Restore first, or the
+    // snapshot records this service's own output as the chart's originals and
+    // turning high contrast off never gets the page back.
+    const wasApplied = this.highContrastApplied;
+    if (wasApplied) {
+      this.restoreOriginalColors();
+    }
+
     this.figure = figure;
     this.originalColorInfo = null;
     // Invalidate the trace-element cache so getAllTraceElements() rebuilds from
@@ -384,6 +398,8 @@ export class HighContrastService implements Disposable {
         this.applyPatternsToElements(highContrastElInfo);
       }
     }
+
+    this.highContrastApplied = true;
   }
 
   /**
@@ -439,6 +455,8 @@ export class HighContrastService implements Disposable {
       this.patternService.dispose();
       this.patternService = null;
     }
+
+    this.highContrastApplied = false;
   }
 
   // ========== Helper Methods ==========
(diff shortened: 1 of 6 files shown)


jooyoungseo:
Both points addressed in `29e1976`.

**The re-render question was the right one to ask,** and I would rather answer it with a test than an argument, so there is now one. You are correct that if the host redraws the chart into fresh SVG nodes before `setFigure` runs, the restore writes to elements that are no longer in the document and reaches nothing.

That turns out to be harmless, but not for an obvious reason, which is exactly why it deserved pinning. The restore exists to stop the capture recording this service's own palette as the chart's originals. Nodes the host has just drawn were never recoloured by us, so capturing them immediately already reads their own colours. Both paths end with a truthful snapshot; only the same-nodes path is self-evident. The new case replaces the chart's `<svg>` with a freshly built one, repoints the display at it, calls `setFigure`, turns high contrast off, and asserts the redrawn mark still carries its own fill.

I checked the ordering you flagged rather than assuming it. `replaceFigure` at `controller.ts:409` builds a new `Figure` from the maidr JSON and resolves elements out of the live DOM, and `setFigure` runs immediately after, so whether the nodes are the same ones is genuinely up to the host. That is why the branch is worth covering rather than reasoning away.

**The `dispose()` asymmetry is fixed too.** `HighContrastService.dispose()` now drops `sharedCanvasCtx`, matching `PatternService`. You are right that it makes no practical difference while the whole service is discarded, but the value of the symmetry is that the next person does not have to work out whether the difference was deliberate.

Re-validated after both changes: lint and type check clean, 503 suites and 6733 unit tests passing, up one from the new case.

---

PR #1231, opened 2026-09-06
fix(anychart): stamp multi-series marks and heatmap cells where they actually are

jooyoungseo:
# Pull Request

## Description

Ten defects across the AnyChart, Observable Plot and Frappe adapters, found in a codebase audit. Each was reproduced with a test that failed first.

Most of them point the reader at the wrong mark. A multi-series line or scatter partitioned its markers with the wrong stride, so series two onwards were stamped against the wrong points. A heatmap stamped its cells by flat DOM index rather than by grid position. A Frappe scatter with more than one dataset matched every dataset's circles while reading only the first, so the outline landed on a mark from a series the reader is never told about.

Two take the whole chart down rather than one mark: an unguarded `series.seriesType()` aborted the entire bind, and a Frappe container id was interpolated into a selector without CSS escaping, so an id starting with a digit threw.

## Related Issues

None.

## Changes Made

- **anychart**: the multi-series marker partition uses the right stride.
- **anychart**: `drawsCategoriesReversed` asks the chart whether it is horizontal, not the series.
- **anychart**: heatmap cells are stamped by grid position. The layer now emits a `(string | null)[][]` selector grid rather than one prefix string.
- **anychart**: `series.seriesType()` is guarded, so one odd series no longer aborts the bind.
- **anychart**: temporary box and scatter stamping diagnostics are removed.
- **observable**: a hexbin layer emits one selector per bin rather than one for the lattice.
- **observable**: `bandIntervals` is built once instead of on every `valueAtPixel` call.
- **frappe**: the container id is CSS-escaped.
- **frappe**: a scatter's selector is scoped to `.dataset-0`, so it stops matching circles belonging to datasets it never read, and it warns when a chart has more than one. Multi-series scatter remains unsupported.
- **frappe**: an empty `datasets` array throws a named error rather than an opaque `TypeError`.

N/A

## Checklist

## Additional Notes

**One change reaches beyond this adapter and deserves the closest look.** Escaping the Frappe id did not actually fix the fatal case until `cssEscape` itself was fixed: its Node and SSR fallback in `src/adapters/shared/selectorUtil.ts` left a digit-led id such as `2024-sales` unescaped. That helper is shared with the D3, Recharts, Victory, Vega-Lite and amCharts adapters. It only changes output for ids that start with a digit or a hyphen-digit, so every adapter keeps the selectors it emits today, and jsdom implements no `CSS` object at all, so this fallback is the path every test already exercises.

**Two existing tests were updated, both keeping their intent.** The AnyChart inverted-scale test modelled a horizontal chart by giving its series the type `bar` while leaving the chart itself unnamed, a disagreement no real chart can produce, since a `bar` series only exists inside `anychart.bar()`. Its fixture now names the chart type. The converter test pinned the heatmap's flat prefix selector and now pins the same panel scoping on the per-cell grid.

**Several findings were deliberately left**, and two of those refusals are worth stating. The box-plot performance findings could not be reproduced at all under jest: reaching that code needs `collectIqCandidates` to return a box, and that filter reads a computed `fill`, which jsdom does not resolve from an SVG presentation attribute. No test could fail first, so nothing was changed on faith. And draining the tracked chart set on teardown contradicts the documented contract of `stopObservablePlots`, that charts already bound stay bound; changing that is a decision about public behaviour, not a minimal fix.

**An adjacent bug was found that is not in this branch and needs its own finding:** with `datasets: []`, the Frappe pie and percentage builders sum over an empty array and emit a value of 0 for every label, which is a reading the chart never drew. Same class as #1191.

`npm run lint:fix` and `npm run type-check` pass. Tests were run as `npx jest --selectProjects unit --coverage=false` (505 suites, 6746 tests) and the esm project under `--experimental-vm-modules` (12 suites, 172 tests), both clean.

### DIFF
--- a/src/adapters/anychart/converters.ts
+++ b/src/adapters/anychart/converters.ts
@@ -1376,12 +1376,15 @@ function collectLineMarkerCandidates(
  *   1. For each line-like series, determine the expected `pointCount`.
  *   2. Run {@link collectLineMarkerCandidates} once over the SVG and sort
  *      candidates left-to-right (matching data-point order).
- *   3. Single-series charts: assign the first `pointCount` candidates as
- *      `0-0 … 0-(N-1)`.
- *   4. Multi-series charts: offset-partition (`candidates[s*N … s*N+N]`)
- *      and emit a one-time warning recommending an explicit `selectors`
- *      entry, because precise per-series attribution requires matching
- *      point coordinates against axis scale transforms (out of scope here).
+ *   3. Hand the candidates out in series order, each line-like series taking
+ *      the next `pointCount` of them and leaving the cursor where it stopped.
+ *      A series this loop skips — the bars of a combined chart, one that
+ *      cannot name its type — consumes nothing, and a series shorter than its
+ *      neighbour does not shift the ones after it.
+ *   4. Multi-series charts also emit a one-time warning recommending an
+ *      explicit `selectors` entry, because precise per-series attribution
+ *      requires matching point coordinates against axis scale transforms
+ *      (out of scope here).
  *
  * The stamp is idempotent — re-running on a chart that has already been
  * stamped is a no-op.
@@ -1404,6 +1407,12 @@ function stampLineAttributes(
   candidates.sort((a, b) => a.x - b.x);
 
   let multiSeriesWarned = false;
+  // How many candidates the series before this one took. A running cursor
+  // rather than `s * pointCount`: `s` counts the series this loop skipped —
+  // the bars of a combined chart, a series that cannot name its type — none
+  // of which consumed a candidate, and `pointCount` is this series' length,
+  // which is a stride only while every series is the same length.
+  let consumed = 0;
 
   for (let s = 0; s < seriesCount; s++) {
     const series = chart.getSeriesAt(s);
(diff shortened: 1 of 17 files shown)


jooyoungseo:
You caught a real error in my description, and I have corrected it.

I checked `buildScatterLayer` before changing anything. It still calls `soleDataset(data, 'scatter plot')`, converts only the first dataset, warns when there are more, and scopes the selector to `.dataset-0`. Multi-series scatter is not supported and this branch does not add it. My "keeps every series and scopes each to its own" was simply wrong, and it would have set up exactly the wrong expectation for whoever read it next.

The description now says what the fix does: the selector stops matching circles belonging to datasets the layer never read. That matters because `ScatterTrace` groups its elements by position rather than rejecting a count mismatch, so the surplus was absorbed silently and the outline landed on a mark from a series the reader is never told about.

**On the lone hyphen in `cssEscape`,** you are right that the fallback does not handle it. `-` alone should escape to `\-`, and the current regex only escapes a hyphen when a digit follows. I have left it: no finding in this sweep reaches it, no adapter generates a single-hyphen id, and widening a shared helper used by six adapters without a case that demonstrates the failure is how a hardening change turns into a regression. It is worth its own finding, and it now has one here in the thread.

---

Issue #961, opened 2026-08-17
A stacked Google Charts layer announces its value and its band under the same word, "Level"


jooyoungseo:
`buildSegmentedLayer` in `src/adapters/google-charts/converters.ts` emits:

```ts
axes: {
  x: { label: dt.getColumnLabel(0) || undefined },
  y: { label: 'Level' },
},
```

`'Level'` is the **core's own default label for the `z` axis** — the band/series axis. Putting it on `y` means it labels the magnitude instead, and since the layer emits no `z` at all, the core falls back to `'Level'` there too. Both quantities then come out under the same word.

Measured by running the emitted layer through the real `TraceFactory` (`Quarter` × `Alpha`/`Beta`, values 1 and 100):

| emitted `axes` | value announced as | band announced as |
|---|---|---|
| **today** — `{x: 'Quarter', y: 'Level'}` | **`Level is 1`** | **`Level is Alpha`** |
| `'Level'` moved to `z`, `y` unset | `Y is 1` | `Level is Alpha` |
| a real `y` label plus a named `z` | `Revenue is 1` | `Series is Alpha` |

A reader moving through the chart hears "Level is Alpha … Level is 1" — one word for the series and for its magnitude. Nothing is *false*, so this is milder than #955, but it makes the two announcements indistinguishable by name, which is the thing the labels exist to do.

For comparison, the Recharts adapter emits `z: { label: 'Series' }` alongside a real `y` label, and reads correctly.

## What the fix costs

Moving `'Level'` from `y` to `z` is a one-line change and removes the collision — but leaves the magnitude under the core's generic `'Y'`, because a stacked DataTable has **no** single value-column label to use: its data columns are the series names, and those correctly become the `z` *values* (`Alpha`, `Beta`).

Naming the value axis properly needs `vAxis.title` / `hAxis.title` from the chart's draw options, which the adapter is not currently handed — `createMaidrFromGoogleChart` receives a reading config (`chartType`, `thresholdOptions`, …), not the options the chart was drawn with. Threading those in is a larger change with its own design questions, so I have not made either call here.

Same applies to `buildBarOrSegmentedLayer`'s dodged and diverging paths, which share this builder.

Found while probing series-label pairing across adapters after #955. The pairing itself is correct in Google Charts, Recharts and Victory — each point's `z` names the series its value came from, including through `VictoryStack`'s child reversal — so that part is a clean negative.

jooyoungseo:
[...]

PR #1232, opened 2026-09-06
fix(echarts): keep a segmented bar's gaps, so later categories are not shifted left

jooyoungseo:
# Pull Request

## Description

Fourteen defects across the ECharts and Google Charts adapters, found in a codebase audit. Every one was reproduced with a test that failed first.

The worst of them give the reader numbers that were never in the chart. A segmented bar layer **dropped gap data**, shifting every later category one place left and announcing wrong totals, so a reader walking the bars was told the wrong category's value throughout. A category axis declared with per-label objects read its heatmap rows and columns as `[object Object]`. A multi-series Google Charts line named its whole y axis after the first series.

Several others take highlighting away entirely: an area series' fill counted as a data mark, so any chart mixing an area line with bars or scatter lost all highlighting; a nested pie lost it too; and an unescaped container id threw and took the whole figure down.

One made the tab hang: quadratic point-marker matching on a volcano or Manhattan plot.

## Related Issues

None.

## Changes Made

**ECharts**

- A segmented bar keeps its gaps, so payloads are rectangular and later categories stay where they belong.
- An area series' fill is no longer counted as a data mark.
- The container id is escaped before it reaches a selector.
- Whole-chart series share one mark count, so a nested pie highlights.
- A category axis declared with per-label objects is read by label.
- One stacked series among several no longer makes every bar series read as one stack.
- A sunburst's node walk is built once per conversion rather than twice.

**Google Charts**

- The funnel layer marks the bars it read, rather than series zero's transparent padding.
- Point-marker matching is no longer quadratic.
- Multi-series line, area, bump and their stacked variants leave `axes.y.label` unset rather than naming it after the first series.
- The reversed-category correction reaches stacked, dodged and diverging layers.
- `getChartLayoutInterface()` is guarded, so a chart without one no longer aborts the conversion.
- `markScatterElements` withdraws a partial or out-of-order mark set instead of shipping it.
- Bar and segmented marking no longer rescans every SVG rect per bar.

N/A

## Checklist

## Additional Notes

**A gap is now `NaN`, never zero.** That is the Tableau adapter's established sentinel and `isMeasured` keeps it out of ranges. Zero would be a reading the chart never drew, which is the failure mode this whole sweep keeps finding.

**Two existing tests were updated**, both pinning the old reading rather than the intended one: the Google Charts area and bump suites asserted a y-axis label taken from the first series of several. The new behaviour follows the rule `buildSegmentedLayer` (#961) and `buildSurvivalLayer` already apply.

**Behaviour changes worth a reviewer's eye.** A mixed or grouped-stack ECharts bar chart is now announced as dodged rather than stacked; only the announced type changes, since both route to `SegmentedTrace`. `markScatterElements` now highlights nothing where it used to highlight the wrong point, which is the safer failure but is a visible difference. `GoogleChart.getChartLayoutInterface` became optional in the adapter's types, which is honest about the material, Calendar, Gauge and Sankey packages; no call site broke. And `markBarElements`' fourth parameter changed meaning from a series count that was always 1 to the index of the series to mark.

`npm run lint:fix` and `npm run type-check` pass. Tests were run as `npx jest --selectProjects unit --coverage=false` (501 suites, 6741 tests) and the esm project under `--experimental-vm-modules` (12 suites, 170 tests), both clean.

---

### DIFF
--- a/src/adapters/echarts/converters.ts
+++ b/src/adapters/echarts/converters.ts
@@ -617,23 +617,34 @@ function barLayer(
   }
 
   const stacked = bars.some(seriesModel => Boolean(seriesModel.get('stack')));
-  const data: SegmentedPoint[][] = bars.map((seriesModel, order) => {
+  // A gap keeps its column rather than being dropped from the row.
+  // `SegmentedTrace` pairs the series by column index -- its summary row reads
+  // `barValues.map(row => row[i])` and takes the category off `points[0][i]`
+  // -- so a row one short puts every later category against another series'
+  // value, announces a total no bar on the page adds up to, and drops the last
+  // category from the summary altogether.
+  //
+  // The magnitude is `NaN` and never `0`: a zero sounds like a real low
+  // reading, can be reached as the row's minimum and pulls the range every
+  // other bar's pitch is scaled against, which is what `isMeasured` keeps a
+  // gap out of (#1002).
+  const rows = bars.map((seriesModel, order) => {
     const list = seriesModel.getData();
     const fill = authoredName(seriesModel) || `Series ${order + 1}`;
     const points: SegmentedPoint[] = [];
+    const drew: boolean[] = [];
     for (let index = 0; index < list.count(); index++) {
       const value = magnitudeOf(list, index, axes.horizontal);
-      if (!measured(value)) {
-        continue;
-      }
+      const magnitude = measured(value) ? value : Number.NaN;
       const position = positionOf(list, index);
       points.push(
         axes.horizontal
-          ? { x: value, y: position, z: fill }
-          : { x: position, y: value, z: fill },
+          ? { x: magnitude, y: position, z: fill }
+          : { x: position, y: magnitude, z: fill },
       );
+      drew.push(measured(value));
     }
-    return points;
+    return { points, drew };
   });
 
   return {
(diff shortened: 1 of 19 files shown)


jooyoungseo:
Good catch, and it was a real gap. Fixed in `56a8679`.

I checked before changing anything: Google Charts had **twenty** raw `container.id` interpolations across `converters.ts` and `selectors.ts`, and zero uses of `cssEscape`, while this same branch had just escaped the equivalent site in the ECharts adapter. Fixing one and not the other, in a branch that changes 487 lines of the Google Charts converter under exactly this bug class, was not a defensible place to stop. All seventeen selector-building sites now escape, and `selectors.ts` gained the shared import it had no imports at all before.

There is a test, since the fix arrived without one. It sets the container id to `:r0:`, builds a layer and asserts both that the emitted selector carries the escaped form and that `querySelectorAll` accepts it. It fails without the change.

**One case I deliberately left out of that test, having tried it and watched it fail.** I also wrote a case for an id starting with a digit, `2024-sales`, and it did not pass. That is not this fix falling short: `cssEscape`'s own non-browser fallback is `value.replace(/([^\w-])/g, '\\$1')`, which leaves a leading digit alone, so escaping at the call site cannot help. That fallback is fixed on #1231, which hardens the shared helper. Rather than duplicate that change here and set up a conflict between two of my own branches, I dropped the case and recorded the reason in the test's docblock, so the next person meets the explanation rather than the surprise.

Re-validated after the change: lint and type check clean, 501 suites and 6742 unit tests passing, up one from the new case.

---

PR #1233, opened 2026-09-06
fix(ui): stop the extrema dialog corrupting numbers and double-announcing

jooyoungseo:
# Pull Request

## Description

Ten defects in the UI layer, found in a codebase audit. Each was reproduced with a test that failed first. This is the layer where accessibility is either delivered or lost, so the weighting reflects that.

Extrema labels were rewritten with `replaceAll`, so formatting the x value **corrupted every other number in the label that happened to match it**. They were also always formatted with the x-axis formatter regardless of orientation, so a horizontal trace applied the value-axis format to a category.

The extrema listbox announced each option twice, once from focus and once from a live region that duplicated it, and that region was only mounted for traces with x-value search, so other trace types got no boundary or search messages at all. Every option was its own tab stop rather than a roving tabindex, so a reader tabbed through the whole list instead of arrowing within it.

Chat messages were `role="listitem"` inside a section rather than a list, which leaves the items without the container that gives them meaning.

## Related Issues

None.

## Changes Made

- **extrema**: only the x value is rewritten, after the label's final ` at `, and it is formatted with the trace's main axis.
- **extrema**: the live region moved to dialog level so every trace type has one, and it no longer repeats what focus already announced. Boundary and search-result messages are kept.
- **extrema**: roving tabindex on both the list and the x-value dropdown.
- **extrema**: the dialog no longer reaches through the ViewModel into the model to navigate the trace itself.
- **settings**: a pitch range that is not audible or not rising falls back to the defaults at save time, and the inputs declare their bounds.
- **settings**: the dialog no longer imports a service directly; the probe moved to a hook.
- **chat**: messages are a list.
- **chat**: the typing animation no longer re-runs the whole markdown pipeline every 10ms tick.

N/A

## Checklist

## Additional Notes

**The change I would most like a second opinion on** is the `visibilitychange` handler, which is deleted outright rather than repaired. It rebuilt the controller when a tab regained focus, silently discarding the reader's position, their text and braille output, and the chat history. Nothing in the tree needed that rebuild, and the audio service already defers its cues behind `resume()`, which is what the rebuild appears to have been working around. But this changes tab-return behaviour for every chart, so it is worth someone confirming there is no case it was quietly serving.

**Two boundary crossings are fixed and one is documented rather than fixed.** `GoToExtremaViewModel.activeContext` was removed, since it was the view's only route into the model and nothing else used it. The settings dialog's probe moved into a hook. But `@service/modelVersions` stays where it is: `scripts/check-model-catalog.mjs`, the model-catalog workflow and its own test all name it by path, so moving it is a repo-wide rename rather than a minimal fix. `test/ui/serviceImports.test.ts` now bans service-class imports from `src/ui` and documents that one carve-out, so the rule is enforced rather than remembered.

Mid-animation the chat bubble now shows the raw prefix as pre-wrapped text and renders markdown once when typing finishes, so half-parsed markdown no longer flickers. The live region was already silent while typing, so nothing announced changes.

One existing fixture was extended, not weakened: the extrema view-model test's trace state now carries `text.mainAxis`, which the production code reads. No existing assertion changed.

`npm run lint:fix` and `npm run type-check` pass. Tests were run as `npx jest --selectProjects unit --coverage=false` (503 suites, 6740 tests) and the esm project under `--experimental-vm-modules` (15 suites, 177 tests), both clean.

---

### DIFF
--- a/src/state/hook/useCredentialProbe.ts
+++ b/src/state/hook/useCredentialProbe.ts
@@ -1,0 +1,101 @@
+import type { Llm } from '@type/llm';
+import { LlmValidationService } from '@service/llmValidation';
+import { useEffect, useState } from 'react';
+
+/** How long typing settles before a probe is sent, in milliseconds. */
+const DEBOUNCE_MS = 500;
+
+/**
+ * What the settings dialog knows about an LLM credential.
+ */
+export interface CredentialProbe {
+  /** True while a probe is in flight. */
+  isValidating: boolean;
+  /** Whether the credential works, or null before anything has been probed. */
+  isValid: boolean | null;
+  /**
+   * Why the last probe failed, as the probe itself described it, or null when
+   * there is nothing more specific to say than "invalid". Kept so a rate limit
+   * or a provider outage is not reported as a bad credential.
+   */
+  error: string | null;
+  /**
+   * Models this credential can actually reach, from the provider's models API
+   * (or the local Ollama server). Replaces the curated suggestion list.
+   */
+  models: string[];
+}
+
+const IDLE: CredentialProbe = {
+  isValidating: false,
+  isValid: null,
+  error: null,
+  models: [],
+};
+
+/**
+ * Probes an LLM credential as the user types it.
+ *
+ * This lives in the state layer rather than in the settings dialog because a
+ * network side effect is not a component's to own — see `rules/ui.md`. It is
+ * the sibling of {@link useOllamaModels}, which reaches the same service for
+ * the same reason.
+ *
+ * The debounce only spaces out request starts; it cannot cancel a request
+ * already in flight, so a superseded cycle (a newer keystroke, a provider
+ * switch, or unmount) discards its own response. A slow early probe can
+ * therefore never overwrite the state of a newer one.
+ *
+ * @param modelKey - Which provider the credential belongs to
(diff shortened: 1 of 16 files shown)


jooyoungseo:
Took your first point and left the second, in `1071025`.

**The silent catches are fixed.** You are right that the repo's own TypeScript rules say never to swallow an error, and this is a case where swallowing costs something concrete: the reader sees the dialog close and nothing move, which is indistinguishable from choosing a target that was already current. Both catches now log with context before returning to trace scope, matching the pattern `chatViewModel` already uses.

**I left the discarded boolean alone,** and deliberately. `moveToXValue` returning `false` means the value was not found, but the dialog only ever offers values sourced from `getAvailableXValueOptions()`, so a `false` here would mean those two disagreed. Reacting to it would mean choosing a behaviour for a state that should not arise, and I would rather not invent one inside a fix for something else. If those two ever can disagree, that is a bug in the pairing rather than something for the caller to paper over, and it deserves its own change with a case that demonstrates it.

On the `visibilitychange` deletion: thank you for looking for a caller and saying you found none. That matches what I found, and it is the change I would most want a human to confirm before release, since it alters tab-return behaviour for every chart.

Re-validated: lint and type check clean, 503 suites and 6740 unit tests passing.

---

PR #1234, opened 2026-09-06
fix(tableau): leave a cell the view never drew as null, not zero

jooyoungseo:
# Pull Request

## Description

Ten defects across the Recharts and Tableau adapters, found in a codebase audit. Each was reproduced with a test that failed first.

The theme running through most of them is a reading the chart never drew. A Tableau heat map emitted `0` for a cell the view left empty, which is the exact spelling `HeatmapData` forbids. A Recharts histogram with no bin configuration announced a fabricated bin range of "constant 0". A composed chart emitted segmented and histogram trace types carrying flat bar data, so a single-series stacked area announced a running total equal to its own value and a 100% share at every point.

Two others lose whole worksheets or crash: Tableau trusted its own mark type without checking the layer could be built, and the Recharts adapter threw inside React render for empty data.

## Related Issues

None.

## Changes Made

**Tableau**

- A cell the view drew nothing at carries `null`. `buildHeatData` widens to `(number | null)[][]`, which is what `HeatmapData` already declared.
- A mark type whose reading cannot be built degrades to the ladder with a warning instead of being honoured, so worksheets that were dropped entirely now appear.
- `isCompleteGrid` counts pairs, not just rows, so a view with duplicates no longer passes as a rectangle.
- `ladderTraceType` stops scanning every dimension once the point-cloud rung cannot fire.

**Recharts**

- Composed mode resolves its layer type through the existing helper.
- A histogram with no bin configuration refuses rather than inventing a range.
- Empty data emits zero layers instead of throwing inside render.
- `stepDirection` and the reversed-axis flags are no longer dropped.
- Subplot mode reads each panel's reversed axis rather than the grid-wide orientation.
- The conversion no longer re-runs on every parent render.

N/A

## Checklist

## Additional Notes

**Two existing tests were updated**, both in the Recharts converter suite and both explained in the commit that changes them. One pinned the fabricated bin range this branch removes; it is replaced by a zero-width-bin case and a refusal case. The other is about a selector and its fixture now states its bins.

**A behaviour change to weigh:** a Recharts histogram with no bin configuration and non-numeric labels now throws, following the D3 binder's precedent for the identical data shape. Numeric labels become zero-width bins. That turns a silent bad reading into a hard failure, which is right for a reader but is the change most likely to surface elsewhere.

**Four findings were investigated and deliberately not fixed.** Two would need a threshold chosen on the maintainers' behalf, and I would rather name that than pick one: rectangularising a segmented Tableau view really can build 90,000 points, but every principled guard I could design either refuses a legitimate grouped bar or contradicts the existing fallback that `DODGED` exists to serve, leaving only a fill-ratio or cell cap. The summary-data read queue likewise needs two design decisions, a read deadline and whether the queue stays module-scoped, and the second is the half that stops one frozen visualisation taking an independent one down with it.

**One finding is the cause of this session's build-machine trouble and is worth surfacing on its own.** `jest.config.ts` sets `collectCoverage: true` unconditionally, and nothing consumes the result. Collecting coverage over the whole tree needs roughly 10 GB across the three workers, which does not fit on a 16 GB machine running anything else, so `npm test` stalls partway and never finishes. Every branch in this sweep was validated with `--coverage=false` for that reason. Turning it off would change what `npm test` means for every contributor and CI, and `.claude/rules/testing.md` documents the current behaviour at some length, so it is a maintainer decision rather than a contained fix. Flagging it here rather than doing it.

`npm run lint:fix` and `npm run type-check` pass. Tests were run as `npx jest --selectProjects unit --coverage=false` (505 suites, 6754 tests) and the esm project under `--experimental-vm-modules` (12 suites, 172 tests), both clean.

---

### DIFF
--- a/src/adapters/recharts/MaidrRecharts.tsx
+++ b/src/adapters/recharts/MaidrRecharts.tsx
@@ -142,32 +142,49 @@ export function MaidrRecharts({
   selectorOverride,
   children,
 }: MaidrRechartsProps): JSX.Element {
+  // Read out of the chart subtree on every render, and cheaply: each walk is
+  // depth-first and stops at the first element it wants. What must NOT be a
+  // dependency of the conversion below is `children` itself — the parent
+  // builds a fresh element tree every time it renders, so its identity is
+  // never stable and the memo would never hold. These three scalars are the
+  // whole of what the schema takes from it, so they are what it depends on;
+  // the per-panel answers travel as their serialised form for the same reason.
+  //
+  // Simple and composed mode read the one chart's axes; subplot mode reads
+  // each panel's own, since a panel is its own chart with its own axes and
+  // the walk would otherwise stop at whichever axis it met first. Each panel
+  // is asked with its own orientation, since a panel may override the grid's
+  // and that is what decides which axis its categories are on.
+  const categoryAxisReversed = subplots
+    ? undefined
+    : categoryAxisReversedFor(children, orientation === Orientation.HORIZONTAL);
+  const perPanel = subplots
+    ? categoryAxisReversedPerPanelFor(
+        children,
+        normalizeRechartsSubplotGrid(subplots, columns).flat(),
+        orientation,
+      )
+    : undefined;
+  // Serialised rather than joined so an empty grid stays empty: `''.split(',')`
+  // is `['']`, which would claim a panel that is not there.
+  const perPanelKey = perPanel === undefined ? undefined : JSON.stringify(perPanel);
+  // The convention is on the `<Line type>` the chart already declares, so a
+  // step chart need not say it twice. An explicit one still wins, which is how
+  // the reading is corrected when the walk finds a curve that is not the one
+  // meant. Subplot mode abstains for the same reason the axis walk does above
+  // -- one verdict would read the first panel's curve onto every other -- so a
+  // grid of step charts declares the convention, on the panel or on the grid.
+  const resolvedStepDirection = subplots
+    ? stepDirection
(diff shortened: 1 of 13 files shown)


jooyoungseo:
Checked your second point, since it was the one that could actually block this.

You asked whether any example or docs snippet in the repo passes a non-numeric bin label without `binConfig`, which the new refusal would now break. None does. `examples/recharts/examples/HistogramExample.tsx` passes `binConfig={{ xMinKey: 'xMin', xMaxKey: 'xMax' }}`, and `docs/recharts.md` already states the requirement twice, in the chart-type table ("requires `binConfig`") and again in the histogram section. So the documented contract and the shipped example both already sit on the safe side of this change, and it breaks nothing here.

On the other two: the render-cost point is fair and I have not pinned an upper bound on the walk. I would rather leave it unpinned than add a timing assertion, which is flaky on a shared machine and tends to pin the machine rather than the algorithm. If that walk becomes a real cost the right test counts node visits, the way the tactile ring change in #1230 counts raster reads, and that belongs with a change that has a measurement motivating it.

The coverage configuration is deliberately left alone here, and it is worth restating why, because it caused real trouble today rather than being theoretical. Collecting coverage over the whole tree needs roughly 10 GB across the three workers, which does not fit alongside anything else on a 16 GB machine, so `npm test` stalls partway and never finishes. Every branch in this sweep was validated with the two `--coverage=false` projects instead. Turning it off changes what `npm test` means for every contributor and for CI, and `.claude/rules/testing.md` documents the present behaviour at length, so the docs would have to move with it. That is a maintainer's call, not something to slip into an adapter fix.

---

PR #1235, opened 2026-09-06
refactor(model): enforce the model layer's boundary with a lint rule, and clear it

jooyoungseo:
# Pull Request

## Description

`CLAUDE.md` and `.claude/rules/model.md` both say the model layer must never import from `src/service/`, `src/state/` or `src/ui/`: the model notifies, it does not call. Nothing checked that, and four imports had crossed.

The valuable part of this change is not the four fixes. It is that the rule is now **enforced rather than remembered**: an eslint `no-restricted-imports` block scoped to `src/model/**` fails the build on the next one.

## Related Issues

None.

## Changes Made

**The guard**, written first, before any fix. It rejects `@service/`, `@state/` and `@ui/` in both alias and relative spellings, using a regex rather than a path group so `../service/x`, `../../service/x` and `../../src/service/x` are all caught, plus `hotkeys-js` by name. `@type/state` and `@util/stack` are deliberately not matched: model files import those constantly and a substring rule would be reverted within a week.

**What the rule found.** Running it before touching anything flagged exactly three files and four imports: `abstract.ts`, `candlestick.ts` and `context.ts` twice. That is the audit's five findings collapsed to their real extent, with nothing missed and nothing invented.

**The crossings, all removable, no exceptions needed:**

- `src/service/navigation.ts` is **deleted**. Nothing in the service layer ever constructed it, not the controller and not any service; it was reachable only from the model. It was a model helper filed in the wrong directory. Its pure coordinate helpers moved verbatim to `src/util/navigation.ts`, and `stepTraceInSubplot` became `Subplot.switchLayer`, since all its collaborators were model objects.
- `KeybindingService` gained a public `setScope`, so it stays the single writer to the `hotkeys` global. `Context` now exposes `onScopeChange`, and the controller owns the subscription and disposes it.
- `Candlestick` no longer redeclares and reconstructs the `navigationService` it inherits, which was discarding the base one on every live-data rebuild.

N/A

## Checklist

## Additional Notes

**On tests, stated plainly rather than dressed up.** None of these crossings caused user-visible misbehaviour: hotkeys stayed in sync either way, and the x-value helpers were pure. So there was no bug to reproduce and no failing-first test to write for the moves themselves, and I would rather say that than manufacture one. What *is* pinned is the part that can silently break: `test/controller/scopeChange.test.ts` fails on the pre-change code and covers the toggle, the `replaceFigure` rebuild, the deliberate same-shape no-op that lets braille survive a live update, and subscription disposal. If that subscription were ever dropped, every keypress after a mode switch would land in the wrong scope. Behaviour across the two moves is carried by the pre-existing assertions in the relocated tests, unchanged.

The guard itself is tested too: `test/model/layerBoundary.test.ts` lints a fixture through the real config, the same technique `test/ui/restrictedSyntax.ts` already uses, and asserts both the flagged and the permitted specifiers as one list. It fails without the rule.

**New public surface:** `Context.onScopeChange` and `KeybindingService.setScope`.

**File moves, which will conflict with any branch touching these:** `src/service/navigation.ts` to `src/util/navigation.ts`, split with `src/model/plot.ts`; and the two navigation test files into `test/model/`. `plot.ts`, `abstract.ts`, `context.ts`, `candlestick.ts` and `controller.ts` are all touched.

A side benefit worth noting: five model tests no longer need `jest.mock('hotkeys-js')` just to construct a `Context`, and those mocks are gone.

`npm run lint:fix` and `npm run type-check` pass. Tests were run as `npx jest --selectProjects unit --coverage=false` (515 suites, 6816 tests) and the esm project under `--experimental-vm-modules` (12 suites, 172 tests), both clean with `main` merged in.

---


jooyoungseo:
Taken, in `f1323d2`. `register()` now calls `this.setScope(initialScope)`.

You called it cosmetic and it nearly is, but the reason to do it is not tidiness. The comment on `setScope` claims the service is the single writer to the `hotkeys` global, and this whole branch exists to make that kind of claim enforceable rather than aspirational. Leaving one call inside the class that bypasses its own setter would have left the assertion false in exactly the place a reader checks it first. There is now one `hotkeys.setScope` call in `src/`, so replacing the library later means finding one line.

Thank you for diffing the moved code against `origin/main` byte for byte rather than taking the description's word that it was a verbatim move. That is the check that matters most in a change like this, since a refactor that quietly alters behaviour while claiming not to is far harder to catch later than an ordinary bug.

Re-validated after the change: lint and type check clean, 515 suites and 6816 unit tests passing.

---

PR #1237, opened 2026-09-06
fix(model): rebuild cached pointer geometry when the viewport moves

jooyoungseo:
# Pull Request

## Description

A pointer hover resolves to a mark by comparing the pointer's viewport coordinates against geometry read from `getBoundingClientRect()`. `LineTrace` and `ScatterTrace` cache that geometry and already drop it on scroll and resize. **Five other traces cache it and never drop it**, and four more do not cache it at all and re-measure every mark on every `pointermove`.

Both halves are fixed by the rule line and scatter already follow, extracted so it is stated once. The correctness half is why this is a `fix`.

### The bug

Heatmap, box, violin-box, violin-KDE and candlestick measure every mark once **in the constructor** and keep the result forever. The pointer coordinates those centres are compared against — `event.clientX` / `clientY` from `PointerGuidanceCommand` — are always current, so a chart built below the fold and then scrolled into view answers against centres that have moved. A heatmap scrolled by one cell height announces the row above the one under the pointer; `isPointInBounds` then measures the wrong element live and reports `onCurve: false`, and the guidance beep steers toward a centre no longer on screen. The announcement that follows is a real cell with a real value, which is the quiet kind of wrong.

### The cost

`findNearestPoint` runs on every `pointermove` DOM event with nothing throttling it on the model side — only the guidance beep is throttled, in the audio service. Each `getBoundingClientRect()` is a forced layout read. Measured as layout reads per hover:

| Trace | Before | After |
|---|---|---|
| Ridgeline, 11 groups × 11 samples | 121 | 4 |
| Hexbin, per hover after the first | one per bin | 1 |
| Gantt, per hover after the first | one per interval | 1 |
| Word cloud, per hover after the first | up to one per term | 0 |

## Related Issues

None.

## Changes Made

- **`src/util/viewport.ts`** (new): `watchViewport(onViewportChange)` returns a release function. `scroll` is watched in the **capture** phase because a scroll event does not bubble — a chart inside its own scrolling container fires on that container and nothing else. `resize` covers a responsive chart changing shape. It is a no-op where there is no `window`.
- **heatmap, box, violin, violinBox, candlestick**: the constructor-built centre caches are marked stale on a viewport change and rebuilt on the next hover.
- **hexbin, wordCloud, gantt**: geometry is measured once and reused, filled on the first hover rather than in the constructor so a chart nobody points at pays nothing at load.
- **ridgeline**: measures once per group rather than once per sample. `pairWith` gives every sample of a group the same path — a ridgeline draws one curve per group — so a row's samples all answer with the identical rect and only the first could ever win the strict comparison. The resolved sample is the group's first, exactly as before.
- **line, scatter**: their hand-rolled copies of the listener pair fold into `watchViewport`. Behaviour is identical — same two events, same capture phase, same `typeof window` guard, released the same way. Only the spelling moves.

N/A

## Checklist

## Additional Notes

**One invalidation rule, in one place.** A cached client rectangle lives until the viewport moves — page scroll, any ancestor container scroll, or a window resize — and the next hover measures again. Rebuilding on the next hover rather than on the event itself keeps a scroll free of layout reads. A redraw or a live-data update needs no event: it replaces the trace, and the new one measures for itself. Every trace that takes a watch releases it in `dispose()`. That rule is not new — line and scatter have followed it since #1220 — but it lived in two hand-copied blocks, so a third trace adopting it meant copying it a third time. Ten traces now share the one statement of it.

**Candlestick was found by review, not by the sweep.** It caches under the same field name as the other four, in a file the pass over pointer geometry did not reach. `test/model/pointerCentersAfterScroll.test.ts` is table-driven, so it now pins the rule for five traces; three of the six candlestick cases fail on the unfixed code.

**The trade-off worth naming:** a chart that animates its marks into place after load is no longer re-measured mid-animation. The reader is hovering after that, and line and scatter already worked this way.

**What was deliberately left alone.** `BoxenTrace.findNearestPoint` has the same shape but scans one element per distribution, so it is bounded by the number of categories. And the ridgeline's sub-sample resolution is not guessed at: a shared element carries no per-sample geometry to choose between, and `RIDGELINE` is non-oriented, so inventing one would be a behaviour change dressed as a speed one.

**The tests count measurements rather than timing them** — a second hover with nothing moved measures at most once; the first hover after a scroll or resize measures every mark again — and each invalidation case is written to fail on the unfixed code.

`npm run lint:fix` and `npm run type-check` pass. Tests were run as `npx jest --selectProjects unit --coverage=false` (530 suites, 6949 tests) and the esm project under `--experimental-vm-modules` (15 suites, 177 tests), both clean on this head with `main` merged in.


jooyoungseo:
Both points taken; both are now in the branch.

**`CandlestickTrace`** — confirmed, and it is a real one. `src/model/candlestick.ts:246` caches in the constructor and `findNearestPoint` reads it with no staleness check, no dirty flag, and nothing in `dispose()`. It was missed because it caches under the same field name in a file the sweep did not reach, not for any reason that would justify leaving it. Fixed the same way as the other four in `afe0e32`, and the candlestick cases join the existing table-driven suite — three of the six (`after the page scrolls`, `after the window resizes`, `stop being watched once the trace is disposed`) fail on the unfixed code and pass with it. The suite now pins the rule for five traces.

**The `scatter.ts` / `line.ts` duplication** — agreed, and already done in `ff221ba`, which landed a minute before your review started; the re-review you have queued on that commit should see it. Both traces now take `watchViewport` and release it in `dispose()`. Behaviour is unchanged — same two events, same capture phase on `scroll`, same `typeof window` guard — so it is only the spelling that moves. That collapses what was three hand-copies into one statement of the rule, now shared by ten traces.

On the verification note: lint, type-check, unit and esm were run locally on this head — 530 suites / 6949 unit tests and 15 / 177 esm, all green — and CI is the check to trust for the build.

---

PR #1238, opened 2026-09-06
perf(model): stop paying quadratic and duplicated work while building a trace

jooyoungseo:
# Pull Request

## Description

Seven fixes to the work done before a reader hears anything. Construction is not a one-off cost: the controller rebuilds the figure on every live-data append, so each of these is paid again on every update.

Every finding was confirmed by a spy count on the unfixed code **before** any source change, and every fix is pinned by the same count rather than by wall clock:

| What | Before | After |
|---|---|---|
| `indexOf` calls, 50 scatter points | 50 | 0 |
| `findIndex` calls, 41 treemap nodes | 41 | 0 |
| `Array.sort` calls per scatter construction | 6 | 4 |
| `Array.sort` calls, line chart of 50 columns | 50 | 0 |
| Vertex x reads reconciling 200 / 400 line points | 20,002 / 80,000 | 796 / 1,596 |
| `getComputedStyle` calls, two series of four points | 8 | 2 |
| `querySelectorAll` calls, two violins on one shared selector | 2 | 1 |
| `getBoundingClientRect` calls at construction, line and scatter | 4 each | 0 |

The two worth reading twice: `reconcilePathCoordinates` was **quadratic in a way that scales with the data, not the edge case** — four times the work for twice the points — and it is reached by any dense line, because Plotly's `line.simplify` and matplotlib's `path.simplify` both drop vertices by default. And a scatter of a few hundred thousand points (a Manhattan plot) spent roughly a third of a second on two sorts that duplicated two others exactly.

## Related Issues

None.

## Changes Made

- **scatter**: a point's slot in its column is looked up rather than searched for; the points are ordered once per axis instead of twice.
- **treemap**: a node records its place among its siblings instead of searching the parent's children for itself.
- **line**: the segment search carries forward across a simplified path; `buildGraph` wires only the links the line family reads.
- **line, scatter**: pointer-hover centres are left for the first hover to measure.
- **line, smoothSvgXY, violin**: a series' highlight markers are drawn in one pass — one computed style, one fragment, one insertion — via the new `Svg.createCircleElements`.

N/A

## Checklist

## Additional Notes

**Three behaviour points a reviewer should know.**

1. **`buildGraph` now leaves `up`, `down`, `left` and `right` null.** Those four are the links `MovableGraph.moveOnce` and `MovableGraph.isMovable` step along, and this family never asks it to: `LineTrace` overrides both for every direction, moving between series by comparing y values at the cursor's own x and along a series by the column index. No subclass — area, bump, contour, parallel, radar, smooth, step, survival — delegates back, and `buildGraph` is private to each class that has one. What the graph is actually read for is `moveToExtreme`'s `top`/`bottom`/`start`/`end`, which stay wired. The docstring says which links exist and that a subclass wanting `MovableGraph`'s stepping must wire the others first.

2. **`Svg.createCircleElement` becomes `Svg.createCircleElements`**, taking a whole series. As a side effect of computing the centres before drawing anything, a series whose coordinates turn out to be unusable no longer leaves already-inserted markers behind — those orphans outlived the trace that made them and were matched by the next resolution of the same selector. That is what the `hiddenCloneDisposal` suite's stated intent wants, and it is now asserted.

3. **Column tie-breaking is preserved exactly** — the last of a column's joint highest is its top, the first of its joint lowest is its bottom — and pinned by two tests. A row with a gap at that column still gets no top or bottom, so an extreme jump from a gap is still refused. `Array.prototype.sort` is stable in both the old and new forms, so points agreeing on x and y keep the order the layer listed them in.

**No cache added here outlives a constructor**, so nothing needs invalidating across a live-data rebuild. The one pre-existing cache reused is `highlightCentersDirty`, which now simply starts stale — the state a scroll already puts it in — so it reuses #1237's rebuild path exactly rather than adding a second one.

**Five findings were deliberately not taken**, and two of them are worth naming because they are real and remain open:

- `Heatmap`'s per-cell selector grid issues one `document.querySelector` per cell. Confirmed, but the cell selectors are opaque strings from five different adapters; batching them loses which element answered which selector, and re-testing with `element.matches` is the same quadratic. A real fix is an adapter-level protocol change, outside this theme.
- Candlestick's `mapToSvgElements` creates roughly 6N throwaway SVG elements, and its read/write interleaving forces up to 2N synchronous layouts. Both were skipped because `candlestick.ts` was being rewritten on another branch while this work ran. That branch has since merged, so these are unblocked but not done.

`npm run lint:fix` and `npm run type-check` pass. Tests were run as `npx jest --selectProjects unit --coverage=false` (536 suites, 6981 tests) and the esm project under `--experimental-vm-modules` (15 suites, 177 tests), both clean with `main` merged in.

---


jooyoungseo:
Thanks — the `violin.ts` nit is right and is fixed in `dc18a4a`. `candidatesOf` filters by element type in preference order (use, then path, then polygon) and returns the first kind that answers; it never sorts, so the comment described work the code does not do. It now says "resolved and narrowed once", which also explains why there can be up to three passes.

On the `buildGraph` JSDoc length: leaving it. Dropping four links from a navigation graph is exactly the kind of change that looks like dead-code removal a year from now, and the reason it is safe — that `LineTrace` overrides `moveOnce`/`isMovable` while `moveToExtreme` is *not* overridden and still reads `top`/`bottom`/`start`/`end` — is not recoverable from the code alone. Your own trace-through independently reconstructing that chain is the argument for keeping it written down.

No other changes. CI on the previous head was green through Type Check and Lint Code with the rest in flight; this push restarts it.

---

PR #1239, opened 2026-09-07
perf(model): read a chart's derived geometry before drawing any of it

jooyoungseo:
# Pull Request

## Description

Two construction-cost fixes for candlestick, box and violin-box, plus the fix for a regression the first attempt at the second one introduced — which is the part worth reading.

`Svg.createLineElement` and `Svg.createWhiskerElement` each read geometry (`getBBox`, `getComputedStyle`) and then immediately insert the element they built. Called in a loop — twice per candle, six times per box — every read after the first sits behind a DOM write, so each one is a forced synchronous layout. Separately, `CandlestickTrace.mapToSvgElements` pre-filled arrays with placeholder SVG elements that the very next loop overwrote.

Measured per construction (jsdom, spies on `getBBox`, `getComputedStyle`, `createElementNS`, `insertAdjacentElement`; "forced layout" = a read with a write between it and the previous read):

| | before | after |
|---|---|---|
| Candlestick 200 candles — `getBBox` | 400 | 200 |
| Candlestick 200 candles — `getComputedStyle` | 400 | 200 |
| Candlestick 200 candles — forced layouts | 400 | 1 |
| Box 30 distributions — `getBBox` / `getComputedStyle` | 180 / 60 | 120 / 30 |
| Box 30 distributions — forced layouts | 120 | 1 |
| Violin-box 30 — forced layouts | 120 | 1 |
| Candlestick placeholders (`createEmptyElement`), 200 candles | 1800 | 0 |
| …with selectors matching nothing | 2800 | 1000 (one genuine fallback per cell) |

The unfixed 400 for 200 candles is exactly 2N, which confirms the original finding's "up to 2N" is exact. Construction is not a one-off: the controller rebuilds the figure on every live-data append.

## Related Issues

None.

## Changes Made

- **`src/util/svg.ts`** — `createLineElement`/`createWhiskerElement` become `buildLineElements`/`buildWhiskerElements`, which do a READ pass (memoised per anchor) and a BUILD pass and return marked-owned but **detached** nodes. A new `Svg.insertDerived(anchor, ...nodes)` performs the write, one `insertAdjacentElement('afterend', node)` per node in creation order — byte-for-byte what the old per-call code did.
- **`src/model/box.ts`, `src/model/violinBox.ts`** — measure and build up front; insert the derived edges at the exact point in the existing phase where `createLineElement` used to stand, and the whiskers inside `offerGeometry` in `[min, max]` order.
- **`src/model/candlestick.ts`** — same split; and the placeholder pre-fills are gone, since every cell is written by the loop below.
- Comment-only reference updates in two adapters and one doc so nothing names a method that no longer exists.

N/A

## Checklist

## Additional Notes

**A regression was found in review and fixed; this is the part to check.** The first version of the batching also moved the *writes*, grouping them into one `anchor.after(...)` per anchor. That is not equivalent when one DOM element serves as more than one box part, and two shipped adapters emit exactly that:

- `src/adapters/victory/selectors.ts:676-677` emits `iq: sel('q1')` **and** `q1: sel('q1')` — one node. Its own comment, "harmless: direct q1/q3 below bypass iq-edge derivation", is precisely the assumption that stopped holding.
- `src/adapters/plotly/extractor.ts:3108-3115` gives a violin box `min`, `iq`, `q2` and `max` all as one `<path>`.

Batching moved the derived lines to before the clones instead of between the anchor and the clones. That flips `previousElementSibling`, which `AbstractTrace.getAllOriginalElements()` uses to pair a hidden clone back to its visible original — and that set is the sole input to `HighContrastService.getAllTraceElements()`, which drives high-contrast recolouring and pattern assignment. Measured on the Victory shape: `getAllOriginalElements()` went from `[min0, q2_0, q3_0, max0]` to `[min0, q1_0, q2_0, q3_0, max0]`, so a low-vision reader's palette was re-derived over a different set. **No existing test caught it** — the full unit suite passed on both sides.

The fix keeps the batched reads, which is where the entire forced-layout win comes from, and restores the write positions. A differential across 28 trace shapes — full `svg.outerHTML`, `getAllOriginalElements()`, `getAllHighlightElements()`, `getGeometryElements()`, and the post-dispose DOM — is now **byte-identical to `main`**, while the pre-fix commit differs on 18 of them. The read counts in the table above are unchanged by the fix.

**`getGeometryElements()` order is preserved** — whiskers still reported lower-cap-first, pinned by the untouched `test/model/boxGeometry.test.ts` and by a new explicit case, because that order feeds the tactile display.

**The placeholder removal rests on an invariant**, verified structurally rather than by sampling: `precomputeSortedSegments` returns `['volatility', ...ohlc]` with four OHLC entries when `hasOpen` and three otherwise, and `candlestickSectionsOf` returns 5 or 4 off the same flag — so a candle's navigation order always has exactly as many entries as the chart has rows. The new test asserts **every cell is defined**, not merely that the allocation count dropped, across nine configurations including N=0 and N=1.

**Not done here, deliberately:** the adjacent pre-existing defect where `collectElements` clones before later selectors resolve (#1004). It is a separate bug and widening this PR to it would mix two changes.

`npm run lint:fix` and `npm run type-check` pass. Tests were run as `npx jest --selectProjects unit --coverage=false` (539 suites, 7003 tests) and the esm project under `--experimental-vm-modules` (15 suites, 177 tests), both clean. I also confirmed the six `sharedAnchorWrites` cases fail on the pre-fix commit and pass on the tip.

---

### DIFF
--- a/.claude/rules/model.md
+++ b/.claude/rules/model.md
@@ -101,7 +101,7 @@ every one of them at once.
 
 **Every code path that creates a highlight element must mark it.** The `Svg`
 helpers (`selectAllElements` / `selectElement` with clone,
-`createEmptyElement`, `createCircleElement`, `createLineElement`,
-`createHighlightElement`) already do. If you clone or synthesize an element by
+`createEmptyElement`, `createCircleElements`, `buildLineElements`,
+`buildWhiskerElements`, `createHighlightElement`) already do. If you clone or synthesize an element by
 hand, wrap it in `Svg.markOwned(...)`. Elements selected with
 `shouldClone = false` are live chart geometry — never mark or remove them.
--- a/.github/instructions/model.instructions.md
+++ b/.github/instructions/model.instructions.md
@@ -103,7 +103,7 @@ every one of them at once.
 
 **Every code path that creates a highlight element must mark it.** The `Svg`
 helpers (`selectAllElements` / `selectElement` with clone,
-`createEmptyElement`, `createCircleElement`, `createLineElement`,
-`createHighlightElement`) already do. If you clone or synthesize an element by
+`createEmptyElement`, `createCircleElements`, `buildLineElements`,
+`buildWhiskerElements`, `createHighlightElement`) already do. If you clone or synthesize an element by
 hand, wrap it in `Svg.markOwned(...)`. Elements selected with
 `shouldClone = false` are live chart geometry — never mark or remove them.
--- a/docs/VIOLIN_PLOT_SPEC.md
+++ b/docs/VIOLIN_PLOT_SPEC.md
@@ -393,7 +393,7 @@ The IQ box (`iq`) is a `<rect>` SVG element spanning Q1 to Q3. The frontend extr
 - **Vertical orientation**: Q1 = bottom edge, Q3 = top edge
 - **Horizontal orientation**: Q1 = left edge, Q3 = right edge
 
-The frontend asks `Svg.createLineElements` for the `'bottom'` and `'top'` edges of every box's `iq` element in one batch, and creates the line elements from that.
+The frontend asks `Svg.buildLineElements` for the `'bottom'` and `'top'` edges of every box's `iq` element in one batch, and creates the line elements from that.
 
 **Selector Array Requirements:**
 - `lowerOutliers` and `upperOutliers` can be empty arrays `[]` when no outliers exist.
--- a/docs/VIOLIN_PLOT_SPEC.md
+++ b/docs/VIOLIN_PLOT_SPEC.md
@@ -393,7 +393,7 @@ The IQ box (`iq`) is a `<rect>` SVG element spanning Q1 to Q3. The frontend extr
 - **Vertical orientation**: Q1 = bottom edge, Q3 = top edge
 - **Horizontal orientation**: Q1 = left edge, Q3 = right edge
 
-The frontend calls `Svg.createLineElement(iqElement, 'bottom')` and `Svg.createLineElement(iqElement, 'top')` to create line elements at the edges.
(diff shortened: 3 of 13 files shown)


jooyoungseo:
All three findings were right and are fixed in `1fd2aa2`.

**1. `.claude/rules/model.md`** — fixed, and it was worse than you found. The list named `createCircleElement`, which #1238 had already renamed to `createCircleElements`, so it was stale before this branch touched anything; and it never listed `createWhiskerElement`/`createWhiskerElements` at all, though that marks ownership too. It now reads `createEmptyElement`, `createCircleElements`, `buildLineElements`, `buildWhiskerElements`, `createHighlightElement`, and `.github/instructions/model.instructions.md` is re-synced via `npm run sync:copilot` (`sync:copilot:check` reports 11 files in sync).

Worth noting why CI could not have caught this: the **Copilot Instructions In Sync** check compares the rule file against its generated twin, and both were equally stale. It verifies they agree with each other, not that either agrees with the code.

**2. `docs/VIOLIN_PLOT_SPEC.md:396`** — fixed to `Svg.buildLineElements`. You're right that this is the worse of the two typos, since that line was touched on this branch specifically to fix a stale reference and I introduced a name that has never existed.

**3. `candlestickDerivedLines.test.ts` docblock** — fixed. It now says `buildLineElements` measures and builds while `insertDerived` inserts, which is the whole point of the split and exactly the distinction the regression fix turns on.

On your note about not being able to run the toolchain: understood, and your manual trace matched what I ran locally. For the record on this head — `lint:fix`, `type-check`, `sync:copilot:check` and the affected suite all clean; the full run before this commit was 539 suites / 7003 unit tests and 15 / 177 esm. I also confirmed independently that the six `sharedAnchorWrites` cases fail on the pre-fix commit `c926630` and pass on the tip, so the regression guard is real rather than a tautology.

---

PR #1240, opened 2026-09-07
perf(model): leave pointer-hover centres for the first hover in five more traces

jooyoungseo:
# Pull Request

## Description

#1237 gave several traces a viewport-invalidation mechanism — a `highlightCentersDirty` flag, a `watchViewport` field that sets it, a rebuild inside `findNearestPoint`, and a release in `dispose()`. #1238 then used that mechanism in `line.ts` and `scatter.ts` to stop measuring in the constructor at all: the flag simply starts `true` and the cache starts `null`, so the first hover measures through the exact path a scroll already used.

#1238 left heatmap out, saying "Heatmap has the same eager call but no stale-marking machinery to reuse". That was already untrue when it was written — #1237 had merged first and had given heatmap exactly that machinery. The same is true of box, violin, violin-box and candlestick. This finishes the job.

Each eager call is one `getBoundingClientRect` per mark, and a layout read straight after the DOM writes that inserted those marks is a **forced synchronous layout**. Measured with a spy on `Element.prototype.getBoundingClientRect`, counting construction only:

| Trace | fixture | before | after |
|---|---|---|---|
| Heatmap | 53×7 calendar | **371** | 0 |
| Heatmap | 2×2 grid | 4 | 0 |
| Candlestick | 4 candles × 5 sections | 20 | 0 |
| Box | 1 box, 5 sections | 5 | 0 |
| Violin-box | 1 violin | 5 | 0 |
| Violin | 2 curves × 2 points | 4 | 0 |

Construction is not a one-off: the controller rebuilds the figure on every live-data append, so a calendar heatmap paid those 371 forced layouts again on each update — before the reader heard anything, for a question only a pointer asks.

## Related Issues

Finishes the work started in #1237 and #1238.

## Changes Made

- **`src/model/heatmap.ts`, `box.ts`, `violin.ts`, `violinBox.ts`, `candlestick.ts`** — `highlightCentersDirty` starts `true` and the constructor assigns `highlightCenters = null` instead of calling `mapSvgElementsToCenters()`. Twelve lines per trace, and no new mechanism: the first hover takes the branch a scroll already took.

N/A

## Checklist

## Additional Notes

**The one risk, checked per trace rather than assumed.** Deferring the build is only safe if `findNearestPoint` is the *only* reader of the cache — otherwise some earlier caller gets `null` where it used to get geometry. Verified across the whole of `src/`: in all five traces every read of `highlightCenters` sits inside the rebuild-then-scan block of `findNearestPoint` (heatmap 565-595, box 666-695, violin 731-760, violin-box 786-815, candlestick 1389-1419), plus the declaration and the constructor's `null`. Nothing outside the owning class touches it, and none of the five has a subclass. The tactile path reads the separate `geometry` field via `getGeometryElements()`, which this does not touch.

**The tests assert the resolved element, not only the call count.** A pure count test would pass if the function started returning the wrong mark quickly. For each of the five, the first hover after the change resolves to the same element it did before — heatmap, box, violin-box and candlestick all to `m3`, violin to row 1 col 1 at `cx="30"` — and the calendar case pins that the 371 reads simply move to the first hover rather than disappearing.

**Also confirmed per trace:** `dispose()` still releases the viewport watch; a trace nobody ever hovers performs zero `getBoundingClientRect` at construction; and the existing `pointerCentersAfterScroll` suite from #1237 still holds for all five.

**Nothing new is cached and nothing new needs invalidating** — this only changes *when* the existing cache is first filled, reusing the existing invalidation rule unchanged.

`npm run lint:fix` and `npm run type-check` pass. Tests were run as `npx jest --selectProjects unit --coverage=false` (536 suites, 6992 tests) and the esm project under `--experimental-vm-modules` (15 suites, 177 tests), both clean with `main` merged in.

---

### DIFF
--- a/src/model/box.ts
+++ b/src/model/box.ts
@@ -72,8 +72,13 @@ export class BoxTrace extends AbstractTrace {
    * however far the chart moved, and the guidance beep points at a section
    * that is no longer there. Rebuilding on the next hover rather than on the
    * event keeps a scroll itself free of layout reads.
+   *
+   * It starts stale, so the first hover measures rather than the constructor.
+   * Measuring there is a forced layout right after the writes that inserted
+   * the marks, once per trace, before the first announcement -- and a
+   * keyboard reader, who is the primary audience, never asks the question.
    */
-  private highlightCentersDirty = false;
+  private highlightCentersDirty = true;
 
   private readonly stopViewportWatch = watchViewport((): void => {
     this.highlightCentersDirty = true;
@@ -143,7 +148,10 @@ export class BoxTrace extends AbstractTrace {
       this.highlightValues?.reverse();
     }
 
-    this.highlightCenters = this.mapSvgElementsToCenters();
+    // Left for the first hover to measure. `highlightCentersDirty` starts
+    // true, so findNearestPoint builds them the same way it does after a
+    // scroll.
+    this.highlightCenters = null;
     this.movable = new MovableGrid<number[] | number>(this.boxValues, { row: 0 });
   }
 
--- a/src/model/candlestick.ts
+++ b/src/model/candlestick.ts
@@ -166,8 +166,13 @@ export class Candlestick extends AbstractTrace {
    * however far the chart moved, and a hover resolves to a candle that is no
    * longer there. Rebuilding on the next hover rather than on the event keeps
    * a scroll itself free of layout reads.
+   *
+   * It starts stale, so the first hover measures rather than the constructor.
+   * Measuring there is a forced layout right after the writes that inserted
+   * the marks, once per trace, before the first announcement -- and a
+   * keyboard reader, who is the primary audience, never asks the question.
    */
-  private highlightCentersDirty = false;
+  private highlightCentersDirty = true;
 
   private readonly stopViewportWatch = watchViewport((): void => {
     this.highlightCentersDirty = true;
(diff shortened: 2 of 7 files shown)


jooyoungseo:
Nit confirmed and fixed in `c15baf1` — extending the suite from four traces to five rewrapped that sentence and the opening `*` of `*viewport*` was lost in the move. Restored, and the line rewrapped cleanly rather than left short. It's the sentence that says why the whole file exists, so worth reading correctly.

Thanks for re-deriving the safety claim rather than taking the description's word for it — that's the part of this change that actually carries risk, and your grep matched mine on all four points (single reader inside the `findNearestPoint` rebuild block, no subclasses, `dispose()` still releasing the watch, `mapSvgElementsToCenters()` free of side effects the constructor relied on).

One addition to your `gantt.ts`/`hexbin.ts` note: `wordCloud.ts` is the third already-lazy trace, and `line.ts`/`scatter.ts` were made lazy in #1238 — so with these five, all ten traces that own the invalidation machinery now defer the first measurement.

---

jooyoungseo:
You were right and my previous comment was wrong — `c15baf1` did not land the fix, and I should not have claimed it did.

The cause is worth recording, because it would have bitten the next attempt too. `c15baf1` genuinely did write `* *viewport* coordinates`. But that landed at the start of a wrapped JSDoc line, which already begins with `*`, so `eslint --fix` read `* *viewport*` as a doubled line marker and dropped one — putting the sentence back exactly as it was. My verification then read the file **before** running `lint:fix` rather than after, so I checked a version that never got committed.

Fixed in `ec42455` by rewording so the emphasis sits mid-line, where there is no ambiguity to normalise away, rather than fighting the formatter:

```
 * `findNearestPoint` calls without measuring again. What they keep are the
 * coordinates *as the viewport saw them*, and the pointer coordinates they
 * are compared against (`event.clientX` / `clientY`) are
```

This time I read the line back after `lint:fix`, then confirmed it against the pushed remote rather than my local copy. The suite still passes (31 tests).

---

PR #1241, opened 2026-09-07
feat(ui): split the settings dialog into tabs

jooyoungseo:
# Pull Request

## Description

`Ctrl+,` opened one flat scroll of roughly thirty setting rows. Reaching the tactile display picker or an API key meant paging past the echo sliders and the high-contrast colours first; at a 900px viewport only the first twelve rows were on screen at once.

The rows now sit on six tabs. The grouping is by the modality a setting reaches the reader through — what someone looking for a setting already knows about it — rather than by the service that happens to own it.

| Tab | Rows |
| --- | --- |
| **General** | Autoplay Duration, ARIA Mode, Hover Mode |
| **Audio** | Volume, Min/Max Frequency, 3D Echo Count, Echo Volume, Echo Duration |
| **Visual** | Outline Color, High Contrast Mode / Levels / Light / Dark |
| **Braille & Tactile** | Braille Display (presets or manual), Tactile Graphics Display |
| **AI** | The four LLM providers, Expertise Level, Custom Instructions |
| **About** | maidr.js version, Loaded From, Browser, OS, Copy diagnostics |

Every panel now fits without scrolling, and four of the six leave the dialog at exactly the same height, so switching tabs does not resize it under a pointer or a magnifier.

No setting was added, removed, renamed, or given a new default. Save, Close, Reset, `Escape` and the `Alt+S` / `Alt+C` accelerators still commit every tab at once.

## Related Issues

None; no open issue tracked this.

## Changes Made

Three commits: the split, then two rounds of fixes that review turned up on it. They are kept separate because each is only legible against the one before it.

### `813481a` — the split

- A `Tabs` tablist named "Settings sections", outside `DialogContent` so it stays put while a panel scrolls. Manual activation (arrow moves focus, `Enter` selects), scrollable with scroll buttons on a narrow viewport.
- Each panel is a `tabpanel` named by its tab through `aria-labelledby`, and focusable, so entering one announces which section you are in — the job the removed `h3`s used to do. About needs that for a second reason: its first four rows are static text, so without it `Tab` skips straight to the copy button.
- The three `h3` section headings and the dividers between them are gone; the `h2` "Settings" title is the only heading left.
- `minHeight: min(400px, 45vh)` on the content area — a floor under the shortest panels, with a viewport term so it cannot squeeze the footer off a short screen.
- Save is disabled while the custom instruction is under 10 characters, and that field can now be several tabs away, so the AI tab takes a badge (decorative icon plus visually hidden text — the reason never rests on colour) and the footer names the tab.

### `66fe44e` — what the split cost, put back

Four things it took away, all invisible to a sighted pointer user and all landing on the reader this dialog exists for:

- **Unmounting an inactive panel was the root of three.** `useCredentialProbe` keys its result to the hook instance, so returning to the AI tab reset it to "unknown": the key went back over the wire, and the model dropdown the reader had already been given went inoperable for the length of a probe, with no action on their part. Live regions fared worse — a region that is not mounted cannot announce, so a **tactile display disconnecting while the reader was on another tab was announced nowhere at all** (`dotPadSession` pushes that state unsolicited, from the vendor SDK callback), and the AI panel's assertive warning re-fired on every visit. Panels now **mount on first visit and stay**, which keeps the part worth having: nothing of the AI panel exists, and no provider is contacted, for a reader who never opens it. Hidden with `display: none`, because `Grid`'s own `display: flex` outranks the UA `[hidden]` rule.
- **The footer hint could not announce**, for the reason this file already documents twelve lines away — it was created holding its message, and a live region has to be in the DOM before its text changes. Now rendered while empty. Its only trigger is a tab switch (the fields that invalidate the instruction are on the tab that suppresses the hint), so being created with the text was the only path it had.
- **`Alt+S` while blocked** returned without preventing the default or saying anything: a reader who used the shortcut the dialog advertises got no response of any kind. It now opens the tab holding the field.
- **The chat's "Open Settings" button** exists to reach an API key and was landing on General, which has no credential field and nothing pointing onward. Openers can now name a page — `SettingsSection` moved to `@type/settings` so the view model can carry one without importing the view.

Plus four smaller ones: the tabs had **no focus indicator at all** (`ButtonBase` clears the outline, `Tab` replaces it with nothing — and with arrows moving focus without selecting, there was nothing to see); selection was carried by colour alone, which a forced-colours mode overrides; switching tabs kept the previous page's scroll offset; and "Copied to clipboard" reappeared on a later visit to About as though it had just happened.

### `12ef5a1` — making the refused save perceivable

The automated review caught that the `Alt+S` fix above only *selected* the AI tab and never moved focus, and it was worse than reported: with panels now persisting, the AI panel is already mounted by the time a save is refused — that is where the instruction was typed — so the switch mutated no live region and announced nothing on the *first* press either, not just the second. The shortcut answered silently again, which is what it had just been changed to stop doing.

Focus now moves to the tab, whose name already reads "AI needs attention". Two details carry weight: it happens in an effect rather than in the key handler, so focus arrives *after* the render that marks the tab selected (focusing first announces it as a tab the reader merely landed on); and it is keyed on a **count** of refusals rather than a flag, which would have fired once and swallowed every refusal after it — the same shape as the bug being fixed.

Also corrects the rationale comment on `aria-controls`. The behaviour was already right; the stated reason no longer was, since a visited panel *is* in the document now. It is named only on the selected tab because a hidden panel is out of the accessibility tree, so pointing at it would be no better than pointing at one that was never rendered.

### Tests and docs

- `test/ui/settings.tabs.test.tsx` (new, 13 cases): the tab/panel naming contract; that no unselected tab names an absent panel; that edits to both the general and the LLM slice survive a tab the reader has left; that a visited panel is the *same* element on re-entry and an unvisited one is absent entirely; that arrows move focus without selecting; that the footer hint arrives as a **mutation of a region that was already there**, observed with a `MutationObserver` rather than asserted on its text; and both halves of the blocked-save affordance, including focus asserted **twice in a loop**, since one pass would have passed against the flag version.
- `dialogAccessibility.spec.ts`: a new case walking the exact sequence `docs/BRAILLE.md` promises — `Tab`, three `ArrowRight`, `Enter` — in a real browser. It is here and not in jsdom because `Enter` reaches a `<button>` as a synthesised click, which jsdom does not model.
- The three existing settings component tests now open the tab holding what they assert on.
- `e2e_tests/utils/constants.ts`: `SETTINGS_MENU_TITLE` was the "General Settings" section heading; it is now the dialog's own title.
- `docs/BRAILLE.md`: the multiline setup steps said "Press Tab to reach the Braille Display options", which no longer lands there.

Captured with Playwright against `examples/barplot.html`. Described rather than embedded — I have no way to upload image attachments from this environment, and a broken `` is worse than a description.

**Before** — one scrolling list: Volume, 3D Echo Count, Echo Volume, Echo Duration, Outline Color, four High Contrast rows, the Braille Display radios, then the scrollbar, with the tactile display, the frequency range, ARIA Mode, Hover Mode, all four LLM providers and the whole About section below the fold.

**After** — six tabs under the "Settings" title, above a panel that fits. Dialog height is 606px on General, Audio, Visual and About; 643px on Braille & Tactile; 722px on AI — identical whether a tab is being opened for the first time or revisited, since hidden panels take no space. Nothing scrolls on any tab. Arrowing the tablist draws a 2px outline on the focused tab while the selected one keeps its underline. Under emulated forced colours the selected tab keeps a `Highlight` underline while the rest stay bare. When the custom instruction is too short, the AI tab carries a red dot and the footer reads, in red above the buttons, "Custom instructions on the AI tab must be at least 10 characters long". At 420×720 the tablist collapses to scroll buttons and every tab stays reachable.

## Checklist

## Additional Notes

Verified in a real browser as well as in jsdom, because most of what broke here is invisible to jsdom:

- **Accessible names read out of Chromium's own accessibility tree**, not a library's re-implementation: dialog `"Settings"`, tablist `"Settings sections"`, exactly one `tabpanel` (the visited-but-hidden ones are out of the tree), and the badged tab computing as `"AI needs attention"` — the visible label extended, not replaced.
- **Keyboard**: `Tab` reaches the tablist, arrows move focus without activating, `Enter` selects, `Tab` then enters the panel. `Escape` closes. `Alt+S` from About persisted an echo count typed on Audio. Three consecutive refused saves, each from a different starting focus, each landed focus on the AI tab with `aria-selected="true"` already set; a valid `Alt+S` still saves and closes from any tab.
- **The regressions, each re-checked after the fix**: the AI panel's key field is the same element on re-entry; an unvisited panel is absent from the DOM; scroll returns to 0 on a switch; the copy status clears; the chat button lands on AI.

`npm run lint`, `npm run type-check`, `npm run build` and `npm test` (7,204 tests) all pass. Playwright passes in full on Chromium (534 tests, including the new keyboard case) — Firefox and WebKit could not run here, as this container ships only a Chromium binary, so CI is the authority for those two.

**One thing I could not verify and would value a second pair of hands on:** whether NVDA and JAWS announce the footer hint. The mechanism is now the one this codebase already relies on for its copy status — text mutating inside a region that was mounted empty — rather than the insertion-of-a-populated-node form the automated review rightly flagged as historically divergent between screen readers. But I tested against Chromium's accessibility tree, not against a real screen reader.

**Two things deliberately left alone**, both predating this PR, so as not to widen the diff — each worth a follow-up: the AI panel's `Alert` is MUI's default `role="alert"`, which interrupts assertively for what is field validation; and `Save` uses `disabled` rather than `aria-disabled`, so it is out of the tab order and its `title` reaches nobody.

https://claude.ai/code/session_018QUbQ9yPxyGENZ4z5QxzEj

### DIFF
--- a/docs/BRAILLE.md
+++ b/docs/BRAILLE.md
@@ -986,7 +986,8 @@ Users can set up multiline braille display support by following these steps:
 
 1. Open any plot in MAIDR.
 2. Press `Ctrl+,` (`Command+,` on macOS) to open the settings dialog.
-3. Press Tab to reach the Braille Display options.
+3. Press Tab to reach the list of settings tabs, then press Right Arrow until you hear "Braille & Tactile" and press Enter to open it.
+4. Press Tab to reach the Braille Display options.
 	- If your braille display model is listed, choose "Single line" or "Multi-line", then tab to the Single-Line Display or Multi-Line Display dropdown and select your braille display model.
 	- If your braille display model is not listed, choose "Configure manually", then enter Braille Display Size (cells per row) and Braille Display Lines (rows). Set the number of lines to 1 for a single-line display.
-4. Press `Alt+S` to save and close the settings dialog.
+5. Press `Alt+S` to save and close the settings dialog. It saves every tab, so you do not need to return to the one you started on.
--- a/e2e_tests/specs/dialogAccessibility.spec.ts
+++ b/e2e_tests/specs/dialogAccessibility.spec.ts
@@ -170,6 +170,38 @@ test.describe('dialog accessibility tree', () => {
     expect(structure.headings[0]).toBe('H2:Settings');
   });
 
+  // `docs/BRAILLE.md` tells blind readers exactly this sequence to reach the
+  // braille display options. It rests on manual activation — an arrow key
+  // moves focus, Enter selects — and on Enter reaching a `<button>` as a
+  // synthesised click, which jsdom does not model. So it is asserted here
+  // rather than in the component tests.
+  test('the settings tabs work by keyboard the way the docs describe', async ({ page }) => {
+    const barPlotPage = await setupBarPlotPage(page);
+    await barPlotPage.openSettingsMenu();
+
+    await page.keyboard.press('Tab');
+    await expect(page.getByRole('tab', { name: 'General' })).toBeFocused();
+
+    for (let i = 0; i < 3; i++) {
+      await page.keyboard.press('ArrowRight');
+    }
+    const brailleTab = page.getByRole('tab', { name: 'Braille & Tactile' });
+    await expect(brailleTab).toBeFocused();
+    // Arrowing past a tab must not open it: on the AI tab that would mean
+    // contacting a provider the reader was only passing over.
+    await expect(page.getByRole('tab', { name: 'General' })).toHaveAttribute(
+      'aria-selected',
+      'true',
+    );
+
+    await page.keyboard.press('Enter');
+
+    await expect(brailleTab).toHaveAttribute('aria-selected', 'true');
+    await expect(page.getByRole('tabpanel')).toHaveAccessibleName('Braille & Tactile');
+    // Focus stays on the tab, so the next arrow continues along the list.
+    await expect(brailleTab).toBeFocused();
+  });
+
   test('a select menu inside the settings dialog resolves by role', async ({ page }) => {
     const barPlotPage = await setupBarPlotPage(page);
     await barPlotPage.openSettingsMenu();
--- a/e2e_tests/specs/dialogAccessibility.spec.ts
+++ b/e2e_tests/specs/dialogAccessibility.spec.ts
@@ -174,6 +174,10 @@ test.describe('dialog accessibility tree', () => {
     const barPlotPage = await setupBarPlotPage(page);
     await barPlotPage.openSettingsMenu();
 
+    // The dialog opens on "General", which has no select at all; the tactile
+    // display picker is the first enabled one, on the tab beside it.
+    await page.getByRole('tab', { name: 'Braille & Tactile' }).click();
+
     // A Select's menu is a modal too, and it is nested deeper still, so it
     // needs the same container scoping the dialog does.
     await page
--- a/e2e_tests/utils/constants.ts
+++ b/e2e_tests/utils/constants.ts
@@ -226,7 +226,9 @@ export abstract class TestConstants {
    * Settings Menu Identifiers
    */
 
-  static readonly SETTINGS_MENU_TITLE = 'General Settings';
+  // The dialog's own title. It used to be the "General Settings" section
+  // heading, which the tabs replaced — every section is now named by its tab.
+  static readonly SETTINGS_MENU_TITLE = 'Settings';
 
   /**
    * Chat Menu Identifiers
--- a/src/state/viewModel/settingsViewModel.ts
+++ b/src/state/viewModel/settingsViewModel.ts
@@ -2,7 +2,7 @@ import type { PayloadAction } from '@reduxjs/toolkit';
 import type { SettingsService } from '@service/settings';
 import type { Disposable } from '@type/disposable';
 import type { DotPadState, DotPadTransport } from '@type/dotPad';
-import type { Settings } from '@type/settings';
+import type { Settings, SettingsSection } from '@type/settings';
 import type { AppStore } from '../store';
 import { createSlice } from '@reduxjs/toolkit';
 import { dotPadSession } from '@service/dotPadSession';
@@ -35,6 +35,9 @@ const { update, reset } = settingsSlice.actions;
  */
 export class SettingsViewModel extends AbstractViewModel<SettingsState> {
   private readonly settingsService: SettingsService;
+
+  /** Set by the caller that opens the dialog; cleared on the next toggle. */
+  private openOnSection: SettingsSection | null = null;
 
   /**
    * Creates a new SettingsViewModel instance and loads initial settings.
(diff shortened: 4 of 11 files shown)


jooyoungseo:
Both points from the automated review are addressed in `66fe44e`, along with several more that a dedicated accessibility pass and code review turned up. Details are in the updated PR description; the two you raised specifically:

**2 — the footer hint's live region.** You were right, and it was worse than "worth a manual pass": the region could not announce **at all**, in any screen reader. It was created already holding its message, and the only path to it was a tab switch — the fields that invalidate the instruction live on the AI tab, where the hint is suppressed — so there was never a moment when a mounted region's text changed. It is now rendered while empty and re-texted, the same shape the copy-status region twelve lines below already uses for exactly this reason. `test/ui/settings.tabs.test.tsx` pins it with a `MutationObserver` across the AI → General switch rather than by asserting on the text, since the text alone reads the same either way.

That said, your underlying caution stands and I have carried it into the PR description: I verified against Chromium's accessibility tree, not against NVDA or JAWS. The mechanism is now the one this codebase already depends on rather than the historically divergent one, but a real screen-reader pass would still be worth having before merge.

**1 — other specs opening the settings dialog.** Checked empirically rather than by reading: the full Playwright suite passes on Chromium, 534 tests including the new keyboard case. The only specs that open the dialog go through `showSettingsMenu` / `closeSettingsMenuWithEscape` / `verifySettingsMenuIgnoresBackdropClick` in `base-page.ts`, which assert on the dialog title and the footer buttons — all outside the panels. `dialogAccessibility.spec.ts` was the one place reaching for a control on a specific page. Firefox and WebKit could not run in my container (only a Chromium binary is installed), so CI is the authority for those two.

Two smaller notes for the record, since both reviews raised them and I did **not** act on either:

- The `Alert` in the AI panel is `role="alert"` (MUI's default), which interrupts assertively for what is a field-validation warning. Now that panels persist it no longer re-fires on every visit, and it predates this PR, so I have left it rather than widen the diff. Worth a follow-up.
- `Save` uses `disabled` rather than `aria-disabled`, so it is out of the tab order and its `title` reaches nobody. Also predates this PR. The blocked-`Alt+S` fix in `66fe44e` gives that path a response, but the button itself is unchanged.

---

jooyoungseo:
Good catch on the focus, and it was worse than the re-visit case you described — fixed in `12ef5a1`.

**The finding was right, and broader than stated.** You framed it as "the *first* blocked `Alt+S` at least mounts the `role="alert"`, so only the second-and-later press is silent". With panels now persisting, even the first one is silent in the ordinary case: the AI panel is mounted as soon as the reader has been there, and that is where the instruction gets typed, so by the time a save is refused the alert has been sitting in the DOM for a while. The tab switch mutated nothing at all. So the branch answered silently again — the behaviour the commit it lives in says it stopped.

Focus now moves to the AI tab, whose name already reads `"AI needs attention"`. Two details in how, both load-bearing:

- **In an effect, not in the key handler.** Focusing synchronously would land on the tab before the render that marks it selected, so a screen reader would announce it as a tab the reader had merely arrived at rather than the one now open.
- **Keyed on a count of refusals, not a boolean.** A flag would fire once and swallow every refusal after it — the exact shape of the bug you found.

Verified in Chromium across three consecutive refusals, each from a different starting focus: `document.activeElement` is the AI tab with `aria-selected="true"` already set every time, and a valid `Alt+S` from another tab still saves and closes. `test/ui/settings.tabs.test.tsx` pins the repeat specifically — it asserts focus twice in a loop, since one pass would pass against the flag version.

**Also fixed:** the stale `aria-controls` comment you flagged. The behaviour was already right; the reason given for it was not, since a visited panel *is* in the document now. It now says what actually justifies it — a hidden panel is out of the accessibility tree, so naming it would be no better than naming one that was never rendered.

**One correction to the review, for the record:** `useCredentialProbe` does not poll. It is a single debounced `setTimeout` inside a `useEffect` keyed on `(provider, apiKey, enabled)` — `src/state/hook/useCredentialProbe.ts:62-72` — so a mounted-but-hidden AI panel is idle, not making repeated background requests. The trade-off you flagged is real but smaller than "keeps polling for the rest of the dialog's lifetime": one probe per credential change, and none while nothing changes.

`npm run lint`, `npm run type-check`, `npm run build`, `npm test` (7,204) and the Playwright suite on Chromium (534) all pass locally on `12ef5a1`.

---

PR #1242, opened 2026-09-07
fix(ui): make the settings dialog's blocked save explain itself

jooyoungseo:
# Pull Request

## Description

Follow-up to #1241. Two accessibility defects were found during that review, judged out of scope because they predate the tab split, and recorded there as follow-ups. This is that follow-up. Both are about the same moment — the reader cannot save, and the dialog does a poor job of telling them why.

Neither was introduced by #1241, but the tab split raised the stakes on both: the field that blocks Save is now on a tab the reader may not be looking at.

## Related Issues

Follows up on #1241 (the two items listed under "Two things deliberately left alone" in its description).

### `c6cf32f` — an unavailable Save that can be reached

`disabled` takes a button out of the tab order. A reader working through the footer met Reset, then Close, then wrapped — never learning that Save existed, that it was unavailable, or why. The `title` carrying the reason could not help either: it needs a hover a keyboard user cannot perform, and `aria-label` outranks it in the name computation regardless.

It is now `aria-disabled`. The button keeps its place in the tab order, announces its own unavailable state, and carries the reason as a description. That description is a separate node from the footer hint on purpose — the footer hint is a live region that deliberately goes quiet on the AI tab, and a description has to be there on every tab, because the button is.

Pressing it now answers the way `Alt+S` does rather than doing nothing. Both go through one `handleSaveRequest`, so a refusal cannot come out differently depending on how the reader asked for it.

### `0443388`, `f3e4911` — a warning that is polite *and* still heard

The warning under the custom instruction is a MUI `Alert`, which defaults to `role="alert"` and so interrupts assertively — too much for a field the reader is in the middle of typing, and against the repo's own rule reserving assertive for the result of the reader's own navigation.

**Simply asking for `role="status"` instead would have made it worse.** The warning appears and disappears as the instruction crosses the length bar, and a status region created already holding its text is routinely not announced at all, where an alert on insertion is. So the region is now the part that stays: rendered whether or not there is anything to say, with its text changing. That is what a live region announces on, and it is the shape the copy status and the footer hint in this same file already use.

`f3e4911` finishes that thought. The region was still inside the block gated on the Custom expertise level, which put it back in the shape it was reshaped to avoid — **picking "Custom" from the dropdown is the ordinary way into this state**, the instruction is empty at that moment, so the region was mounting alongside its own first message. It now sits outside that block and lives as long as the panel.

The field also identifies itself as the one at fault — `aria-invalid`, plus a description pointing at that region — rather than leaving a reader to discover the warning sitting underneath it.

### `ebf9ff7`, `fbfd2a0` — comments that outlived what they described

Four comments in the tab tests were written against behaviour that has since changed, and are the kind a later reader trusts. Three described an inactive panel as unmounted, from before panels started mounting on first visit and staying — they asserted the opposite of what the code does. The fourth called Save "disabled" one commit after it stopped being, which is exactly the distinction the tests around it now turn on, since one of them asserts it is deliberately *not* `disabled`.

The assertions were already right in every case; only the stated reasons were stale.

No visual change. `Save & Close` renders exactly as it did — greyed out, reading as unavailable — because the disabled palette is applied through `sx` instead of coming free with the `disabled` attribute. The warning still sits 20px under the textarea, left-aligned with it, and its region occupies no space while empty: the dialog measures the same height on the AI tab with a valid instruction as with an invalid one.

## Checklist

## Additional Notes

Read back out of Chromium's own accessibility tree, since the whole point is what assistive technology is handed:

```
AX name       : "Save & Close Settings"
AX description: "Custom instructions on the AI tab must be at least 10 characters long"
AX disabled   : true
AX focusable  : true
```

`disabled` and `focusable` both true is the goal state — unavailable, and still reachable. Tabbing from Close now lands on Save, which it did not before; pressing Enter there keeps the dialog open and moves focus to the tab reading "AI needs attention".

The live region was watched on the running page rather than only in jsdom:

```
region exists before Custom is picked : yes, empty
mutations on picking Custom           : ["Custom instructions must be at least 10 …"]
mutations on typing across the bar    : ["", "Custom instructions must be at least 10 …"]
```

Three test cases were added, each pinning a property rather than the markup that produces it: that Save keeps its tab stop and answers a press; that the warning region exists *before* it has anything to say, across a typed valid→invalid transition; and that it exists before **Custom is picked**, which is the path `f3e4911` fixes. A fourth property — that the field's `aria-describedby` resolves to that region — is asserted inside an existing case.

### What the live region does and does not carry

The claim is that the warning announces because its region outlives its content. After `f3e4911` that holds for every transition *within* a mounted AI panel — typing across the length bar, and picking "Custom" from the dropdown.

It does not hold for the panel's own first mount, which can arrive with the warning already inside it whenever the dialog opens on saved settings that are already `custom` and too short. Two routes reach that: a refused `Alt+S` from another tab, and simply opening the AI tab. **An earlier version of this description named only the first; the review was right that it is both.**

On neither route is the live region the mechanism, and on neither is the reader left without the reason:

- The tab is named **"AI needs attention"**, so selecting it — however they got there — announces that something on it wants attention.
- The field carries `aria-invalid` and a description resolving to the same text, so reaching it states the requirement.
- On the refused-save route, focus is moved to that tab deliberately, which is what makes the refusal perceptible at all.

So the live region is what covers changes; the tab name and the field description are what cover arrival. `should open the tab holding the reason when a blocked Alt+S is pressed` exercises the first-mount case.

**Still not verified, and unchanged from #1241:** none of this has been through NVDA or JAWS. That matters more for the warning region than for the Save button, since the reason the region is mounted-and-mutated rather than inserted is precisely a screen-reader behaviour I can reason about but not observe here.

**The third item from #1241's follow-up list is not here.** That was the NVDA/JAWS pass itself, which needs a real screen reader.

`npm run lint`, `npm run type-check`, `npm run build` and `npm test` (7,207) all pass, as does the Playwright suite on Chromium (534). Firefox and WebKit could not run here — this container ships only a Chromium binary — so CI is the authority for those two.

https://claude.ai/code/session_018QUbQ9yPxyGENZ4z5QxzEj

### DIFF
--- a/src/ui/component/Settings.tsx
+++ b/src/ui/component/Settings.tsx
@@ -559,6 +559,7 @@ const Settings: React.FC = () => {
   const copyStatusId = `${id}-copy-status`;
   const tactileLabelId = `${id}-tactile-label`;
   const tactileStatusId = `${id}-tactile-status`;
+  const saveBlockedId = `${id}-save-blocked`;
   const tactileMenu = useModalContainer();
   const contentRef = React.useRef<HTMLDivElement>(null);
   // `HTMLDivElement` because that is what MUI declares `Tab`'s ref as, even
@@ -819,6 +820,31 @@ const Settings: React.FC = () => {
     = llmSettings.expertiseLevel !== 'custom'
       || llmSettings.customInstruction.length >= MIN_CUSTOM_INSTRUCTION_LENGTH;
 
+  /**
+   * Answers a save that cannot go through.
+   *
+   * Doing nothing would leave a reader with no response at all — no sound, no
+   * text, nothing saying why. Opening the tab that holds the field is the
+   * answer to "why not", and puts them where the fix is; the effect above
+   * then moves focus there, which is the part they can actually perceive.
+   */
+  const refuseSave = useCallback((): void => {
+    setActiveTab('ai');
+    setVisitedTabs(previous =>
+      previous.has('ai') ? previous : new Set(previous).add('ai'));
+    setRefusedSaves(previous => previous + 1);
+  }, []);
+
+  // The single entry point for "the reader asked to save", so the button and
+  // the shortcut cannot answer a refusal differently.
+  const handleSaveRequest = useCallback((): void => {
+    if (!isCustomInstructionValid) {
+      refuseSave();
+      return;
+    }
+    handleSave();
+  }, [isCustomInstructionValid, refuseSave, handleSave]);
+
   // Dialog-scoped: KeybindingService's hotkeys.filter blocks shortcuts while
   // focus is in a non-MAIDR <input>, which would silently break Alt+s / Alt+c
   // inside the manual cells/lines fields.
@@ -831,26 +857,13 @@ const Settings: React.FC = () => {
       const key = e.key.toLowerCase();
       if (key === SAVE_SHORTCUT_KEY) {
         e.preventDefault();
-        if (!isCustomInstructionValid) {
-          // Returning silently would leave a reader who pressed the shortcut
-          // the dialog advertises with no response at all — no sound, no
-          // text, nothing saying why. Opening the tab that holds the field is
-          // the answer to "why not", and puts them where the fix is; the
-          // effect above then moves focus there, which is the part they can
-          // actually perceive.
-          setActiveTab('ai');
-          setVisitedTabs(previous =>
-            previous.has('ai') ? previous : new Set(previous).add('ai'));
-          setRefusedSaves(previous => previous + 1);
-          return;
-        }
-        handleSave();
+        handleSaveRequest();
       } else if (key === CANCEL_SHORTCUT_KEY) {
         e.preventDefault();
         handleClose();
       }
     },
-    [isCustomInstructionValid, handleSave, handleClose],
+    [handleSaveRequest, handleClose],
   );
 
   return (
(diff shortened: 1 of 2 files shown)


jooyoungseo:
Agreed on the edge case, and worth pinning down precisely since it qualifies a claim in the PR description.

The path is real: open the dialog with saved settings that are already invalid (`expertiseLevel: 'custom'` and a short instruction), never visit the AI tab, press `Alt+S`. The panel mounts for the first time with the `Alert` already inside it — the "inserted with content" shape, which is exactly what the live region was reshaped to avoid.

Your read of why it does not need fixing is right, and it is the one case where the live region was never the mechanism. Focus moves to the tab, whose accessible name is `"AI needs attention"`, and that focus move is what the reader perceives. `should open the tab holding the reason when a blocked Alt+S is pressed` covers that exact path — it starts with the AI tab unvisited, so the panel's first mount *is* what it exercises.

So the guarantee is narrower than the description implies, and it is the narrower one that matters: **the region announces every change after the panel exists** — which is every transition the reader can cause by typing, since typing is only possible on a mounted panel. The first mount is covered by focus instead. I have left the code alone and am noting the distinction here rather than in the description, since it is a property of the interaction rather than of the change.

One unrelated push since your review: `fbfd2a0` fixes a line in the test file's own docblock that still called Save "disabled" — the distinction the tests around it now turn on, since one of them asserts it is deliberately *not* `disabled`. Same class of staleness as `ebf9ff7`, missed on the first pass.

---

jooyoungseo:
The dropdown-switch case is a real gap and I have fixed it in `f3e4911`, rather than leaving it for the screen-reader pass.

You framed it as "beyond what the PR description already flags," and on reflection it is worse than that framing: it is not an edge case at all. **Picking "Custom" from the dropdown is the ordinary way a reader reaches this state.** The instruction is empty at that moment, so the warning is true immediately, and the region was mounting alongside its own first message — precisely the shape the whole commit exists to avoid. Where a refused save is rescued by the focus move to the tab, a dropdown change has nothing.

So the region no longer lives inside the `expertiseLevel === 'custom'` block. It sits outside it, lives as long as the panel, and every arrival of the warning is now a mutation of something already standing. Confirmed in Chromium with a `MutationObserver` on the live page:

```
region exists before Custom is picked : yes, empty
mutations on picking Custom           : ["Custom instructions must be at least 10 …"]
mutations on typing across the bar    : ["", "Custom instructions must be at least 10 …"]
aria-describedby resolves             : yes
```

The description still resolves because the region now outlives the field rather than the other way round. Layout is unchanged: the warning sits 20px under the textarea, left-aligned with it, and the region costs no space while empty.

`should announce the warning when Custom is picked, not spring it into being` pins it, and it is the case your review correctly noted had no coverage.

**What that leaves.** One path still mounts the region with content: the AI panel's very first mount, when the dialog opens on already-invalid saved settings and the reader presses `Alt+S` without visiting the tab. That one is covered by the focus move, as we agreed, and it is now the *only* such path rather than one of two.

`npm run lint`, `type-check`, `build`, `npm test` (7,207) and Playwright on Chromium (534) all pass on `f3e4911`.

---

PR #1244, opened 2026-09-07
feat(keybinding): answer an unassigned key instead of doing nothing

jooyoungseo:
# Pull Request

## Description

A key that no shortcut in the current scope claims used to produce nothing at all — no sound, no announcement. A reader cannot tell that apart from a chart that has stopped responding, and nothing about it reveals that a list of the real keys exists. Guessing was the only way in, and guessing wrong taught nothing.

Such a press now answers **"Invalid key. Press Control Slash for keyboard help."** together with the existing warning tone.

The chord is spelled out in words on purpose: a screen reader at default verbosity commonly skips punctuation, so `Ctrl + /` can reach the listener as just "control" — half an instruction. And it says **Command Slash** on macOS, because `command+/` is the chord actually bound there.

## Related Issues

None filed; raised directly.

## Changes Made

**`src/command/invalidKey.ts`** (new) — `InvalidKeyCommand`. Decides whether an unclaimed press was aimed at MAIDR, then notifies and plays the warning tone. Constructed directly by `KeybindingService`, the way `Mousebindingservice` constructs `PointerGuidanceCommand`: `CommandFactory` maps a key *of* a scope's keymap to a command, and this one answers for every key that is not in it.

**`src/service/keybinding.ts`** — registers one `*` handler per scope that binds the help chord, and marks the event a binding claimed.

- **Which scopes warn.** Only those that bind `HELP_CHORD`. The advice has to be advice the reader can act on, and inside settings, chat, the command palette or the help dialog itself, help is not reachable and the keystroke is likelier to be text than a missed shortcut. That list is derived from the keymaps rather than hardcoded, and the chord is now a named constant the six keymaps share, so rebinding help moves both halves at once.
- **How "unclaimed" is decided.** By hotkeys-js itself, not by a second copy of its key parsing that could drift from the keymaps. It dispatches every `*` handler *before* the handlers bound to that specific key, so the check waits on a microtask for the rest of the dispatch to finish and then compares event identity.
- Nothing calls `preventDefault`, so no key changes what it did before.

**`src/util/platform.ts`** — `Platform.ctrlSpoken` (`Control` / `Command`), the spoken counterpart to `Platform.ctrl`'s hotkeys-js token.

**What stays silent.** Anything held with Ctrl, Cmd or Alt — that is where the browser, the OS and the screen reader keep their own commands, VoiceOver owning every Ctrl+Option among them, and answering "invalid key" to a working screen-reader command would be both wrong and unstoppable. Also the modifiers themselves, Caps Lock, Tab, the lock keys, Insert, F1–F12, IME/dead keys, and the repeats of a held key. Shift is deliberately *not* on that list: Shift and a letter is still someone typing a key at the chart.

Not applicable — the change is an announcement and a tone.

## Checklist

## Additional Notes

**Tests.** `test/command/invalidKey.test.ts` covers the message and every ignore rule. `test/service/keybindingInvalidKey.test.ts` runs against the real hotkeys-js in jsdom — an unassigned key warns, a key that runs a shortcut does not, and a scope without the help chord stays quiet — which is what pins the wildcard-then-microtask ordering this depends on.

**Verified locally:** `npm run lint:fix`, `npm run type-check`, `npm run build` all clean; `npm test` green at 550 unit suites / 7592 tests plus 15 ESM suites / 177 tests.

**Docs.** `docs/CONTROLS.md` gains a short "Unassigned Keys" section covering the message and what is left alone.

https://claude.ai/code/session_01U977UMv7uEhebax7npqVX7

---
_Generated by [Claude Code](https://claude.ai/code/session_01U977UMv7uEhebax7npqVX7)_

### DIFF
--- a/docs/CONTROLS.md
+++ b/docs/CONTROLS.md
@@ -46,9 +46,13 @@ A key that no shortcut in the current mode uses answers with a short warning —
 — together with the warning tone, so a guess is never met with silence.
 
 Keys held with Control, Command, or Option are left alone, because that is
-where the browser and your screen reader keep their own commands. Inside the
-settings, chat, and command palette dialogs there is no warning either: help is
-not reachable from them, and what you type there is usually text.
+where the browser and your screen reader keep their own commands.
+
+The warning appears only where the help shortcut itself works — reading a
+chart, in braille mode, and in label mode. Anywhere it does not, such as the
+dialogs, review mode, and inside a grid cell, an unassigned key stays silent:
+there would be nothing useful to point you at, and what you press there is
+often text rather than a missed shortcut.
 
 ## Monitor Mode (Live Charts)
 
--- a/docs/CONTROLS.md
+++ b/docs/CONTROLS.md
@@ -39,6 +39,17 @@ Below is a detailed list of keyboard shortcuts for various functions:
 | Open/Close keyboard shortcut help       | Control + /                 | Command + /                 |
 | Open AI chat (API key set in Settings)  | Shift + / (?)               | Shift + / (?)               |
 
+## Unassigned Keys
+
+A key that no shortcut in the current mode uses answers with a short warning —
+"Invalid key. Press Control Slash for keyboard help." (Command Slash on a Mac)
+— together with the warning tone, so a guess is never met with silence.
+
+Keys held with Control, Command, or Option are left alone, because that is
+where the browser and your screen reader keep their own commands. Inside the
+settings, chat, and command palette dialogs there is no warning either: help is
+not reachable from them, and what you type there is usually text.
+
 ## Monitor Mode (Live Charts)
 
 On charts configured with `live: true`, press **M** to toggle monitor mode. While monitoring is on, every newly streamed data point is automatically sonified and announced by your screen reader without moving your current position. See the [Live & Streaming Data](LIVE_DATA.html) guide for details.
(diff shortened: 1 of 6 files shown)


jooyoungseo:
Thanks — nit 1 was right and is fixed in f37ed9d.

Naming settings, chat and the command palette listed three of the nine silent scopes, so a reader could reasonably expect a warning in the go-to-extrema dialog, review mode or a grid cell and not get one. `docs/CONTROLS.md` now states the rule instead of a partial list: the warning appears where the help shortcut itself works — reading a chart, braille mode, label mode — and stays silent anywhere it does not. That stays accurate if a scope is added later.

Nit 2 (the `STOP_AUTOPLAY` hotkeys-js#172 wildcard treating any Ctrl/Cmd keydown as claimed) I left alone, as you suggested — it is pre-existing and cannot produce a wrong warning, since `isUnassignedPress` excludes every Ctrl/Cmd/Alt chord before it gets that far. Worth fixing on its own, not by widening this PR.

On what you could not run in your environment, verified locally since:

- `npm run lint:fix`, `npm run type-check`, `npm run build` — clean.
- `npm test` — 550 unit suites / 7592 tests, plus 15 ESM suites / 177 tests, all passing.
- The full chromium e2e suite — 540 passed.
- In a real browser on `examples/barplot.html`: `q` and `7` announce "Invalid key. Press Control Slash for keyboard help."; `Ctrl + /` then actually opens the help dialog; `ArrowRight`, `b` and `t` announce what they always did; `Ctrl + C` and keys pressed inside the help and settings dialogs announce nothing.

---

PR #1245, opened 2026-09-07
feat(ui): read the description table the way the chart's author asked

jooyoungseo:
# Pull Request

## Description

A chart layer may declare a `format` per axis — currency, a percentage, a date — and every spoken and braille announcement has honoured it for a long time. Walking `examples/vertical-candlestick.html` announces **"Jan 3"** and **"$180"**.

The data table in the same `d` dialog did not. It printed `2023-01-03` and `180.2534`, because nothing told it which column sat on which axis. One dialog described one value two ways, with nothing to say which was meant — and a reader comparing the table against what they had just heard had no way to tell whether they had misheard, or whether the chart disagreed with itself.

This was left as an explicit follow-up on #1243: *"Fixing it needs per-column axis identity on `DescriptionState.dataTable` and a change in every trace that builds one — worth doing, too large to ride along here."*

## Related Issues

Follow-up to #1243.

## Changes Made

**Per-column axis identity.** `DescriptionState.dataTable` gains an optional `columnAxes: (AxisType | undefined)[]`, index-aligned with `headers`. `DescriptionService` reads it and routes each column through the layer's own formatter.

**Gated on the format being *authored*, not merely resolved.** An axis with no `format` still has a formatter — the default — so the service asks `FormatterService.hasAuthoredFormat` first. A chart nobody formatted reads exactly as it did before, rather than gaining the word `missing` in every empty cell of a sparse table.

**Blanks stay blank.** An axis format is a reading for a value, and two kinds of cell hold none. An empty cell is how a sparse table shows an absent value, and a chart's own `format.function` sees it as text rather than absence: `Number(value).toFixed(1)` meeting `''` yields `0.0`, inventing a reading for a cell that has none. A non-finite number has to reach the dialog as a number for its own blanking check. Both fall through to the rounding the column would have used with no format at all, so declaring a format changes how values read and never how their absence does.

**All 32 traces that build a table now declare their columns.** A category column takes its axis like any other — the bar example maps `"Sat"` to `"Saturday"` through an x format function, and a category name is an axis value. A column the chart *computes* rather than measures takes none: a share, a running total, a rank, a delta, a duration, a joined ancestry, a trend word. Gantt's `Start` and `End` take none either, because `atTime` has already run the author's formatter over them and naming the axis would run it twice.

Where a family swaps which axis carries its categories, the column list is built from the same expression that swaps the headers; where a column is conditional, from the same condition. The two cannot drift.

Not visual — the change is in what the dialog's table says. For the candlestick example, before and after:

| Column | Announced | Table before | Table after |
| --- | --- | --- | --- |
| Date | `Jan 3` | `2023-01-03` | `Jan 3` |
| Open | `$180.25` | `180.2534` | `$180.25` |
| Volatility | `$4.62` | `4.617` | `$4.62` |
| Trend | `Bull` | `Bull` | `Bull` |

## Checklist

## Additional Notes

**Verified locally:** `npm run lint:fix` (no changes), `npm run type-check` (clean), `npm run build` (15/15), unit **549 suites / 7746 tests**, ESM **15 suites / 177 tests**. The full `npm test` was run as its constituent projects rather than in one invocation, because the runner's default worker count exhausts memory on this machine.

**On the tests.** `test/service/descriptionAxisFormat.test.ts` drives a real `Figure`, `Context`, `FormatterService` and `DescriptionService` rather than mocks — the defect lived in the seam between them, and a mocked formatter is exactly the thing that cannot prove it closed. It also pins the *unformatted* chart to what it printed before, which is the half of the contract that would otherwise regress silently.

`test/model/descriptionContract.test.ts` gains a test asserting `columnAxes` is either absent or exactly as long as its headers, plus sixteen cases for the *conditional* shapes — a candlestick with no open, a candlestick carrying volume, an error bar with groups, an error bar drawing bounds only, a treemap of a pure hierarchy, a network of several components, a horizontal histogram, a horizontal gantt. A table whose width depends on its data has a second shape the ordinary case never builds, and a column list that drifts there is invisible until someone reads that chart. Each of those cases was checked to produce a genuinely different table width from its sibling, so none of them is a case that passes by describing the same shape twice.

**Two judgement calls worth a reviewer's eye:**

1. **Candlestick's `Volatility` column takes the price axis.** It is `high − low`, a difference rather than a price — but `text` announces it through `crossAxis`, which *is* the price axis, so the table and the cursor agree. A currency format renders it `$4.62`, which reads correctly as a range in dollars.
2. **A stacked area's `Total` column takes no axis**, while `TextService` announces `state.stack.value` through the y formatter. So a layer with an authored y format announces the total formatted while the table prints the bare rounded float. A running total is computed rather than measured, which is the rule everywhere else here; changing it would be a decision about the *announcement*, not a fix to these files. Flagged rather than silently resolved.

https://claude.ai/code/session_019qQxbADuB9BZPwjv46ntaR

---
_Generated by [Claude Code](https://claude.ai/code/session_019qQxbADuB9BZPwjv46ntaR)_

### DIFF
--- a/src/controller.ts
+++ b/src/controller.ts
@@ -174,6 +174,7 @@ export class Controller implements Disposable {
       this.context,
       this.displayService,
       this.rotorNavigationService,
+      this.formatterService,
     );
     this.helpService = new HelpService(this.context, this.displayService);
     this.chatService = new ChatService(
--- a/src/model/area.ts
+++ b/src/model/area.ts
@@ -421,6 +421,10 @@ export class AreaTrace extends LineTrace {
 
     return {
       headers: [...table.headers, TOTAL],
+      // The line's own columns keep their axes; the total is summed here out
+      // of every band at that x rather than read off an axis the layer
+      // declared, so it carries none and keeps the dialog's own rounding.
+      columnAxes: table.columnAxes && [...table.columnAxes, undefined],
       // Looked up by the x in column 0 of the row the line built, never by
       // index, for the reason `columnTotals` gives. A column reached only by
       // gaps carries NaN, which the dialog blanks rather than printing a stack
--- a/src/model/bar.ts
+++ b/src/model/bar.ts
@@ -353,12 +353,20 @@ export abstract class AbstractBarPlot<T extends BarPoint> extends AbstractTrace 
       return [main, isMeasured(value) ? value : MISSING_TEXT];
     });
 
+    // The category column sits on whichever axis carries the names and the
+    // magnitude column on the other one, swapping with the headers above, so a
+    // layer that formats its magnitudes as currency -- or spells "Sat" out as
+    // "Saturday" -- reads the same way here as when the reader walks the bars.
+    const columnAxes: DescriptionState['dataTable']['columnAxes'] = isVertical
+      ? ['x', 'y']
+      : ['y', 'x'];
+
     return {
       chartType: this.getChartTypeLabel(),
       title: this.title,
       axes: this.getDescriptionAxes(),
       stats,
-      dataTable: { headers, rows },
+      dataTable: { headers, columnAxes, rows },
     };
   }
 
--- a/src/model/box.ts
+++ b/src/model/box.ts
@@ -1,7 +1,7 @@
 import type { BoxplotSectionType } from '@type/boxplotSection';
 import type { BoxPoint, BoxSelector, MaidrLayer } from '@type/grammar';
 import type { Movable } from '@type/movable';
-import type { AudioState, BrailleState, DescriptionState, TextState } from '@type/state';
+import type { AudioState, AxisType, BrailleState, DescriptionState, TextState } from '@type/state';
 import type { Edge, LineRequest, WhiskerRequest } from '@util/svg';
 import type { Dimension, NearestPoint } from './abstract';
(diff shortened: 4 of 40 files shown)


jooyoungseo:
Checked the one open question from the review — **`flow.ts` giving both `From` and `To` the `x` axis** (`columnAxes: ['x', 'x', 'y']`). It is correct as written, and the announcement is why.

`FlowTrace.text` returns `main: { label: this.nodeLabel, value: node.name }` with `mainAxis: 'x'` (`src/model/flow.ts:556-558`), and `TextService` formats `text.main.value` through `mainAxisType` (`src/service/text.ts:175-183`). So a Sankey whose author declares an x format **already** has its node names formatted in every move announcement, and has for as long as that formatter has existed. The table's two node columns now say the same thing the cursor does, which is the whole point of the change.

Two columns sharing an axis is not unusual here once you look for it — a box plot's table puts five section columns and the outliers on the value axis, and boxen and violin do the same. What makes a column's axis unusual is not how many share it but whether the value in it was *measured* on that axis, and a ribbon's two ends are both node names of exactly one kind.

No change made. The other two points in the review are the judgement calls the PR body already flags, and I've left them as flagged rather than resolved.

---

PR #1246, opened 2026-09-11
feat(i18n): add a language setting and speak Korean everywhere

jooyoungseo:
# Pull Request

## Description

MAIDR now has a **Language** setting and a localization layer behind it, and ships **Korean (한국어)** as the first language beyond English. When Korean is chosen, every announcement, the chart description dialog, the AI chat, the Go To dialog, the keyboard shortcut help, the command palette, mode notifications, and the settings dialog itself are read in Korean. Chart data (axis labels, category names, titles) is never translated; only MAIDR's own words around it are.

The default is **Browser default**: a reader whose browser asks for Korean hears Korean without opening any dialog, and everyone else keeps English exactly as before. English output is byte-identical to `main`, which is why the existing test suite (7806 unit tests, 178 ESM tests) passes unchanged.

## Related Issues

None filed; requested directly.

## Changes Made

**Mechanism** (`src/util/i18n/`)
- `t(key, params)` renders a message in the active locale; `setLocale`/`onLocaleChange`/`resolveLocale` manage it. `MessageKey` is `keyof typeof en`, and the Korean dictionary is typed `Record<MessageKey, string>`, so a key added to English without a Korean rendering fails `npm run type-check` rather than falling back at runtime. 1193 keys in each language, split by area (`text`, `keybinding`, `rotor`, `notification`, `description`, `dialogs`, `settings`, `model`, `llm`, `tactile`, `common`).
- Korean particles (조사) depend on the word before them, and that word is usually chart data. A template writes `{label|은는}` and the right form is attached at render time from the actual word (`josa.ts`), including digits and Latin letters by their Korean reading.
- `useLocale()` hook (`src/state/hook/useLocale.ts`) subscribes React components to locale changes so open dialogs re-render in the new language.

**Setting**
- `general.language: 'auto' | 'en' | 'ko'` (default `auto`). `SettingsService` applies it on load, save, and reset; an unknown stored value falls back to `auto`.
- A **Language** select is the first row of the General tab. Each language is named in itself, with a matching `lang` attribute per option.
- The `<article>` wrapping each chart carries `lang={locale}` so screen readers switch voice.

**Routed through the dictionary**
- Model layer: chart type names, plot type words, every description statistic label and composed value phrase, section labels (box plot, candlestick, waterfall, error bar, …), rotor filter units, extrema target labels, the activation instruction (and its pre-activation twin in `maidr-component.tsx`, now built from the same keys). Lookup tables store `MessageKey`s and translate on lookup; nothing calls `t()` at module load.
- Text service and commands: every trace/layer/subplot announcement in both terse and verbose modes, label and position announcements, boundary cues, review mode, invalid-key hint.
- Keybindings: `KeybindingEntry.description` is now a `MessageKey`; the help menu rebuilds when the locale changes and the command palette resolves descriptions on each open.
- Rotor, autoplay, audio, braille, monitor, high contrast, candlestick reference-line comparison, tactile display, DotPad session, LLM validation and chat errors.
- Dialogs: settings, help, command palette, chart description, chat, Go To extrema, reference-line picker.
- The AI system prompt gains "Respond in Korean (한국어)." when the locale is not English; nothing is appended for English.

**Cleanup**
- The `Constant.IS/ARE/THROUGH` connectives and the seven `Constant.*_MODE` rotor names are gone; the dictionary owns those words.
- Chat reducers no longer read locale state: the placeholder and error text are rendered in `prepare`, keeping reducers pure.

**Docs and tests**
- `docs/LOCALIZATION.md` (linked from the README) explains the setting, the particle syntax, and how to add a language.
- New tests: `test/util/i18n/*` (interpolation, particles, locale resolution), `settingsLanguage`, `settings.language` (picker), `textKorean`, `instructionKorean`, `descriptionKorean`, and a help-menu relocalization case.

Not applicable; the change is spoken output and dialog text.

## Checklist

## Additional Notes

- Verified locally: `npm run lint:fix` clean, `npm run type-check` clean, `npm test` 558/558 unit suites and 15/15 ESM suites passing.
- Rotor navigation modes are still identified by their (now translated) label, compared `t() === t()` within one keystroke, which is consistent; a future locale where two mode names collide would need a stable key next to the label. Left as is to keep the change to localization.
- Model version names, provider names, device names, MAIDR, Bluetooth, USB, and key names stay untranslated.

https://claude.ai/code/session_0117eUjVDozabMwNNBBVmigY

---
_Generated by [Claude Code](https://claude.ai/code/session_0117eUjVDozabMwNNBBVmigY)_


jooyoungseo:
Review follow-up, all pushed:

- **Finding 1 (stored language applied after the first paint)**: fixed in bb9ed98. `useMaidrController` now applies the stored language during its first render rather than in an effect, so `getInitialInstruction()` and the article's `lang` are already in the reader's language on the first render. `test/ui/initialInstructionLanguage.esm-test.tsx` pins this with a server render, which runs no effects at all, so the old effect-based version fails it.
- **Finding 2 (rotor modes identified by translated label)**: tracked in #1247 with the proposed stable-key shape, as suggested.
- **Early finding (root component importing `plotTypeLabel` from the model)**: fixed in b0c90cb; the lookup now lives in `src/util/plotTypeLabel.ts` and both the model and the root component import it from there.
- The two earlier CI reds were this PR's too and are fixed: the docs site build needed a title and description for the new `docs/LOCALIZATION.md` page (ec321c4), and the dialog-accessibility E2E spec opened the first settings select on another tab, which is now the language picker on General (1cb0cbe).

---

PR #1252, opened 2026-09-11
feat(i18n): ship every language but English as a locale pack

jooyoungseo:
# Pull Request

## Description

Each language added in #1246 and #1250 grew `maidr.js` by about 70 KB for every page, whether or not that page ever spoke it. This change moves every dictionary but English out of the core bundle into a **locale pack**, a small file beside `maidr.js`, and has MAIDR fetch a pack itself when a language is chosen and the page has not loaded it.

| Bundle | Before | After |
| ------ | ------ | ----- |
| `dist/maidr.js` | 2.25 MB (nine languages built in) | 1.71 MB (English only) |
| `dist/locale-<code>.js` | — | 58 to 91 KB each, loaded only when needed |

Adding a tenth language now costs nothing to pages that do not use it.

## Related Issues

Follows #1246, #1248, and #1250.

## Changes Made

**Registry** (`src/util/i18n/index.ts`, `src/util/i18n/pack.ts`)
- `MESSAGES` holds English and whatever packs have registered. `registerLocale(locale, messages)` adds a dictionary; `isLocaleLoaded` and `dictionary` report what is present. `t()` for an active locale whose pack has not arrived renders English, and registration re-renders once it does.
- Packs meet the core through `globalThis.maidrLocales`, the `dataLayer` pattern: a plain array before the core runs, an object whose `push` registers immediately after. Either load order works, and a second MAIDR bundle on the same page chains onto the first so both hear every pack.
- `getLocaleRevision()` advances on a locale switch and on a pack registering for the active locale. `useLocale()` subscribes to the revision rather than the locale string, since a late pack changes every message while leaving the locale unchanged; the pre-activation instruction memo keys on it too.

**Loader** (`src/util/i18n/localePack.ts`)
- `ensureLocalePack(locale)` resolves the pack's URL beside `maidr.js` (the same way `maidr-math.css` is found), or from `window.maidrLocaleBaseUrl`, injects one script per locale, and resolves once the pack registers. A pack already registered, in flight, or present as the author's own tag is not fetched again. When no location can be found it warns once and English stays.
- `SettingsService` calls it after every locale switch (page load, save, reset), so the switch never waits on the network.

**Packs** (`src/locale/<code>.ts`, `src/locale/all.ts`)
- Each pack is three lines: import its dictionary, register it. It imports nothing from the core, which is what keeps it to the size of its dictionary. `all.ts` loads every pack for a page that would rather have one file.

**Build and publishing** (`scripts/build.js`, `package.json`, `cdnjs/maidr.json`)
- `LOCALE_PACKS` generates a build per language emitting `locale-<code>.js` (classic script) and `locale-<code>.mjs`. Flat names rather than a `locale/` directory because each pack builds in its own worker and the merge step treats a directory two workers both emit as a collision. npm exports `maidr/locale/*`; cdnjs mirrors the eight classic-script packs.

**Tests**
- `test/util/i18n/pack.test.ts`: packs queued before or pushed after the core, English until the active pack arrives then a switch with listeners told, no notification for an unused locale, unknown locales ignored with a warning, two adopters both served.
- `test/util/i18n/localePack.test.ts`: URL resolution beside the bundle and from the override, one script per locale, resolution on load, failure on error, no refetch of an author's tag, warning when the bundle cannot be located.
- `test/state/hook/useLocale.test.tsx`: a component re-renders on a locale change and when the active locale's pack arrives.
- Existing Korean and multi-language tests load the packs they need; `dictionaries.test.ts` loads all of them and asserts every known locale is present.
- `e2e_tests/specs/languageAutoDetect.spec.ts` now exercises the runtime path the example pages take: `maidr.js` alone, then the pack fetched from beside it, for all nine browser locales.
- `test/cdnjs/manifest.test.ts` checks every pack is mirrored by name.
- `test/util/i18n/navigatorLanguages.ts`: a shared stub for the browser's languages whose restore removes the own properties it added, since Jest's node environment shares one `navigator` across the files in a worker; the i18n and settings language suites use it.

**Docs**: `docs/LOCALIZATION.md` explains the packs, both load orders, `window.maidrLocaleBaseUrl`, and the steps to add a language; the README mentions the packs.

Not applicable.

## Checklist

## Additional Notes

- Behaviour change to be aware of: a page that loads only `maidr.js` now hears one English announcement cycle before the fetched pack registers. A page that wants the very first announcement in the reader's language loads the pack itself (`<script src=".../locale-ko.js">` before or after the bundle, or `import 'maidr/locale/ko'`); `test/ui/initialInstructionLanguage.esm-test.tsx` pins that case.
- Key compaction (packing dictionaries as arrays aligned to the English key order) was considered and left out: with packs opt-in it saves only a few KB gzipped per language, and it would couple each pack to the exact key order of one core version, so a CDN serving a newer core with an older pack would render the wrong strings.
- Verified locally: `npm run lint:fix` and `npm run type-check` clean; the i18n, pack, loader, hook, Korean, cdnjs, and build-config suites pass; `languageAutoDetect` and `dialogAccessibility` E2E specs pass in Chromium against a fresh core and pack build.

https://claude.ai/code/session_0117eUjVDozabMwNNBBVmigY

---
_Generated by [Claude Code](https://claude.ai/code/session_0117eUjVDozabMwNNBBVmigY)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -200,8 +200,10 @@ jobs:
       - name: Install Playwright chromium
         run: npx playwright install --with-deps chromium
 
-      - name: Build core bundle
-        run: node scripts/build.js core
+      # The example pages fetch a locale pack from beside maidr.js when the
+      # browser asks for a language, so the packs have to be in dist too.
+      - name: Build core bundle and locale packs
+        run: node scripts/build.js core locales
 
       - name: Run e2e specs
         id: e2e
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ For detailed encoding schemes for each plot type, see the [Braille Generation do
 
 ## Languages
 
-Every announcement and dialog can be read in English, 한국어, 日本語, 中文, Español, Deutsch, Français, Italiano, or हिन्दी. Choose the language on the **General** tab of Settings; by default MAIDR follows the browser's language. Chart data is read as the author wrote it; only MAIDR's own words are translated.
+Every announcement and dialog can be read in English, 한국어, 日本語, 中文, Español, Deutsch, Français, Italiano, or हिन्दी. Choose the language on the **General** tab of Settings; by default MAIDR follows the browser's language. English is built in and each other language is a small locale pack beside `maidr.js` that MAIDR fetches when needed (or `import 'maidr/locale/ko'`). Chart data is read as the author wrote it; only MAIDR's own words are translated.
 For how the dictionaries work and how to add a language, see the [Languages documentation](docs/LOCALIZATION.md).
 
 ## Examples
--- a/cdnjs/maidr.json
+++ b/cdnjs/maidr.json
@@ -25,7 +25,15 @@
         "files": [
           "maidr.js",
           "maidr.css",
-          "maidr-math.css"
+          "maidr-math.css",
+          "locale-ko.js",
+          "locale-ja.js",
+          "locale-zh.js",
+          "locale-es.js",
+          "locale-de.js",
+          "locale-fr.js",
+          "locale-it.js",
+          "locale-hi.js"
         ]
       }
     ]
--- a/docs/LOCALIZATION.md
+++ b/docs/LOCALIZATION.md
@@ -27,6 +27,36 @@ Chart data is never translated. Axis labels, category names, and titles are
 read exactly as the chart's author wrote them; only MAIDR's own words around
 them change. The AI chat is asked to answer in the chosen language.
 
+## Loading a language
+
+English is built into `maidr.js`. Every other language is a **locale pack**, a
+small file beside the bundle, so a page pays only for the languages it uses.
+
+MAIDR fetches the pack itself when a language is chosen, from the directory
+`maidr.js` was loaded from: a page that loads
+`https://cdn.jsdelivr.net/npm/maidr/dist/maidr.js` fetches
+`https://cdn.jsdelivr.net/npm/maidr/dist/locale-ko.js` the moment Korean is
+needed. Until the pack arrives, announcements are in English; the pack
+re-renders them as soon as it registers.
+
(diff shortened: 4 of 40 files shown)


jooyoungseo:
Review follow-up, pushed in 54493b1:

- **1. Later adopter misses earlier packs**: fixed. The adopted queue now keeps every pack it has seen, and a bundle adopting later replays that list before chaining, so both bundles hear every pack whenever it arrived. `pack.test.ts` gains "should catch a later bundle up on packs an earlier one already took".
- **3. Author's own tag resolves early**: fixed. The loader now treats a pack tag the author placed like its own and settles on that tag's `load`/`error` rather than answering before it has run; the corresponding test now asserts the promise resolves true only after the tag loads and registers.
- **2. No `types` condition on `./locale/*`**: left as is, deliberately. The packs export nothing, so there is no declaration to ship; a bare side-effect import resolves without one under the default TypeScript settings (`noUncheckedSideEffectImports` is off). If a consumer with that flag on reports it, a one-line `export {}` declaration per pack is the fix.

Also in this round: acb8407 guards the diagnostics module's `instanceof HTMLScriptElement` against a DOM that has `document` but not the element constructors, which the recharts DOM tests stand up and which this PR put on every chart's import path.

---

jooyoungseo:
Second-round follow-up, pushed in 23aefaf:

- **Loader placement**: moved to `src/util/i18n/localePack.ts`, beside the registry it feeds and matching the KaTeX stylesheet loader precedent; the test moved with it. `SettingsService` imports it from there.
- **Redundant template literal**: removed.
- **E2E red on the previous heads**: this PR's own doing. The E2E job built only the core bundle, so the packs the example pages fetch did not exist on the runner. `scripts/build.js` gains a `locales` shorthand for every pack build and the job now runs `node scripts/build.js core locales`; a test pins the shorthand's expansion.

---

PR #1260, opened 2026-09-17
fix(plotly): scope a choropleth's regions to its own map, not every map

jooyoungseo:
## Description

`choroplethRegionSelectors` scoped each region by `.geolayer > g.geo.${geoId}`, which is a class **token** match. Plotly writes a geo subplot's class attribute as `"geo " + id`, so the first subplot's attribute is the doubled string `class="geo geo"` — whose `classList` is the single token `["geo"]`, a token every other geo subplot carries as well (`class="geo geo2"` → `["geo", "geo2"]`).

So `.geolayer > g.geo` and `.geolayer > g.geo.geo` are the same selector. Measured in Chromium on three maps — `geo`, `geo2`, `geo10` — carrying two regions each:

| selector | matched | `__data__.loc` |
|---|---|---|
| `…g.geo.geo …:nth-of-type(1)` | **3** | `['USA', 'FRA', 'IND']` |
| `…g.geo.geo2 …:nth-of-type(1)` | 1 | `['FRA']` |
| `…g.geo.geo10 …:nth-of-type(1)` | 1 | `['IND']` |
| `…g[class='geo geo'] …:nth-of-type(1)` | 1 | `['USA']` |

Only the doubled id collapses to an ambiguous token, so only the layer drawn on `geo` is affected — and `geo10` collides with neither of the ids it is prefixed by, because attribute matching is whole-string.

**What it cost was that layer's whole highlight.** `ChoroplethTrace.mapToSvgElements` compares the resolved elements against the declared regions and returns `null` on a mismatch, so an N-region layer resolving N elements in each of M maps is withdrawn outright rather than outlining the wrong country. The symptom is one map on a figure silently losing its outline while its neighbours keep theirs — which is also why nothing upstream flagged it.

## Related Issues

No issue filed. Found while giving the Python emitter the same selector (xability/py-maidr#640), where the ambiguity was measured first and the exact-attribute form shipped in xability/py-maidr#798. This brings the JS adapter to the same conclusion; the two paths are independent, since py-maidr emits its own selectors rather than using these.

## Changes Made

- `src/adapters/plotly/selectors.ts` — the scope becomes `g[class='geo ${geoId}']`, plus a doc block recording the measurement and why this family alone needs it.
- `test/adapters/plotly/choroplethSubplotScope.test.ts` (new) — a multi-map figure, which is the case nothing covered.
- `test/adapters/plotly/extractor.test.ts` — the two existing assertions that pinned the old string.

**Swept the other subplot families rather than assuming.** Measured beside geo, a polar subplot's attribute is the bare id — `class="polar"`, `class="polar2"` — and a ternary's likewise, so `g.polar` already names exactly one subplot and `g[class='polar polar']` matched nothing at all. `polarSeriesSelectors` scopes by token and is **correct as written**; the doubling is specific to geo. That is recorded in the doc block so the next reader does not "fix" polar to match.

The new test builds its fixture with `setAttribute` rather than `classList`, because collapsing `"geo geo"` to one token is precisely the doubling the test exists to reproduce — built through the token list, it would pass against the buggy selector too.

## Testing

- `npx jest test/adapters/plotly/choroplethSubplotScope.test.ts test/adapters/plotly/extractor.test.ts` → **226 passed**
- `npm run type-check` → clean
- `npx eslint` on the three changed files → clean

The browser measurements in the table above were taken in Chromium against real plotly figures (topojson served locally so no network was needed).

**Not yet reported here:** the full `npm test` run was still in flight when this PR was opened. I will confirm it, and a mutation check that the new test goes red against the old token-form selector, in a follow-up comment — and push a fix if either turns up a problem.

## Checklist

- [ ] I have tested the changes locally following `ManualTestingProcess.md`, and all tests related to this pull request pass. — targeted suites, type-check and lint pass; full suite confirmation to follow as noted above.
- [ ] I have updated the documentation, if applicable. — none applies; the reasoning lives in the doc block and the test.

## Additional Notes

The fix is one line; the rest is the record of why, since a selector that silently stops matching is invisible to every other test in the repo — the highlight goes and the reading, sonification and braille stay correct.

https://claude.ai/code/session_01Fb22qFkq8uRkgsK2WXDQiT

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fb22qFkq8uRkgsK2WXDQiT)_


jooyoungseo:
Closing the two gaps the PR description flagged as still in flight. Both are now done and verified.

**Mutation check — the new test does catch the bug.** Reverting only the one-line scope back to the token form (`g[class='geo ${geoId}']` → `g.geo.${geoId}`) turns **4 of the 8** new tests red:

```
✓ gives the first geo subplot a doubled class attribute
✓ matches every subplot from the token form the selector used to use
✓ leaves the ids that are not doubled unambiguous
✕ resolves each map's selectors to that map's regions alone
✕ gives each region exactly one element
✕ scopes by the whole class attribute rather than by a token
✓ still resolves every region, the ordinary case
✕ does not collide with the ids it is prefixed by
```

and the jsdom failure reproduces the browser measurement exactly — `nth-of-type(1)` comes back with two elements:

```
Received object: [<path class="choroplethlocation" data-loc="USA" />,
                  <path class="choroplethlocation" data-loc="FRA" />]
```

Worth noting which one stays green: **`still resolves every region, the ordinary case`** (a single-map figure) passes against the old form too. That is the point — the fixture reproduces the real asymmetry rather than failing indiscriminately, so it pins the actual defect instead of just the string.

**Full suite.** `npx jest --selectProjects unit --coverage=false` → **564 suites, 7880 tests, all passing**; the esm project → 204 passing; `npm run build` → clean.

Under `npm test` (which adds coverage) one unrelated suite, `test/model/flowNetworkPointIndices.test.ts`, had its jest worker `SIGKILL`ed — a sandbox memory limit, not an assertion failure. Verified it is not this change's: it passes standalone in 0.48 s (10/10), and it contains zero references to `geo` or `choropleth`. CI will be the real word on it.

One correction to something implied in the PR description, since the measurement contradicted the first guess: **only the layer drawn on the default `geo` subplot lost its highlight** — `g.geo.geo2` requires *both* tokens and so already matched exactly one container, meaning `geo2`, `geo3` and `geo10` were correct under the old form. So the symptom was one map on a figure silently losing its outline while its neighbours kept theirs, which is the PR body's table and is why nothing upstream flagged it. No *wrong* region was ever outlined: the count gate fails and withdraws the highlight rather than pointing at another country.

---

PR #1265, opened 2026-09-18
feat(keybinding): let a reader change any shortcut from the help menu

### LINKED ISSUE #189: feat: editable keyboard bindings
# Feature Request

## Motivation

The keybindings are currently fixed to specific keys. We have to introduce edit buttons in the react keyboard shortcut menu and record the user keybindings to map the new bindings to existing actions.

## Proposed Solution

1. Provide unique id to each action in the backend
2. Add editable keys on focus/hover of the keyboard shortcut row.
3. Record the new keys. Since we have migrated completely to react, there are a whole lot of libraries to carry of this function. [This](https://react-hotkeys-hook.vercel.app/docs/api/use-record-hotkeys) is one such example, but feel free to suggest any other robust one.
4. Use the id of the action to map the keybindings dynamically.
5. Update the keyboard shortcut to show the new keys.

## Checklist

- [x] I have searched for similar feature requests and confirmed that this is a new request.
- [x] I have provided a clear and concise description of the feature.
- [x] I have explained the motivation behind this feature request.
- [x] I have outlined a proposed solution or ideas for implementing this feature.
- [ ] I have provided any additional context or screenshots if applicable.

jooyoungseo:
# Pull Request

## Description

The shortcuts were fixed. A reader whose screen reader, browser or hands already used a key MAIDR wanted had no way to move MAIDR's binding, and the help menu that listed the keys could only be read. #189 asked for editable bindings in the keyboard shortcut menu: a unique id per action, an edit control on each row, the new keys recorded, the mapping applied dynamically, and the menu updated.

This does that, in the help menu, on every row but the help chord:

- **Change** on each row, named after its action (`Change shortcut for Toggle Braille Mode`) so a screen reader says which shortcut it changes. Pressing it prompts for the new shortcut in the dialog's live region.
- The next key or chord is **captured rather than run**. The keydown is stopped before it reaches hotkeys-js on the document and before MUI's dialog sees it, so Escape keeps the old shortcut without closing the dialog, Backspace restores the default, a modifier on its own is waited through, and Tab is let through so the reader is never trapped.
- The live region says what happened: `Toggle Braille Mode is now shift + b.`, or `b is already used by Toggle Braille Mode.` when the key is taken, or `That key cannot be used as a shortcut.` for Tab, the function keys and the lock keys.
- A changed row shows `shift + b (custom, default b)` and gains **Restore default**; **Restore all default shortcuts** appears above the list when anything is changed.
- The new key works **immediately**, without a reload, and **survives** one.

## Related Issues

Closes #189.

## Changes Made

- `src/util/keyCombo.ts` (new): the one spelling of a shortcut. `comboFromKeyboardEvent` turns a keydown into the string hotkeys-js binds, naming the physical key (`event.code`) so Shift and the slash key give `shift+/` and fire again on any layout; `normalizeCombo` makes `Control+Shift+P`, `shift+ctrl+p` and `ctrl+shift+p` compare equal; `combosOf` splits a multi-key binding such as `=, shift+=, num_add` while keeping a lone `,` as the comma key; `formatCombo` spells a lone symbol in words for the help menu.
- `src/type/settings.ts`: `general.keybindings`, a record of command key to shortcut, default `{}`. Keyed by **command** rather than by key, which is the "unique id" the issue asked for and what makes a command bound in several scopes (the arrows in trace and braille mode) move in all of them at once.
- `src/service/keybinding.ts`: `resolveOverrides` drops anything from localStorage that is not a string shortcut for a rebindable command; `getKeymapForScope(scope, overrides)` applies them; `findBindingConflict` checks the wanted key against the effective keymap of every scope the command lives in, hidden bindings included, so nothing can shadow Escape or another override; `KeybindingService` now implements `Observer<Settings>` and, when the overrides differ, unbinds and rebinds in the scope in force. `TOGGLE_HELP` is the one fixed command: the unassigned-key warning spells that chord out, and a reader who moved it and forgot where would be told to press a key that no longer opens anything.
- `src/service/help.ts`: rows carry `commandKey`, and `defaultKey`/`isCustom` when changed; `rebind`, `resetBinding` and `resetAllBindings` save through `SettingsService` and return the message to announce. The menu cache now also invalidates when the overrides change. The third constructor argument is optional so the service still works read-only where a test builds it without settings.
- `src/state/viewModel/helpViewModel.ts`: `status` and `statusRevision` in the slice, and `rebind`/`resetBinding`/`resetAllBindings`/`announce` that call the service and refresh the list.
- `src/ui/component/Help.tsx`: the buttons, the recording state, the keydown capture and a `role="status"` live region re-mounted per status so a refusal met twice is said twice. The Close button stays the only button in `DialogActions`, which the e2e helpers select by position.
- `src/controller.ts`: registers the keybinding service as a settings observer, passes the saved overrides at registration, hands the settings service to the help service, and removes the observer on dispose.
- `src/util/i18n/*/keybinding.ts`: fourteen messages in all nine locales.
- `docs/CONTROLS.md`: a "Customizing Shortcuts" section.
- Tests: `test/util/keyCombo.test.ts` (14), `test/service/keybindingOverrides.test.ts` (13, driving hotkeys-js in jsdom: the new key runs the command, the old one warns, a settings change rebinds in place, an untouched settings change does not, the scope in force is kept), `test/service/helpRebind.test.ts` (12), `test/ui/helpRebind.test.tsx` (10, the accessible names, the prompt, capture, Escape, Backspace, refusal, the live region), and `e2e_tests/specs/customShortcuts.spec.ts` (5). One existing assertion in `test/service/help.test.ts` loosened from exact shape to `objectContaining`, since rows now carry the command.

Measured end to end in Chromium against `examples/dot.html`:

| Step | Result |
| --- | --- |
| Change shortcut for Toggle Braille Mode, press `x`, close | status: `Toggle Braille Mode is now x.` |
| press `x` | braille display appears |
| press `b` | stays up; `Invalid key. Press Control Slash for keyboard help.` |
| reload, activate, press `x` | braille display appears |
| Change shortcut for Toggle Text Mode, press `b` | status: `b is already used by Toggle Braille Mode.` |
| Restore all default shortcuts, close, press `b` | braille display appears |

## Checklist

- [x] I have tested the changes locally following `ManualTestingProcess.md`, and all tests related to this pull request pass. (`npm run lint:fix`, `npm run type-check`, `npm run build`, `npm test`: 7975 unit and 215 ESM tests pass; the new and the dot e2e specs pass in Chromium.)

## Additional Notes

Decisions worth a maintainer's eye:

- **The help chord cannot be moved.** Everything else can. Easy to relax by removing it from `FIXED_COMMANDS` if the unassigned-key warning is taught to read the effective chord.
- **A conflict is refused, not swapped.** The issue did not say; refusing and naming the other action keeps the reader in control and never takes a shortcut away silently. A swap could be added later.
- **A multi-key default (`=, shift+=, num_add`) becomes a single key when changed.** The recorder captures one chord; the default's alternatives come back with Restore default.
- **No settings-dialog tab.** The issue put the editing in the shortcut menu, and the settings dialog's tab list is pinned by `settings.tabs.test.tsx`; the overrides live in the settings object so they travel with the other preferences and reset with them.

https://claude.ai/code/session_01HNMgVtK5UEmuRhsNra2v4k

---
_Generated by [Claude Code](https://claude.ai/code/session_01HNMgVtK5UEmuRhsNra2v4k)_


jooyoungseo:
Both findings were real, and both are fixed in 6c770f4.

1. **Stop Autoplay** joins `FIXED_COMMANDS`. Its wildcard handler fires on a bare modifier press and never reads the entry's hotkey, so an override could only ever add a listing while Control kept working. The row no longer offers a change, and `resolveOverrides` drops a saved override for it. Pinned in `test/service/keybindingOverrides.test.ts`.
2. **Backspace on a row at its default** now answers "Shortcut unchanged.", the same words Escape gets, rather than claiming Backspace is a key no shortcut may take. Pinned in `test/service/helpRebind.test.ts`.

The ViewModel calling the service to act is, as you note, the pattern `toggle()` already had; left as is so this PR does not restructure the help view model.

---

jooyoungseo:
Fixed in 94d1d78. `findBindingConflict` now prefers, among the commands claiming a key, a listed and movable one over a hidden one, and a fixed one (`STOP_AUTOPLAY`, the help chord) only when nothing else holds the key. A bare `up` is refused as used by Navigate Up rather than by Stop Autoplay; Escape, which only hidden bindings claim, is still named. Pinned in `test/service/keybindingOverrides.test.ts`.

---

PR #1266, opened 2026-09-18
feat(tableau): follow a mark a user clicks in the viz with MAIDR's cursor

### LINKED ISSUE #938: feat(tableau): sync the MAIDR cursor from a user's mark selection
Follow-up to #932.

Navigation currently flows one way: MAIDR's cursor drives Tableau's mark selection through `selectMarksByValueAsync`. Clicking a mark in the viz does not move MAIDR's cursor, so a sighted colleague pointing at a bar and a screen-reader user reading the chart can end up on different marks.

**Work:** listen for `MarkSelectionChanged`, read `event.detail.getMarksAsync()`, and move MAIDR's cursor to the mark the user clicked.

**Why it was deliberately left out of #932 — two real blockers, not just effort:**

1. **Re-entrancy.** The adapter's own `selectMarksByValueAsync` calls fire `MarkSelectionChanged` too, and the API documents no way to tell a programmatic selection from a user one. The `guard.programmatic` flag in `selection.ts` exists in anticipation of this but has never been load-bearing, because #932 does not listen for the event at all. A naive listener loops.
2. **The reverse lookup is lossy.** `MarkInfo` carries no stable mark id, so a position has to be reconstructed from field values — which is ambiguous whenever two rows share the same dimension values, exactly the detail-dimension case the point-cloud path already treats as unaddressable.

Design both before writing the listener.

jooyoungseo:
# Pull Request

## Description

Navigation in the Tableau adapter flowed one way: MAIDR's cursor drove Tableau's mark selection, but a click on a mark in the viz moved nothing, so a sighted colleague pointing at a bar and a screen-reader user reading the chart could be on different marks. #938 asked for the reverse direction and named the two reasons it had been left out of #932: the adapter's own `selectMarksByValueAsync` calls fire the same `MarkSelectionChanged` event, with no way to tell them from a user's click, and a mark carries no stable id, so a position has to be reconstructed from field values and is ambiguous wherever two rows share them.

This does the listener, and settles both problems in the selection module rather than in the binder. It also adds the general capability underneath, because "move the reader to the mark the host's chart was clicked on" is not Tableau-specific: `window.maidrLive.navigateTo` / `navigateMaidr` is the inbound half of `onNavigate`, spelled in the same currency, usable by any adapter or page.

What a reader gets:

- A colleague clicks a mark in the viz. If the reader is inside MAIDR, the cursor moves there at once and the mark is announced, with text, braille, audio and highlight together — into another worksheet's subplot if that is where it is.
- Usually the reader is *not* inside at that moment, because the click took the keyboard focus into the iframe and MAIDR's controller was disposed with it. The target is kept and the reader lands on it when they next Tab in — after the instruction, so the arrival is the first navigation and turns announcements on.
- A click on empty space withdraws a kept target; a data change discards it.
- The colleague's click is not undone. Focus leaving the figure used to clear the selection unconditionally; now it clears only what MAIDR still holds, and when focus left *into the viz* it waits a second for the click's own report, which releases the worksheet.

## Related Issues

Closes #938.

## Changes Made

- `src/type/grammar.ts`: `NavigationTarget` — `{ layerId, row, col }` or `{ layerId, pointIndex }`, the inbound counterpart of `NavigateCallback`.
- `src/type/navigation.ts`: `PointCloudAddressable` (`positionOfDataIndex`) with a guard; `ScatterTrace` already had the method and now declares it.
- `src/model/context.ts`: `Context.canNavigateTo(target)` and `Context.navigateTo(target)`. The target is resolved first — the layer located across every subplot, a point index translated through `positionOfDataIndex`, the cell checked against the grid — and refused without moving anything when it does not resolve or a virtual layer is on top of the stack. Otherwise the plot stack is rebuilt silently the way `enterSubplot` / `selectTrace` build it — the layer selected within its subplot before the figure's cursor moves, figure at the bottom for a multi-panel figure, the figure's own cursor moved under `runSilently` — and the trace's `moveToIndex` makes the one announcement.
- `src/controller.ts`: `Controller.navigateTo` asks `canNavigateTo` first, so a refused target leaves every mode as it was; only for a target that will be taken does it close the candlestick delta layer silently, leave grid-cell scope and reset the rotor to data mode (as `updateData` does), then delegate.
- `src/service/liveData.ts`: `LiveDataManager.register` takes a `navigator`; `navigateTo(target, { id })` routes to it (`null` withdraws), with the same sole-chart default and warnings as `appendData`; `navigateMaidr` and `MaidrLiveApi.navigateTo`.
- `src/state/hook/useMaidrController.ts`: registers a navigator that moves a live controller or keeps the target in a ref; applies a kept target on focus-in after the instruction; drops it on any data change.
- `src/index.tsx`, `src/react-entry.ts`: expose `navigateTo` / `navigateMaidr` / `NavigationTarget`.
- `src/adapters/tableau/types.ts`: `TableauMarksCollection`, `TableauMarksSelectedEvent`, mirrored from the contract and exported from the adapter index.
- `src/adapters/tableau/selection.ts`: `positionOfMarks` (marks → the one position they were read from, by field name, `null` for an ambiguous mark or a selection spanning positions), the issued-selection ring on the bridge with `takeEcho` (oldest match consumed), `owned` on the bridge, `clearOwnedSelections`, and `handleMarkSelection`, which fetches the marks, drops echoes, releases the worksheet on a user's change, withdraws a kept target when nothing resolves, and hands a position to the navigator.
- `src/adapters/tableau/binder.tsx`: listens for `markselectionchanged` and delivers through `liveDataManager.navigateTo(…, { id: figureId })`; the focusout clear is ownership-aware and, when focus went into the viz, deferred by `SELECTION_HANDOFF_MS` (1 s), cancelled by a focusin; a pre-read clear releases everything; all of it removed on dispose.
- `docs/tableau.md`: "Following a click" section, the event table row, the Limitations bullet rewritten from "one-directional" to what remains undecidable. `docs/LIVE_DATA.md`: `navigateTo` for the script-tag and React APIs.
- Tests: `test/model/contextNavigateTo.test.ts` (13: single panel, across panels from the lobby and from inside one, across layers, point cloud by data index, refusals leave everything untouched, `canNavigateTo` moves nothing, virtual layer), `test/controller/navigateTo.test.ts` (2, on a real `Controller`: a refused target leaves the delta layer, scope and announcement untouched; a good one closes the layer and lands), `test/service/liveData.test.ts` (+5 registry cases), `test/adapters/tableau/selection.test.ts` (+22: reverse lookup by field name, no worksheet named, wrong worksheet, missing field, unknown value, empty selection, point cloud, ambiguous address via a declared point cloud, multi-mark selections, date by the day; echoes late and in flight, a click in another worksheet while a call is in flight, a repeat of an echoed mark, deselect, echo of a clear, unresolved, refused, a failing `getMarksAsync`; ownership), `test/adapters/tableau/binder.test.tsx` (+8: click moves the cursor through the registry, echo ignored, deselect withdraws, the handoff keeps a clicked mark, clears after the handoff when no click followed, clears at once elsewhere, clears nothing when holding nothing, listener gone on dispose; one existing case now navigates before it leaves), `test/state/hook/maidrControllerNavigateTo.esm-test.tsx` (4: moves at once while inside, kept and landed on at focus-in, withdrawn, refused — the whole chain through the real controller). Fakes: `fakeMarks`, `fakeMarkSelection`.

Not applicable: the Tableau side needs a live Tableau Public connection, which the CI and this environment do not have. The behaviour is pinned against the same duck-typed fakes the rest of the adapter is tested with, and the full chain from `navigateMaidr` to the announcement runs against the real controller in the ESM test.

## Checklist

- [x] I have tested the changes locally following `ManualTestingProcess.md`, and all tests related to this pull request pass. (`npm run lint`, `npm run type-check`, `npm run build`, `npm test`.)

## Additional Notes

Decisions worth a maintainer's eye:

- **Echoes are recognised by what they selected, not by timing.** The event arrives through the iframe's message channel after the issuing promise has settled, so a flag raised around the call would classify nothing — and the one flag the bridge has is shared by every worksheet it binds, so it would read a click in one worksheet as the echo of a call in flight to another. The ring of issued positions, recorded before each call is made, is what makes both the late and the early echo safe; its one cost is that a user's click on the very mark MAIDR is on reads as an echo — and the cursor is already there.
- **An ambiguous mark moves nothing.** Rather than picking the first of two positions that share an address. Reachable only on a point cloud the page declared with `overrides[name].traceType` over rows that repeat a detail value, since the heuristics treat a repeated value as a category, not a point.
- **A kept target survives the reader's absence.** The selection is still visible in the viz, so starting there matches the visual state. It is dropped whenever the data changes, and by a click on empty space.
- **The handoff is one second.** Long enough for Tableau's report of the click to arrive, short enough that a click on a filter control leaves MAIDR's highlight in place for no longer than that. Pinned in the binder tests with fake timers.
- **`navigateTo` is general.** Other adapters that report `onNavigate` (Chart.js, amCharts) could wire their own click handlers to it; that is not done here.

https://claude.ai/code/session_01HNMgVtK5UEmuRhsNra2v4k


jooyoungseo:
The finding was real, and all three points are addressed in c032e45.

1. **A refused target no longer touches any mode.** `Context.canNavigateTo` answers whether the target names a position the figure has, without moving; `Controller.navigateTo` asks it first and returns `false` before the delta layer, grid cell or rotor are touched. Pinned in `test/controller/navigateTo.test.ts` on a real `Controller`: with the delta layer active, an off-grid cell and an unknown layer are both refused with no scope change and nothing new announced, and a good target closes the layer and lands on the mark.
2. **The subplot-entry fallback** now selects the layer within its subplot *before* the figure's cursor moves, so the one step that could still refuse does so with the lobby untouched, and the recovery path that could not honour "nothing moves" is gone.
3. **The shared guard flag is out of the echo check.** You are right that one flag across every bound worksheet would read a click in worksheet B as the echo of a call in flight to A. The ring records a selection before its call is made, so it already covers an echo that beats the promise, and it keys by worksheet and position. Pinned both ways in `selection.test.ts`: an echo arriving while its own call is in flight is an echo; a click in another worksheet while that call is in flight is a click.

---

jooyoungseo:
The asymmetry stays as is, on purpose: `canNavigateTo` answers for the target alone so the controller can ask it *while* the delta layer is still active and then close that layer before moving. Folding `isOnRealLayer` into it would make the controller unable to leave the delta layer for a good target at all. Its docstring names the narrowing, and `Controller.navigateTo` is the one caller.

---

PR #1268, opened 2026-09-18
feat(pie): carry a pie's start angle and drawn direction through the grammar

jooyoungseo:
# Pull Request

## Description

`PieTrace` laid every pie out from 12 o'clock going clockwise, on the assumption (stated in `src/model/pie.ts` and in #780) that producers all draw that way. They do not: matplotlib and base R's `pie()` start at 3 o'clock and go counterclockwise, and plotly goes counterclockwise. For those charts Right stepped the wrong way round the dial, the audio panned each slice to the side it is not on, and `p` named a clock position the slice is not at. xability/py-maidr#800 hit exactly this on py.maidr.ai's own pie example (Sat drawn from 12 o'clock down the left was panned hard right and announced "from 12 o'clock to 4 o'clock") and fixed it producer-side for matplotlib; this is the renderer-side follow-up it names, so a producer can declare the geometry instead of rewriting its document.

## Related Issues

#780 (the "clockwise from 12" assumption it left open), #784, xability/py-maidr#800.

## Changes Made

- `src/type/grammar.ts`: two optional layer keys, read by the pie trace and ignored elsewhere.
  - `startAngle`: where the first slice begins, in degrees clockwise from 12 o'clock (default 0). matplotlib's `startangle=0` is `90` here; Chart.js's `rotation` carries over unchanged.
  - `direction`: `'clockwise'` (default) or `'counterclockwise'`, the way successive slices follow one another from `startAngle`. New `PieDirection` enum.
- `src/model/pie.ts`: the walk is clockwise from `startAngle` whichever way the pie was drawn. A `'counterclockwise'` layer walks `data` in reverse (the slice drawn last is the first one clockwise from the start), with the highlight elements reversed alongside so element k is still slice k; the description's table follows the walk. `startAngle` offsets the pan. The trace state carries `startAngle` (new optional field in `src/type/state.ts`). The comment claiming matplotlib lays a pie out clockwise is corrected.
- `src/command/describe.ts`: `p` places a slice from `startAngle`; a slice that straddles the top reads "from 11 o'clock to 1 o'clock" rather than ending before it starts.
- `src/adapters/plotly/extractor.ts` + `types.ts`: declare plotly's direction (its default is counterclockwise) and where the ring starts. That is not `rotation` alone for a counterclockwise pie: plotly ends the *first* wedge at `rotation` and draws it clockwise of there, then lays the rest out counterclockwise (`pie/plot.js`, `setCoords`), which is how a default pie's largest slice sits to the right of 12 o'clock. The ring therefore starts at `rotation` plus the first wedge's sweep; a clockwise pie starts at `rotation`.
- `src/adapters/chartjs/extractor.ts`: carry a non-zero `options.rotation` over as `startAngle` (Chart.js is already clockwise from the top).
- `docs/SCHEMA.md`: document both keys in the pie section.
- Tests: pie model (start angle offsets and wraps, reversed walk keeps label/value/share/pan/table aligned, layer left untouched), pie highlight (reversed walk outlines the wedge drawn for the slice), position announcement (start angle, straddling 12), plotly (direction, start from the first wedge's end, rotation added) and Chart.js extractors.

Driven in headless Chromium with the rebuilt bundle on `examples/plotly-pie.html` (values 30/50/20/15, plotly's defaults, so drawn Bananas, Apples, Cherries, Dates counterclockwise with Bananas to the right of 12 o'clock):

| Step | Announced | Outline lands on | `p` |
|---|---|---|---|
| Right ×1 | Dates 13.0% | the bottom wedge | from 5 o'clock to 7 o'clock |
| Right ×2 | Cherries 17.4% | lower left | |
| Right ×3 | Apples 26.1% | upper left | |
| Right ×4 | Bananas 43.5% | right | from 12 o'clock to 5 o'clock |

Before this PR the same page walked Bananas → Apples → Cherries → Dates, i.e. counterclockwise, and placed Bananas "from 12 o'clock to 5 o'clock" only by coincidence of it being first.

## Checklist

## Additional Notes

- `npm run lint:fix`, `npm run type-check`, `npm run build` and `npm test` (8057 unit + 219 ESM) all pass.
- Backwards compatible: a layer that declares neither key behaves exactly as before. The plotly adapter now always declares `direction`, so a plotly pie's walk turns round to clockwise, which is the fix; the first slice announced is the one clockwise-adjacent to the ring's start rather than the largest.
- Not covered here: the d3 binder (`d3.pie()` defaults to clockwise from the top, matching the grammar default, but a custom `startAngle`/`endAngle` is not read), and the R producers. py-maidr will emit `startAngle` for matplotlib pies and `direction`/`startAngle` for its plotly pies as the follow-up to xability/py-maidr#800.

https://claude.ai/code/session_01UCkYH6b3mVkPdHrry7AmWZ


jooyoungseo:
Both points taken, pushed as one commit:

- `docs/SCHEMA.md` now says what the grammar JSDoc and the adapter say: only a clockwise plotly pie starts at `rotation`; a counterclockwise one (plotly's default) ends its first wedge there and draws it clockwise of the top, so the ring starts at `rotation` plus that wedge's sweep, with the 50-of-100 → `startAngle: 180` example spelled out.
- `extractPieLayer` leaves `startAngle` undeclared when the first drawn wedge is unknown (no calcdata and `sort` left on), the same condition under which it already withholds the selectors, rather than measuring the start off `slices[0]`; `direction` is still declared. Pinned by a new extractor test for exactly that combination.

---

jooyoungseo:
On the `NaN` nit: leaving it as is for this PR. The trace's `toStartAngle` already reads a non-finite start as the top, which is also what the adapter declares when the first wedge is unknown, so a malformed authored value lands in the same place either way; a stricter filter in the adapter can ride the next change to that extractor.

---

Issue #1273, opened 2026-09-21
Line-family traces never highlight an SVG `<polygon>`: ggplot2/gridSVG area, stacked area and polygon layers announce every point and outline nothing


jooyoungseo:
## Summary

`LineTrace.mapViaPathParsing` (`src/model/line.ts`) reads vertices only from an `SVGPathElement` (`d`) or an `SVGPolylineElement` (`points`). An `SVGPolygonElement` falls through both branches with no coordinates, `createCircleElements` gets nothing to place, and the series is reported as unusable, so the layer has no highlight while navigation and speech keep working.

gridSVG, which r-maidr renders through, exports every `geom_area()`, `geom_ribbon()`, stacked / `position = "fill"` area and `geom_polygon()` as a `<polygon points="…">`, and Base R `cdplot()` the same. Measured in headless Chromium against 4.9.0 (Tab, Right Arrow twice), reading `svg [data-maidr-owned]` with `visibility` not hidden:

| chart | layer | selectors | visible clone |
|---|---|---|---|
| `ggplot(df, aes(x, y)) + geom_area()` | `area` | `["#GRID\\.polygon\\.1633\\.1 polygon"]` | none |
| `geom_area(aes(fill = g))` | `stacked_area` (2 series) | one per series | none |
| `geom_area(position = "fill")` | `stacked_normalized_area` | one per series | none |
| `geom_polygon(aes(group = id))` | `line` (2 series) | one per series | none |
| Base R `cdplot(factor(am) ~ mpg)` | `stacked_normalized_area` | one per band | none |
| `geom_line()` (control) | `line` | `<polyline>` | 1, fill `rgb(3, 200, 9)` |

The polygon's `points` are the top edge (one vertex per sample) followed by the baseline edge back, so a plain area carries `2n` vertices. `AreaTrace`'s `withoutBaseline` (`src/model/area.ts`) only recognises matplotlib's `3n - 1` / `3n` closed-path shapes, so even once the vertices are read, the `2n` case would need its return journey dropped before `reconcilePathCoordinates` (which today interpolates when there are fewer vertices than points and would be handed twice as many).

## Suggested fix

- In `mapViaPathParsing`, read `points` from an `SVGPolygonElement` exactly as from a polyline.
- In `withoutBaseline`, also accept `2 * expected` and keep the first `expected` vertices (the drawn edge); for a stacked band the top edge is what the announced values sit on.

[...]

PR #1275, opened 2026-09-23
feat(model): keep reading the selector shapes producers already ship, and warn when a layer loses its highlight

jooyoungseo:
# Pull Request

## Description

Three 4.x changes to what a layer reads from `selectors` (#750 in 4.0, #991 in 4.4, #1135 in 4.5) were each correct for the producers in this tree, shipped as `fix:`, and between them took the highlight off r-maidr's and py-maidr's charts: every point still announced, nothing outlined, nothing in the console. py-maidr loads the latest maidr.js by default, so **every installed py-maidr up to 1.24 has had no scatter, strip, eventplot or rug highlight since 4.0**, whatever version the user has; only a change here can fix those copies.

The root cause: before 4.0 a flat list reaching a model that reads one string went to `querySelectorAll` as it was, and the DOM joined it into a selector list, so `["#bars rect"]` worked as `"#bars rect"`. Producers relied on that without knowing it. #750 stopped handing non-strings to the DOM (to stop `[]` taking the whole figure down), #991 gave a bar list a one-per-bar meaning, and #1135 made segmented layers series-major by default, which silently flipped r-maidr's `<rect>` bars onto the wrong bar.

## Related Issues

Follows the analysis behind xability/r-maidr#316 / #317 and xability/py-maidr#807; completes the protection those binding-side fixes could not give already-installed copies.

## Changes Made

- **`fix(model)`, legacy lists** (`src/util/selectors.ts`, `bar`, `segmented`, `heatmap`, `scatter`, `pie`): a flat list of non-blank strings that is not one entry per bar is joined into the selector list it stood for, and the producer is told once in the console to emit a string. A segmented list that declares no `domMapping.order` and whose marks are `<rect>` is walked category by category, as before #1135 (read series by series it outlines a bar other than the one announced). An empty list, a blank entry, a grid and a per-bar list keep today's meaning, so #750's protection stands.
- **`feat(model)`, the warning** (`TraceFactory.create`): once per layer, when a layer declares selectors and the built trace has nothing to highlight, the console says so, naming the three causes (no match, a shape the type does not read, a payload attached before the chart finished drawing) and linking the contract. This alone would have made the regression visible on day one.
- **`docs`, the contract**: `docs/SCHEMA.md` gains a *Selectors* section with a table of every declarable type, the shapes it reads and how matches pair with data (derived from each model, with file-and-line evidence), the pre-4.0 lists still read, and the rule for changing any of it; a test fails when a type is missing or listed twice. The grammar's `selectors` JSDoc, which said an array is one element per point for every type, now points at it. `.claude/rules/git-workflow.md` (and its Copilot copy) says a change producers depend on keeps the old shape working with a warning for a major version, carries `!`, and passes the binding spec.
- **`test(e2e)`, binding output in CI**: 14 real r-maidr and py-maidr pages in `e2e_tests/fixtures/bindings/` (6 in the shapes they shipped before 4.x, 8 in today's), pointed at this build, with provenance in a README. `bindingOutput.spec.ts` presses the arrow keys and checks that a mark is outlined, that it moves with the reader, and on bar charts that the outlined heights rank the way the announced values do.

Measured in headless Chromium:

| | shipped 4.9.0 | this branch |
|---|---|---|
| `bindingOutput.spec.ts` (14 pages) | 7 fail: the 6 legacy pages and the gridSVG area of #1273 | 14 pass |
| r-maidr 0.5.0 pages (29): bar, hist, dot, lollipop, pie, point, facets, patchwork, Base R dodged/stacked | no outline | outline on the announced mark, bar heights ranking with values; identical to 3.75.1 on the ggplot2 dodged/stacked/normalized pages |
| py-maidr 1.24 pages: scatter, hue scatter, strip, eventplot, rug | no outline | outline follows the announced point |

Still unhighlighted, and not this repository's to fix: a pre-#317 r-maidr page whose two bar layers each named both layers' bars, and one whose mosaic had no selectors (both fixed in r-maidr); py-maidr swarm plots, whose selectors name an attribute py-maidr never writes (being fixed in py-maidr).

## Checklist

## Additional Notes

`npm run lint:fix`, `npm run type-check` and `npm run build` pass; the new and affected unit suites pass (`legacySelectorList`, `unresolvedSelectorWarning`, `selectors`, `schemaSelectors`, `barSelectorConsumption`, the Recharts orientation test). Three existing tests needed changing: the segmented "declines a flat list" test now pins the restored reading, and the Recharts orientation test reads only its own `[BarTrace]` warning since its undrawn chart correctly earns the new one. Locally the e2e spec ran against Chromium 1194 through a throwaway config because this sandbox cannot install the build Playwright 1.59 pins; CI runs it as configured.

Not `!`: nothing that works today stops working; the change restores shapes that stopped working.

https://claude.ai/code/session_01LAa1qJGAZtXYYmg9ntVNqK

---
_Generated by [Claude Code](https://claude.ai/code/session_01LAa1qJGAZtXYYmg9ntVNqK)_


jooyoungseo:
Thanks for rebuilding the pre-PR bundle and reproducing the six legacy failures; that is the check that matters here. On the three notes:

- **heatmap `selector` / `flatSelector`:** fair, the grid branch should read one variable. It is correct as written (a joined list is never an array), so it rides the next push that touches `heatmap.ts` rather than starting one.
- **`reported` growing:** one short key per layer that actually failed, and only failing layers add one; a page would need thousands of distinct broken layers for it to matter. Left as is.
- **early attach:** no trace re-resolves its selectors after construction (`highlightValues` is built once in the constructor; `watchViewport` only re-measures geometry of elements already held), so a payload attached before the chart is drawn really does leave that layer with no outline for the life of the figure. The warning is then the true statement, which is why the docs name that cause. py-maidr's Plotly path had exactly this shape of bug (xability/py-maidr#809), and the warning would have pointed at it.

---

PR #1278, opened 2026-09-23
feat(tactile): keep chart shapes intact and centred when zooming, and buzz at the ends of the braille line

jooyoungseo:
# Pull Request

## Description

DotPad tactile graphic improvements, from reading charts on a device:

1. **Bars broke apart when zooming in.** Each step centred on the middle of the focused bar. Its top and baseline fell off the pins, and from 6× on every pin was down.
2. **The cursor should stay in the middle of the pins while zooming.** Every zoom step, in and out, now centres on the focused mark.
3. **The braille line should not speak while scrolling.** Function 1/4 now move silently. Reaching either end vibrates the device with **one 300 ms pulse**, deliberately different from the double pulse the DotPad gives on connection.
4. **The same zoom breakage in other chart types.** All 213 examples were driven through a recorded DotPad at every zoom level (1× to 12× and back), and the frames were inspected. The fixes are listed below.

## Related Issues

None filed. This came from hands-on testing with a DotPad.

## Changes Made

**Bar-chart zoom** (`fix(tactile): quantise…`, `fix(tactile): centre the focused mark…`)
- Each zoom step centres on the focus.
- A tall bar is held by its value end: the edge the other bars don't share.
- Projected dot coordinates are quantised to 1/1000 of a pin. Otherwise float32 path noise splits one edge across two pin rows and a bar's top edge comes out as a staircase.

**Braille line** (`feat(tactile): buzz…`, `fix(tactile): mark the end…`)
- Scrolling is silent. The edge is marked by `requestVibrator(device, 300, 0, 1)`.
- On an SDK without vibration, the edge is spoken instead.

**Other chart types** (`fix(tactile): keep every chart's shapes intact…`)
- **No blank frames on zoom.** Before this, 41 of the examples tested went fully blank at some zoom level: funnel, treemap, icicle, sunburst, pie, hexbin, waterfall, ridgeline, sankey, wordcloud, live-line/candlestick and others.
  - A mark that doesn't fit the window is held by the nearest point of its own outline. The window may only move along the axis where the mark doesn't fit, so a floating waterfall bar is held by its top or bottom, not a long side.
  - A zoom that would still leave every pin down moves to the nearest mark.
- **Multi-piece SVG paths.** Each subpath is sampled between its own ends. Piece boundaries are measured with a detached path, so arcs and relative commands work. Closed pieces are still filled together, even-odd.
  - This fixes the error bar drawn as a "Z", and the slashes through plotly box plots and candles.
- **`<use>` / `<defs>`.** A `<use>` is followed to the shape it places. An element inside `<defs>` is never drawn directly. matplotlib violins are drawn this way.
- **Point markers stay points.** A closed, roughly square mark that is a disc at 1× stays the same disc at every zoom. Before, line vertices and scatter points grew into ellipses covering the line. Error bars and slim bars keep their shape.
- **No solid slabs.** A focused mark is outlined heavily once it covers ≥ 75% of the display both ways (was 90%).

**Violin** (`fix(violin): …`)
- The density layer reports each violin's drawn outline through `getGeometryElements`, as line traces do.
- Before, the tactile display drew the highlight circles synthesised along the curve, which became a ladder of loops when zoomed.

**Docs:** `docs/TACTILE_DISPLAY.md` updated (Zooming, The braille line).

Not included. The frame dumps were checked by eye across all examples.

## Checklist

## Additional Notes

- Checks run: `npm run lint:fix`, `npm run type-check`, `npm run build`. Unit tests: 583 suites / 8210 tests pass. ESM tests (via `node scripts/test.js --selectProjects esm`): 219 pass.
- Browser sweep: 197 examples drew and none produces a blank zoom frame. Some examples could not be exercised:
  - amCharts draws to canvas and plotly parcoords to WebGL, so there is nothing to draw on the tactile display. That is a pre-existing limitation.
  - `highcharts-choropleth` and `tableau-bar` did not load in the harness.
- Not yet checked on physical DotPad hardware.

https://claude.ai/code/session_01BtTi6X4uK5ZqTYRFhQhwLQ


jooyoungseo:
Re the review above:

1. **Zero-length subpath:** confirmed and fixed in 5b05e80. Each subpath start now carries its offset into the path data, so a bare `M` moves the next piece's start instead of shifting every later slice. A regression test (`M … M 2 2 M … Z`) fails without the fix.
2. **Hardware gate:** agreed that this can't be settled by the test suite. It is still unverified on a physical DotPad and is left to the maintainers before merge.

---

jooyoungseo:
Re the latest review: all three bugs reproduced and fixed. Each fix has a test that fails without it.

1. **`nearestOutlinePoint` flattened multi-piece paths:** fixed in 0d24d3d. It now walks `ring.parts` piece by piece, as the renderer does. Without the fix, the regression test (an error bar's two caps, drawn with no stem) lands on "50% down; nothing is in view".
2. **The settle pass undid a pan:** fixed in 4bcea9c. Only the first canvas pass follows the focus; the settled pass redraws with `'none'`.
3. **`vibrate()` reported success before delivery:** fixed in 4328782. `vibrate(onFailure)` calls back when the queued `requestVibrator` throws or rejects, and `signalLineEdge` speaks the fallback then.
4. **The duplicate docs sentence and the stale "Line part 1" comment** are removed in dced2cb.

Not changed:
- **`CANVAS_STROKE_PX` and narrow bars:** a bar narrower than 8 CSS px is indeed raised whole. From pixels alone, though, a narrow bar and a thick line are the same run. Outlining such runs brings back the double lines on thick strokes, which is the more common case on these charts. I left it documented rather than trading one artifact for the other.
- **Performance:** not measured beyond the example sweep, where no per-key lag was noticeable. It is fair to profile on the largest canvas charts.

---

jooyoungseo:
Re the latest review: all four findings were confirmed against the code and are fixed.

1. **A refused vibration repainted the display:** fixed in a23ce32. The refusal is logged and spoken but no longer rethrown, so it doesn't reach `onWriteFailure`. The test now also asserts that `onWriteFailure` is not called.
2. **The canvas-cover check measured the overlap against the canvas itself:** fixed in fdf2ca9. It is now measured against the chart's area.
3. **Border and letterbox pins read outside the plot area:** fixed in cfaa967. Each pin's patch is clipped to the plot area, with 1 px of grace so an axis line drawn on its edge survives. `TactileCanvas` takes the largest patch as the pin size. The new test puts a title band just above the plot area, and it fails without the clip.
4. **`captureClean` ran on every frame:** fixed in 6258bff. The tactile display sets `data-maidr-pixel-reader` on the overlay layer while it reads pixels, and the service clears it on dispose. Chart.js copies the canvas only while that attribute is set, and drops a stale copy otherwise.
   - Trade-off: the very first reading of a chart falls back to the visible canvas, because no clean copy exists yet.

Also:
- **Docs:** a4d62e1 documents the refusal fallback, and that an accepted-but-silent buzz can't be detected.
- **Browser re-check:** I re-ran the recorded-DotPad sweep on 7 examples (Chart.js bar, line, pie and scatter; amCharts XY and pie; plotly-parallel). No frame is blank. At 1×, the only pins that dropped were chart chrome outside the plot area.

Not changed:
- **Per-subpath samples:** the count is bounded by `MAX_SAMPLES + N`. Each subpath gets `ceil(samples × share)`, so it is not unbounded.
- **Buzz queued behind the frame:** this is deliberate, so the buzz follows what the key produced.

---

jooyoungseo:
Re the latest review:

1. **Tooltip in the first canvas read:** fixed in c76c8b6. The overlay watches for `data-maidr-pixel-reader` and repaints the chart through `chart.draw()` as soon as a reader arrives with no copy yet. The service's settled pass then reads the clean copy, so the tooltip is on the pins for at most the first 500 ms. A new test covers the repaint.
2. **A short square bar stayed a disc:** fixed in f5ce456. A mark whose own SVG outline is an axis-aligned rectangle is drawn as a shape, not a marker, and has a render test.
   - The canvas path is left unchanged: its focus is always a highlight box, rectangular for points too. Without that exception, the Chart.js scatter's focused point grew at deep zoom (121 → 336 pins).
   - An SVG sweep over 59 point, line and bar examples changed only the histograms, whose focused short bars now show their sides and baseline at 12×.
3. **`willReadFrequently`:** fixed in c76c8b6. It is now passed on the clean canvas's first `getContext`.
4. **Nits, fixed in 62e04c9:**
   - `requestVibrator` is typed `void | Promise<void>`.
   - A vibration queued for a display that has since gone is dropped, with no speech.
   - A test now covers an asynchronously rejected vibration.

Not changed:
- **`subpathStarts` quadratic cost:** it is paid once per distinct `d`, and the paths that have many moves (maps, hatching) are drawn once and cached. I'd rather cap it when a real chart shows the cost than guess a limit now.
- **Legend-swatch cover test:** the non-clean-canvas path composites through a 2D context that jsdom doesn't provide, so a unit test can't tell the old check from the new one. The browser sweep ran with the fix and was unchanged for Chart.js, amCharts and plotly-parallel.

---

jooyoungseo:
Re the latest review:

1. **Stale amCharts plot area:** confirmed, and it was worse than described. A move that resolved no targets returned before the plot area was written at all. Fixed in de9aa2e: the plot area is written first, and the lobby branch clears it. `applyHighlight` isn't exported or unit-tested, so this was checked with the browser sweep.
2. **Unrelated canvas taken for the chart:** fixed in 6183f33. The overlap must now be at least 50% of the chart *and* at least 50% of the canvas, so a large decorative canvas the chart sits on no longer qualifies.
   - Re-swept all 36 Chart.js, amCharts and plotly-parallel examples. Every one that loads still draws, and none adds a blank zoom frame.
3. **Queued vibration after a disconnect:** left silent on purpose. The previous review asked for the opposite, since there's no longer a display whose edge could be marked. The disconnect is announced on its own, so speaking "End of the line" for a session that's gone would be the misleading part.
4. **Quadratic `subpathStarts`:** capped in d869da2. A path with more than 256 moves is sampled as one run, as it was before this PR.

**Performance of the canvas read:** agreed it should be profiled on a device. In the harness, no per-key lag was noticeable on the largest canvas examples, but that's not the same as a real device. I'm leaving that to the hardware check that's still outstanding.

---

